PR #287, opened 2017-03-24
Additions to mod lists

acesnik:


EngineLayer/Mods/missing.txt
```diff
@@ -1,126 +1,147 @@
...
 //
-ID   G
+ID   Missing Glycine (G)
 PP   Anywhere.
 TG   G
-MT   missing
+MT   Missing
 MM   -57.02146372069
+CF   C-2 H-3 O-1 N-1
+//
```
acesnik: It was missing the end here, so that last one wasn't processed.

EngineLayer/Mods/missing.txt
```diff
@@ -1,126 +1,147 @@
 Missing
-ID   W
+ID   Missing Tryptophan (W)
```
acesnik: Clarifies that these are missing. This is what I'm using in ProteoformSuite.

If you like the conciseness of before, maybe you could at least put "-W" or "W-1" instead of "W."

stefanks:
Will not add these additions, since it is not clear where they could be localized to. I don't like just saying "they can be inserted anywhere". I changed the id of removals to have a negative sign

acesnik:
Sounds good!

Issue #302, opened 2017-03-28
Button to cancel processing without crashing MM


trishorts:
I frequently find that I want to change some setting and restart a search. Now. I have to crash MM and then reload everything. Be nice to just stop processing and be able to amend search paramters. Then start over.

acesnik:
I put 15 mins into this while waiting to get coffee earlier. Just aborting the EverythingEngine thread doesn't work. The CPU was still working hard. What other tasks need to be shut down?

stefanks:
All the parallel for loops it seems!

https://msdn.microsoft.com/en-us/library/dd460721%28v=vs.110%29.aspx

acesnik:
It turns out this is pretty tough. I tried to propagate aborting the thread that hosts the EverythingRunnerEngine through all of the Parallel.For loops, but this isn't a great solution, since you need to sense the change to the containing thread from within the parallel loop. There's also this [CancelationToken structure](http://stackoverflow.com/questions/6052699/how-can-i-kill-all-threads-spawned-by-parallel-foreach-from-my-main-thread) that I haven't tried, but it also depends on testing within the loop. It seems like the Parallel loops should have a list of the threads spawned, perhaps one with the active threads, or at least some way of accessing that information, but that doesn't exist. 

We might be able to achieve halting these processes efficiently if we do the thread managing ourselves instead of relying on the System.Threading.Tasks.

acesnik:
My changes are here: https://github.com/smith-chem-wisc/MetaMorpheus/pull/304, but I'm planning to abandon this issue for now.

acesnik:
Thanks to whomever removed the nested Parallel.For/Each loops, I was able to implement this fairly easily. It'll be in my next pull request.

zrolfs:
Neat! Can you have a window that pops up to make sure the user didn't accidentally click it?

acesnik:
Sure, good thought!

rmillikin:
done in #1280

PR #304, opened 2017-03-29
Cancel all tasks button

acesnik:
See https://github.com/smith-chem-wisc/MetaMorpheus/issues/302


stefanks:
Anthony, did you verify that it works, i.e. cpu usage goes to zero even when cancelling a process with nested parallel for loops? E.g. the calibration task?

acesnik:
Good thought. I only tested canceling the classic search. I'll look into how it works with calibration.

PR #335, opened 2017-04-06
Add peptide Length textbox

lonelu:
There is one part I am not sure. In order to parse Int? Input from TextBox in GUI/SearchTaskWindow.xaml.cs, I add the following code. 
+            int temp;
+           TheTask.MinPeptideLength = int.TryParse(txtMinPeptideLength.Text, out temp) ? temp : (int?)null;
+           TheTask.MaxPeptideLength = int.TryParse(txtMaxPeptideLength.Text, out temp) ? temp : (int?)null;

What you did for Another Int input is 
+      TheTask.MaxMissedCleavages = int.Parse(missedCleavagesTextBox.Text, CultureInfo.InvariantCulture);

### DIFF
--- a/EngineLayer/Analysis/AnalysisEngine.cs
+++ b/EngineLayer/Analysis/AnalysisEngine.cs
@@ -18,6 +18,8 @@ namespace EngineLayer.Analysis
         private const int max_mods_for_peptide = 3;
         private readonly double binTol;
         private readonly int maximumMissedCleavages;
+        private readonly int? minPeptideLength;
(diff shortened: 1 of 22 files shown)


stefanks:
Nice!
Could you rearrange it a bit so the peptide objects are not even created if the length for not match? With your change these peptides are created, but then filtered after the fact.

stefanks:
Length does not match*

stefanks:
Looks good! Merging

Issue #253, opened 2017-03-14
Protein PTM output


zrolfs:
I wrote some code yesterday to output the number of sites on each protein where a given PTM occurs. I think adding that output to MetaMorpheus would be a fast, easy way for users to visualize the PTM sites observed for each protein. I'll clean up the code and do a pull request sometime, if you think it would be worthwhile?

stefanks:
Zach, please coordinate this with Rob, since this feature would be closely intertwined with his parsimony code. It sounds useful, but I need more detail to understand the visualization part better!

acesnik:
It could also be useful to note where coverage wasn't observed, and thus where PTM sites could not be observed -- clearly annotating missing information.

rmillikin:
How about something like this?

mgkgtpsfgkrhnkshtlcnrcgrrSFHVQKKTCSSCGYPAAKtrsynwgakakrrHTTGTGRmrylkhvsrrFKN[Deamidation of N]GFQTGSASKasa

lowercase = not observed residue
uppercase = observed residue
brackets = mod at that location, same notation as PSM output

zrolfs:
I like Rob's suggestion, but that can still be cumbersome for a user to wade through to find differences in observed PTMs between different samples/conditions. I think we would still benefit by implementing the PTMs with the observed residues to provide an overview of what PTMs were found, but perhaps also have an additional column with an output like:

aa76v:[Deamidation of N] | aa101:[Acetylation] | etc.

Where:
"aa#" is position of PTM
"v:" signifies that the given residue was observed both with and without that PTM
"[*]" is the mod at that location, same notation as PSM output

acesnik:
You could put an estimated occupancy ratio (e.g. by PSM count) inside the brackets. xxxxxxxXXXX[mod1|info:occupancy=1]XXXX[mod2|info:occupancy=0.25][mod3|info:occupancy=0.5]xxXXXXXXXXX

In this example,
Mod1 is observed in all PSMs
Mod2 and mod3 are observed at the same site, which is unmodified 25% (1-0.25-0.5) of the time
Mod2 is observed 25% of the time
Mod3 is observed 50% of the time

zrolfs:
I like the occupancy idea. It would still be nice to have a list of modifications with their indexes, though, so that you can easily compare what PTMs were observed where without having to scan through each protein.

[...]

Issue #237, opened 2017-03-02
New task windows have elements that float out the window, extra whitespace, etc


stefanks:

stefanks:
Done in https://github.com/smith-chem-wisc/MetaMorpheus/pull/359

PR #352, opened 2017-04-12
#253 PTM output

lonelu:
#253 PTM output #237 GUI


lonelu:
Sorry, this one doesn't work with the new changes. Just delete it.

PR #479, opened 2017-07-20
Added Complementary Ion option for Search Tasks

zrolfs:


EngineLayer/ModernSearch/ModernSearchEngine.cs
```diff
@@ -153,12 +161,38 @@ protected override MetaMorpheusEngineResults RunSpecific()
...
+                //If ETD
+                if (lp.Contains(ProductType.C) | lp.Contains(ProductType.Zdot))
+                {
+                    for (int i = 0; i < spectrum.MassSpectrum.Size; i++)
+                    {
+                        experimentalPeaks.Add(new MzPeak((thePrecursorMass - spectrum.MassSpectrum[i].Mz + Constants.protonMass * 3), (spectrum.MassSpectrum[i].Intensity / 100)));
+                    }
+                }
+            }
+            IEnumerable<IMzPeak> sortedPeaksMZ = experimentalPeaks.OrderBy(x => x.Mz);
```
stefanks: shouldnt the sorting only happen in doing compIons?

stefanks: seems a bit inefficient the way it is done here, with creation of peak objects.. can you reorganize this so that if the compIons is turned off, the matching is as efficient as before?

zrolfs:
Haven't added unit tests yet. Working on them now. Trying to figure out how other parts of MetaMorpheus work to do so.

zrolfs:
Alrighty, it passed the unit tests and should be ready to go!

PR #490, opened 2017-07-24
changed putative fatty acid mods to more likely surfactant adducts

acesnik:


stefanks:
can't merge because it's not compiling

acesnik:
got it to compile

PR #534, opened 2017-08-14
Bug Fixes for downstream of NonSpecific Search and for Partitions

zrolfs:


EngineLayer/Proteomics/PeptideWithSetModifications.cs
```diff
@@ -124,7 +129,13 @@ public int NumMods
 
         public int MissedCleavages
         {
-            get { return modPep.MissedCleavages; }
```
stefanks: Why do we need a new constructor?

EngineLayer/Proteomics/CompactPeptideWithModifiedMass.cs
```diff
@@ -27,7 +27,9 @@ public CompactPeptideWithModifiedMass(CompactPeptideBase cp, double Monoisotopic
 
         public void AssignCorrectMass()
```
stefanks: This method is so unintuitive. Maybe rename it? I have no idea what it does

EngineLayer/Proteomics/PeptideWithSetModifications.cs
```diff
@@ -190,7 +201,7 @@ public PeptideWithSetModifications Localize(int j, double massToLocalize)
             }
 
             vvv.Add(j + 2, new ModificationWithMass(null, null, null, TerminusLocalization.Any, massToLocalize + massOfExistingMod, null, new List<double> { 0 }, new List<double> { massToLocalize + massOfExistingMod }, null));
-            var hm = new PeptideWithSetModifications(modPep, vvv, numFixedMods);
+            var hm = new PeptideWithSetModifications(vvv, numFixedMods, this, OneBasedStartResidueInProtein, OneBasedEndResidueInProtein);
```
stefanks: Why the change?

EngineLayer/Psm.cs
```diff
@@ -267,15 +267,25 @@ public void AddOrReplace(CompactPeptideBase compactPeptide, double score, int no
...
-
-                var candidatePli = new ProteinLinkedInfo(matching[ok]);
+                compactPeptides[cpKey] = new Tuple<int, HashSet<PeptideWithSetModifications>>(compactPeptides[cpKey].Item1, matching[cpKey]);
+                var candidatePli = new ProteinLinkedInfo(matching[cpKey]);
+                if (MostProbableProteinInfo == null || FirstIsPreferable(candidatePli, MostProbableProteinInfo))
+                    MostProbableProteinInfo = candidatePli;
+            }
+            List<CompactPeptideBase> duplicateKeysToRemove = new List<CompactPeptideBase>();
+            foreach (KeyValuePair<CompactPeptideBase, Tuple<int, HashSet<PeptideWithSetModifications>>> kvp in compactPeptides)
+                if (kvp.Value.Item2 == null)
```
stefanks: This should have never been null in the first place. If this condition is true it means there is a bug somewhere before this point.

EngineLayer/Proteomics/PeptideWithSetModifications.cs
zrolfs: For NonSpecific search modes, you had previously requested a removal of modPep from newly generated PWSM. When modPep is null, modPep.Protein throws a null pointer exception. The new constructor generates a modPep.Protein by taking this.Protein instead.

EngineLayer/Proteomics/PeptideWithSetModifications.cs
zrolfs: When modPep==null, there is a null pointer exception.

EngineLayer/Psm.cs
zrolfs: It's an equal method issue, where there are two identical keys in the dictionary (which should never happen, but occurs when we modify the monoisotopic masses.

EngineLayer/Psm.cs
```diff
@@ -287,6 +286,14 @@ public void MatchToProteinLinkedPeptides(Dictionary<CompactPeptideBase, HashSet<
             Notch = Resolve(compactPeptides.Select(b => b.Value.Item1)).Item2;
         }
 
+        public bool CompactPeptidesContainsKey(CompactPeptideBase key)
+        {
+            if (compactPeptides.Keys.ToList().Contains(key))
```
stefanks: Inefficient. Just use  compactPeptides.ContainsKey

EngineLayer/Proteomics/PeptideWithSetModifications.cs
```diff
@@ -51,6 +51,11 @@ internal PeptideWithSetModifications(PeptideWithSetModifications modsFromThisOne
             this.allModsOneIsNterminus = modsFromThisOne.allModsOneIsNterminus.Where(b => b.Key >= proteinOneBasedStart - modsFromThisOne.OneBasedStartResidueInProtein && b.Key <= proteinOneBasedEnd - modsFromThisOne.OneBasedEndResidueInProtein).ToDictionary(b => b.Key, b => b.Value);
         }
 
+        internal PeptideWithSetModifications(Dictionary<int, ModificationWithMass> allModsOneIsNterminus, int numFixedMods, PeptideWithSetModifications modsFromThisOne, int proteinOneBasedStart, int proteinOneBasedEnd):base(modsFromThisOne.Protein, proteinOneBasedStart, proteinOneBasedEnd)
```
stefanks: Just pass the Protein object here instead of PeptideWithSetModifications

EngineLayer/Proteomics/PeptideWithSetModifications.cs
```diff
@@ -124,7 +129,13 @@ public int NumMods
 
         public int MissedCleavages
```
stefanks: change to int?

EngineLayer/Proteomics/PeptideWithSetModifications.cs
```diff
@@ -124,7 +129,13 @@ public int NumMods
...
         {
-            get { return modPep.MissedCleavages; }
+            get
+            {
+                if (modPep == null)
+                    return 0;
```
stefanks: return null

stefanks:
Actually, let's just remove the  private readonly PeptideWithPossibleModifications modPep from PWSM! And only populate the int? MissedCleavages property when it is passed in the constructor. This will conserve memory.

EngineLayer/Proteomics/PeptideWithSetModifications.cs
zrolfs: it is int?

stefanks: haha
Literally, "int?"

stefanks: nullable

EngineLayer/Proteomics/PeptideWithSetModifications.cs
zrolfs: can't. int is nonnullable value type

acesnik: "int?" is nullable :) https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/nullable-types/

zrolfs: aha! so smart.

stefanks: Sorry if I wasn't clear!

EngineLayer/Psm.cs
```diff
@@ -290,6 +289,14 @@ public void MatchToProteinLinkedPeptides(Dictionary<CompactPeptideBase, HashSet<
             Notch = Resolve(compactPeptides.Select(b => b.Value.Item1)).Item2;
         }
 
+        public bool CompactPeptidesContainsKey(CompactPeptideBase key)
+        {
+            if (compactPeptides.ContainsKey(key))
```
stefanks: just do return compactPeptides.ContainsKey(key)

EngineLayer/Proteomics/PeptideWithSetModifications.cs
```diff
@@ -34,23 +34,30 @@ public class PeptideWithSetModifications : Peptide
...
         {
-            this.modPep = everythingElseFromThisOne.modPep;
+            this.missedCleavages = everythingElseFromThisOne.missedCleavages;
             this.allModsOneIsNterminus = modsFromThisOne.allModsOneIsNterminus;
             this.numFixedMods = modsFromThisOne.numFixedMods;
         }
 
         internal PeptideWithSetModifications(PeptideWithSetModifications modsFromThisOne, int proteinOneBasedStart, int proteinOneBasedEnd) : base(modsFromThisOne.Protein, proteinOneBasedStart, proteinOneBasedEnd)
         {
+            this.missedCleavages = modsFromThisOne.missedCleavages;
```
stefanks: this is incorrect, just leave as null

EngineLayer/Psm.cs
```diff
@@ -315,7 +319,7 @@ public override string ToString()
                 sb.Append('\t' + Resolve(compactPeptides.SelectMany(b => b.Value.Item2).Select(b => b.BaseSequence)).Item1);
                 sb.Append('\t' + Resolve(compactPeptides.SelectMany(b => b.Value.Item2).Select(b => b.Sequence)).Item1);
                 sb.Append('\t' + Resolve(compactPeptides.SelectMany(b => b.Value.Item2).Select(b => b.NumVariableMods)).Item1);
-                sb.Append('\t' + Resolve(compactPeptides.SelectMany(b => b.Value.Item2).Select(b => b.MissedCleavages)).Item1);
+                sb.Append('\t' + Resolve(compactPeptides.SelectMany(b => b.Value.Item2).Select(b => b.missedCleavages ?? -1)).Item1);
```
stefanks: oh, no need for "?? -1" here!!

PR #543, opened 2017-08-15
removed duplicate protein output in psms file; bar delimiters instead of " or "

rmillikin:


EngineLayer/Psm.cs
```diff
@@ -401,14 +401,14 @@ internal void Add(Psm psmParent)
         private Tuple<string, double?> Resolve(IEnumerable<double> enumerable)
         {
             double doubleTol = 1e-6;
-            var list = enumerable.ToList();
+            var list = enumerable.Distinct().ToList();
```
stefanks: I don't think this should be Distinct(). 
The whole idea is to either collapse all into one, or display all the options! Say you have 3 ambiguous peptides with proteins 1,1,2 
In this case just saying 1,2 loses the assignment information

stefanks:
Maybe spaces around bars, for readability? 
Like this: " | "

EngineLayer/Psm.cs
rmillikin: it does display all the options, though. It displays all proteins that those peptides could have come from. It just removes duplicates. It gets especially confusing when a PSM is ambiguous and one option is a shared peptide. then you have peptides 1,2 with proteins A,B,C - there is no indication of which peptide comes from what protein anyway

rmillikin: and I think the delimiter should be one character; Excel does allow more than one character when splitting a column into multiple columns using a delimiter

rmillikin: does not allow*

stefanks: It's in order!  
There is never a case where there are two peptides (sequences?) corresponding to three proteins in the output right now.

stefanks:
What's the best delimiter character, anyway? 
Excel provides these default choices: tab, semicolon, comma, space. We can't do tab or space, but what about the others? Why pipe?

EngineLayer/Psm.cs
rmillikin: there certainly should be if there isn't... at least it's an edge case that is possible. a peptide sequence can be shared between multiple proteins, and parsimony won't always be able to select one. if an ambiguous PSM includes one of those, there could be 2+ peptides with 3+ proteins

stefanks: Of course, but then in the output we would repeat the sequence again, so that 3 sequences are written. And 3 proteins.

Sure, one of the sequences is repeating, but this way there is no confusion which of the sequences corresponds to 2 proteins and which to 1.

rmillikin: IMO it is more confusing to have the same protein repeated several times per peptide, and it makes it look like we don't know how to parse a database. it's a matter of readability/usability. maximum information vs least confusion.

stefanks:
I would be okay changing the delimiter, if there is concrete evidence it won't break things:
That modification types, ids, protein full names and genes do not have that character.

EngineLayer/Psm.cs
stefanks: I respectfully disagree.

zrolfs: If it's a democracy I'm voting with rmilikin.

lschaffer2: How would you know which peptide goes with which protein if you don't repeat it in the same order?

stefanks: Same protein repeated several times per peptide is an extremely rare occurrence. This means that somehow there is a repeating sequence, and that is interesting in itself. I wouldn't want to hide it!

rmillikin: that's Stefan's argument and I think it's a valid one; it's a difference of opinion on which is less confusing

rmillikin: maybe we could only do that if the PSM is ambiguous?

stefanks: That's already the only time it is done. In case of ambiguity.

lschaffer2: I just think for the stop and start or's it wouldn't be easy to convey the info simply without listing the protein each time

rmillikin: there are cases when the PSM repeats the same base sequence and the same protein. like so:


MFGMIK or MFGMIK or MGFMIK or MGFMIK DECOY_P04803  DECOY_P04803 Q06834  Q06834

zrolfs: MFGMIK   [DECOY_P04803-Q06834] ?

stefanks: Something is different between the first and the second match. Look in other columns.

lschaffer2: Probably the start and stop, which is why I think we need to repeat the names, to make those make sense

stefanks:
Chars in gene name, protein full name, modification ids:

n 172186
r 157645
i 141431
y 129701
o 128750
a 125323
  120232
e 113133
m 105400
p 87179
t 85952
s 79293
1 49378
l 49343
c 47477
A 34180
- 33280
d 30716
2 30630
P 27938
C 25725
h 23626
u 22438
g 22214
R 21286
3 20198
b 20148
S 18588
f 17819
T 17477
N 16538
4 15007
M 13899
F 13714
D 13492
I 12961
L 12744
5 12424
H 11778
B 11695
0 11506
G 11120
6 10465
O 10214
K 10027
E 9989
k 9144
7 8773
x 8380
8 7943
9 7153
U 5143
Z 4014
X 3676
v 3676
( 3509
) 3509
, 2549
/ 2531
V 2416
Y 2099
z 2047
W 1712
Q 1594
q 1203
w 1010
J 891
> 813
. 728
j 460
: 368
' 325
[ 259
] 259
+ 115
_ 67
* 2

Based on these it seems that the pipe | is a good choice.

Issue #559, opened 2017-08-22
"Peptides Sharing Same Peaks" Column exceeds excel character limit


zrolfs:
Add check for this as is done for other columns

stefanks:
How is this possible? Isn't it just an int?

zrolfs:
It seems like it should be just an int, but there are " or " delimiters making there way in there.

PR #560, opened 2017-08-22
Fix issue #559

zrolfs:


EngineLayer/Psm.cs
```diff
@@ -533,6 +533,14 @@ internal void Add(Psm psmParent)
             }
         }
 
+        private string Resolve(string s)
```
stefanks: Could you rename this? Something like TrimStringForExcel. Also, codacy! Thanks

PR #577, opened 2017-08-24
call retention time calibration function in FlashLFQ

rmillikin:


stefanks:
Would you please add the parameter to the task5 in the examples folder? And to any existing unit test? Thanks!

rmillikin:
didn't even know we had an examples file, will do!

stefanks:
These serve as examples, but also are run on appveyor and Jenkins, separately from unit tests

rmillikin:
rewrote the quant unit test to use the sliced raw file 8)

PR #613, opened 2017-09-01
Implemented complimentary tolerances for classic search

zrolfs:
Changes to unit tests are due to changes in intensity calculation.

### DIFF
--- a/EngineLayer/MetaMorpheusEngine.cs
+++ b/EngineLayer/MetaMorpheusEngine.cs
@@ -82,10 +82,11 @@ namespace EngineLayer
                 if (productMassTolerance.Within(currentExperimentalMz, currentTheoreticalMz))
                 {
                     matchedIonMassesList.Add(currentTheoreticalMass);
-                    productMassErrorDa.Add(currentExperimentalMz - currentTheoreticalMz);
-                    productMassErrorPpm.Add((currentExperimentalMz - currentTheoreticalMz) * 1000000 / currentTheoreticalMz);
+                    double currentExperimentalMass = currentExperimentalMz - Constants.protonMass;
+                    productMassErrorDa.Add(currentExperimentalMass - currentTheoreticalMass);
+                    productMassErrorPpm.Add((currentExperimentalMass - currentTheoreticalMass) * 1000000 / currentTheoreticalMass);
 
-                    currentTheoreticalIndex++;
+                   currentTheoreticalIndex++;
                     if (currentTheoreticalIndex == TotalProductsHere)
                         break;
                     currentTheoreticalMass = sortedTheoreticalProductMassesForThisPeptide[currentTheoreticalIndex];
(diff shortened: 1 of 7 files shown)


zrolfs:
Besides Codacy not understanding what the named tuples are, I think it's good to go.

stefanks:
What's the motivation for removing ReplaceObservedWithComplementaryPeaks, and instead duplicating the match ions code for the addComp case?

zrolfs:
Tolerances must be recalculated for complementary ions. To do this for observed ions will increase search time for classic searches, so there is some duplication in code to prevent that.

EngineLayer/MetaMorpheusEngine.cs
```diff
@@ -172,19 +238,21 @@ public static double CalculateClassicScore(IMsDataScan<IMzSpectrum<IMzPeak>> thi
...
-                // Else if for sure passed a theoretical
+                // Else if for sure did not reach the next theoretical yet
                 else if (currentExperimentalMz > currentTheoreticalMz)
                 {
                     // Move on to next index and never come back!
                     currentTheoreticalIndex++;
                     if (currentTheoreticalIndex == TotalProductsHere)
                         break;
-                    currentTheoreticalMz = sortedTheoreticalProductMassesForThisPeptide[currentTheoreticalIndex] + Constants.protonMass;
+                    currentTheoreticalMass = sortedTheoreticalProductMassesForThisPeptide[currentTheoreticalIndex];
```
stefanks: These masses are removed on purpose! For efficiency, no need to bring them back

EngineLayer/MetaMorpheusEngine.cs
```diff
@@ -200,18 +268,89 @@ public static double CalculateClassicScore(IMsDataScan<IMzSpectrum<IMzPeak>> thi
                         testTheoreticalIndex++;
                         if (testTheoreticalIndex == TotalProductsHere)
                             break;
-                        testTheoreticalMz = sortedTheoreticalProductMassesForThisPeptide[testTheoreticalIndex] + Constants.protonMass;
+                        testTheoreticalMass = sortedTheoreticalProductMassesForThisPeptide[testTheoreticalIndex];
```
stefanks: Same with this one

EngineLayer/MetaMorpheusEngine.cs
zrolfs: Didn't realize that change had been made; I was using the old code. Thanks for letting me know.

EngineLayer/MetaMorpheusEngine.cs
```diff
@@ -72,25 +67,26 @@ public static void MatchIons(IMsDataScan<IMzSpectrum<IMzPeak>> thisScan, Toleran
...
             for (int experimentalIndex = 0; experimentalIndex < numExperimentalPeaks; experimentalIndex++)
             {
                 double currentExperimentalMz = experimental_mzs[experimentalIndex];
                 // If found match
                 if (productMassTolerance.Within(currentExperimentalMz, currentTheoreticalMz))
                 {
-                    MatchingProductsHere++;
-                    MatchingIntensityHere += experimental_intensities[experimentalIndex];
-
                     matchedIonMassesList.Add(currentTheoreticalMass);
```
stefanks: I believe this line doesn't work now, since the currentTheoreticalMass is not being updated anymore. Please either restore the code that was there in the MatchIons (with the masses), or fix this! My previous comment was about the more highly optimized CalculateClassicScore which should not have this extra mass variables.

zrolfs: Great catch. I got overzealous with optimizing there.

Issue #656, opened 2017-09-14
Inconsistent handling of identical peptides between indexed and classic searches


zrolfs:
When ambiguity is turned off, a single theoretical is assigned to a spectra.
In the case where a peptide is present in both target and decoy search spaces (something that shouldn't occur in the first place), there is inconsistency in that:
For Classic, the peptide assigned is a always a decoy
For indexed searches, the peptide assigned is always a target.

To satisfy a TDA, there should be a random coin toss to determine which ID is assigned in both cases.

zrolfs:
As an update, it doesn't appear that turning off ambiguity does anything to output of classic search, so that should be fixed, too.

stefanks:
Is this fixed now?

PR #657, opened 2017-09-15
Function to violate FDR not quite as badly when not reporting ambiguity.

zrolfs:
Issue #656


zrolfs:
New merge strategy makes it more difficult to discern which changes have been made IMO.

stefanks:
I'll look at this pull request closely after lunch. 

The new merge strategy exists so that the commit history of the master branch is easy to parse and follow. Now it does does not include a huge number of commits that were a part of debugging/development efforts, and only includes clean commits. 

If you want to avoid having a huge old commit history in your pull requests, do not keep developing on a branch that has been previously merged into the master.

stefanks:
Ah, actually it is a small pull request! I'm uneasy about it though. I would rather we overestimate the FDR than underestimate it!

Anyone else have a strong opinion about this?

zrolfs:
It is a dramatic overestimation of FDR with the previous code and a marginally slight underestimation with these changes. Users who select ambiguity (the default) will continue to experience the overestimation, however they will be aware of them.

zrolfs:
I disagree with codacy, unless you think it's valid.

stefanks:
Good point!

PR #677, opened 2017-09-22
Fix CompactPeptideBase Crash

zrolfs:
I know we talked about making them all separate, but I like this way better, since it only requires a maximum of 4 comparisons for any situation. It additionally allows for more flexibility for "weird" comparisons. In the case of N1.null, N2.true, C1.true, C2.true, it will still compare the C ions. It may be important to limit these comparisons, as I had noticed that my C-searches were about 5% faster than my N-searches. I hypothesize that was because C ions were checked before N ions were in this method, and thus the number of comparisons does make a difference on search time.


stefanks:
I'll take it, but I doubt that that's the source of degraded performance

zrolfs:
Stefan accepttttttt

PR #726, opened 2017-10-16
Ms2 decon search

stefanks:


rmillikin:
I totally agree! The code that is there is pretty preliminary, I just wanted to get it out in the open in a pull request. I think it would slightly slow down bottom-up searches, but it would be more accurate (fewer false-positives as Zach mentioned, and of course more correct mass fragments). We should reevaluate the "top 200" ion strategy as well, since deconvoluting with only the 200 most intense peaks in a scan may give some deconvolution errors. I'm not super familiar with the localization and scoring code either; would this need to be rewritten to use the new MS2ScanWithSpecificMass, assuming it has a list of fragment masses attached to it? i.e., do these methods re-read the scan and thus have to be rewritten to use the deconvoluted masses instead?

stefanks:
All these methods (scoring and localization) could easily switch over to using a sorted list of masses instead of a list of m/z

stefanks:
And I am open to reevaluating the top 200 strategy too. The idea is to remove noise peaks. Maybe there is a better way of doing this.

zrolfs:
I still think we should remove all peaks in a mass bin. If we're only looking at b and y ions, we shouldn't expect more than 2 correct peaks every 57 Da (at a minimum).

stefanks:
What about co-isolation?

In any case, this is a different issue! Let's tackle the less controversial one first.

PR #746, opened 2017-10-26
Fragment ion max mass crash fix

zrolfs:


zrolfs:
Commit 1: Max fragment mass was set to 30000, making the highest possible bin 29999.999. Any peak that was under 30000 but rounded up to 30000 caused an index out of bounds.

Commit 2: Max fragment max was enabled only for modern search, but not for nonspecific search which uses same indexing engine. A cleaner fix may be to never disable it and implement functionality into classic search as well.

PR #750, opened 2017-10-30
small nonspecific search bugfix

zrolfs:


stefanks:
This change seems to fail the TestSemiSpecificEnzymeEngineSingleN unit test!

PR #810, opened 2017-12-05
MultiScore

stefanks:


stefanks:
Need to put in e-value as a feature

PR #849, opened 2018-01-16
Allow users to send email report upon crash

dippman:
During runtime crashes, user has the option to send crash info to mm_support@chem.wisc.edu from their local email client.


GUI/MainWindow.xaml.cs
```diff
@@ -427,8 +428,26 @@ private void EverythingRunnerExceptionHandler(Task obj)
...
-                MessageBox.Show(message);
+                var messageBoxResult = System.Windows.MessageBox.Show(message + "\n\nWould you like to report this crash?", "Runtime Error", MessageBoxButton.YesNo);
                 outRichTextBox.AppendText(message + Environment.NewLine);
+                Exception exception = e;
+
+
+                if(messageBoxResult == MessageBoxResult.Yes)
+                {
+                    string body = exception.Message + "%0D%0A" + exception.Data +
+                       "%0D%0A" + exception.StackTrace +
```
stefanks: What's %0D%0A

TaskLayer/SearchTask/SearchTask.cs
```diff
@@ -1094,6 +1092,7 @@ protected override MyTaskResults RunSpecific(string OutputFolder, List<DbForTask
 
             var psmsGroupedByFile = allPsms.GroupBy(p => p.FullFilePath);
 
+            Psm.ModstoWritePruned = SearchParameters.ModsToWriteSelection;
```
stefanks: No, revert this change please!

TaskLayer/TaskLayer.csproj
```diff
@@ -18,6 +18,10 @@
     <ProjectReference Include="..\EngineLayer\EngineLayer.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="System.Windows.Forms" />
+  </ItemGroup>
+
```
stefanks: Is this needed in the task layer?

GUI/MainWindow.xaml.cs
dippman: These are the characters for 'newline' needed to format with the line endings that emails use. For some reason '\n' , '\r\n', and 'Environment.newline'  don't work for email line endings.

TaskLayer/TaskLayer.csproj
dippman: woops, no

PR #897, opened 2018-01-29
Fix xml exclusion list

zrolfs:


EngineLayer/GlobalVariables.cs
```diff
@@ -67,22 +88,6 @@ static GlobalVariables()
...
-        public static IEnumerable<Modification> UniprotDeseralized { get; }
-        public static UsefulProteomicsDatabases.Generated.obo PsiModDeserialized { get; }
-        public static IReadOnlyDictionary<string, Protease> ProteaseDictionary { get; }
-        public static IEnumerable<Modification> AllModsKnown { get { return allModsKnown.AsEnumerable(); } }
-
-        #endregion Public Properties
-
         #region Public Methods
 
         public static void AddMods(IEnumerable<Modification> enumerable)
```
stefanks: Add the mod types known here as well!

PR #898, opened 2018-01-29
fix

zrolfs:


zrolfs:
hold up, this prevents crashes in gui, but loading old tomls creates a crash since the variable is now stored as [] instead of "" when blank.

PR #934, opened 2018-02-07
.fa extension compatibility,GUI Tolerance Type fix, and TSV Excel Character limit checking

dippman:


EngineLayer/GlobalVariables.cs
```diff
@@ -104,6 +104,14 @@ public static void AddMods(IEnumerable<Modification> enumerable)
             }
         }
 
+        public static string CheckLengthOfPSM(string psmString)
```
rmillikin: please change name to CheckLengthOfOutput

PR #966, opened 2018-02-22
Update README.md

rmillikin:


README.md
```diff
@@ -7,77 +7,76 @@
...
 
 ## Typical Usage (Windows GUI)
-1. Open MetaMorpheusGUI.exe, and drag and drop the raw spectra files and the compressed Uniprot XML database on the GUI.
-2. Select "New Calibrate Task" tab and enter the appropriate search parameters, using slightly liberal mass tolerances (10-20 ppm). Then "Add the Calibration Task".
-3. Select "New GPTMD Task" tab. Use tighter parent mass tolerance than you would in your typical search. Specify the G-PTM-D modifications that you think may be present in your sample. Many typical modifications are pre-selected. Then "Add the GPTMD Task".
-4. Select "New Search Task" tab. Specify the search paramters. High-resolution data that has been calibrated can frequently use a parent mass tolerance of 5ppm or less. Specify the Post-Search-Parameters (e.g. protein parsimony, quantification). Then "Add the Search Task".
+1. Open MetaMorpheusGUI.exe, and drag and drop your raw spectra files and protein database into the GUI.
+2. Select "New Calibrate Task" tab and enter the appropriate search parameters, using slightly liberal mass tolerances (20 ppm recommended). Then "Add the Calibration Task".
```
zrolfs: Remove recommended tolerance. If we recommend something, it should be the default.

README.md
```diff
@@ -7,77 +7,76 @@
...
+3. Select "New GPTMD Task" tab. Use tighter parent mass tolerance than you would in your typical search (e.g. 2-5 ppm). Specify the G-PTM-D modifications that you think may be present in your sample. Many typical modifications are pre-selected. Then "Add the GPTMD Task".
```
zrolfs: Why is this tighter than a typical search? Default is not. This is also outdated, as fileSpecific tolerances from calibration will override this.

README.md
rmillikin: 20ppm is the default and is thus recommended here, the "slightly liberal mass tolerance" is just an elaboration/justification for the 20ppm number

README.md
rmillikin: I agree

PR #977, opened 2018-02-23
alphabetize protein group names

rmmiller22:


rmillikin:
- alphabetize proteins within a protein group by accession

PR #983, opened 2018-02-23
Empty Database Warning

dippman:


GUI/MainWindow.xaml.cs
```diff
@@ -297,7 +297,13 @@ private void AddXML_Click(object sender, RoutedEventArgs e)
             if (openPicker.ShowDialog() == true)
                 foreach (var filepath in openPicker.FileNames)
                 {
-                    AddAFile(filepath);
+                    FileInfo info = new FileInfo(filepath);
```
rmillikin: I'd prefer checks in the loading protein database section of the code, not the GUI

please make sure none of these cause crashes:
[bad fastas.zip](https://github.com/smith-chem-wisc/MetaMorpheus/files/1757075/bad.fastas.zip)

rmillikin: also please make unit test(s) for those fastas

rmillikin: and also please try to make some messed up XML databases (incorrectly placed tags, empty database, correct tags with no sequence, etc...) and make tests for those and make sure there are no crashes

PR #989, opened 2018-02-24
fix suggested calibration ppm tolerances

rmillikin:


TaskLayer/CalibrationTask/CalibrationTask.cs
```diff
@@ -296,8 +296,8 @@ protected override MyTaskResults RunSpecific(string OutputFolder, List<DbForTask
                     var tomlFileName = Path.Combine(OutputFolder, Path.GetFileNameWithoutExtension(currentDataFile) + "-calib.toml");
                     FileSpecificTolerances f = new FileSpecificTolerances
                     {
-                        PrecursorMassTolerance = precTol,
-                        ProductMassTolerance = prodTol
+                        PrecursorMassTolerance = bestPrecursorTol,
+                        ProductMassTolerance = bestProductTol
```
rmillikin: 299-300 is the crux of the problem... it was writing the ppm tolerances for the round of calibration that came after the best one (i.e., the tolerances gave worse results than the best results)

PR #1005, opened 2018-02-27
Gptmd fix

zrolfs:


TaskLayer/GPTMDTask/GPTMDTask.cs
```diff
@@ -152,7 +152,7 @@ protected override MyTaskResults RunSpecific(string OutputFolder, List<DbForTask
 
             allPsms = allPsms.Where(b => b != null).OrderByDescending(b => b.Score).ThenBy(b => b.PeptideMonisotopicMass.HasValue ? Math.Abs(b.ScanPrecursorMass - b.PeptideMonisotopicMass.Value) : double.MaxValue).GroupBy(b => new Tuple<string, int, double?>(b.FullFilePath, b.ScanNumber, b.PeptideMonisotopicMass)).Select(b => b.First()).ToList();
 
-            new FdrAnalysisEngine(allPsms, searchMode.NumNotches, false, new List<string> { taskId }).Run();
+            new FdrAnalysisEngine(allPsms, tempSearchMode.NumNotches, false, new List<string> { taskId }).Run();
```
rmillikin: this is going to get fuzzy if we eventually allow file-specific notches but it's ok for now

PR #1009, opened 2018-02-28
Append TOML to Error Reporting

dippman:
Add task toml of crash and append it to email text


GUI/MainWindow.xaml.cs
```diff
@@ -500,17 +500,24 @@ private void EverythingRunnerExceptionHandler(Task obj)
                 var messageBoxResult = System.Windows.MessageBox.Show(message + "\n\nWould you like to report this crash?", "Runtime Error", MessageBoxButton.YesNo);
                 outRichTextBox.AppendText(message + Environment.NewLine);
                 Exception exception = e;
-
+                //Find Output Folder
+                string outputFolder = e.Data["folder"].ToString();
+                var toml = Directory.GetFiles(outputFolder, "*.toml");
```
rmillikin: check to make sure the directory exists here before continuing, by using Directory.Exists

GUI/MainWindow.xaml.cs
```diff
@@ -500,17 +500,24 @@ private void EverythingRunnerExceptionHandler(Task obj)
...
+                //will only be 1 toml per task
+                string tomlText = File.ReadAllText(toml[0]);
```
rmillikin: make sure that toml.Any() before doing this

PR #1044, opened 2018-03-24
classic search changes, gui changes, update mzlib

rmillikin:


zrolfs:
Why change Psm to PeptideSpectralMatch? I felt like Psm was easier on the eyes...

rmillikin:
PSM isn't descriptive enough for me personally. I think it should be spelled out. it's more accessible for new coders/users

PR #1089, opened 2018-04-11
XL crosslinker with different amino acid 

lonelu:
Also include unit test


lonelu:
OK to be merged. This also solve the #1117.

PR #1145, opened 2018-04-27
Make PSM ToString code more maintainable and add SequenceVariations to it

acesnik:


rmillikin:
I don't really understand how an array of strings is more maintainable than the tab-separated header; I get that it's difficult to line the header up with the value but I don't think this makes it any easier?

rmillikin:
I think the ideal would be a dictionary where the keys are static and the values are object-specific but I'm not sure that's possible

acesnik:
The arrays got rid of the "\t" anywhere and everywhere. I like this idea about the dictionaries.

PR #1153, opened 2018-05-04
 mzML consistent naming

zrolfs:


GUI/MainWindow.xaml.cs
```diff
@@ -410,16 +410,16 @@ private void AddAFile(string draggedFilePath)
                             "please uninstall and reinstall MetaMorpheus", null));
                     }
 
-                    goto case ".mzml";
+                    goto case ".mzML";
 
-                case ".mzml":
+                case ".mzML":
```
rmillikin: this is a problem because we are allowing file extensions to be named with variable casing (i.e., a file is valid input if it's named .mzml, .mzML, .MZML, etc etc. 

line 386 converts all of these variant cases to ".mzml" (with .ToLowerInvariant()) for easy parsing. please change this back to ".mzml"

GUI/MainWindow.xaml.cs
```diff
@@ -410,16 +410,16 @@ private void AddAFile(string draggedFilePath)
...
                     RawDataForDataGrid zz = new RawDataForDataGrid(draggedFilePath);
                     if (!SpectraFileExists(spectraFilesObservableCollection, zz)) { spectraFilesObservableCollection.Add(zz); }
                     UpdateFileSpecificParamsDisplayJustAdded(Path.ChangeExtension(draggedFilePath, ".toml"));
                     UpdateOutputFolderTextbox();
                     break;
 
-                case ".mzml.gz":  // not implemented yet
+                case ".mzML.gz":  // not implemented yet
```
rmillikin: same as above

PR #1154, opened 2018-05-04
Add "possible PSMs" to results output

zrolfs:
MM's co-isolation feature allow us to identify more PSMs than the number of MS2 spectra acquired. By including the number of possible spectra searched against, the user has a metric for roughly determining how many peptides are being ionized. Additionally, it will hopefully diminish confusion if the number of identified PSMs is greater than the number of MS2 spectra. It feels a little dishonest to say we identified 15000 PSMs out of 20000 spectra when a few thousand of those are from chimeras.

I'm not super attached to this change, as it might just introduce more confusion, so you can veto.


rmillikin:
I think it's a useful metric but I agree it could introduce confusion especially since many? most? people assume 1 MS2 scan = 1 possible PSM. Perhaps calling it something like "Number of precursor species fragmented"? Since this is really what arrayOfMs2ScansSortedByMass attempts to represent

PR #1160, opened 2018-05-07
Include only sequence variants in range of peptide in PSMs results

acesnik:


acesnik:
I ran a manual test on this, and it's getting the expected behavior.

PR #1188, opened 2018-05-25
Mzlib update

zrolfs:


EngineLayer/Calibration/CalibrationEngine.cs
```diff
@@ -132,24 +132,24 @@ private void CalibrateSpectra(IPredictorModel<double> ms1predictor, IPredictorMo
                   {
                       var scan = myMsDataFile.GetOneBasedScan(i);
 
-                      if (scan is IMsDataScanWithPrecursor<IMzSpectrum<IMzPeak>> ms2Scan)
+                      if (scan is MsDataScan ms2Scan)
```
rmillikin: need to check if ms order == 2

PR #1198, opened 2018-05-31
Add mgf reading capabilities

zrolfs:
Mgf is not compatible with FlashLFQ, because there are no MS1 scans.
Chimeric spectra can't be found, but can still be analyzed from the single precursor m/z.

I converted a mzml to mgf with MSConvert 3.0.11392 and searched both.
Mzml 1% FDR PSMs w/ decon: 15405 Time: 5:31
Mzml 1% FDR PSMs w/o decon: 11965 Time: 4:36
Mgf 1% FDR PSMs w/ decon: 10652 Time: 5:25
Mgf 1% FDR PSMs w/o decon: 10652 Time: 5:25

MSConvert assigns new scan numbers, so comparing results is difficult. I am not sure why the mgf performed poorer and took longer.


TaskLayer/MyFileManager.cs
```diff
@@ -95,10 +96,15 @@ internal MsDataFile LoadFile(string origDataFile, int? topNpeaks, double? minRat
 
             // By now know that need to load this file!!!
             lock (fileLoadingLock) // Lock because reading is sequential
+            {
                 if (Path.GetExtension(origDataFile).Equals(".mzML", StringComparison.OrdinalIgnoreCase))
                 {
                     myMsDataFiles[origDataFile] = Mzml.LoadAllStaticData(origDataFile, filter);
                 }
+                else if (Path.GetExtension(origDataFile).Equals(".mgf"))
```
rmillikin: `.Equals(".mgf", StringComparison.OrdinalIgnoreCase))`

GUI/MainWindow.xaml.cs
```diff
@@ -454,14 +454,21 @@ private void AddAFile(string draggedFilePath)
 
                     goto case ".mzml";
 
+                case ".mgf":
```
rmillikin: maybe something like:
".mgf files lack MS1 spectra, which are needed for quantification and searching for coisolated peptides. All other features of MetaMorpheus will function."

TaskLayer/SearchTask/SearchTask.cs
```diff
@@ -78,6 +78,12 @@ public static MassDiffAcceptor GetMassDiffAcceptor(Tolerance precursorMassTolera
 
         protected override MyTaskResults RunSpecific(string OutputFolder, List<DbForTask> dbFilenameList, List<string> currentRawFileList, string taskId, FileSpecificParameters[] fileSettingsList)
         {
+            //disable quantification if a .mgf is being used
+            if (SearchParameters.DoQuantification && currentRawFileList.Any(x => Path.GetExtension(x).Equals(".mgf")))
```
rmillikin: this is ok for now but at some point I'll correct FlashLFQ to bail on quantifying files without MS1 spectra. just making a mental note here.

you probably want to use OrdinalIgnoreCase when checking for ".mgf"

Test/MsDataFileTest.cs
```diff
@@ -0,0 +1,45 @@
...
+            };
+            List<(string, MetaMorpheusTask)> taskList = new List<(string, MetaMorpheusTask)>
+            {
+                ("task1", task1),
+            };
+            //run!
+
+            var engine = new EverythingRunnerEngine(taskList, new List<string> { mgfName }, new List<DbForTask> { new DbForTask(xmlName, false) }, Environment.CurrentDirectory);
+            engine.Run();
+            //Just don't crash! There should also be at least one psm at 1% FDR, but can't check for that.
```
rmillikin: we can read the psm file in and check but this is ok for now

TaskLayer/MyFileManager.cs
zrolfs: I've only ever seen .mgf lowercase, but I can add it.

PR #1203, opened 2018-06-01
Refactored CommonParameters

rmmiller22:
CommonParameters have been refactored so that default arguments are set in the constructor


EngineLayer/CommonParameters.cs
```diff
@@ -8,115 +8,146 @@ namespace EngineLayer
...
+        public bool UseDeltaScore { get; private set; }
+        public bool CalculateEValue { get; private set; }
 
         #endregion Public Properties
 
         #region Public Methods
 
         public CommonParameters Clone()
         {
-            return (CommonParameters)this.MemberwiseClone();
```
zrolfs: Doesn't the old code accomplish the same thing more succinctly?

EngineLayer/CommonParameters.cs
```diff
@@ -8,115 +8,146 @@ namespace EngineLayer
...
+                PrecursorMassTolerance: this.PrecursorMassTolerance,
+                DeconvolutionMassTolerance: this.DeconvolutionMassTolerance,
+                MaxThreadsToUsePerFile: this.MaxThreadsToUsePerFile,
+                DigestionParams: this.DigestionParams,
+                ListOfModsVariable: this.ListOfModsVariable,
+                ListOfModsFixed: this.ListOfModsFixed
+            );
+        }
+
+        public void SetTaskDescriptor(string TaskDescriptor)
```
zrolfs: This set method is unnecessary since the variable still has a public set.

zrolfs: You could just leave the others two as public sets and remove the other to set methods, too.

EngineLayer/CommonParameters.cs
rmillikin: MemberwiseClone is a "shallow copy", it just duplicates the reference to the object. "Deep clone" (copying the properties of an object to a new object) is missing from C#. A deep copy is preferable in most cases. this revised method gets closer to that but it's still a bit problematic because it's still copying references to the same DigestionParams, etc and not making new ones

Issue #1233, opened 2018-06-26
matched ion 'number and type' in output + 'Matched Ion Intensities' documentation


ftwkoopmans:
I'm composing transition lists directly from MetaMorpheus output with a simple R script. Most information I need is present in .psmtsv already, but I could use some pointers (and possibly extra output columns) with the last few pieces of data.

For each matched ion I need:

1. Fragment number and type (eg; "y7")
2. Fragment intensity
    - #1205 'Matched Ion Intensities' probably contains this data, but I cannot find any documentation on the ordering of these nested arrays (I assume they represent b and y, or y and b, series? But which fragment numbers are these? Maybe you can add a fragment type+number array aligned with the other 'Matched Ion' arrays? )
3. Fragment m/z (theoretical)
4. Fragment Charge

Optionally, I could get 3 and 4 by re-computing with some R package (but getting everything out of the search engine would be ideal).

zrolfs:
1. I think this is a great idea! I'll get on that and make a new column. It'll use the same nested arrays used in #2. I plan to include charge information here as well, so instead of y7, it'll be y7+z.
2. Matched Ion Intensities is a new feature that should achieve what you're interested in. The order of the nested arrays is the same as for the Fragment m/z (theoretical), which is b-y-c-zdot. Adding in #1 will hopefully alleviate the confusion.
3. Fragment m/z (theoretical) is present in "Matched Ion Masses", in the form of masses rather than m/zs. I'll change this to reflect theoretical m/zs, which makes more sense.
4. MetaMorpheus currently only searches for z=1 fragment ions. We're interested in searching for z=2 fragment ions for z=3+ precursors, but we haven't gotten around to it yet.

trishorts:
Zach. will you please edit the related entries in the wiki once your new code is added? https://github.com/smith-chem-wisc/MetaMorpheus/wiki/Glossary-of-Terminology

current headings that need improvement are: 

Matched Ion Counts

Matched Ion Mass Diff (Da) -- The absoulte mass difference between the identified peptide theoretical mass and the mass of the experimental reported as the BestMassInPrecursor.

Matched Ion Mass Diff (Ppm) -- The ppm mass difference between the identified peptide theoretical mass and the mass of the experimental reported as the BestMassInPrecursor.

Matched Ion Masses

rmillikin:
I think we should retool this output to contain maximal information in a machine-parseable format. 

Something like:
[...]

PR #1234, opened 2018-06-26
Add ion series info

zrolfs:


rmillikin:
for matched ion series output I get this:
[[b2+1, b9+1, b10+1, b11+1, b12+1];[y1+1, y2+1, y3+1, y4+1, y5+1, y6+1, y7+1, y8+1, y9+1, y10+1, y11+1, y12+1];]

I assume it is supposed to be b3, b10, b11, etc ?

zrolfs:
The b ions are working as intended from line 208 in MetaMorpheusEngine.cs

           //matchedIonSeries assumes 1-n, but bnob1 has no b1, so we need to ++ each series number
            if (productType == ProductType.BnoB1ions)
            {
                for (int i = 0; i < matchedIonSeries.Count; i++)
                {
                    matchedIonSeries[i]++;
                }
            }

rmillikin:
capture

I think you are trying to add one to the ion series number which is fine but in the output it is not format like you intend (I think?)

zrolfs:
It looks as intended, I'm not sure what you mean.
The +1 is from the charge. This is a standard notation.

zrolfs:
The charge addresses 4. from #1233

rmillikin:
Ah I see, I understand. looks good

PR #1235, opened 2018-06-26
new mods lists from frey

trishorts:


acesnik:
Can we make the capitalization of all of the mod files consistent (Pascal case)?

Issue #1228, opened 2018-06-23
How to set the number of used CPU in GUI?


01joy:
When run MetaMorpheus, all CPUs will be used by it. How to set the number of used CPU or processor in GUI?

I see a discussion in this issue: https://github.com/smith-chem-wisc/MetaMorpheus/issues/279, but I can't find the MaxDegreeOfParallelism  option in GUI.

The version of MetaMorpheus is 0.0.282.

Thanks.

rmillikin:
This is currently not possible, unfortunately - we can add this soon. It should be a fairly easy thing to implement; I would guess it will be done in ~1 week.

01joy:
Thank you @rmillikin for your quick response, and look forward to your next version with this feature.

Just another related quick question, I saw a parameter `MaxThreadsToUsePerFile = 31` in XLSearchTaskconfig.toml file, Is it the max number of CPU used? Can I customize it and run MetaMorpheus in command line, if yes, how to run in command line.

Thank you.

rmillikin:
The `MaxThreadsToUsePerFile` is indeed the setting you want to change. It's defined internally, but it is not settable in the GUI. I tried altering a .toml to set this field to 1, but it still used all of my threads, so this does not appear to be a viable workaround, unfortunately.

rmillikin:
Just a quick update, we're working on this in #1254 . Should be done in a day or two. I will release a new version of MetaMorpheus next week with this change.

rmillikin:
This is in the newest release (0.0.284)! Please give that a try. It's "Max Threads" in the GUI.

I've noticed that mzML loading is still maximally parallelized. The next update will include restricting the number of threads when loading mzML files. I'm going to close this and make a new issue about that.

PR #1254, opened 2018-07-02
Select how many processors to use

ianhirsch:
#1228


EngineLayer/MetaMorpheusEngine.cs
```diff
@@ -18,14 +18,17 @@ public abstract class MetaMorpheusEngine
             { DissociationType.ETD, 2*Constants.protonMass }
         };
 
+        protected readonly CommonParameters commonParameters;
```
acesnik: For future reference, try to make field variable names PascalCase

PR #1257, opened 2018-07-02
added some diagnostic ions and neutral losses to the common biological modifications text file

trishorts:


EngineLayer/Mods/CommonBiological.txt
```diff
@@ -1,15 +1,31 @@
...
+ID   Phosphotyrosine
+TG   Y
+PP   Anywhere.
+NL   H0 or H3 O4 P1
+MT   Biological
+CF   H1 O3 P1
+DI   216.042
+//
+ID   Phosphothreonine
 TG   S or T or Y
```
rmillikin: `TG   T` ?

EngineLayer/Mods/CommonBiological.txt
```diff
@@ -1,15 +1,31 @@
...
+MT   Biological
+CF   H1 O3 P1
+//
+ID   Phosphotyrosine
+TG   Y
+PP   Anywhere.
+NL   H0 or H3 O4 P1
+MT   Biological
+CF   H1 O3 P1
+DI   216.042
```
rmillikin: I don't think you want this here

EngineLayer/Mods/CommonBiological.txt
trishorts: Good  catch. Just T. I'll change tomorrow

EngineLayer/Mods/CommonBiological.txt
acesnik: Why not?

EngineLayer/Mods/CommonBiological.txt
```diff
@@ -1,15 +1,31 @@
...
+//
+ID   Phosphothreonine
+TG   T
 PP   Anywhere.
 NL   H0 or H3 O4 P1
 MT   Biological
 CF   H1 O3 P1
 //
-ID   Acetyl_Internal
+ID   Acetyl_Internal_Lysine
```
acesnik: Do we allow spaces in these names?

How about "Acetyllysine, Internal"?

PR #1258, opened 2018-07-02
Cancel button, mzLib proteomics digestion stuff, lots of refactoring

acesnik:


acesnik:
This is building and passing, now! I had to copy `proteases.tsv` back into MetaMorpheus. It turns out it's fairly easy to copy text files to NuGet packages (mzLib), but it's tough to copy it to the build folder! I think we would have to use the mzLib.targets file (one for net471 and a new one for netstandard2.0), but I tried and failed to get those to even print "Hello World" during/after an MSBuild event.

PR #1268, opened 2018-07-06
Implicit parameter defaults

acesnik:
Review after capitalization and cancel button PRs


acesnik:
The idea here is that setting parameter defaults with the implicit constructors reduces duplication in those classes. It's a little clearer what has defaults set, too.

PR #1280, opened 2018-07-10
Cancel

acesnik:
Added cancel button to main window and functionality to cancel Calibration, Search, and GPTMD tasks. Cancel functionality was added to XL tasks, but the effectiveness was not tested. Neo was ignored.

### DIFF
--- a/GUI/MainWindow.xaml.cs
+++ b/GUI/MainWindow.xaml.cs
@@ -569,6 +569,9 @@ namespace MetaMorpheusGUI
 
         private void RunAllTasks_Click(object sender, RoutedEventArgs e)
         {
+            GlobalVariables.StopLoops = false;
+            CancelButton.IsEnabled = true;
+
             // check for valid tasks/spectra files/protein databases
             if (!StaticTasksObservableCollection.Any())
             {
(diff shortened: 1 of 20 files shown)


GUI/MainWindow.xaml.cs
```diff
@@ -1255,6 +1258,17 @@ private bool SpectraFileExists(ObservableCollection<RawDataForDataGrid> rDOC, Ra
             return false;
         }
 
+        private void CancelButton_Click(object sender, RoutedEventArgs e)
+        {
+            string grammar = StaticTasksObservableCollection.Count <= 1 ? "this task" : "these tasks";
```
ianhirsch: nice addition

acesnik: Thanks!

GUI/MainWindow.xaml.cs
```diff
@@ -1255,6 +1258,17 @@ private bool SpectraFileExists(ObservableCollection<RawDataForDataGrid> rDOC, Ra
...
+            if (MessageBox.Show("Are you sure you want to cancel " + grammar + "?", "Cancel Tasks", MessageBoxButton.OKCancel) == MessageBoxResult.OK)
+            {
+                GlobalVariables.StopLoops = true;
+                CancelButton.IsEnabled = false;
+                notificationsTextBox.AppendText("Canceling...\n");
```
zrolfs: I like this, but it didn't go away after it finished when I canceled during protein loading. Maybe another message saying "Canceled!" or just remove the "Canceling..." once it's finished?

zrolfs:
If you're looking to spend even more time on this, you could change the message next to the task progress from "Done!" to "Canceled!" and make the bar red instead of green :)

acesnik:
@zrolfs and @ianhirsch Thanks for testing the feature! Would you like to commit to this branch to add/fix those features?

zrolfs:
Not at the moment; I'm working on my projects 🔥

PR #1281, opened 2018-07-11
Average target-decoy control (unfinished)

acesnik:
@trishorts Feel free to pull this branch and build on this progress


acesnik:
@zrolfs The one test this is failing is strangely complementary ions, and I couldn't figure out why back in June. That's your code, so maybe you can figure out what's going on there.

zrolfs:
From the talk (https://www.pathlms.com/asms/events/1111/video_presentations/109388), it seems like the advantage of aTDC is simply reproducibility in FDR, particularly when database size and/or # of spectra are low. It doesn't seem like it would necessarily improve nonspecific searches (#1370) any more than a shuffled database. Rather, as was asked in the Q&A, having to search N decoy databases increases the search time by a factor of (N+1)/2. "The situations in which aTDC is most important, are the situations in which is least expensive (small database, few spectra)."

acesnik:
I think @trishorts was saying it would remove the possibility of having decoys that have the same sequence as a target peptide.

acesnik:
Oh yes, I see you said "any more than a shuffled database." Yes, I think both this approach and using a shuffled database would have that benefit.

acesnik:
I think @trishorts had a take on why this still might be of interest to you, @zrolfs.

acesnik:
I'm also not sure that comment in the talk about increasing the search time by (N+1)/2 is entirely correct. That seems to assume you're generating N decoy databases and searching them all, and that's not exactly the case here.

Another benefit is that protein FDRs might be more solid, since you're evaluating peptides against a distribution of peptides modeled from the same protein.

zrolfs:
I understood it as you needed to generate N decoy databases. How did you interpret it?

I'm not sure I grasp the connection between decoy peptide origin and fidelity of protein FDR. Can you elaborate?

Issue #1217, opened 2018-06-12
Nonspecific tomls are being reset to defaults whenever opened.


zrolfs:

zrolfs:
They appear to be reset every time, regardless of if it's opened or not, so the lengths can't be changed from the default of 8-25...

rmillikin:
Just making a quick note that I tried to reproduce this bug and couldn't. The search task window saves settings, saves the toml, and MM reads the toml all ok as far as I can tell

zrolfs:
Classic Search -> non-specific protease -> modify the max length and save
Reopen, max length is default of 25

zrolfs:
I'm working on a fix right now

rmillikin:
any update on this? maybe Ian can finish it up?

rmillikin:
done in #1284

PR #1284, opened 2018-07-12
Non tomls

ianhirsch:
#1217


GUI/SearchTaskWindow.xaml.cs
```diff
@@ -600,7 +599,7 @@ private void NonSpecificUpdate(object sender, TextChangedEventArgs e)
                 try
                 {
                     System.Windows.Controls.TextBox textBox = (TextBox)sender;
-                    if (textBox.Name.Equals("txtMaxPeptideLength")) //if maxPeptideLength was modified
+                    if (textBox.Name.Equals("MaxPeptideLengthTextBox")) //if maxPeptideLength was modified
```
zrolfs: Change this back. You're trying to compare a string and a textbox, which won't work.

Issue #1274, opened 2018-07-09
Scan Number (different then in .mgf ?)


gvinterhalter:
Hi,

I have an .mgf that I converted to .mzML (so only MS2).
The resulting Scan Numbers in .psmtsv or .mzid  seem to be different then the original .mgf "SCANS".
Am I missing something? I need the number exactly as is in .mgf

Tnx in advance

trishorts:
will look into this shortly

trishorts:
which converter did you use?

gvinterhalter:
1. MSconvetGUI default options
2. I get an index out of bound of the array error

zrolfs:
Are you able to share your .mgf file with us? It will help us more quickly identify a fix.

gvinterhalter:
Here is the  [mgf and the search settings](https://drive.google.com/open?id=1kr99ROaK7AmXfbGDcQTZ4brszuzNV1Wf)

zrolfs:
The scans don't appear to be in order. The first 5 MS2 scan numbers are:
3585
3458
5041
9672
8147

Do you know why this is?
It is the cause of both problems 1 and 2.

MSConvert is handling this well, so it's something we can address on our end. I'll keep you posted and we should have a fix in 0.0.284.

If it's time sensitive, I can sort the mgf for you and send it back?
Thanks for letting us know!

gvinterhalter:
I see. 
So they should just be ordered by the scan number? 
I can do that, tnx.

mgf was exported from Prognosis (not by me)

gvinterhalter:
Sorting the mgf fixes 2 but I still think something is wrong with the output.
In the resulting files I'm getting a couple of Scan numbers that are not included in the original .mgf,
for example `2693`.

gvinterhalter:
No, these are the ones I can't find in the mgf, but as you say they are all probably off.

[unknown_PSM_psmtsv.txt](https://github.com/smith-chem-wisc/MetaMorpheus/files/2179055/unknown_PSM_psmtsv.txt)
 (I had to make .txt because of the github)

--

gvinterhalter:
Actually yes, it's an offset BUT:

1. using (sorted) mgf and mzml gives different resutls
2. mgf produces much less rows (498) and missing scans are offseted by +1 (as far as I can see)
3. using mzml (converted froms sorted mgf) produces more rows (1174) but scan numbers are completely random.

[...]

PR #1285, opened 2018-07-12
Update to mzlib 328

zrolfs:


GUI/GUI.csproj
```diff
@@ -50,47 +50,47 @@
     <ApplicationManifest>app.manifest</ApplicationManifest>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Chemistry, Version=1.0.0.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\packages\mzLib.1.0.325\lib\net471\Chemistry.dll</HintPath>
+    <Reference Include="Chemistry, Version=1.0.0.0, Culture=neutral, processorArchitecture=AMD64">
```
acesnik: Are you building with x64? I believe MSIL is for "Any CPU" so that MM can work on x32 computers...

zrolfs: image
There's no option to make the other to "Any CPU"

zrolfs:
This should fix #1274 by throwing an informative exception if a datafile has unique duplicate scans.

GUI/GUI.csproj
zrolfs: I couldn't get it to work, so I manually edited GUI.csproj in notepad :DumpsterFire:

acesnik: Thanks for working on that, Zach.

zrolfs:
This is ready...

PR #1297, opened 2018-07-16
Refactor common parameters to use default initializers

acesnik:
This change makes it so we don't have to add a new parameter in 3 places every time we add a parameter to the ComonParameters class. I also think using default initializers on the parameters makes it easier to see which have defaults set.

### DIFF
--- a/EngineLayer/CommonParameters.cs
+++ b/EngineLayer/CommonParameters.cs
@@ -2,133 +2,83 @@
 using Proteomics.ProteolyticDigestion;
 using System;
 using System.Collections.Generic;
+using System.Reflection;
 
 namespace EngineLayer
 {
+    /// <summary>
+    /// Parameters used across engines within a certain task
+    /// </summary>
     public class CommonParameters
     {
-        // this parameterless constructor needs to exist to read the toml.
-        // if you can figure out a way to get rid of it, feel free...
-        public CommonParameters() : this(digestionParams: null)
-        {
-        }
+        // Note:
+        //
+        // Any new property must not be nullable (int?) or else if it is null,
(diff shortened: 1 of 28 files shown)


EngineLayer/CommonParameters.cs
```diff
@@ -2,133 +2,83 @@
...
-                productMassTolerance: this.ProductMassTolerance,
-                precursorMassTolerance: this.PrecursorMassTolerance,
-                deconvolutionMassTolerance: this.DeconvolutionMassTolerance,
-                maxThreadsToUsePerFile: this.MaxThreadsToUsePerFile,
-                digestionParams: this.DigestionParams,
-                listOfModsVariable: this.ListOfModsVariable,
-                listOfModsFixed: this.ListOfModsFixed
-            );
-        }
+        public DigestionParams DigestionParams { get; set; } = new DigestionParams();
```
rmmiller22: I know from when I refactored Common Parameters a few months ago, eliminating the interface, that Rob wants the all the variables to be as secure as possible. This meaning only being set with the constructor or with a private set method(the initial goal was to have all variable be read only, but there are certain cases where the ability to set the variable outside of the constructor is necessary). You may want to discuss this with him further.

rmillikin: yeah I really don't want these properties to be easily settable. IMO they should be only settable in the constructor (either readonly, or private with no set method). The settings should not change in the middle of a run and I want to prevent that from accidentally happening.

acesnik:
Gotcha, @rmillikin and @rmmiller22. I'll make a smaller refactor, then.

PR #1305, opened 2018-07-20
File specific overwrote all other parameters to default

zrolfs:


zrolfs:
There's a hefty unit test, because it seems like this is the third time I've had to fix a parameter overwrite issue...

EngineLayer/CommonParameters.cs
```diff
@@ -104,15 +104,5 @@ public CommonParameters()
         public bool TrimMsMsPeaks { get; private set; }
         public bool UseDeltaScore { get; private set; }
         public bool CalculateEValue { get; private set; }
-
-        public CommonParameters Clone()
```
rmillikin: plz put this back

TaskLayer/MetaMorpheusTask.cs
```diff
@@ -181,7 +188,28 @@ public static CommonParameters SetAllFileSpecificCommonParams(CommonParameters c
                 zDotIons: zdotIons,
                 precursorMassTolerance: precursorMassTolerance,
                 productMassTolerance: productMassTolerance,
-                digestionParams: fileSpecificDigestionParams);
+                digestionParams: fileSpecificDigestionParams,
+
+                //NEED THESE OR THEY'LL BE OVERWRITTEN
```
rmillikin: 👍

EngineLayer/CommonParameters.cs
zrolfs: Is there a reason? I didn't see it being used for anything.

rmillikin: it's not used right now, but I'd like to use it to duplicate a task in the GUI

Issue #1290, opened 2018-07-13
mzML loading uses maximum threads regardless of Max Threads setting


rmillikin:
probably same for mgf. Thermo loading is not parallelized because of restrictions from MSFileReader.

This change needs to be made in mzLib.

rmillikin:
We should also check mzLib for other `Parallel` loops to make sure the number of threads is limited appropriately

acesnik:
Maybe make a static object in mzLib (`ParallelizationVariables`?) with the number of threads and also the Boolean for stopping loops, and then the cancel button could be used to stop flashLFQ and such.

rmillikin:
not sure when this was done but it's in #1406 at least

PR #1310, opened 2018-07-23
MetaMorpheus Add for maxThreads in MzLib change

ianhirsch:
#1290


ianhirsch:
Build eroor occurs in overload method from mzLib in the LoadFile method of MyFileManager.cs

PR #1315, opened 2018-07-24
Dumb down calibration (An Opinion Piece)

zrolfs:
This is a rework of calibration because the old calibration:
-Took too long to calibrate (opinion)
-Wasn't completely reproducible (some random seed thing going on)
-Felt like a black box with the random forest (main reason)

This PR yields consistent calibration results for the two raw files I tested with comparable (<1% difference) PSMs, peptides, and proteins. It requires ~500 ms between pre- and post-searches to calibrate.

All it does is take the relative mass errors for a scan and average them. It then smooths the mass errors over 50 scans to prevent over fitting. Each peak in the scan is then shifted according to the mass error.


acesnik:
What do you think about keeping the original method intact and making your new one a different option? That way we wouldn't have to commit to removing the old method right away, and we could rather test it over time.

A bit of duplication is fine. Maybe make a new `RelativeMassErrorCalibrationEngine` class.

zrolfs:
I am more hesitant on that. When the old option is left as the default, the new method tends not to get used and just collects dust/clutters the GUI (looking at Delta Scores here). The idea is that eventually it will get tested and phased in, but I haven't seen that happen with other features.

EngineLayer/Calibration/CalibrationEngine.cs
```diff
@@ -1,332 +1,249 @@
...
+            //impute missing values
+            //not all scans are guarenteed to contain data points. We can infer these data point with nearby points.
+            for (int index = 0; index < relativeErrors.Length; index++)
             {
-                for (int j = 0; j < myInputs.First().xValues.Length; j++)
+                if (relativeErrors[index] == 0) //if there were no points, then the value should be perfectly zero (the double default)
                 {
-                    myXValueMatrix[i, j] = myInputs[i].xValues[j];
+                    int startingBlankIndex = index;
+                    for (; index < relativeErrors.Length && relativeErrors[index] == 0; index++) { };
```
acesnik: This is pretty opaque. Any way you can do the math without a loop?

EngineLayer/Calibration/CalibrationEngine.cs
```diff
@@ -1,332 +1,249 @@
...
-            // parameter ranges for the optimizer
-            var parameters = new ParameterBounds[]
+            //get correction factor for each scan, where half of the smooth comes from both directions
+            double[] smoothedErrors = new double[relativeErrors.Length];
+            int leftIndex = 0; //starting scan index used for numbers less than the current scan
+            int rightIndex = 1; //
+            double smoothedCorrectionFactor = 0;
+            //for scan #1 (index 0)
+            //no left scans, because we're at the beginning of the file. Just populate the first N number of scan errors
+            for (; rightIndex < numberOfScansUsedForSmoothingOnEachSide && rightIndex < relativeErrors.Length; rightIndex++)
```
acesnik: again here

EngineLayer/Calibration/CalibrationEngine.cs
zrolfs: I need to figure out the next index at which there is a datapoint, so it has to loop through until it hit  a scan with a data point... I could comment it?

EngineLayer/Calibration/CalibrationEngine.cs
zrolfs: This loop is also needed, since we're getting the sum of all errors for the first "numberOfScansUsedForSmoothing". I'll put another commit with detailed comments. I incorrectly assumed it was an obvious implementation.

EngineLayer/Calibration/CalibrationEngine.cs
acesnik: I think a while loop would be clearer. These empty `for` loops without an initializer statement are really strange, and they're effectively while loops.

`while (index < relativeErrors.Length && relativeErrors[index] == 0) { index++; }`

EngineLayer/Calibration/CalibrationEngine.cs
acesnik: A while loop would be cleaner here, too.

EngineLayer/Calibration/CalibrationEngine.cs
```diff
@@ -1,332 +1,252 @@
...
 namespace EngineLayer.Calibration
 {
     public class CalibrationEngine : MetaMorpheusEngine
     {
-        private const double MaximumFracForTraining = 0.70;
-        private const double MaximumDatapointsToTrainWith = 20000;
-        private const int TrainingIterations = 30;
-        private readonly int RandomSeed;
-
+        private const int NumberOfScansUsedForSmoothingOnEachSide = 100;
```
rmillikin: I like this PR

EngineLayer/Calibration/CalibrationEngine.cs
```diff
@@ -1,332 +1,252 @@
...
         private readonly MsDataFile MyMsDataFile;
         private readonly DataPointAquisitionResults Datapoints;
+        public MsDataFile CalibratedDataFile { get; private set; }
```
rmillikin: RAM issues maybe? does doubling-up on storing the spectra causes 2x the RAM usage? or did we do this before?

TaskLayer/CalibrationTask/CalibrationTask.cs
```diff
@@ -332,7 +337,7 @@ private DataPointAquisitionResults GetDataAcquisitionResults(MsDataFile myMsData
 
             new FdrAnalysisEngine(allPsms, searchMode.NumNotches, CommonParameters, new List<string> { taskId, "Individual Spectra Files", fileNameWithoutExtension }).Run();
 
-            List<PeptideSpectralMatch> goodIdentifications = allPsms.Where(b => b.FdrInfo.QValueNotch < 0.01 && !b.IsDecoy && b.FullSequence != null).ToList();
+            List<PeptideSpectralMatch> goodIdentifications = allPsms.Where(b => b.FdrInfo.QValueNotch < 0.001 && !b.IsDecoy && b.FullSequence != null).ToList();
```
rmillikin: any particular reason for this?

zrolfs: It produced better calibration, since there was a lot of noise introduced by the lower confidence PSMs. Average MS1 PPM error went from -0.059 to -0.031 and standard deviation went from 0.91 to 0.87.
It's not a huge difference, so we can scrap it if you prefer, but it seemed that using higher confident data points achieved better calibration. It doesn't add any time, either.

I would think the advantage to using 0.01 instead of 0.001 would be for files with fewer PSMs in them.

EngineLayer/Calibration/CalibrationEngine.cs
zrolfs: :)

EngineLayer/Calibration/CalibrationEngine.cs
zrolfs: I agree that this would take up double the RAM, but the RAM of an MS file isn't massive (the most I've seen is 3GB). We can get around this by modifying the existing spectra (as was done previously), but that is a little ugly and opens the door for continuous integration issues. I am leaning that we make a new calibrated spectrum instead of modifying an old one.

zrolfs: I'm willing to go back to modifying the spectra instead of creating new spectra, but then this will have to wait until mzLib gets updated in MetaMorpheus.

Issue #1223, opened 2018-06-20
Non-specific crash


zrolfs:
Not sure what's up, but I'll fix it this weekend.

rmillikin:
any update?

zrolfs:
No, this slipped my mind. It's a bit tricky, so probably something I have to do and not something we can outsource to Ian.

zrolfs:
I couldn't reproduce this in 0.0.286.

zrolfs:
Outsourcing from graduate students to undergraduate students was the subtext, but I suppose it's insourcing if your frame of reference is to the group.

zrolfs:
Alright, I got the crash again. Still an issue. Still working on it...

zrolfs:
It was whoever added  PostSearchAnalysisParameters. MetaMorpheusTask already has a CommonParameters as a property, but it was also added in the PostSearchAnalysisParameters. The other SequencesToActualPeptides was updated, but not the nonspecific one...

rmillikin:
you mean there is a CommonParameters object in the PostSearchAnalysisParameters and the MetaMorpheusTask, when there should only be one in the MetaMorpheusTask?

zrolfs:
Yes

rmillikin:
fixed in #1329

PR #1329, opened 2018-08-03
Fix NSESTAP

zrolfs:
#1223


EngineLayer/NonSpecificEnzymeSearch/NonSpecificEnzymeSequencesToActualPeptides.cs
```diff
@@ -108,7 +108,9 @@ protected override MetaMorpheusEngineResults RunSpecific()
                                         if (index >= 0 && index >= digestionParam.MinPeptideLength)
                                         {
                                             //generate correct sequence
-                                            PeptideWithSetModifications tempPWSM = new PeptideWithSetModifications(pwsm, pwsm.OneBasedStartResidueInProtein, pwsm.OneBasedStartResidueInProtein + index - 1);
+                                            Dictionary<int, ModificationWithMass> allModsOneIsNTerminus = pwsm.AllModsOneIsNterminus
```
rmillikin: the idea here is to just remove mods that can't be on the clipped peptide right? for your non specific search?

rmillikin: wait I'm confused, is this a bug in mzLib? is it putting mods where it shouldn't? or is this some funky nonspecific thing

zrolfs: Yes to the first. The second one is bug in mzlib, because digestion params isn't carried over in the constructor, which causes a parsimony crash now because multiprotease searches require them.

Issue #1346, opened 2018-08-15
Provide a contaminants database in the GUI


trishorts:
rather than making the user create their own. This is the one from MaxQuant

http://lotus1.gwdg.de/mpg/mmbc/maxquant_input.nsf/7994124a4298328fc125748d0048fee2/$FILE/contaminants.fasta

rmillikin:
there's one in the MetaMorpheus box folder: https://uwmadison.box.com/v/MetaMorpheusPublic
actually two, one is fasta, one is xml.

not exactly sure where that comes from. It's labeled UniProt.

also I think this suffices? https://www.thegpm.org/crap/

I don't like the idea of distributing protein databases... but my mind could be changed. I know MaxQuant does it.

rmillikin:
though I do like Zach's implementation. a button that says "add default contaminants" is nice

zrolfs:
The database I used is an xml version of the crapome (the gpm crap)

trishorts:
I looked again at the GPM one. Doesn't seem like it's been updated in a while. And, there is some unusual stuff there. There is stuff at CRAPome.org, but that is also problematic for other reasons. I spent time yesterday and today looking around and I think the MaxQuant one might be the best thing going for standard proteomics searches. I ended up re-visiting this because Brett Phinney asked me what we used. I recommend that we all switch over to this. Since this is a small file, I'd like to make it available directly. It's just not all that convenient to stash and use (and it's also a nice service to provide). Users can complain about its contents and we can update accordingly. Crowdsourcing composition might be nice.

zrolfs:
The MaxQuant file has some weird stuff in it, too, with proteins from TrEMBL, Ensembl, RefSeq, and a Custom (source unspecificed).

rmillikin:
done in #1347

PR #1347, opened 2018-08-15
Add contaminant database

zrolfs:
#1346


rmillikin:
this is semi-nitpicky but it feels a little weird to rip MaxQuant's contaminants database and call it MetaMorpheusContaminants. What was the original name of the file?

PR #1351, opened 2018-08-17
remove Neo

rmillikin:
I'm ok having Neo in MetaMorpheus (or interacting with MM the way MM interacts with FlashLFQ) but I don't want to have a bunch of code in the repo that's not used and is untested. And it's probably pretty out of date anyway?


acesnik:
This should get Zach's approval, but it looks fine to me.

rmillikin:
well I view it as just temporary, whenever you want to put your updated code in you're welcome to

PR #1361, opened 2018-08-21
Install contaminants

zrolfs:


rmillikin:
This builds and the unit tests pass, codecov seems to be down though, which is why the appveyor build failed

Issue #1374, opened 2018-08-23
"ProteinTermMod: Met loss + Acetylation" is present when there aren't methionine


zrolfs:
This mod should require an M OR should not be a modification. This mod is only useful if the protein methonine setting is set to retained, which is pretty uncommon.

zrolfs:
Also, "ProteinTermMod: Acetylation + Methionine" currently has two neutral losses, "0", and "173.05". There shouldn't be a neutral loss?

rmillikin:
I'd vote to get rid of this mod. This situation is taken care of by setting M cleavage to variable and having protein N-term acetylation checked, right?

rmillikin:
done in #1396

Issue #1371, opened 2018-08-23
Crash in non-specific search


rmillikin:
Katie reported this crash. It occurs in v288 and v287. Unsure about previous versions.

[SearchTaskconfig.zip](https://github.com/smith-chem-wisc/MetaMorpheus/files/2315050/SearchTaskconfig.zip)
[results.zip](https://github.com/smith-chem-wisc/MetaMorpheus/files/2315060/results.zip)

Data files are on bison; ask me for the folder

zrolfs:
This is looking like an mzlib issue. The key is present in the dictionary, but the program isn't calling them as equal (same values, but not the same object). I presume that the Equals() override was deleted, but I can't tell from the history because the old CompactPeptideWithModifiedMass is no longer present in mzlib.

PR #1375, opened 2018-08-23
Fix nonspecific crash when weird mods are used

zrolfs:
I'm honestly not sure what's going on here. I expect this will have to change when compact peptide is refactored.
The notes from this headache are that:
There can be multiple neutral losses on a single N-terminus? This caused the whole N-terminal mass to be removed previously.
There needs to be a unit test that runs the whole non-specific (index, search, and post search), because lots of crashes are happening only when these steps interface with each other.
#1374 #1371


zrolfs:
Here are some notes so that reviewing isn't so painful, since a couple of lines were "changed" (but they weren't, thanks github).
-Previously, when there were multiple neutral losses, only the last one was used. The indexing uses both (or at least the first one), so a different mass was produced here, excluding the peptide from the match. The new code searches for each one of the neutral masses separately to ensure the correct one is not overlooked. I think this is treating a symptom of a different problem, but here's the bandaid.
This fix has to be applied twice, because there are N and C terminal subtelties when using different ion types.

rmillikin:
does this fix #1371 ?

zrolfs:
This fixes #1371. It does NOT fix #1374

PR #1376, opened 2018-08-24
Unit tests for Search Task

ianhirsch:


Test/SearchTaskTest.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            try
+            {
+                result = SearchTask.GetMassDiffAcceptor(searchTask.CommonParameters.PrecursorMassTolerance, MassDiffAcceptorType.Custom, "TestCustom Test 5");
+            }
+            catch(MetaMorpheusException)
+            {
+                return;
+            }
+            Assert.Fail();
```
rmillikin: Assert that exception is thrown instead of unit test fail

Test/SearchTaskTest.cs
```diff
@@ -0,0 +1,93 @@
...
+                CommonParameters = new CommonParameters()
+            };
+
+            string outputFolder = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestSameSettingsOutput");
+            string myFile = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\PrunedDbSpectra.mzml");
+            string myDatabase = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\DbForPrunedDb.fasta");
+            DbForTask db = new DbForTask(myDatabase, false);
+
+            searchTask.RunTask(outputFolder, new List<DbForTask> { db }, new List<string> { myFile }, "SettingsTest");
+        }
```
rmillikin: Read in the PSM results file and then assert that there are some PSMs

EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
```diff
@@ -87,7 +89,7 @@ private void DoFalseDiscoveryRateAnalysis(FdrAnalysisResults myAnalysisResults)
             //determine if Score or DeltaScore performs better
             if (UseDeltaScore)
             {
-                const double qValueCutoff = 0.01; //optimize to get the most PSMs at a 1% FDR
```
zrolfs: Please leave the hard coded 0.01. The default qValue right now is 1.0, which doesn't make sense to use here, since delta and normal scores would produce the same results

PR #1379, opened 2018-08-28
Return key to close task windows

zrolfs:


GUI/CalibrateTaskWindow.xaml.cs
```diff
@@ -252,18 +252,21 @@ private void SaveButton_Click(object sender, RoutedEventArgs e)
             DialogResult = true;
         }
 
-        private void Row_DoubleClick(object sender, MouseButtonEventArgs e)
```
zrolfs: I removed this method, because it wasn't called anywhere. I also don't understand why there would be rows in the calibration window.

PR #1380, opened 2018-08-28
Unit test for XLSearch

ianhirsch:


Test/SearchTaskTest.cs
```diff
@@ -181,5 +181,12 @@ public static void PrunedDbWithContaminantsTest()
             Assert.That(File.ReadAllLines(Path.Combine(folderPath, @"DbForPrunedDbproteinPruned.xml")).Length > 0);
             Assert.That(File.ReadAllLines(Path.Combine(folderPath, @"DbForPrunedDbPruned.xml")).Length > 0);
         }
+
+        [Test]
```
rmillikin: remove? or finish?

Test/XLTest.cs
```diff
@@ -273,6 +273,46 @@ public static void XlTest_DiffCrosslinkSites()
...
+            XLSearchTask xlSearchTask = new XLSearchTask();
+            string myFile = Path.Combine(TestContext.CurrentContext.TestDirectory, @"XlTestData\BSA_DSS_23747.mzML");
+            string myDatabase = Path.Combine(TestContext.CurrentContext.TestDirectory, @"XlTestData\BSA.fasta");
+            string folderPath = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestXLSearch");
+            DbForTask db = new DbForTask(myDatabase, false);
+
+            //creates .params files if they do not exist
+            xlSearchTask.RunTask(folderPath, new List<DbForTask> { db }, new List<string> { myFile }, "normal");
+            //tests .params files
+            xlSearchTask.RunTask(folderPath, new List<DbForTask> { db }, new List<string> { myFile }, "normal");
```
rmillikin: can you write the results of these two tasks to two different directories and then check that the output is equal?

PR #1382, opened 2018-08-28
Fix nses indexing

zrolfs:
There wasn't a check that the index was within the maximum allowed range.


zrolfs:
I was tipped off to this issue, because a proteolysis product outside of the max peptide length allowed was generated and caused a crash. That still needs to be fixed in mzlib. But this will prevent any crashes for users that don't specify length constraints.

PR #1386, opened 2018-08-30
Remove referenced methods, Add XL Test

ianhirsch:


Test/XLTest.cs
```diff
@@ -136,6 +136,37 @@ public static void XlTest_XlPosCal()
             Assert.AreEqual(crosslinkerModSitesAll, "STC");
         }
 
+        [Test]
+        public static void DeadendPeptideTest()
+        {
```
rmillikin: this is OK for now but I would like to get a new spectra file with a real dead end peptide and make sure that we can correctly identify it

Issue #1364, opened 2018-08-21
point new users to wiki and youtube to avoid common mistakes


trishorts:
in the message box
OR in the installer message (new/revised features).

rmillikin:
done in #1391

PR #1391, opened 2018-08-31
Add youtube and wiki link

bintiibrahim:
Issue #1364

### DIFF
--- a/GUI/MainWindow.xaml
+++ b/GUI/MainWindow.xaml
@@ -225,17 +225,6 @@
                         Height="5" />
                 <GroupBox Header="Notifications" Grid.Row="2">
                     <RichTextBox x:Name="notificationsTextBox" VerticalScrollBarVisibility="Visible" IsReadOnly="True" IsDocumentEnabled="True" >
-                        <FlowDocument>
-                            <Paragraph>
-                                Visit our 
-                                <Hyperlink NavigateUri="https:" RequestNavigate="WikiHyperlink">Wiki page</Hyperlink>
-                                or
-                                <Hyperlink NavigateUri="https:" RequestNavigate="YoutubeHyperlink">Youtube</Hyperlink>
-                                for more information.
-                                <LineBreak/>
-                                <LineBreak/>
-                            </Paragraph>
-                        </FlowDocument>
                         <RichTextBox.Resources>
                             <Style TargetType="{x:Type Paragraph}">
                                 <Setter Property="Margin" Value="0" />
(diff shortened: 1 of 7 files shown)


GUI/MainWindow.xaml
```diff
@@ -224,7 +224,18 @@
                         Background="Silver"
                         Height="5" />
                 <GroupBox Header="Notifications" Grid.Row="2">
-                    <RichTextBox x:Name="notificationsTextBox" VerticalScrollBarVisibility="Visible" IsReadOnly="True">
+                    <RichTextBox x:Name="notificationsTextBox" VerticalScrollBarVisibility="Visible" IsReadOnly="True" IsDocumentEnabled="True" >
```
rmillikin: can we add some kind of check that writes this only the first time someone opens MetaMorpheus?

bintiibrahim: would clearing the notifications textbox when "run all tasks" is clicked work?

rmillikin: that would still output this every time MetaMorpheus is run - I'd prefer if this was only output the first time it was run. Try adding a `bool` in `GuiGlobalParams`. Maybe like `bool FirstTimeRunning = true;` then when MM is opened, set it to `false`. I think it will reset every time MetaMorpheus is updated but that's fine

bintiibrahim: how would the bool value translate to the GUI? i read something about data triggers but i'm not entirely sure.

bintiibrahim:
I had to make some changes relating to the GuiGlobalParams access (to non internal). not so sure if it will hugely affect the rest of the code. let me know what u think.

GUI/MainWindow.xaml.cs
```diff
@@ -69,7 +69,16 @@ public MainWindow()
             FileSpecificParameters.ValidateFileSpecificVariableNames();
 
             // LOAD GUI SETTINGS
-            GuiGlobalParams = Toml.ReadFile<GuiGlobalParams>(Path.Combine(GlobalVariables.DataDir, @"GUIsettings.toml"));
+
+            try
+            {
+                GuiGlobalParams = Toml.ReadFile<GuiGlobalParams>(Path.Combine(GlobalVariables.DataDir, @"GUIsettings.toml"));
+                notificationsTextBox.Document.Blocks.Clear();
```
rmillikin: probably not a good idea to clear the notifications on startup, since there can be important warning messages. instead of coding the startup message into the xaml, write it here if the GUISettings.toml doesn't exist

PR #1396, opened 2018-09-05
remove some mods

rmillikin:
#1374


zrolfs:
This will fail appveyor. You need to update  the "FilesLoading()" test

zrolfs:
Oops, that's in mzlib and not MetaMorpheus. My mistake.

PR #1423, opened 2018-09-27
added neutralLosses and diagnosticIons to Mods.txt

trishorts:


EngineLayer/Mods/Mods.txt
```diff
@@ -96,11 +96,19 @@ MT   Mod
...
+MT   Mod
+CF   H1 O3 P1
+//
+ID   Phospho
+TG   Y
 PP   Anywhere.
 NL   H0 or H3 O4 P1
 MT   Mod
 CF   H1 O3 P1
+DI   216.0420
```
rmillikin: I think these neutral losses and diagnostic ions are associated with particular fragmentation methods right? HCD/CID?

EngineLayer/Mods/Mods.txt
```diff
@@ -277,8 +374,22 @@ MT   Mod
...
+//
+ID   DiMethyl (asymmetric)
+TG   R
+PP   Anywhere.
+NL   ETD:45.0579
+MT   Mod
+CF   H4 C2
+//
+ID   DiMethyl (symmetric)
+TG   R
```
rmillikin: this particular case might be problematic (2 different dimethyl mods targeting the same residue). unless you're doing ETD there is no way to distinguish these, which means every PSM will be ambiguous between these two mods for HCD/CID searches

I understand the reasons for separating these into two mods. But I think it might be better to have a single mod Dimethyl with NL 45 or 31 Da maybe? it would avoid (imo unnecessary) ambiguity but would make interpretation of ETD spectra  bit more manual

acesnik:
image

PR #1434, opened 2018-10-02
Category Fdr for Full/Semi/Non

zrolfs:


EngineLayer/ClassicSearch/ClassicSearchEngine.cs
```diff
@@ -17,11 +17,11 @@ public class ClassicSearchEngine : MetaMorpheusEngine
         private readonly List<Protein> Proteins;
         private readonly List<Modification> FixedModifications;
         private readonly List<Modification> VariableModifications;
-        private readonly PeptideSpectralMatch[] PeptideSpectralMatches;
+        private readonly PeptideSpectralMatch[][] PeptideSpectralMatches;
```
rmillikin: this seems overly complex... do you really need a 2d array of PSMs for the search? aren't you just categorizing the PSM results by cleavage specificity in the FDR (post-processing?)

GUI/SearchTaskWindow.xaml
```diff
@@ -307,6 +307,13 @@
                                                     </TextBlock>
                                                 </ComboBox.ToolTip>
                                             </ComboBox>
+                                            <CheckBox x:Name="checkBoxClassicSemiSpecific" Content="Semi-Specific Digestion" Margin="6" DockPanel.Dock="Top" ToolTipService.ShowDuration="999999" ToolTipService.InitialShowDelay="500">
```
rmillikin: what is this for? we already have a semi-specific search button?

GUI/SearchTaskWindow.xaml.cs
```diff
@@ -353,6 +352,19 @@ private void SaveButton_Click(object sender, RoutedEventArgs e)
             }
 
             Protease protease = (Protease)proteaseComboBox.SelectedItem;
+            if (classicSemiSpecific)
```
rmillikin: I assume you're including options for "traditional" semi specific and your semi specific search algorithm. This seems pretty confusing. If a user wants to do a semi-specific search will it be clear in the GUI what the difference is and what they're supposed to do?

I'd rather we just give one option instead of providing one recommended one and one subpar one with the same names

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -859,5 +870,102 @@ private void WritePeakQuantificationResultsToTsv(FlashLfqResults flashLFQResults
 
             FinishedWritingFile(peaksPath, nestedIds);
         }
+
+        private void DetermineBestCategoryPsms()
+        {
```
rmillikin: I don't really get this. In the FDR engine, can you just do `AllPsms.GroupBy(p => p.BestPeptide.Peptide.CleavageSpecificity)` in the FDR analysis engine, and then for each group, calculate group q-values? Seems much simpler, what am I missing?

How do you want to handle ambiguity? Like if a spectrum matches to a semi-specific peptide and a full specificity peptide, what category does it belong in for FDR purposes?

EngineLayer/ClassicSearch/ClassicSearchEngine.cs
zrolfs: I agree that this is overly complex. This was something @acesnik brought up, in that the highest scoring peptide from each category (full/semi/nonspecific) should be saved so that they don't have to compete against each other using the MetaMorpheus score. Instead, they each get a q-value from their category, and then the psms between categories "compete" based on q-value, instead of on score. In this way, ambiguity between a full and a semi is resolved based on q-value.

GUI/SearchTaskWindow.xaml
zrolfs: We have a semi-specific search mode button, but this is a semi-specific protease button. The idea being that if you click this, trypsin turns into semi-trypsin. As such, I've removed semi-trypsin from the protease list. Using a button was cleaner to me than adding a bunch of semi-proteases for comparison (semi-GluC, semi-AspN, etc). I wanted to leave the classic semi-specific search mode intact to compare with for the paper. We can phase it out once published.

GUI/SearchTaskWindow.xaml.cs
zrolfs: I completely agree. I need both, for now, to compare between them for the paper.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
zrolfs: We're not getting the best psms from each category, but rather a single psm for each scan and discriminating based on the category specific q-value.
Because of this, when a spectrum matches to a semi- and a fully-specific peptide, the one with the lower q-value wins out.

zrolfs:
Still working on this; should be done in a few hours.

zrolfs:
Alright, this is pretty close to ready. Just a few GUI things left that I need to manually check, but the guts of it are finished.

TaskLayer/SearchTask/SearchTask.cs
```diff
@@ -169,37 +169,50 @@ protected override MyTaskResults RunSpecific(string OutputFolder, List<DbForTask
                 // nonspecific search
                 else if (SearchParameters.SearchType == SearchType.NonSpecific)
                 {
-                    for (int currentPartition = 0; currentPartition < combinedParams.TotalPartitions; currentPartition++)
+                    List<CommonParameters> paramsToUse = new List<CommonParameters> { combinedParams };
```
rmmiller22: Why are you making paramsToUse instead of just using combinedParams, especially if combined params is the only thing within the paramsToUse list?

zrolfs: For semi, I need to search two different params: one for N-terminal specific cleavages and one for C-terminal. If it's semi, the list gets modified (see next line).

rmmiller22: Ok. I understand now

zrolfs:
Hey, do we still do codacy? I don't see that on here as a check.

Issue #1424, opened 2018-09-27
wipe temporarily stored data in unit tests to prevent future fails


trishorts:
SemiSpecificTest generated fragment and precursor indexes several days ago. These indexes were not updated during current unit testing. The old style digestionparams caused current mm tests to fail. I spent a long time chasing a bug that didn't exist. Maybe, let's wipe these and other folders following each unit test.

rmillikin:
I think we should probably add a line at the end of each unit test that wipes that unit test's output. 

I also really prefer to have each unit test write output to a folder with that unit test's name. Keeps things nice and organized when you're debugging and looking for unit test output.

rmillikin:
fixed in #1436

Issue #1413, opened 2018-09-24
Output all .toml to one directory instead of one .toml to each task directory


rmillikin:

rmillikin:
done in #1436

PR #1436, opened 2018-10-03
Consolidate toml

ianhirsch:
#1413 #1424


TaskLayer/MetaMorpheusTask.cs
```diff
@@ -216,7 +216,12 @@ public MyTaskResults RunTask(string output_folder, List<DbForTask> currentProtei
         {
             StartingSingleTask(displayName);
 
-            var tomlFileName = Path.Combine(output_folder, GetType().Name + "config.toml");
+            var path = output_folder;
+            if (!output_folder.Contains("netcoreapp2.0") && !(output_folder == ""))
```
rmillikin: how much does this break if you don't check for this? I'm not really a fan of different code for unit tests vs. "normal" runs

TaskLayer/MetaMorpheusTask.cs
```diff
@@ -216,7 +216,12 @@ public MyTaskResults RunTask(string output_folder, List<DbForTask> currentProtei
...
+            {
+                path = Directory.GetParent(output_folder).FullName;
+            }
+            var tomlFileName = Path.Combine(path, "Task Settings", GetType().Name + "config.toml");
```
rmillikin: use task name instead of task type

PR #1465, opened 2018-10-11
semi/nonspecific fixes

zrolfs:


zrolfs:
Fixed two bugs:
1. If the peptide was correctly digested during indexing, the precursor mz was being indexed instead of the mass. Mass is now indexed.
2. Peptides with the minimum length were being removed because of the fragment index lookup. "-1" has been added to the loop to prevent their exclusion.

A unit test has been added covering both of these situations.

zrolfs:
This is passing all unit tests in visual studio, but failing appveyor here. Anybody know why?

GUI/SearchTaskWindow.xaml.cs
```diff
@@ -144,6 +144,7 @@ private void PopulateChoices()
 
         private void UpdateFieldsFromTask(SearchTask task)
         {
+            proteaseComboBox.SelectedItem = task.CommonParameters.DigestionParams.SpecificProtease; //needs to be first, or nonspecific overwrites for missed cleavages/max length
```
rmillikin: I don't understand why this is moved

GUI/SearchTaskWindow.xaml.cs
```diff
@@ -524,7 +524,8 @@ private void SaveButton_Click(object sender, RoutedEventArgs e)
                 TheTask.SearchParameters.CustomMdac = customkMdacTextBox.Text;
             }
 
-            //determine if semi or nonspecific with a specific protease. Full is already added by default.
+            //determine if semi or nonspecific with a specific protease.
+            TheTask.SearchParameters.LocalFdrCategories = new List<FdrCategory> { FdrCategory.FullySpecific }; //set the default again to reset the list
             if(searchModeType == CleavageSpecificity.Semi || protease.CleavageSpecificity==CleavageSpecificity.Semi)
```
rmillikin: i'd like to change this structure a bit. instead of adding to a list, can we just set the list based on conditions? seems less error-prone.
```
if(semispecific)
{
   SearchParameters.LocalFdrCategories = new List<>{ Full, Semi };
}
if(nonspecific)
{
   SearchParameters.LocalFdrCategories = new List<>{ Semi, Non };
}
else
{
   SearchParameters.LocalFdrCategories = new List<>{ Full };
}
```

Also, it seems weird to me to have semi-specific FDR category as a separate FDR category for the nonspecific search. If I'm doing a "true" nonspecific search and there is no digestion motif at all, what does "semispecific" even mean?

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -47,14 +47,17 @@ public MyTaskResults Run()
             }
 
             //update all psms with peptide info
-            Parameters.AllPsms = Parameters.AllPsms.Where(psm => psm != null).ToList();
-            Parameters.AllPsms.ForEach(psm => psm.ResolveAllAmbiguities());
+            if (Parameters.SearchParameters.SearchType != SearchType.NonSpecific) //if it hasn't been done already
```
rmillikin: I don't really like that q-values are calculated in the nonspecific search without using the FdrAnalysisEngine. it's a code duplication thing. can we edit FdrAnalysisEngine to do FDR for each FDR category instead of having a separate duplicated FDR calculation method in NonSpecificEnzymeSearchEngine?

GUI/SearchTaskWindow.xaml.cs
zrolfs: For non-specific searches with nonspecific protease, the number of missed cleavages is always updated as the maximum length -1 and vice versa. If the protease is not actually non-specific (like you're doing chymotrypsin), but you want to look for non-specific peptides (because chymotrypsin is awful), then you can do a non-specific search, but keep chymotrypsin as a the protease for FDR purposes. If protease isn't set first, then the missed cleavages is updated to the maximum length -1 (50-1=49). That would make sense in nonspecific searches, but we interpret it as allowing 49 chymotrypsin cleavages, which is an insane number.

Would you like me to make that explanation a comment?

GUI/SearchTaskWindow.xaml.cs
zrolfs: I'll make the first change.
You can do nonspecific searches of specific proteases  (rare in trypsin, common in chymotrypsin). In that case, full, semi, and non are all stratified.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
zrolfs: The nonspecific method uses the fdranalysis engine for each category. A separate method is needed to aggregate them, though, which is what this does.

GUI/SearchTaskWindow.xaml.cs
rmillikin: ok, so Non-specific search mode with "chymotrypsin" selected as the protease is a valid option?

what happens if I select "Non-specific" as the protease with Non-specific search mode checked? Is there only effectively one FDR category then?

zrolfs: Correct.
If you do non-specific with non-specific, there are still three categories, but only the non-specific one ever gets used.

Issue #1435, opened 2018-10-03
Search for mods by name


rmillikin:

rmillikin:
done in #1484. it's a little slow but it works

PR #1484, opened 2018-10-22
search mods box

bintiibrahim:
#1435


GUI/CalibrateTaskWindow.xaml.cs
```diff
@@ -11,6 +11,8 @@
 using TaskLayer;
 using Proteomics.ProteolyticDigestion;
 using MassSpectrometry;
+using System.Windows.Threading;
+using MetaMorpheusGUI.Util;
```
rmillikin: remove `using MetaMorpheusGUI.Util;`
potentially remove the other using - what is this for?

GUI/CalibrateTaskWindow.xaml.cs
```diff
@@ -270,5 +272,50 @@ private void KeyPressed(object sender, KeyEventArgs e)
                 CancelButton_Click(sender, e);
             }
         }
+
+        internal SearchModifications SearchMod = new SearchModifications();
```
rmillikin: change `FilterTree` into a static method in `SearchModifications` class; no need to initialize a `SearchModifications` object

GUI/CalibrateTaskWindow.xaml.cs
```diff
@@ -270,5 +272,50 @@ private void KeyPressed(object sender, KeyEventArgs e)
...
+
+        private void TextChanged_Fixed(object sender, TextChangedEventArgs args)
+        {
+            SearchMod.FixedSearch = true;
+            SetTimer();
```
rmillikin: remove all this timer stuff, I assume that's for debug purposes?

GUI/CalibrateTaskWindow.xaml.cs
```diff
@@ -270,5 +272,50 @@ private void KeyPressed(object sender, KeyEventArgs e)
                 CancelButton_Click(sender, e);
             }
         }
+
+        internal SearchModifications SearchMod = new SearchModifications();
+
+        private void TextChanged_Fixed(object sender, TextChangedEventArgs args)
+        {
+            SearchMod.FixedSearch = true;
```
rmillikin: Call `SearchModifications.FilterTree` here directly and don't use the bools etc

GUI/CalibrateTaskWindow.xaml.cs
```diff
@@ -270,5 +272,50 @@ private void KeyPressed(object sender, KeyEventArgs e)
...
+            SetTimer();
+        }
+
+        private void TextChanged_Var(object sender, TextChangedEventArgs args)
+        {
+            SearchMod.VarSearch = true;
```
rmillikin: Call SearchModifications.FilterTree here directly

GUI/CalibrateTaskWindow.xaml.cs
```diff
@@ -270,5 +272,50 @@ private void KeyPressed(object sender, KeyEventArgs e)
...
+            SetTimer();
+        }
+
+        private void SetTimer()
```
rmillikin: remove

GUI/CalibrateTaskWindow.xaml.cs
```diff
@@ -270,5 +272,50 @@ private void KeyPressed(object sender, KeyEventArgs e)
...
+        {
+            if (!SearchMod.TimerCreated)
+            {
+                SearchMod.Timer.Tick += new EventHandler(TextChangeTimerHandler);
+                SearchMod.TimerCreated = true;
+            }
+            SearchMod.SetTimer();
+        }
+
+        private void TextChangeTimerHandler(object sender, EventArgs e)
```
rmillikin: I don't think you need this method if you call `SearchModifications.FilterTree` above

GUI/GPTMDTaskWindow.xaml
```diff
@@ -260,7 +260,7 @@
                                 </Grid.ColumnDefinitions>
                                 <StackPanel>
                                     <Label Content="Fixed" />
-                                    <TreeView x:Name="fixedModsTreeView" ItemsSource="{Binding}">
+                                    <TreeView x:Name="fixedModsTreeView" ItemsSource="{Binding}" Height="400">
```
rmillikin: does this need to be fixed height?

GUI/GPTMDTaskWindow.xaml
```diff
@@ -279,7 +279,7 @@
                                 </StackPanel>
                                 <StackPanel Grid.Column="1">
                                     <Label Content="Variable" />
-                                    <TreeView x:Name="variableModsTreeView" ItemsSource="{Binding}">
+                                    <TreeView x:Name="variableModsTreeView" ItemsSource="{Binding}" Height="400">
```
rmillikin: same

GUI/Util/SearchModifications.cs
```diff
@@ -0,0 +1,58 @@
+﻿using System;
+using System.Collections.ObjectModel;
+using System.Linq;
+using System.Windows.Controls;
+using System.Windows.Threading;
+
+namespace MetaMorpheusGUI.Util
```
rmillikin: change namespace to MetaMorpheusGUI

GUI/Util/SearchModifications.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    class SearchModifications
+    {
+        public DispatcherTimer Timer;
```
rmillikin: remove timer stuff

GUI/Util/SearchModifications.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        public void SetTimer()
+        {
+            Timer.Stop(); // Resets the timer
+            Timer.Start();
+        }
+
+        public void FilterTree(TextBox textbox, TreeView tree, ObservableCollection<ModTypeForTreeView> collection)
+        {
```
rmillikin: the major feature I'd like to see here is I want the mod type visible along with the expanded node and the modifications that meet the search criteria. e.g.:

Currently:
```
[] Mod1
[] Mod2
[] Mod3
```

I'd like:
```
[] ModType1
   [] Mod1
   [] Mod2
[] ModType2
   [] Mod3
```

GUI/CalibrateTaskWindow.xaml.cs
bintiibrahim: I’m using the timer to keep track of the user typing. Without it, the tree gets filtered immediately after each keystroke and makes the search slow. Should I make a “go” button instead?

rmillikin: that's fine then - can you put a comment in there explaining what that's for?

bintiibrahim: yeap. I’ll add comments and make the other edits first thing tomorrow!

GUI/CalibrateTaskWindow.xaml.cs
bintiibrahim: System.Windows.Threading is for the timer. should i just use `System.Windows.Threading.DispatcherTimer` ?

GUI/CalibrateTaskWindow.xaml.cs
bintiibrahim: so my idea behind SearchModifications is to reduce code redundancy for each **window** (CalibrateTask, SearchTask, XLSearchTask, ...) - so each instance of SM would set the timer, track which search box is used (fixed/var), and filter the mods. what would be the alternative?

lonelu: I like the idea of keeping one.

GUI/CalibrateTaskWindow.xaml.cs
rmillikin: no this is fine as is

GUI/CalibrateTaskWindow.xaml.cs
rmillikin: what you described is fine and is a good idea (centralizing and not duplicating code). I just want it to be a static method

GUI/CalibrateTaskWindow.xaml.cs
```diff
@@ -22,25 +23,24 @@ public partial class CalibrateTaskWindow : Window
...
-        {
-            InitializeComponent();
-            PopulateChoices();
-            
-            TheTask = new CalibrationTask();
-            UpdateFieldsFromTask(TheTask);
+        private readonly SearchModifications SearchMod = new SearchModifications();
 
-            this.saveButton.Content = "Add the Calibration Task";
+        public CalibrateTaskWindow() : this(null)
```
bintiibrahim: @rmillikin i simplified the constructors for all the task windows cause I noticed code redundancy. let me know if u prefer having it the old way.

GUI/Util/SearchModifications.cs
```diff
@@ -0,0 +1,61 @@
...
+        public static void FilterTree(TextBox textbox, TreeView tree, ObservableCollection<ModTypeForTreeView> collection)
+        {
+            string key = textbox.Text.ToLower();
+            if (string.IsNullOrEmpty(key))
+            {
+                tree.DataContext = collection; // shows full tree if nothing is searched
+                return;
+            }
+
+            var modTypesWithMatchingMods = collection.Where(p => p.Children.Any(c => c.DisplayName.ToLower().Contains(key))); // parent of child mods that match key
```
rmillikin: i edited this region but it's still basically the same.

I'm wondering if it's faster to show/hide nodes in the tree rather than recreating a whole new one and copying over nodes from the old one to the new one. Can you try showing/hiding nodes in the tree rather than creating/setting a new one? I'm guessing it'll be more RAM efficient and faster but I've been reading that visibility of nodes in a treeview is hard to edit. plz check that out

bintiibrahim: i figured out a way to do that, but it does add more code so maybe it's slower. each ModForTreeView would have a "Show" property. and because we'll be toggling between hide/show, we'll have to reset the collection if textbox is empty/null. what do u think?

rmillikin: give it a shot and see if it's faster

bintiibrahim: I do like this new approach better.

bintiibrahim: feels about the same but can't really tell. the thing is the each mod type's children nodes are binded to a collection, so for this to work i'd still have to have create another collection for display purposes

bintiibrahim: something like this:

public ObservableCollection<ModForTreeView> Children { get; }
```

        public ObservableCollection<ModForTreeView> DisplayChildren
        {
            get
            {
                var children = new ObservableCollection<ModForTreeView>();
                foreach (var child in Children)
                {
                    if (child.Show)
                    {
                        children.Add(child);
                    }
                }
                return children;
            }
        }
```

PR #1494, opened 2018-10-24
Fix nses gui

zrolfs:
-Fixed a GUI bug that caused non-specific to be treated like semi if the task was reopened
-Converted a parellelized .order() into an unparellelized .order(), which _decreased_ computation time from 6s to 3s.


GUI/CalibrateTaskWindow.xaml.cs
```diff
@@ -125,7 +125,12 @@ private void PopulateChoices()
                 proteaseComboBox.Items.Add(protease);
             }
 
-            proteaseComboBox.SelectedIndex = 12;
+            proteaseComboBox.SelectedIndex = 1;
```
rmillikin: I'm not really a fan of this either... can we just do `proteaseComboBox.SelectedItem = "trypsin";` ? any issues with that?

Issue #1419, opened 2018-09-26
Output database indexes in a single folder


zrolfs:
For Modern search, a unique folder is produced for each partition of the database. It would be cleaner if these were all put in a single folder. Additionally, a more descriptive name than just the date/time would be nice so that it's readily apparent that the index folder is not the results folder.

rmillikin:
done in #1500

PR #1500, opened 2018-10-26
indexes output folder

bintiibrahim:
#1419

### DIFF
--- a/TaskLayer/MetaMorpheusTask.cs
+++ b/TaskLayer/MetaMorpheusTask.cs
@@ -85,6 +85,8 @@ namespace TaskLayer
 
         public CommonParameters CommonParameters { get; set; }
 
+        public const string IndexFolderName = "DatabaseIndex";
+
         public static IEnumerable<Ms2ScanWithSpecificMass> GetMs2Scans(MsDataFile myMSDataFile, string fullFilePath, CommonParameters commonParameters)
         {
             var ms2Scans = myMSDataFile.GetAllScansList().Where(x => x.MsnOrder > 1).ToArray();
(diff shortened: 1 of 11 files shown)


TaskLayer/MetaMorpheusTask.cs
```diff
@@ -530,22 +530,42 @@ private static string GetExistingFolderWithIndices(IndexingEngine indexEngine, L
...
-                // Look at every subdirectory...
+                // Look at every subdirectory to find indexes folder...
                 foreach (DirectoryInfo possibleFolder in directories)
                 {
-                    if (File.Exists(Path.Combine(possibleFolder.FullName, "indexEngine.params")) &&
-                        File.Exists(Path.Combine(possibleFolder.FullName, "peptideIndex.ind")) &&
-                        File.Exists(Path.Combine(possibleFolder.FullName, "fragmentIndex.ind")) &&
-                        (File.Exists(Path.Combine(possibleFolder.FullName, "precursorIndex.ind")) || !indexEngine.GeneratePrecursorIndex) &&
-                        SameSettings(Path.Combine(possibleFolder.FullName, "indexEngine.params"), indexEngine))
+                    string pathToIndexes = Path.Combine(possibleFolder.FullName, "Indexes");
```
rmillikin: We can just create a single folder in the same folder containing the protein database and just call it `[DatabaseName]_Indices_[$DATETIME]`, no need to go looking for it in subdirectories

rmillikin: actually, I guess you still have to check for indexEngine.params...

PR #1507, opened 2018-10-30
Allow SNES to look at notches (round 2)

zrolfs:


zrolfs:
Hmm. The search took 60% longer than before, but got 3% more PSMs, so it's now better than the classic?? I'm pretty confused here.

zrolfs:
Oh, the indexes being in a new folder made it remake the indexes. So it's not actually 60% slower! (But there are still more PSMs!).

acesnik:
Can N64 look at notches, yet?

zrolfs:
It was an intentional choice :)

zrolfs:
This is ready!

EngineLayer/ModernSearch/ModernSearchEngine.cs
```diff
@@ -70,7 +70,7 @@ protected override MetaMorpheusEngineResults RunSpecific()
                     // note that this is the OPPOSITE of the classic search (which calculates experimental masses from theoretical values)
                     // this is just PRELIMINARY precursor-mass filtering
                     // additional checks are made later to ensure that the theoretical precursor mass is acceptable
-                    var notches = MassDiffAcceptor.GetAllowedPrecursorMassIntervals(scan.PrecursorMass);
+                    var notches = MassDiffAcceptor.GetAllowedPrecursorMassIntervalsFromTheoreticalMass(scan.PrecursorMass);
```
rmillikin: does this need to be `GetAllowedPrecursorMassIntervalsFromObservedMass`?

zrolfs: Yes, it should be. Good catch.
Curious that this wasn't caught in the modern/classic comparison before. Perhaps no missed mono in that test?

zrolfs: Ahh, The "Observed" code was just happening below it. I've simplified it now.

PR #1514, opened 2018-11-09
re-add peptidomics protease, fix some issues with custom mod window

rmillikin:


acesnik:
did you add that protease to mzLib, too?

rmillikin:
no; I'm not sure why you were having issues with multiple proteases.tsv files but the one in ProteolyticDigestion works fine for me

PR #1521, opened 2018-11-13
MetaDraw for crosslinked peptide annotation

lonelu:
TsvResultReader can be less restricted in the future.


GUI/MetaDraw.xaml.cs
```diff
@@ -155,7 +156,7 @@ private void DrawPsm(int oneBasedScanNumber, string fullSequence = null)
             mainViewModel.DrawPeptideSpectralMatch(msDataScanToDraw, psmToDraw);
 
             // draw annotated base sequence
-            DrawAnnotatedBaseSequence(psmToDraw);
```
rmillikin: is this available for XL yet? maybe check to see if the PSM is a crosslink, and draw it if it isn't?

rmillikin:
looks very nice 👍 just need that one change to the base sequence annotation

GUI/MetaDraw.xaml.cs
lonelu: Ready to merge.
The DrawAnnotatedBaseSequence will be changed in the future for glyco or crosslink peptide.

rmillikin:
GitHub is giving me a weird error trying to update the branch from master... I'll try again later today/tmw. PR looks good 👍

Issue #1473, opened 2018-10-15
Custom fragmentation/ion types


rmillikin:
Before the revamp, we let users select between b, y, z, and c ions. We need to re-enable this.

rmillikin:
done in #1523

PR #1523, opened 2018-11-16
custom fragmentation

bintiibrahim:
issue #1473


GUI/CustomFragmentationWindow.xaml.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+        }
+
+        public CustomFragmentationWindow(List<ProductType> list)
+        {
+            InitializeComponent();
+            // update checkboxes
+            if (list != null)
+            {
+                bCheckBox.IsChecked = list.Contains(ProductType.b);
```
rmillikin: this PR works but I think it can be more concise and a bit more flexible:
1. Instead of reading checkboxes to a list, then storing the list to a toml, then reading the toml to another list, can you just directly go from checkboxes -> add the `ProductType`s to the `DissociationType.ProductsFromDissociationType[Custom]` list? then in the common params constructor, you can just set `CustomIons = DissociationType.ProductsFromDissociationType[Custom]` without needing to worry about modifying the constructor parameters, reading/writing .tomls, etc.
2. Instead of hard-coding checkboxes for b/y/c/z ions, I'd like a dynamic list of checkboxes that expands to include all known product types. `DissociationTypeCollection.ProductsFromDissociationType.Values.SelectMany(p => p).Distinct()` will get you all known product types.
3. Need a unit test to ensure custom product types are handled correctly between tasks. So if task1 uses b/y ions, and task2 uses c/z ions, and task3 uses y/c/z ions, make sure these are read and searched correctly from the .tomls

bintiibrahim: ok, working on it

rmillikin:
this is looking really good but I did notice something sort of weird - when I open metamorpheus in visual studio, add a search task, then close MM, visual studio doesn't stop debugging. like there's some process that's not killed still running in the background. this behavior doesn't happen if you don't open the search task window (i didn't check other tasks). also doesn't happen with the current master branch code. can you look into that?

bintiibrahim:
> this is looking really good but I did notice something sort of weird - when I open metamorpheus in visual studio, add a search task, then close MM, visual studio doesn't stop debugging. like there's some process that's not killed still running in the background. this behavior doesn't happen if you don't open the search task window (i didn't check other tasks). also doesn't happen with the current master branch code. can you look into that?

yeap i'll look into that! i think i have an idea what's happening

bintiibrahim:
> this is looking really good but I did notice something sort of weird - when I open metamorpheus in visual studio, add a search task, then close MM, visual studio doesn't stop debugging. like there's some process that's not killed still running in the background. this behavior doesn't happen if you don't open the search task window (i didn't check other tasks). also doesn't happen with the current master branch code. can you look into that?

should be ok now. what happened was the custom fragmentation window was hidden and not properly closed

GUI/CalibrateTaskWindow.xaml.cs
```diff
@@ -175,6 +175,7 @@ private void PopulateChoices()
         private void CancelButton_Click(object sender, RoutedEventArgs e)
         {
             DialogResult = false;
+            CustomFragmentationWindow.Close();
```
bintiibrahim: totally missed this, sorry

PR #1531, opened 2018-11-26
pdf export

bintiibrahim:
#1433


rmillikin:
this looks really good 👍 nice clean looking pdfs, quick export, small file size. can we export the canvas at the bottom (the base sequence annotation) with it?

rmillikin:
and/or make the base sequence annotation into an oxyplot thing and not a canvas thing. then add it to the pdf.

bintiibrahim:
> and/or make the base sequence annotation into an oxyplot thing and not a canvas thing. then add it to the pdf.

i'll give it a try. i think i'd have to customize the OxyPlot.Pdf source code for that to work because there aren't any built-in options for combining 2 oxyplots. what do you think?

PR #1535, opened 2018-11-30
Peptide FDR analysis

acesnik:


rmillikin:
also line ~460

acesnik:
Line 165 of FdrAnalysisEngine modifies the Q-values for the PSM objects. That would mess with the qvalues that are used in subsequent analysis, e.g. parsimony.

rmillikin:
ah I see. ok. needs a unit test

acesnik:
Would you prefer calculating PSM FDR before and after the results output? That'd be a simpler code change at the cost of doing FDR analysis twice.

acesnik:
It might break nonspecific, though, since Zach's doing FDR analysis separately for that.

rmillikin:
@zrolfs  preference?

zrolfs:
I always favor run speed over code management, so I'd prefer not doing FDR twice. That should prevent nonspecific issues, too. Sorry that it's more work, though...

acesnik:
Less work and preferred for me, too. Thanks, Zach.

acesnik:
now to find a search test to hijack

acesnik:
Test added. It fails if the peptide FDR isn't calculated for AllPeptides.psmtsv.

Test/SearchTaskTest.cs
```diff
@@ -241,5 +242,53 @@ public static void PrunedDbWithContaminantsTest()
             Assert.That(File.ReadAllLines(Path.Combine(folderPath, @"DbForPrunedDbPruned.xml")).Length > 0);
             Directory.Delete(folderPath, true);
         }
+
+        /// <summary>
+        /// Test ensures pruned databases are written when contaminant DB is searched
```
rmillikin: ?

Test/SearchTaskTest.cs
```diff
@@ -241,5 +242,53 @@ public static void PrunedDbWithContaminantsTest()
...
+        /// </summary>
+        [Test]
+        public static void PeptideFDRTest()
+        {
+            SearchTask searchTask = new SearchTask()
+            {
+                SearchParameters = new SearchParameters
+                {
+                    WritePrunedDatabase = true
```
rmillikin: can just use default params; a bit confusing to specify thing you aren't testing

Test/SearchTaskTest.cs
```diff
@@ -241,5 +242,53 @@ public static void PrunedDbWithContaminantsTest()
...
+                },
+            };
+
+            string myFile = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\PrunedDbSpectra.mzml");
+            string myFile2 = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\PrunedDbSpectra2.mzml");
+            if (!File.Exists(myFile2)) { File.Copy(myFile, myFile2); }
+            string myDatabase = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\DbForPrunedDb.fasta");
+            string folderPath = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestPeptideFDR");
+            string filePath = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\ExperimentalDesign.tsv");
```
rmillikin: don't need exp design

Test/SearchTaskTest.cs
```diff
@@ -241,5 +242,53 @@ public static void PrunedDbWithContaminantsTest()
...
+            foreach (string line in File.ReadAllLines(Path.Combine(folderPath, @"AllPeptides.psmtsv")))
+            {
+                string[] lineline = line.Split('\t');
+                if (line.StartsWith("File Name")) // header
+                {
+                    columns = lineline.ToList();
+                }
+                else if (lineline[columns.IndexOf("Decoy/Contaminant/Target")] == "D")
+                {
+                    Assert.AreEqual(++cumDecoys, int.Parse(lineline[columns.IndexOf("Cumulative Decoy")]));
```
rmillikin: I don't really understand this test - you're checking to see if the number of targets/decoys equals the number of targets/decoys if you start counting from the top. isn't that true currently too, even with the buggy FDR? 

I think it would be better to have 3 PSMs, two targets of the same sequence and then the third being a decoy. the q-value of the decoy should be 0.5 if the peptide FDR is calculated incorrectly, and 1.0 if it's calculated correctly, right? q=(targets / decoys)

acesnik: No, it's not true. The target/decoy counts are from PSM FDR analysis. In this test, there are 2 PSMs for every peptide, so with the current code the peptide list goes 1,3,5,7. With these fixes, it goes 1,2,3,4, as expected.

Test/SearchTaskTest.cs
acesnik: updated

Test/SearchTaskTest.cs
acesnik: done

Test/SearchTaskTest.cs
acesnik: deleted

Test/SearchTaskTest.cs
acesnik: I think the Q-value test is reasonable, slightly simpler, and achieves the same result.

On the other hand, this one addresses the counts that are used to calculate the q-value, which were why it became obvious that peptide FDR was broken.

rmillikin: ah i see. ok. I do think it's good to have an assertion there that checks the q-value is computed correctly, since that's the measurement we actually care about. can you add that assertion?

acesnik: Sure, I added an assertion for the final q-value. Intermediate ones have assumptions that make them not always (decoy / target)

Issue #752, opened 2017-11-03
Add PSM count column to unique peptides TSV file


trishorts:

trishorts:
yes. I would love to see this feature added still......

trishorts:
I'd like to promote the urgency of this functionality. I look for it routinely to get a sense for how likely a particular peptide assignment is correct. Do we see it once or a hundred times?

trishorts:
q<=0.01 I think.

rmillikin:
done in #1536

PR #1536, opened 2018-11-30
Psm Count

bintiibrahim:
#752


bintiibrahim:
should i write a unit test for this?

PR #1550, opened 2018-12-19
Assign tasks to threads more equitably.

zrolfs:


zrolfs:
There's about a 10% increase in modern search speed for tryptic HeLa QE-HF

zrolfs:
It felt a little silly doing this for classic and indexing engines, since you don't expect proteins to be sorted by length. However, whenever users attach "custom" peptides or proteins to a database (hello peptidomics community), this is a substantial improvement, because otherwise some threads have only peptides while others are stuck with big proteins.

acesnik:
Good to hear. Thanks for checking.

PR #1551, opened 2018-12-19
Diff frag type

lonelu:
This pull request is for searching MS2-MS2 or MS2-MS3 type of data, which is widely used in crosslinked and glycopeptides.


rmillikin:
I think a more flexible experimental design would be nice. We could have 3 different experiment types: MS2, MS2/MS2, MS2/MS3. For each fragmentation type, you could choose a dissociation type. That way it would be simple to select MS2-HCD/MS3-LowCID, or whatever combination you want, rather than having certain combinations pre-loaded.

I like the idea of parent/child scans. I think we need to best determine how to report child scans (MS3, MS2 with different fragmentation, etc). If the scores are simply added to the parent scan, this affects FDR. Do we want to do this?

lonelu:
Good idea.

Issue #1556, opened 2019-01-22
MetaDraw - export base sequence annotation to PDF along with spectrum annotation


rmillikin:

rmillikin:
done in #1573

PR #1573, opened 2019-02-19
base seq annotation

bintiibrahim:
#1556 
#1583


rmillikin:
the resolution is kind of low, and I think the main drawing area is getting a little crowded. I think it would be better if we could 1) use some kind of vector format for drawing this rather than exporting as an image and 2) putting the base sequence annotation underneath the spectrum annotation. This seems like it may not be possible, though - does the current pdf exporter allow this?

PR #1574, opened 2019-02-20
Stable isotope labeling quantification in MetaMorpheus

zrolfs:


TaskLayer/SearchTask/SearchParameters.cs
```diff
@@ -79,5 +80,6 @@ public SearchParameters()
         public bool WritePepXml { get; set; }
         public bool WriteDecoys { get; set; }
         public bool WriteContaminants { get; set; }
+        public List<SilacLabel> SilacLabels { get; set; }
```
rmillikin: check this out - does this belong here?

Issue #1578, opened 2019-02-27
XL issue: By default DSSO is not cleavable


mriffle:
I don't know this is intentional or not, but the default settings for DSSO are not as a cleavable cross-linker as of the latest version of MetaMorpheus. Users need to set up User Defined cross-linker with DSSO as a cleavable linker.

trishorts:
That's strange. Which version of MetaMorpheus are you using. I looked at the code I have downloaded and it is specified as cleavable there.  How are you noticing this problem on your end? 
![image](https://user-images.githubusercontent.com/16841846/53518249-3deda280-3a96-11e9-9366-37ed9f703f67.png)

mriffle:
Here is the appropriate section of the toml file gets created when using the built-in DSSO linker. Using 0.0.297

[XlSearchParameters]
DecoyType = "Reverse"
CrosslinkerType = "DSSO"
CrosslinkSearchTopNum = 300
CrosslinkerResidues = "K"
CrosslinkerResidues2 = "K"
IsCleavable = false
RestrictToTopNHits = true
WriteOutputForPercolator = false
WritePepXml = true
XlQuench_H2O = true
XlQuench_Tris = true
XlQuench_NH2 = false

trishorts:
ok. I see. we'll get right on it.

rmillikin:
This seems to be an issue writing the .toml, not with the internal definition of DSSO. I performed the search and got the expected number of CSMs, but the .toml did indeed read "Cleavable = false". I dragged this .toml into MetaMorpheus and ran the search, and got the same result as before (correct CSMs, incorrect .toml written).

I'll investigate further, but this doesn't appear to be a major problem in terms of the actual search. Just something funky going on with writing the .toml.

mriffle:
Ah cool, thanks

lonelu:
Currently, the "IsCleavable = false" is only for User-Defined crosslinker. This is confusing. We plan to make change of it #1398. Again, thanks for pointing this out.

lonelu:
fixed

Issue #1398, opened 2018-09-06
Store user-defined crosslinkers in a text file


lonelu:
Make adding user-defined crosslinker like adding a new protease.

rmillikin:
Reading/writing to/from a user-editable .tsv file

bintiibrahim:
do we still want to add this feature?

rmillikin:
definitely

rmillikin:
done in #1584

Issue #1339, opened 2018-08-13
zero length crosslinker configuration / adding


talpa25:
Dear all,

I've tried to configue MM for  zero-length Crosslinkers like DMTMM. The make a covalent bound between K and DE with loss of water. 
But unfortunately MM cannot take a minus in the crosslinker total mass box.

Would you mind providing that feature?

Best
Roman

lonelu:
Definitely! I need to make sure that the fragments are the correct masses with water loss on the correct position. I'll look into this - hopefully in a few weeks we will have this ready.

lonelu:
Hi, the next release of MM will be able to take a minus in the crosslinker total mass box. As you mentioned the mass will be a loss of water. The H2O-dead-end mass will be considered to have a proper parameter setting. 
I am willing to code your crosslinkers into MM if you like it (just need the user-defined crosslinker information).

rmillikin:
done in #1355

rmillikin:
this feature is in v288!

talpa25:
Dear https://github.com/lonelu

thank you very much for your offer. Below you'll find some crosslinkers that are commonly used in the field.

DMTMM crosslinks  K or N-terminus with DE (or C-terminus) by mass change of -18.011 monomass: 140.046 

BMOE crosslinks C with C by mass change of 220.048 monomass: 220.115

SDA cleavable crosslinks KSTY or N-terminus with DE (or C-terminus) by mass change of 82.042 (monomass of 125.059) and a cleaved long fragment mass add of 100.052 and short of 82.042

DHSO cleavable crosslinks DE and C-terminus with DE and C-terminus by mass change of 222.063 (monomass 221.071) and a cleaved long fragment mass add of 100.01 and short of 68.037

CDI cleavable crosslinks K and N-terminus with KSTY and N-terminus by mass change of 25.979 (monomass of 43.006) and a cleaved long fragment mass of 25.97926454 and a short mass of 0.

PDH crosslinks DE and C-terminus with DE and C-terminus by mass change of 152.106 (monomass of 170.117)

ADH crosslinks DE and C-terminus with DE and C-terminus by mass change of  138.09055 (monomass of 156.10111)

If you have question don't hesitate to contact me.

Best

Roman

rmillikin:
[...]

PR #1584, opened 2019-03-06
Read crosslinker file

lonelu:
solve #1578, #1398, #1339


EngineLayer/Data/Crosslinkers.tsv
```diff
@@ -0,0 +1,5 @@
+Name	CrosslinkAminoAcid	CrosslinkerAminoAcid2	Cleavable	CrosslinkerTotalMass	CrosslinkerShortMass	CrosslinkerLongMass	QuenchMassH2O	QuenchMassNH2	QuenchMassTris
+DSSO	K	K	T	158.0038	54.01056	85.982635	176.0143	175.0303	279.0777
+DSS	K	K	F	138.06808	0	0	156.0786	155.0946	259.142
+DSUB	K	K	T	196.0848	85.05276	111.032	214.0954	213.1113	317.1587
+DisulfideBond	C	C	T	-2.01565	33.98772	31.97207	0	0	0
```
rmillikin: are the short/long masses of the disulfide bond correct? Shouldn't it be -1 Da for both? I'm assuming it cleaves the S-S bond, but does it actually cleave the S-C bond?

lonelu: The truth is that it is dissociate-type related. the short one is wrong. it should be -33.98772

PR #1588, opened 2019-03-08
use num windows

trishorts:


GUI/Util/GlobalGuiSettings.cs
```diff
@@ -112,19 +116,46 @@ public static bool CheckTopNPeaks(string text)
...
                 MessageBox.Show("The minimum intensity ratio must be between zero and one. \n You entered " + '"' + text + '"');
                 return false;
             }
             return true;
         }
 
-        public static bool CheckPrecursorMassTolerance(string text)
+        public static bool CheckWindowWidthThomsons(string text)
         {
+            if (double.TryParse(text, NumberStyles.Any, CultureInfo.InvariantCulture, out double windowWidthThomsons))
```
zrolfs: refactor with "!"to make this an if-else, instead of the current if-else if-else

GUI/Util/GlobalGuiSettings.cs
```diff
@@ -112,19 +116,46 @@ public static bool CheckTopNPeaks(string text)
...
+            }
+            else if (windowWidthThomsons < 0 || !string.IsNullOrWhiteSpace(text))
+            {
+                MessageBox.Show("The value for window width in Thomsons must be greater than one. \n You entered " + '"' + text + '"');
+                return false;
+            }
+            return true;
+        }
 
+        public static bool CheckWindowMaxNormalizationValue(string text)
```
zrolfs: same here

GUI/SearchTaskWindow.xaml
```diff
@@ -88,21 +88,45 @@
...
+                                        <Label Content="Number of windows" />
+                                        <TextBox x:Name="NumberOfWindowsTextBox" PreviewTextInput="CheckIfNumber" Width="45" ToolTipService.ShowDuration="999999" ToolTipService.InitialShowDelay="500">
+                                            <ToolTipService.ToolTip>
+                                                <ToolTip Content="Divide the MS2 spectrum into an integer number of evenly sized windows before filtering peaks" />
+                                            </ToolTipService.ToolTip>
+                                        </TextBox>
+                                    </StackPanel>
+                                    <StackPanel Orientation="Horizontal" Margin="5">
+                                        <Label Content="Peak intensity maximum normalization value" />
+                                        <TextBox x:Name="WindowMaxNormalizationValueTextBox" PreviewTextInput="CheckIfNumber" Width="45" ToolTipService.ShowDuration="999999" ToolTipService.InitialShowDelay="500">
```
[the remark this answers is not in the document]
trishorts: changed to bool in mzlib. once that goes through, this will be resolved

GUI/Util/GlobalGuiSettings.cs
trishorts: deleted

GUI/Util/GlobalGuiSettings.cs
trishorts: done

GUI/SearchTaskWindow.xaml.cs
```diff
@@ -473,11 +463,6 @@ private void SaveButton_Click(object sender, RoutedEventArgs e)
                 {
                     windowWidthThompsons = windowWidth;
```
zrolfs: windowWidthThompsons =WindowWidthThomsonsTextBox.Text.Length=0 ? null : Convert.ToInt(WindowWidthThomsonsTextBox.Text);

Test/TestData/AllPSMs.psmtsv
```diff
@@ -0,0 +1,126 @@
+File Name	Scan Number	Scan Retention Time	Num Experimental Peaks	Total Ion Current	Precursor Scan Number	Precursor Charge	Precursor MZ	Precursor Mass	Score	Delta Score	Notch	Base Sequence	Full Sequence	Essential Sequence	PSM Count	Mods	Mods Chemical Formulas	Mods Combined Chemical Formula	Num Variable Mods	Missed Cleavages	Peptide Monoisotopic Mass	Mass Diff (Da)	Mass Diff (ppm)	Protein Accession	Protein Name	Gene Name	Organism Name	Intersecting Sequence Variations	Identified Sequence Variations	Splice Sites	Contaminant	Decoy	Peptide Description	Start and End Residues In Protein	Previous Amino Acid	Next Amino Acid	All Scores	Theoreticals Searched	Decoy/Contaminant/Target	Matched Ion Series	Matched Ion Mass-To-Charge Ratios	Matched Ion Mass Diff (Da)	Matched Ion Mass Diff (Ppm)	Matched Ion Intensities	Matched Ion Counts	Localized Scores	Improvement Possible	Cumulative Target	Cumulative Decoy	QValue	Cumulative Target Notch	Cumulative Decoy Notch	QValue Notch	eValue	eScore
```
zrolfs: Can you rename this file so that we know which test it's for?

Test/XLTest.cs
```diff
@@ -663,29 +663,11 @@ public static void TestMixedMs2Ms2()
             Assert.That(csm.ChildMatchedFragmentIons.First().Value.Count == 17);
             Assert.That(csm.BetaPeptide.ChildMatchedFragmentIons.First().Key == 3);
             Assert.That(csm.BetaPeptide.ChildMatchedFragmentIons.First().Value.Count == 12);
-
-            // write results to TSV
```
zrolfs: Not sure if it's okay that these were deleted...

trishorts:
> The file "Test/DatabaseTests/LowResSnip_B6_mouse_11700_117500.xml" is massive at 35K lines. Can you reduce the size of this file at all while maintaining the fidelity of the test?

I removed 20k lines from the Db and deleted an unused file. I reverted changes to the XL test (lines mistakenly deleted). Changed the Allpsms filename to one more appropriate for the test.

rmillikin:
- When would a user want to enable these settings?
- What is the effect of this type of filtering on the data? If you pick 10 windows+20 peaks per window vs. 200 peaks+1 window, how does this affect the results?
- Same with normalizing peak intensities across the windows - what does this do to the results?
- Is this supposed to be combined with XCorr? Do we have a checkbox to use XCorr scoring? Is that the next step for this?

trishorts:
> * When would a user want to enable these settings?
> * What is the effect of this type of filtering on the data? If you pick 10 windows+20 peaks per window vs. 200 peaks+1 window, how does this affect the results?
> * Same with normalizing peak intensities across the windows - what does this do to the results?
> * Is this supposed to be combined with XCorr? Do we have a checkbox to use XCorr scoring? Is that the next step for this?

I'm not sure of the effect yet. I have always felt that giving the user the choice to filter was good but that just picking a random number of peaks was too crude. The number of peaks in a spectrum can vary lots between scans and the most intense peaks are usually clustered into a smaller window. I simply prefer to allow the user to have more control over the peak filtering so that it could be spread out evenly across the spectrum. 

I guess that there is a great deal of variation between instruments orbi v. iontrap / fusion v. QE and that one filter set-up might not translate. 

As for xcorr, i believe that we are far from optimized. Maybe we're only at 50% of what's possible. Normalization has a surprising effect. I'd like to explore that at some point, and having ready access to the parameters that effect that  (windows count/peak counts/normalization intensity) will make it easier.

I can see how having lots of accessible parameters might be confusing. So, I'm okay with hiding these and others. Ultimately I care less about where they are than that they exist

EngineLayer/CommonParameters.cs
```diff
@@ -86,8 +89,11 @@ public CommonParameters()
         public double ScoreCutoff { get; private set; }
         public DigestionParams DigestionParams { get; private set; }
         public bool ReportAllAmbiguity { get; private set; }
-        public int TopNpeaks { get; private set; }
-        public double MinRatio { get; private set; }
+        public int? NumberOfPeaksToKeepPerWindow { get; private set; }
+        public double? MinimumAllowedIntensityRatioToBasePeak { get; private set; }
+        public double? WindowWidthThomsons { get; private set; }
+        public int? NumberOfWindows { get; private set; }
+        public bool NormalizePeaksAccrossAllWindows { get; private set; }
```
rmillikin: I'm also worried about these being nullable - but I think they all have non-null "no filter" values that make sense.
```
NumberOfPeaksToKeepPerWindow = Infinity
MinimumAllowedIntensityRatioToBasePeak = 0
WindowWidthThomsons = Infinity
NumberOfWindows = 1
```

Does the .toml read/write still work with these values? And does the peak filtering (or lack of it, I guess) still work? And does it make sense in the context of peak filtering settings?

rmillikin: I also don't really understand why there is both `WindowWidthThomsons` and `NumberOfWindows`; they seem to conflict with each other. Can you explain why there are two different settings that effectively set the width of the window? I might have asked that already but I forget the answer...

GUI/MetaDraw.xaml.cs
```diff
@@ -383,7 +383,7 @@ private async void loadFilesButton_Click(object sender, RoutedEventArgs e)
             prgsFeed.IsOpen = true;
             prgsText.Content = "Loading spectra file...";
 
-            var slowProcess = Task<MsDataFile>.Factory.StartNew(() => spectraFileManager.LoadFile(spectraFilePath, null, null, false, false, new CommonParameters()));
+            var slowProcess = Task<MsDataFile>.Factory.StartNew(() => spectraFileManager.LoadFile(spectraFilePath, new CommonParameters()));
```
rmillikin: I think this changes the filtering. The original `LoadFile` method call used no peak filtering (`null, null, false, false`). Now it uses default `CommonParameters `settings, which include peak filtering (top 200 peaks, etc.)

TaskLayer/MyFileManager.cs
```diff
@@ -72,9 +72,10 @@ public static ThermoMsFileReaderVersionCheck ValidateThermoMsFileReaderVersion()
             return ThermoMsFileReaderVersionCheck.DllsNotFound;
         }
 
-        public MsDataFile LoadFile(string origDataFile, int? topNpeaks, double? minRatio, bool trimMs1Peaks, bool trimMsMsPeaks, CommonParameters commonParameters)
```
rmillikin: This is a good change

PR #1644, opened 2019-04-29
compute psm-level p-values

trishorts:


trishorts:
Consider this an alpha-request. I'd like to spend whatever time next week is necessary to get this up to snuff. Let me know what needs improvement.

EngineLayer/FdrAnalysis/FdrAnalysisResults.cs
```diff
@@ -13,13 +13,15 @@ public FdrAnalysisResults(FdrAnalysisEngine s, string analysisType) : base(s)
         public int PsmsWithin1PercentFdr { get; set; }
         public bool DeltaScoreImprovement { get; set; }
         private string AnalysisType { get; set; }
+        public string BinarySearchTreeMetrics { get; set; }
```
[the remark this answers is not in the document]
trishorts: At the time when the ~10% of the data gets chosen for training, another 10% gets chosen for evaluation. Then after training, the effectiveness of the model gets evaluated on the test set. The results of that evaluation are converted to text values called BinarySearchTreeMetrics and this gets written to the results.tsv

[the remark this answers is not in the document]
rmillikin: plz put that in a comment

EngineLayer/EngineLayer.csproj
```diff
@@ -7,18 +7,13 @@
...
-    <PackageReference Include="SharpLearning.Common.Interfaces" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.Containers" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.CrossValidation" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.DecisionTrees" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.GradientBoost" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.InputOutput" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.Metrics" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.Optimization" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.RandomForest" Version="0.26.9" />
+    <PackageReference Include="SharpLearning.Optimization" Version="0.31.1" />
```
rmillikin: packages

EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
```diff
@@ -48,47 +45,16 @@ private void DoFalseDiscoveryRateAnalysis(FdrAnalysisResults myAnalysisResults)
...
-
-                //Calculate delta scores for the psms (regardless of if we are using them)
-                foreach (PeptideSpectralMatch psm in psms)
-                {
-                    if (psm != null)
-                    {
-                        psm.CalculateDeltaScore(ScoreCutoff);
-                    }
-                }
+                ////Calculate delta scores for the psms (regardless of if we are using them)
```
rmillikin: @zrolfs need to talk about this code, we changed some delta score stuff

EngineLayer/FdrAnalysis/PValueAnalysisGeneric.cs
```diff
@@ -0,0 +1,340 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using static Microsoft.ML.DataOperationsCatalog;
+
+namespace EngineLayer
+{
+    public static class PValueAnalysisGeneric
```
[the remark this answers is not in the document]
rmillikin: PValue is still used in a lot of places

EngineLayer/PsmTsv/PsmTsvHeader.cs
```diff
@@ -61,8 +60,9 @@ public static class PsmTsvHeader
         public const string CumulativeDecoyNotch = "Cumulative Decoy Notch";
         public const string QValue = "QValue";
         public const string QValueNotch = "QValue Notch";
-        public const string EValue = "eValue";
-        public const string EScore = "eScore";
+        public const string LongestSeries = "longestSeries";
```
rmillikin: "longestSeries" -> "Longest Ion Series"

EngineLayer/PsmTsv/PsmTsvWriter.cs
```diff
@@ -411,8 +403,8 @@ internal static void AddMatchScoreData(Dictionary<string, string> s, PeptideSpec
             string cumulativeTargetNotch = " ";
             string cumulativeDecoyNotch = " ";
             string qValueNotch = " ";
-            string eValue = " ";
-            string eScore = " ";
+            string length = " ";
```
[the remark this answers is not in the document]
rmillikin: still need to change this

GUI/GUI.csproj
```diff
@@ -74,103 +74,31 @@
     <Prefer32Bit>true</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
```
rmillikin: packages

GUI/packages.config
```diff
@@ -3,19 +3,16 @@
...
-  <package id="SharpLearning.InputOutput" version="0.26.9" targetFramework="net471" />
-  <package id="SharpLearning.Metrics" version="0.26.9" targetFramework="net471" />
-  <package id="SharpLearning.Optimization" version="0.26.9" targetFramework="net471" />
-  <package id="SharpLearning.RandomForest" version="0.26.9" targetFramework="net471" />
+  <package id="SharpLearning.Common.Interfaces" version="0.31.1" targetFramework="net471" />
+  <package id="SharpLearning.Containers" version="0.31.1" targetFramework="net471" />
+  <package id="SharpLearning.DecisionTrees" version="0.31.1" targetFramework="net471" />
+  <package id="SharpLearning.InputOutput" version="0.31.1" targetFramework="net471" />
+  <package id="SharpLearning.Optimization" version="0.31.1" targetFramework="net471" />
+  <package id="SharpLearning.RandomForest" version="0.31.1" targetFramework="net471" />
```
rmillikin: packages

TaskLayer/TaskLayer.csproj
```diff
@@ -7,18 +7,10 @@
...
-    <PackageReference Include="SharpLearning.Common.Interfaces" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.Containers" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.CrossValidation" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.DecisionTrees" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.GradientBoost" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.InputOutput" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.Metrics" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.Optimization" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.RandomForest" Version="0.26.9" />
+    <PackageReference Include="SharpLearning.Optimization" Version="0.31.1" />
```
rmillikin: packages

Test/Test.csproj
```diff
@@ -7,19 +7,10 @@
...
-    <PackageReference Include="SharpLearning.Common.Interfaces" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.Containers" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.CrossValidation" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.DecisionTrees" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.GradientBoost" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.InputOutput" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.Metrics" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.Optimization" Version="0.26.9" />
-    <PackageReference Include="SharpLearning.RandomForest" Version="0.26.9" />
+    <PackageReference Include="SharpLearning.Optimization" Version="0.31.1" />
```
rmillikin: packages

EngineLayer/FdrAnalysis/PValueAnalysisGeneric.cs
```diff
@@ -0,0 +1,348 @@
...
+    {
+        public static string ComputePValuesForAllPSMsGeneric(List<PeptideSpectralMatch> psms)
+        {
+            Dictionary<string, int> accessionAppearances = GetAccessionCounts(psms);
+            Dictionary<string, int> sequenceToPsmCount = GetSequenceToPSMCount(psms);
+
+            MLContext mlContext = new MLContext();
+            IDataView dataView = mlContext.Data.LoadFromEnumerable(CreatePsmData(psms, accessionAppearances, sequenceToPsmCount));
+
+            TrainTestData trainTestSplit = mlContext.Data.TrainTestSplit(dataView, testFraction: 0.1);
```
rmillikin: make training fraction of 0.1 explicit

EngineLayer/FdrAnalysis/FdrAnalysisResults.cs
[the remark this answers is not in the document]
trishorts: done

EngineLayer/FdrAnalysis/PValueAnalysisGeneric.cs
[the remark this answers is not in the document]
trishorts: done

EngineLayer/PsmTsv/PsmTsvWriter.cs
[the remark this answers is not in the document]
trishorts: done

EngineLayer/EngineLayer.csproj
trishorts: should be fixed

trishorts: should be done

EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
trishorts: requested zach to look at it.

EngineLayer/PsmTsv/PsmTsvHeader.cs
trishorts: done

GUI/GUI.csproj
trishorts: should be okay

GUI/packages.config
trishorts: same

EngineLayer/FdrAnalysis/PValueAnalysisGeneric.cs
trishorts: explained in comment

EngineLayer/FdrAnalysis/FdrAnalysisResults.cs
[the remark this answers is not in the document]
rmillikin: I don't see the comment

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -0,0 +1,375 @@
...
+            //
+            //   samplingKeyColumnName:
+            //     Name of a column to use for grouping rows. If two examples share the same value
+            //     of the samplingKeyColumnName, they are guaranteed to appear in the same subset
+            //     (train or test). This can be used to ensure no label leakage from the train to
+            //     the test set. If null no row grouping will be performed.
+            //
+            //   seed:
+            //     Seed for the random number generator used to select rows for the train-test split.
+            TrainTestData trainTestSplit = mlContext.Data.TrainTestSplit(dataView, testFraction: 0.1);
```
rmillikin: make training fraction (0.1?) explicit

EngineLayer/PeptideSpectralMatch.cs
```diff
@@ -126,12 +127,18 @@ public void AddOrReplace(PeptideWithSetModifications pwsm, double newScore, int
...
                 }
             }
-            else if (Score - RunnerUpScore > ToleranceForScoreDifferentiation)
+            else if (newScore - RunnerUpScore > ToleranceForScoreDifferentiation)
             {
                 RunnerUpScore = newScore;
             }
         }
 
+        //P-Value analysis identifies ambiguous peptides with lower probability. These are removed from the bestmatchingpeptides dictionary, which lowers ambiguity.
```
rmillikin: "P-Value analysis" -> "PEP analysis"

EngineLayer/PeptideSpectralMatch.cs
```diff
@@ -126,12 +127,18 @@ public void AddOrReplace(PeptideWithSetModifications pwsm, double newScore, int
...
+        public void RemoveThisAmbiguousePeptide(int notch, PeptideWithSetModifications pwsm)
```
rmillikin: "Ambiguous" (spelling)

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -0,0 +1,375 @@
...
+                        index++;
+                    }
+
+                    foreach (var (notch, pwsm) in bestMatchingPeptidesToRemove)
+                    {
+                        ambiguousScans = ambiguousScans + psm.ScanNumber + "|";
+                        psm.RemoveThisAmbiguousePeptide(notch, pwsm);
+                    }
+
+                    psm.FdrInfo.PEP = pepValuePredictions[0]; //they should all be the same at this point so it doesn't matter which you take. First is good.
```
rmillikin: I think you want
`psm.FdrInfo.PEP = 1 - pepValuePredictions[0];`

the PEP is the probability of being an error. the machine learning algorithm is predicting the probability of being correct, I think?

EngineLayer/FdrAnalysis/FdrAnalysisResults.cs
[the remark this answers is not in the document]
trishorts: It's line 347 in PEPValueAnalysisGeneric where the method is. 
image

lonelu:
1. Why remove Evalue, is it same to PEP?
2. The method we use from Microsoft.ML is what? Is it a totally black box?

trishorts:
Evalue is not the same as PEP. Evalue is more about fitting a tail to a distribution and then looking at the probability of getting a certain score. it doesn't consider any other factors. The PEP currently is the probability that a PSM is false positive given a combination of several different criteria (not just score, in fact, not score at all). 

The microsoft.ml is a general package that lets you do all sorts of different machine learning functions. I am use the binary classification. you can read about microsoft.ml here: https://docs.microsoft.com/en-us/dotnet/machine-learning/how-does-mldotnet-work

EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
```diff
@@ -207,55 +156,34 @@ private void DoFalseDiscoveryRateAnalysis(FdrAnalysisResults myAnalysisResults)
...
+            if (AnalysisType == "Peptide")
             {
-                count = globalMeanCount;
+                Compute_PEPValue_Based_QValue(AllPsms);
             }
+        }
 
-            double probabilityOfScore = 1 - Math.Pow(preValue, count);
-            return count * probabilityOfScore;
+        public static void Compute_PEPValue_Based_QValue(List<PeptideSpectralMatch> psms)
```
lonelu: Is my understanding right? Here PEPValue is just used as a score range from 0-1 instead of probability. Could we use PEP < 0.01 instead of FDR < 0.01?

trishorts: no. it is an actual probability. it is the probability that an individual PSM is not not a true positive. We have not had anything like it in metamorpheus so far. FDR is a groupwise number. 1%FDR means that 1% of the PSMs are not true positives. It is not an individual parameter and several PSMs can have a much higher probability of being false than 1%. I added a second paramter, which is the PEP based Q-value. It's the last column. In that column, you want all PSMs with a q-value less that 0.01. That is a set-wise parameter and comparable to the old Q-value.

lonelu:
Statistical binary classification methods (machine learning):
Decision trees
Random forests
Bayesian networks
Support vector machines
Neural networks
Logistic regression
Probit model

I still couldn't find which one we are using, could you please point it to me?

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -0,0 +1,375 @@
...
+            //
+            //   seed:
+            //     Seed for the random number generator used to select rows for the train-test split.
+            TrainTestData trainTestSplit = mlContext.Data.TrainTestSplit(dataView, testFraction: 0.1);
+            IDataView trainingData = trainTestSplit.TrainSet;
+            IDataView testData = trainTestSplit.TestSet;
+
+            var trainer = mlContext.BinaryClassification.Trainers.FastTree(labelColumnName: "Label", featureColumnName: "Features");
+
+            var pipeline = mlContext.Transforms.Concatenate("Features", "Intensity", "ScanPrecursorCharge", "DeltaScore", "Notch", "PsmCount", "ModsCount", "MissedCleavagesCount", "Ambiguity", "AccessionAppearances", "LongestFragmentIonSeries")
```
zrolfs: Including PSM count as a ML parameter is dangerous and inherently biased for the identification of targets over decoys, such that a false-positive target is more likely to be identified than a decoy.

From [this paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4751987/):
"One of the better documented problems with TD estimation of FDR is its incompatibility with any scoring system that uses protein inference, such as weighting based on the presence of other peptides from the same protein. As more peptides are being matched to the target database, it is much more likely that a random PSM to the target database is to a protein already identified than a decoy PSM being to the same protein as another decoy database match. It was largely for this reason that the Percolator software removed protein inference parameters from its re-scoring."

zrolfs: I would not feel comfortable using PEPs until that parameter is removed.

trishorts:
I believe it is the decisioin tree.  The example I followed is here: https://medium.com/@dvirsegal/ml-net-heart-disease-prediction-6fd73a274eec

You can learn how decision trees work here:
https://www.youtube.com/watch?v=LDRbO9a6XPU

lonelu:
The 'Trainers.FastTree' suggests that we may use decision-tree here.

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
trishorts: there is some potential for bias. however. the training set for decoy is mixed with low scoring trues. i dont have the actual number in from but id say the 40% of the false category is of such time (60% actual decoys). The damage is mitigated somewhat. Also, i did a search of mouse data and added arabadopsis database. After, I compared arabaopsis true to arabadopsis decoy in the results. A ranked set, showed that the two occured with basically the same frequency across the score range. This particular decoy/decoy search also shows that ther is little if any bias. though i'm sure there must be some

trishorts:
yes

lonelu:
I follow Rob's method to create a local pull request, ran the test data got this error. This may be related to some settings. 

e: System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.ML.Data, Version=1.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified.
File name: 'Microsoft.ML.Data, Version=1.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'
   at EngineLayer.PEP_Analysis.ComputePEPValuesForAllPSMsGeneric(List`1 psms)
   at EngineLayer.FdrAnalysis.FdrAnalysisEngine.DoFalseDiscoveryRateAnalysis(FdrAnalysisResults myAnalysisResults) in E:\GitHub\MetaMorpheus\EngineLayer\FdrAnalysis\FdrAnalysisEngine.cs:line 161

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
zrolfs: There is not a good way to measure this bias. Low scoring targets used for training are identified purely from MetaMorpheus score and largely have PSM counts of 1. The consideration of PSM scores allow for those with randomly higher PSM counts (e.g. 2) to be unfairly increased. Entrapment with Arabidopsis won't work to measure this bias, because Arabidopsis proteins aren't present in the mouse sample to unfairly bias the Arabidopsis target sequences, only the mouse target sequences.
Independent of the previously mentioned bias, this parameter is somewhat useless for protein identification and quantification. Although it may results in more PSMs (arguably because the FDR is underestimated), the number of proteins and the number of quantified peptides should not change, because this parameter only favors the identification of PSMs that were already identified. As such, it's more for posturing (we got more PSMs) than for actual benefit.

zrolfs: This parallels the issue with the classic method for protein FDR (not top-picked). In the classic, we identify a large number of target proteins before we start hitting false-positives. When we hit a false positive, there's no longer a 50/50 chance of hitting a target/decoy, because targets have been removed from the space. This allows decoys a higher probability of identification and results in an overestimated FDR. The opposite problem occurs here when using the PSM count parameter. We identified a high number of target PSMs that were actually present in the sample. When we get to a false-positive, there is no longer a 50/50 shot of a target/decoy being matched because identifications are biased toward the targets that were previously identified, resulting in an underestimated FDR.

EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
```diff
@@ -207,55 +156,32 @@ private void DoFalseDiscoveryRateAnalysis(FdrAnalysisResults myAnalysisResults)
...
-
-            // this is the "default" case for when there are no scores except the best hit
-            // it uses a global mean score (all scores across all PSMs) to generate the null Poisson distribution
-            // this will be overriden by the next few lines if there are enough scores in this PSM to estimate a null distribution
-            double preValue = SpecialFunctions.GammaLowerRegularized(globalMeanScore, psm.Score);
-            maximumLikelihood = globalMeanScore;
 
-            // calculate single-spectrum evalue if there are enough hits besides the best scoring peptide
-            if (psm.Score == 0)
+            if (AnalysisType == "Peptide")
```
lonelu: This is not asking for changes. I just feel like the ML input of 'Peptide' should be different from 'PSM'. Maybe adding more feature such as retention time evaluation of psms with the same peptide.

trishorts:
I'm interested in quantifying bias and receptive to performing data experiments that would elucidate it.

lonelu:
> I follow Rob's method to create a local pull request, ran the test data got this error. This may be related to some settings. (re-download everything works.)
> 
> e: System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.ML.Data, Version=1.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified.
> File name: 'Microsoft.ML.Data, Version=1.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'
> at EngineLayer.PEP_Analysis.ComputePEPValuesForAllPSMsGeneric(List`1 psms)
> at EngineLayer.FdrAnalysis.FdrAnalysisEngine.DoFalseDiscoveryRateAnalysis(FdrAnalysisResults myAnalysisResults) in E:\GitHub\MetaMorpheus\EngineLayer\FdrAnalysis\FdrAnalysisEngine.cs:line 16
> 1

It is weird that my local pull request only contains Microsoft.ML in EngineLayer. After I add Microsoft.ML into all other Layers, It worked.

rmillikin:
This PR ran for me without crashing. 

With regard to which parameters to train on, I think we should either discuss it in person, over slack, or in an issue. Meaning, the code runs, provides seemingly reasonable PEPs, and it's not used for FDR ranking or anywhere else in the code anyway, so it seems pretty low-risk for now. But we should certainly discuss how to measure bias and which parameters to train on. I especially don't want to reinvent the wheel if Percolator decided to not train on something - we should learn from their "mistakes" and not repeat them. But I'd like to understand why they decided not to train on something, too.

PR #1647, opened 2019-05-06
Fix nonspecific search bug when file-specific digestion params overwrite it

zrolfs:


zrolfs:
hold on this; there's an mzlib bug, too.

zrolfs:
Also not sure why codecov is 0%...

PR #1658, opened 2019-05-29
301

rmillikin:


EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
```diff
@@ -281,13 +281,29 @@ private static int GetNumPSMsAtqValueCutoff(List<PeptideSpectralMatch> psms, dou
         public void CountPsm()
         {
             // exclude ambiguous psms and has a fdr cutoff = 0.01
-            var psmsGroupedBySequence = AllPsms.Where(psm => psm.FullSequence != null && psm.FdrInfo.QValue <= 0.01 && psm.FdrInfo.QValueNotch <= 0.01)
-                .GroupBy(p => p.FullSequence).ToList();
+            var allUnambiguousPsms = AllPsms.Where(psm => psm.FullSequence != null);
 
-            foreach (var group in psmsGroupedBySequence)
+            var unambiguousPsmsLessThanOnePercentFdr = allUnambiguousPsms.Where(psm =>
```
acesnik: What fraction of PSMs are ambiguous in a normal search?

trishorts: image

acesnik: That's an interesting way to look at this. It's an exponential curve of sorts.

Thinking of this as a histogram, it's probably a pretty similar fraction of PSMs that are ambiguous as what I see in variant searches. ~5 out of every 6 PSMs

acesnik: Do you have that plot before/after your p-value stuff?

trishorts: this is psm level 
image

trishorts: this is accession level
image

trishorts: 20% more with a single accession after p-value filtering

acesnik: That's awesome. Is this new, or is this building off a published method?

acesnik: Could you plot the fractions? `(filtered - unfiltered) / unfiltered`

trishorts: "new" I using a new machine learning package from microsoft that's compatible with .net core. This one does binary classification (Target or Decoy) with probability based on ~8 factors

trishorts: that excel sheet is long gone. when I get back from ASMS

PR #1666, opened 2019-06-11
RawFileReader

rmillikin:
Adds the mzLib version that includes RawFileReader and updates the installer with paths to .dll files. Not used yet because there's a problem getting precursor scan numbers. Need to implement user agreement to Thermo's licence, too.


rmillikin:
fixed the precursor scan issue. works well on one file that I tested (17343 PSMs with RawFileReader vs 17345 PSMs with MSFileReader)

rmillikin:
Capture

Seems ok across a couple of tested files

rmillikin:
still need to add:
1. command-line support for RawFileReader. basically just need to add a way for the user to accept thermo's licence.
2. replace MSFileReader code in FlashLFQ with RawFileReader code. currently MSFileReader is still needed to quantify peptides in .raw files

rmillikin:
added both of those. 

rawfilereader won't run in .NET core apparently, though there are updated versions where it will. I'll look into that when .NET core 3.0 comes out and we migrate over to that

EngineLayer/EngineLayer.csproj
```diff
@@ -64,9 +64,6 @@
     <None Update="ProteolyticDigestion\proteases.tsv">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
-    <None Update="settings.toml">
```
trishorts: why delete the settings.toml

GUI/ThermoLicenceAgreementWindow.xaml.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Windows;
+
```
trishorts: is this a one-time agreement? if so, does the answer get stored?

TaskLayer/MyFileManager.cs
```diff
@@ -1,35 +1,20 @@
...
-
+using IO.MzML;
 using MassSpectrometry;
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.IO;
-using System.Linq;
 using System.IO.Compression;
+using ThermoRawFileReader;
```
trishorts: is there no option to continue using thermofilereader?

trishorts:
This all seems fine, albeit a little circumventable with open source. Should we check it on data with MS3? or mixed mode fragmentation?

EngineLayer/EngineLayer.csproj
rmillikin: it's generated automatically if it doesn't exist now

GUI/ThermoLicenceAgreementWindow.xaml.cs
rmillikin: ya, it's a one-time agreement. it gets stored in settings.toml

TaskLayer/MyFileManager.cs
rmillikin: not in MM. the code to use MSFileReader still exists in mzLib but isn't referenced in MM or FlashLFQ anymore. it can still be referenced by editing the code, but it's not a user-facing option

rmillikin:
I agree with the "circumventable" part.. I'll look into how other open-source packages manage this kind of thing. There is a line in the Thermo agreement that says "If you use this Software, you accept this License." but it seems a little flimsy to rely on that...

I can test this out with MS3 data and mixed mode, Lei has some data that he gave me

rmillikin:
here are the benefits of RawFileReader over MSFileReader if you're wondering:
1. Faster .raw file load times
2. Supported by Thermo into the foreseeable future
3. Does not require installation of MSFileReader
4. Does not require installation of C++ redistributable
5. Can run in Linux (not in this particular PR, but I just need to update RawFileReader in mzLib)

rmillikin:
Capture

Tested MS2 mixed mode and MS2/MS3 with the crosslink search. Same number of PSMs, and same scores and matched ions as far as I could tell

trishorts:
I don't recall if we had unit tests for rawfilereader. if we did, do they still work? For example. if we have a slice, does the count of ms1s, ms2s and ms3s make sense? to the precursors for ms2s and 3s make sense? Does the fragmentation mode for each match? what if missing?

rmillikin:
there are no Thermo unit tests in MetaMorpheus, but there are in mzLib. I used the same data files to create new unit tests for RawFileReader in mzLib, and those work ok. it's also feasible now to create unit tests for RawFileReader in MetaMorpheus, though I didn't do that here.

rmillikin:
looks like it's still not possible to make a unit test to read a .raw file in MM. I get the same error message that I get when I try to open a .raw file in .NET core. Seems OK for now, given that we have Jenkins builds. But at some point I'll see if using the .NET core Thermo .dll files work for that.

MetaMorpheusSetup/Product.wxs
```diff
@@ -257,15 +251,30 @@
       <Component Id="src_System.Xml.XPath.XDocument.dll" Guid="70f63556-f331-4c08-ac9f-3515dacbf0e9">
         <File Id="src_System.Xml.XPath.XDocument.dll" Name="System.Xml.XPath.XDocument.dll" Source="$(var.GUI_TargetDir)System.Xml.XPath.XDocument.dll" />
       </Component>
-      <Component Id="src_OxyPlot.dll">
+      <Component Id="src_OxyPlot.dll" Guid="1B25D408-13DB-4A9E-A49E-B0A7FC8C05A6">
```
lonelu: Something changed here. Is this change necessary?

rmillikin: https://wixtoolset.org/documentation/manual/v3/howtos/general/generate_guids.html

I don't fully understand it, but it seems like each of these components is supposed to have a GUID for Windows installer reasons. Some of these components didn't have one - I corrected that in this PR

PR #1686, opened 2019-06-26
MetaDraw change color of variant containing ions

cameyer6:


EngineLayer/PsmTsv/PsmFromTsv.cs
```diff
@@ -201,5 +209,43 @@ private static List<MatchedFragmentIon> ReadFragmentIonsFromString(string matche
...
+                    int peptideEnd = int.Parse(positionMatch.Groups[2].Value);
+                    int variantResidueStart = int.Parse(variantMatch.Groups[1].Value);
+                    int variantResidueEnd = variantResidueStart + variantMatch.Groups[2].Value.Length - 1;
+
+                    foreach (MatchedFragmentIon ion in MatchedIons)
+                    {
+                        Match ionMatch = IonParser.Match(ion.Annotation);
+                        if (ionMatch.Success && // finds match, variant is within peptide, and ion contains variant (differs if ion is abc or xyz)
+                            (variantResidueEnd >= peptideStart && variantResidueStart <= peptideEnd) &&
+                            ((ion.NeutralTheoreticalProduct.ProductType <= ProductType.c &&
```
rmillikin: this `<= ProductType.c` stuff seems weird

GUI/MetaDraw.xaml.cs
```diff
@@ -476,6 +476,7 @@ private void DrawAnnotatedBaseSequence(PsmFromTsv psm)
             {
                 int residue = ion.NeutralTheoreticalProduct.TerminusFragment.AminoAcidPosition;
                 string annotation = ion.NeutralTheoreticalProduct.ProductType + "" + ion.NeutralTheoreticalProduct.TerminusFragment.FragmentNumber;
+                Color color = psm.VariantCrossingIons.ContainsKey(ion) && psm.VariantCrossingIons[ion] ? Colors.Green : productTypeToColor[ion.NeutralTheoreticalProduct.ProductType];
```
rmillikin: `&& psm.VariantCrossingIons[ion]` doesn't do anything

GUI/MetaDraw/PsmAnnotationViewModel.cs
```diff
@@ -293,7 +293,11 @@ public void DrawPeptideSpectralMatchPdf(MsDataScan msDataScan, PsmFromTsv psm, s
         {
             OxyColor ionColor;
 
-            if (productTypeDrawColors.ContainsKey(matchedIon.NeutralTheoreticalProduct.ProductType))
+            if(psmToDraw.VariantCrossingIons.ContainsKey(matchedIon) && psmToDraw.VariantCrossingIons[matchedIon])
```
rmillikin: ` && psmToDraw.VariantCrossingIons[matchedIon]` doesn't do anything

Test/MetaDrawTest.cs
```diff
@@ -54,5 +55,40 @@ public static void TestMetaDrawReadCrossPsmFile()
...
         }
+
+        [Test]
+        public static void TestLabelingVariantCrossingIons()
+        {
+            string myFile = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\VariantCrossTest.psmtsv");
+            List<string> warnings = new List<string>();
+            List<PsmFromTsv> psms;
+
+            psms = PsmTsvReader.ReadTsv(myFile, out warnings);  //test will fail if order of psms isn't from top to bottom row
```
rmmiller22: Check that you are parsing everything in correctly now that you have added the identified variant field. Add assert statements to ensure that these are as expected.

Test/MetaDrawTest.cs
```diff
@@ -54,5 +55,40 @@ public static void TestMetaDrawReadCrossPsmFile()
...
+            expected.Add(new Dictionary<string, bool>() { { "b4", true }, { "b16", true }, { "y1", false }, { "y7", false } });      // variant at 1st position (9221)
+            expected.Add(new Dictionary<string, bool>() { { "b2", false }, { "b3", false }, { "y1", true }, { "y9", true } });       // variant at last position (6778)
+            expected.Add(new Dictionary<string, bool>() { { "b3", false }, { "b4", true }, { "y34", false }, { "y35", true } });     // peptide comes from multiple proteins (8613)
+            expected.Add(new Dictionary<string, bool>() { { "b1", true }, { "b10", true }, { "y1", true }, { "y6", true } });        // variation spans the whole peptide (8765)
+            expected.Add(new Dictionary<string, bool>() { { "b2", false }, { "b5", false }, { "y1", false }, { "y30", false } });    // variant before peptide (8169)
+            expected.Add(new Dictionary<string, bool>() { { "b2", false }, { "b4", false }, { "y1", false }, { "y14", false } });    // variant after peptide (6532)
+
+
+            Assert.AreEqual(0, psms[0].VariantCrossingIons.Count,
```
rmmiller22: Don't just check for one psm. Ensure that the count is as expected for all the psms

GUI/MetaDraw/PsmAnnotationViewModel.cs
rmmiller22: same response as the comment above

EngineLayer/PsmTsv/PsmFromTsv.cs
```diff
@@ -38,13 +40,16 @@ public class PsmFromTsv
...
         public string OrganismName { get; }
+        public string IdentifiedSequenceVariations { get; }
         public string PeptideDesicription { get; }
         public string StartAndEndResiduesInProtein { get; }
         public string PreviousAminoAcid { get; }
         public string NextAminoAcid { get; }
         public string DecoyContamTarget { get; }
         public double? QValueNotch { get; }
 
+        public Dictionary<MatchedFragmentIon, bool> VariantCrossingIons { get; }
```
rmmiller22: The dictionary should be renamed it is confusing because it seems that this dictionary would only contains fragment ions that cross the variant, which is not the case.

GUI/MetaDraw.xaml.cs
rmmiller22: It does check to see if the entry with the key matched ion crosses the variant site. The dictionary is poorly named because it contains all of the matched ions, those that do and those that do not cross the variant site.

trishorts:
I think that we should add these three psm fields to the Properties window of the metadraw output (Intersecting Sequence   Variations,  Identified Sequence Variations ,  Splice Sites).

image

trishorts:
Also, this is not specific to variants, but I have a hard time seeing the yellow dot at the front of this annotated sequence. maybe a slightly darker color like orange? 
image

EngineLayer/PsmTsv/PsmFromTsv.cs
lonelu: Is that possible to add a property to MatchedFragmentIon, like "bool ionCrossingVariant {get; set;}" ? So no extra dictionary is needed here.

rmmiller22: That was the original plan but Rob wanted to avoid making changes to mzlib, which that required.

EngineLayer/PsmTsv/PsmFromTsv.cs
lonelu: I guess (ProductType <= ProductType.c) is same as (ProductType ==ProductType.a || ProductType ==ProductType.b || ProductType ==ProductType.c). Please make sure it works like what it means.

cameyer6: It now checks if the ProductType is in a list of a, b, and c product types or x, y, and z product types.

Test/MetaDrawTest.cs
cameyer6: I added checks for Intersecting Sequence Variations, Identified Sequence Variations, and Splice Sites.

GUI/MetaDraw.xaml.cs
cameyer6: VariantCrossingIons a List now, so that expression is now `psm.VariantCrossingIons.Contains(ion)`.

Test/MetaDrawTest.cs
cameyer6: The whole unit test works differently now that VariantCrossingIons is a List, but it checks that there aren't any extra ions in the list.

cameyer6:
> I think that we should add these three psm fields to the Properties window of the metadraw output (Intersecting Sequence Variations, Identified Sequence Variations , Splice Sites).
> 
> image

I added those properties and changed the modification color on the annotated base sequence to orange.

trishorts:
seeing some strange things related to variants in metadraw. just loaded a tsv with only two variant psms from a raw file but I see five scans listed. first three are incorrect but last two are correct.
image

in the properties window, variantcrossingions is not displaying the ions, but instead an error 
image

Issue #1688, opened 2019-06-27
Sequence Coverage Column in Protein Group Results


acesnik:
We should change "Sequence Coverage %" to "Sequence Coverage Fraction" and then rows should be a fraction with more precision (e.g. 0.45827). It can come up as zero with long proteins and cause some confusion. Also, removing the % from the row entries will make it less annoying to format.

rmillikin:
Sure. should be pretty easy

rmillikin:
@acesnik can you make a PR with this change?

acesnik:
Task an undergrad with it, por favor. 😄

trishorts:
this is now in a pull request. anthony, would you review it?

acesnik:
Resolved

PR #1690, opened 2019-07-03
Added TMT 11-plex and updated target platform to x64 on most projects

trishorts-smrf:


TaskLayer/GPTMDTask/GPTMDParameters.cs
```diff
@@ -5,13 +5,16 @@ namespace EngineLayer
...
         public GptmdParameters()
         {
             ListOfModsGptmd = GlobalVariables.AllModsKnown.Where(b =>
                 b.ModificationType.Equals("Common Artifact") ||
-                b.ModificationType.Equals("Common Biological") ||
-                b.ModificationType.Equals("Metal") ||
-                b.ModificationType.Equals("Less Common")
+                b.ModificationType.Equals("Common Biological")
+                //|| b.ModificationType.Equals("Metal") 
+                //|| b.ModificationType.Equals("Less Common")
```
acesnik: My main concern was the less common mods. Have you been unchecking metals, too?

TaskLayer/SearchTask/MzIdentMLWriter.cs
```diff
@@ -682,7 +682,7 @@ public static void WriteMzIdentMl(IEnumerable<PeptideSpectralMatch> psms, List<E
                                 accession = "MS:1001093",
                                 name = "sequence coverage",
                                 cvRef = "PSI-MS",
-                                value = proteinGroup.SequenceCoveragePercent.First().ToString()
+                                value = proteinGroup.SequenceCoverageFraction.First().ToString()
```
acesnik: Does MzIdent expect an integer percent here? This might break something down the line if so

Test/gptmdPrunedBdTests.cs
```diff
@@ -50,7 +50,7 @@ public static void TestPrunedGeneration()
             //ensures that protein out put contins the correct number of proteins to match the folowing conditions. 
                 // all proteins in DB have baseSequence!=null (not ambiguous)
                 // all proteins that belong to a protein group are written to DB
-            Assert.AreEqual(20, proteins.Count);
+            Assert.AreEqual(18, proteins.Count);
```
acesnik: Why did this change?

trishorts: The original unit test used 4 different modification types (common biological, less common, common artifact and metal). The default is now changed to common biological and artifact). So there are fewer mods in the pruneddb and fewer proteins.

TaskLayer/SearchTask/MzIdentMLWriter.cs
trishorts: i don't know. i'll have to check

TaskLayer/GPTMDTask/GPTMDParameters.cs
trishorts: i do. This change may go away. I plan to ask many users about this change before it goes through.

acesnik: Sounds good 👍

TaskLayer/SearchTask/MzIdentMLWriter.cs
acesnik: 👍

Test/gptmdPrunedBdTests.cs
acesnik: Makes sense :)

TaskLayer/SearchTask/MzIdentMLWriter.cs
trishorts: the ask for a decimal value. I checked and other terms require integer or positive integer. I also saw examples of decimal value.

Test/gptmdPrunedBdTests.cs
trishorts: comes from the "Less Common" class of mods. Even though they are not written. This is a strangely written unit test. Writes gptmd prunded db but the db has no mods.....

EngineLayer/ProteinParsimony/ProteinGroup.cs
```diff
@@ -18,7 +18,7 @@ public ProteinGroup(HashSet<Protein> proteins, HashSet<PeptideWithSetModificatio
             AllPeptides = peptides;
             UniquePeptides = uniquePeptides;
             AllPsmsBelowOnePercentFDR = new HashSet<PeptideSpectralMatch>();
-            SequenceCoveragePercent = new List<double>();
+            SequenceCoverageFraction = new List<double>();
```
rmmiller22: Just curious as to why this is changed over from percent to fraction?

TaskLayer/GPTMDTask/GPTMDParameters.cs
rmmiller22: I personally keep metals in my GPTMD search, I feel that (at least some of them) go along with common artifacts.

EngineLayer/ProteinParsimony/ProteinGroup.cs
trishorts: Anthony mad this request specifically. I'm agnostic about it but it helps him. https://github.com/smith-chem-wisc/MetaMorpheus/issues/1688

EngineLayer/Mods/tmt.txt
```diff
@@ -1,4 +1,65 @@
...
+#NL   HCD:C6 N1 H15 C{13}2
+DI   HCD:C6 N1 H15 C{13}2
+#NL   HCD:C5 N1 H15 C{13}3
+DI   HCD:C5 N1 H15 C{13}3
+#NL   HCD:C4 N1 H15 C{13}4
+DI   HCD:C4 N1 H15 C{13}4
+#NL   HCD:C4 H15 C{13}4 N{15}1
+DI   HCD:C4 H15 C{13}4 N{15}1
+#NL   HCD:C3 H16 N1 C{13}5
+DI   HCD:C3 H16 N1 C{13}5
```
lonelu: Did you test with real TMT data?

trishorts: image

trishorts: image

TaskLayer/GPTMDTask/GPTMDParameters.cs
acesnik: I kind of believe the metal adducts, too...

acesnik: Have we ever done a group FDR for these different mod groups?

trishorts: not per se. But, using TMT as a fixed mod, which I think would be the general process, means that everything is the group..... I guess.

Issue #1677, opened 2019-06-20
Lots of redundancy in ModsInfo in Protein Groups output


acesnik:
Makes me wonder if we're counting them right if we're encountering the same modification site a bunch of times, each with 1 observation...
![image](https://user-images.githubusercontent.com/16342951/59885324-bf33b280-936f-11e9-98dd-9eb151208401.png)

acesnik:
It also seems like this is only a problem with some modification types...

acesnik:
It happens whenever there are two different modifications at the same site. Every mod after the first one isn't summed. It's just initialized in the dictionary with a count of 1 each time.
![image](https://user-images.githubusercontent.com/16342951/59885926-d83d6300-9371-11e9-9f7b-6f286bb39c0f.png)

acesnik:
We should also sort by the aa# before writing to output

acesnik:
I also don't get why there are no bar characters '|' in this column for protein groups with multiple members.

rmillikin:
@lonelu can you take a look at this bug?

acesnik:
Anyone tackling this bug?

trishorts:
you've added the bars in the occupancy with multiple proteins but there is an extra semicolon. 
![image](https://user-images.githubusercontent.com/16841846/61242496-78856e00-a70b-11e9-89d8-7ae8aa7b813e.png)

rmillikin:
fixed in #1695

PR #1695, opened 2019-07-12
fix redundancy in occupancy stuff

acesnik:
https://github.com/smith-chem-wisc/MetaMorpheus/issues/1677


acesnik:
It fixes the redundancy I described here https://github.com/smith-chem-wisc/MetaMorpheus/issues/1677, and it also sorts the occupancy output by residue.

I flattened the indentation a bit because I'm working on my laptop and it's hard to scroll sideways. It also makes it easier to read.

I also extended one of Rob's tests to make sure the bug was fixed.

trishorts:
image

acesnik:
At the end? That was the behavior before, but I'll change it.

PR #1696, opened 2019-07-12
Variant Results summary

rmmiller22:
This adds a post search analysis step that is triggered if there are variants in the database. When variants are detected, a variant specific psm and peptide output files are generated. Additionally a text file summarizing the number of variants and there types as well as if any present are modified peptides.
 -All Complete!


TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -71,6 +71,10 @@ public MyTaskResults Run()
             WriteQuantificationResults();
             WritePrunedDatabase();
             WritePeptideResults(); // modifies the FDR results for PSMs, so do this last
+            if(Parameters.ProteinList.Any((p => p.GetVariantProteins(4,1).Count()>0)))
```
acesnik: The `(4,1)` should be values from the settings. Max combinitoric variants and minAlleleDepth.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -905,6 +909,128 @@ private void WritePeptideResults()
...
+            Status("Writing variant peptide results...", Parameters.SearchTaskId);
+            string variantPsmFile = Path.Combine(Parameters.OutputFolder, "VariantPSMs.psmtsv");
+            string variantPeptideFile = Path.Combine(Parameters.OutputFolder, "VaraintPeptides.psmtsv");            
+            List<PeptideSpectralMatch> FDRPsms = Parameters.AllPsms
+                .Where(p => p.FdrInfo.QValue <= CommonParameters.QValueOutputFilter
+                && p.FdrInfo.QValueNotch <= CommonParameters.QValueOutputFilter && p.BaseSequence != null).ToList();
+            List<PeptideSpectralMatch> possibleVariantPsms = new List<PeptideSpectralMatch>();
+            foreach (var peptide in FDRPsms)
+            {
+                var variants = peptide.BestMatchingPeptides.SelectMany(p => p.Peptide.Protein.AppliedSequenceVariations.Select(v => PsmTsvWriter.IntersectsWithVariation(p.Peptide, v, true)).ToList()).ToList();
```
acesnik: nice usage of a previous method :)

acesnik: Should check whether this handles stop-gain and stop-loss cases

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -905,6 +909,128 @@ private void WritePeptideResults()
...
+                }
+            }
+
+            string[] variantResults = new string[12];
+            variantResults[0] = "Variant Result Summary";
+            variantResults[2] = "--------------------------------------------------";
+            variantResults[4] = "Number of ambiguous variant peptides at 1% FDR: " + variantPeptides.Count();
+            variantResults[5] = "Number of variant peptides at 1% FDR: " + confidentVariantPeps.Count();
+            variantResults[6] = "Number of SAV variant peptides at 1% FDR: " + savCount;
+            variantResults[7] = "Number of frameshift variant peptides at 1% FDR: " + frameshiftCount;
```
acesnik: I think you can say `frameshift indel peptides`: I don't think there's a way to get a frameshift without an insertion or deletion.

acesnik: Can we separate insertions and deletions in the results?

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -905,6 +909,128 @@ private void WritePeptideResults()
...
+                        
+                    }
+                }
+            }
+
+            string[] variantResults = new string[12];
+            variantResults[0] = "Variant Result Summary";
+            variantResults[2] = "--------------------------------------------------";
+            variantResults[4] = "Number of ambiguous variant peptides at 1% FDR: " + variantPeptides.Count();
+            variantResults[5] = "Number of variant peptides at 1% FDR: " + confidentVariantPeps.Count();
```
acesnik: `Number of uniquely identified variant peptides at ...`

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -905,6 +909,128 @@ private void WritePeptideResults()
...
+                        }
+                        
+                    }
+                }
+            }
+
+            string[] variantResults = new string[12];
+            variantResults[0] = "Variant Result Summary";
+            variantResults[2] = "--------------------------------------------------";
+            variantResults[4] = "Number of ambiguous variant peptides at 1% FDR: " + variantPeptides.Count();
```
acesnik: `Number of identified variant peptides ...`

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -905,6 +909,128 @@ private void WritePeptideResults()
...
+            variantResults[5] = "Number of variant peptides at 1% FDR: " + confidentVariantPeps.Count();
+            variantResults[6] = "Number of SAV variant peptides at 1% FDR: " + savCount;
+            variantResults[7] = "Number of frameshift variant peptides at 1% FDR: " + frameshiftCount;
+            variantResults[8] = "Number of stop gain variant peptides at 1% FDR: " + stopGainCount;
+            variantResults[9] = "Number of inframe indel variant peptides at 1% FDR: " + indelCount;
```
acesnik: Could you move this above stop gain, below frameshifts?

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -905,6 +909,128 @@ private void WritePeptideResults()
...
+            }
+
+            string[] variantResults = new string[12];
+            variantResults[0] = "Variant Result Summary";
+            variantResults[2] = "--------------------------------------------------";
+            variantResults[4] = "Number of ambiguous variant peptides at 1% FDR: " + variantPeptides.Count();
+            variantResults[5] = "Number of variant peptides at 1% FDR: " + confidentVariantPeps.Count();
+            variantResults[6] = "Number of SAV variant peptides at 1% FDR: " + savCount;
+            variantResults[7] = "Number of frameshift variant peptides at 1% FDR: " + frameshiftCount;
+            variantResults[8] = "Number of stop gain variant peptides at 1% FDR: " + stopGainCount;
```
acesnik: Could you add a stop loss category?

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -905,6 +909,128 @@ private void WritePeptideResults()
...
+                        }
+                        if (variant.Description.Description.Contains("frameshift_variant"))
+                        {
+                            frameshiftCount++;
+                        }
+                        if (variant.Description.Description.Contains("stop_gained"))
+                        {
+                            stopGainCount++;
+                        }
+                        if (variant.Description.Description.Contains("conservative_inframe_insertion") || variant.Description.Description.Contains("disruptive_inframe_deletion"))
```
acesnik: Include these and do a case insensitive contains operation:
disruptive_inframe_insertion
disruptive_inframe_deletion
conservative_inframe_deletion
conservative_inframe_insertion
exon_loss_variant

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -905,6 +909,128 @@ private void WritePeptideResults()
...
+                    {
+                        if (variant.Description.Description.Contains("missense_variant"))
+                        {
+                            savCount++;
+                        }
+                        if (variant.Description.Description.Contains("frameshift_variant"))
+                        {
+                            frameshiftCount++;
+                        }
+                        if (variant.Description.Description.Contains("stop_gained"))
```
acesnik: Add another category for stop loss and look for:
`stop_lost`

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
rmmiller22: using the GetVariantProteins(4,1) was problematic. I have changed it to checking if any proteins have an applied variant and if so write the specified results.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
rmmiller22: done

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
rmmiller22: done

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
rmmiller22: There are potential variant peptides. Their base sequence could result from either a variant containing protein or a canonical protein sequence. I will try to figure out a better way to convey this. But it doesn't feel right to me to just call them identified variant peptides.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
rmmiller22: I have separated  inframe insertions and deletions. But I will have to keep framshifts separate in this case because there is no distinction in the description in the frameshift results form an insertion or deletion

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
acesnik: That makes sense. This `GetVariantProteins(4,1)` should still return a single `Protein` if there are no variants.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
acesnik: Can't you check if the applied variant has a longer original or variant sequence?

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
acesnik: That makes sense. I guess `potential variant peptides` makes more sense than `ambiguous`

acesnik:
Looks great, Rachel. I think there may be a few `Count()`s that you might be able to change to `Count` for efficiency, and you could run CodeMaid on the files to standardize all the spacings. But that's pretty small.

For the frameshifts, I think you might be able to check if the applied variation is an insertion or deletion by checking the `OriginalSequence` length against the `VariantSequence` length to see which it is. That might be something for the next iteration, though.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -905,6 +909,147 @@ private void WritePeptideResults()
...
+            {
+                var variantPWSM = peptide.BestMatchingPeptides.FirstOrDefault();
+                var variants = variantPWSM.Peptide.Protein.AppliedSequenceVariations;
+                var culture = CultureInfo.CurrentCulture;
+                foreach (var variant in variants)
+                {
+                    bool intersects = PsmTsvWriter.IntersectsWithVariation(variantPWSM.Peptide, variant, true);
+                    if(intersects == true)
+                    {
+                        if (culture.CompareInfo.IndexOf(variant.Description.Description, "missense_variant", CompareOptions.IgnoreCase)>=0)
```
trishorts: can one peptide have multiple? if not maybe switch/case is better?

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -905,6 +909,147 @@ private void WritePeptideResults()
...
+            }
+
+            int savCount = 0;
+            int insertionCount = 0;
+            int deletionCount = 0;
+            int frameshiftCount = 0;
+            int stopGainCount = 0;
+            int stopLossCount = 0;
+            int exonLossCount = 0; ;
+            confidentVariantPeps.RemoveAll(p => p.FdrInfo.QValue > 0.01 || p.FdrInfo.QValueNotch > 0.01 || p.IsDecoy);
```
trishorts: are we never reporting psms with high q-values

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -905,6 +909,147 @@ private void WritePeptideResults()
...
+            int insertionCount = 0;
+            int deletionCount = 0;
+            int frameshiftCount = 0;
+            int stopGainCount = 0;
+            int stopLossCount = 0;
+            int exonLossCount = 0; ;
+            confidentVariantPeps.RemoveAll(p => p.FdrInfo.QValue > 0.01 || p.FdrInfo.QValueNotch > 0.01 || p.IsDecoy);            
+            List<PeptideSpectralMatch> modifiedVariantPeptides = confidentVariantPeps.Where(p => p.ModsIdentified.Count() > 0).ToList();
+            List<PeptideSpectralMatch> modifiedVariantSitePeptides = new List<PeptideSpectralMatch>();
+            foreach (var entry in modifiedVariantPeptides)
```
trishorts: add comment why treated separately

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -905,6 +909,147 @@ private void WritePeptideResults()
...
+                    }
+                }
+                if (modifiedVariant == true)
+                {
+                    modifiedVariantSitePeptides.Add(entry);
+                }
+            }
+            foreach (var peptide in confidentVariantPeps)
+            {
+                var variantPWSM = peptide.BestMatchingPeptides.FirstOrDefault();
```
trishorts: should this be expanded to all beyond first at some point. can add a TODO.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -905,6 +909,147 @@ private void WritePeptideResults()
...
+                    }
+                }
+            }
+
+            string[] variantResults = new string[15];
+            variantResults[0] = "Variant Result Summary";
+            variantResults[2] = "--------------------------------------------------";
+            variantResults[4] = "Number of potentially variant containing peptides identified at 1% FDR: " + variantPeptides.Count();
+            variantResults[5] = "Number of unqiuely identified variant peptides at 1% FDR: " + confidentVariantPeps.Count();
+            variantResults[6] = "Number of SAV variant peptides at 1% FDR: " + savCount;
```
trishorts: does each psm get labeled as to why kind of variant it is? in one of the columns in the tsv?

Test/VariantSearchTests.cs
```diff
@@ -141,5 +142,55 @@ public void MoreTests(string filename, DecoyType decoyType = DecoyType.None)
...
+
+            // test that output is what's expected
+            Assert.That(missingFiles.Count() == 0 && extraFiles.Count() == 0);
+
+            string[] checkResults = File.ReadAllLines(Path.Combine(thisTaskOutputFolder, "VariantOutput", "VariantAnalysisResultSummary.txt"));
+
+            Assert.AreEqual("Number of potentially variant containing peptides identified at 1% FDR: 1", checkResults[4]);
+            Assert.AreEqual("Number of unqiuely identified variant peptides at 1% FDR: 1", checkResults[5]);
+            Assert.AreEqual("Number of frameshift variant peptides at 1% FDR: 1", checkResults[7]);
+            Assert.AreEqual("Number of modified variant peptides at 1% FDR: 1", checkResults[13]);
```
trishorts: add checks for each of the kinds of variants

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
rmmiller22: No it does not at this time

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
rmmiller22: I am not sure if switch cases can be used when checking for sub-strings, but I will look into it.

Test/VariantSearchTests.cs
rmmiller22: done

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
rmmiller22: I do not believe that searching for a substring works well with switch cases

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
rmmiller22: I can include high q-value psms in output files but filter them out prior to results summary. I want the text results summary to only be 1% FDR.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -905,6 +909,149 @@ private void WritePeptideResults()
...
+                    foreach (var variant in variants)
+                    {
+                        bool intersects = PsmTsvWriter.IntersectsWithVariation(variantPWSM.Peptide, variant, true);
+                        if (intersects == true)
+                        {
+                            if (culture.CompareInfo.IndexOf(variant.Description.Description, "missense_variant", CompareOptions.IgnoreCase) >= 0)
+                            {
+                                savCount++;
+                            }
+                            if (culture.CompareInfo.IndexOf(variant.Description.Description, "frameshift_variant", CompareOptions.IgnoreCase) >= 0)
```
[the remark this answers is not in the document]
rmmiller22: made else if change

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -905,6 +909,149 @@ private void WritePeptideResults()
...
+            Status("Writing variant peptide results...", Parameters.SearchTaskId);
+            string variantPsmFile = Path.Combine(Parameters.OutputFolder, "VariantPSMs.psmtsv");
+            string variantPeptideFile = Path.Combine(Parameters.OutputFolder, "VariantPeptides.psmtsv");            
+            List<PeptideSpectralMatch> FDRPsms = Parameters.AllPsms
+                .Where(p => p.FdrInfo.QValue <= CommonParameters.QValueOutputFilter
+                && p.FdrInfo.QValueNotch <= CommonParameters.QValueOutputFilter && p.BaseSequence != null).ToList();
+            List<PeptideSpectralMatch> possibleVariantPsms = new List<PeptideSpectralMatch>();
+            foreach (var peptide in FDRPsms)
+            {
+                var variants = peptide.BestMatchingPeptides.SelectMany(p => p.Peptide.Protein.AppliedSequenceVariations.Select(v => PsmTsvWriter.IntersectsWithVariation(p.Peptide, v, true)).ToList()).ToList();
```
rmillikin: `bool psmContainsVariant = psm.BestMatchingPeptides.Any(p => p.Peptide.Protein.AppliedSequenceVariations.Any(v => PsmTsvWriter.IntersectsWithVariation(p.Peptide, v, true)));` should be a bit more efficient. 

`IntersectsWithVariation` looks like it belongs more in `PeptideWithSetModifications` in mzLib, rather than the `PsmTsvWriter`. it doesn't have anything to do with PSM writing. it looks like the `Includes` method in `PsmTsvWriter` also belongs in `PeptideWithSetModifications`. maybe also the `SequenceVariantString` method.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -905,6 +909,149 @@ private void WritePeptideResults()
             }
         }
 
+        private void WriteVariantResults()
+        {
+            Status("Writing variant peptide results...", Parameters.SearchTaskId);
+            string variantPsmFile = Path.Combine(Parameters.OutputFolder, "VariantPSMs.psmtsv");
```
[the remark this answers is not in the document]
acesnik: Good catch! I'm interested in using these lists to calculate a group FDR at 1 % global FDR, but I really don't care about the PSM FDR within these files.

What do we think about recalculating the PSM FDR within this context? i.e. (decoy variants / target variants)

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
acesnik: I think this handles stop-gain, but stop-loss isn't included in the XML, yet.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
acesnik: Good point. I think we should do this.

I started it here: https://github.com/smith-chem-wisc/mzLib/pull/508

Moving it makes it much easier to test, too, and some issues are becoming apparent:
1. stop losses aren't being included in the protein XML; we'll need to look into the SnpEff method for making this to see if we can include them.
2. what is a variant peptide identification? It's a subtle point, but I think it needs to cross a part of the sequence that contains a sequence change. Otherwise, it's not really evidence for that sequence variation. An example of this would be an MNP where the sequence changes from "AA" to "AR". If you only cross the first A, it's not a sequence variant peptide identification because it doesn't cross something that has a sequence change. If we agree that's the case, I think we could probably remove the boolean check for "checkForSequenceChange" and just make that our assumption. @rmmiller22, what do you think about this?

rmmiller22: I agree with the point that to be a variant containing peptide the sequence must fully cross the variant.  I feel like being more stringent with what we classify as a variant peptide is a good thing, and making the assumption that the sequence changes is fine.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -905,6 +909,149 @@ private void WritePeptideResults()
...
+                possibleVariantPsms.RemoveAll(b => b.IsDecoy);
+            }
+            if (!Parameters.SearchParameters.WriteContaminants)
+            {
+                possibleVariantPsms.RemoveAll(b => b.IsContaminant);
+            }
+
+            WritePsmsToTsv(possibleVariantPsms, variantPsmFile, Parameters.SearchParameters.ModsToWriteSelection);
+
+            List<PeptideSpectralMatch> variantPeptides = possibleVariantPsms.GroupBy(b => b.FullSequence).Select(b => b.FirstOrDefault()).ToList();
```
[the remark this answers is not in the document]
rmmiller22: I did take this code from the WritePeptideResults() method, assuming that it functioned properly there and would as well in this scenario. I also did not have any issues with this in my testing of the code, but I am not 100% sure if I encountered a case with an ambiguous PSM

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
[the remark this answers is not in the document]
rmmiller22: I think there might be an issue though because, if I understand what Rob is saying, that the PSMs that we are calling 1% FDR are not actually 1% global FDR. It might be close, but there could be some differences. I think that we could remedy this by simply moving the WriteVariantResults() above WritePeptideResults(). I am all for calculating a group FDR in these files, but I think making sure that the psms we use are the actual 1 % global FDR PSMs from the allPSMs file is important.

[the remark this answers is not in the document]
acesnik: Oh, sounds good about writing them above `WritePeptideResults()`! 😄

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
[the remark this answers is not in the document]
acesnik: Could we make sure we have a test for this method with ambiguous PSMs? There may be a lot of them with variant databases.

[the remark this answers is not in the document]
rmmiller22: ambiguous PSMs are being handles correctly. Ran a full search with a full spritz database to confirm.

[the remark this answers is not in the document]
acesnik: I guess this groups all ambiguous peptides together and selects the first one, right? So it more or less eliminates ambiguous peptides.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -905,6 +910,149 @@ private void WritePeptideResults()
...
+            variantResults[5] = "Number of unqiuely identified variant peptides at 1% FDR: " + confidentVariantPeps.Where(p => p.IsDecoy == false &&  p.IsContaminant == false && p.FdrInfo.QValue <= 0.01 && p.FdrInfo.QValueNotch <= 0.01).ToList().Count();
+            variantResults[6] = "Number of SAV variant peptides at 1% FDR: " + savCount;
+            variantResults[7] = "Number of frameshift variant peptides at 1% FDR: " + frameshiftCount;
+            variantResults[8] = "Number of inframe insertion variant peptides at 1% FDR: " + insertionCount;
+            variantResults[9] = "Number of inframe deletion variant peptides at 1% FDR: " + deletionCount;
+            variantResults[10] = "Number of stop gain variant peptides at 1% FDR: " + stopGainCount;
+            variantResults[11] = "Number of stop loss variant peptides at 1% FDR: " + stopLossCount;
+            variantResults[12] = "Number of exon loss variant peptides at 1% FDR: " + exonLossCount;
```
acesnik: Could you remove exon loss? It's not included after all.

acesnik:
I moved the `Includes` and `IntersectsWithVariation` methods over to `PeptideWithSetModifications` in mzLib and tested them. I also changed them to require that they identify a variation with a sequence change or by covering the complete variant. https://github.com/smith-chem-wisc/mzLib/pull/508

I also made some changes to SnpEff to allow stop_gained and stop_lost variations to be written to the protein XML. Most of them were missing. https://github.com/smith-chem-wisc/SnpEff/pull/11 -- I lost my write privileges, so could I have those back, or can someone merge that?

exon_loss variations aren't being written. We can look for those with the isoform detection methods in spritz, anyway. So let's delete that from the summary.

rmmiller22:
The PR is ready to be merged now.

PR #1698, opened 2019-07-16
Economize PEP calculation and add retention time as factor

trishorts:


EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -13,13 +14,33 @@ namespace EngineLayer
 {
     public static class PEP_Analysis
     {
+        private static readonly string filepathSubstitue = "SubstitueFilePath";//this is only for unit test that have psms with no filename.
```
lonelu: If this is only for unit test. Add the filename in unit test is better.

Test/RetentionTimeTest.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Test
+{
+    [TestFixture]
+    public sealed class RetentionTimeTest
+    {
+        private static object[,] _peptides300A;
+
+        [Test]
+        public static void SimpleReversePhaseRetentionTimeTest()
```
lonelu: The purpose of this unit test is not testing real data retention time, but test the round function?

trishorts: this test is only to show that retention time calculation is working inside MM. The tests for retention time calculation are in mzlib.

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
trishorts: have to change too many unit tests.....

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -71,12 +91,8 @@ public static string ComputePEPValuesForAllPSMsGeneric(List<PeptideSpectralMatch
                     //Here we compute the pepvalue predection for each ambiguous peptide in a PSM. Ambiguous peptides with lower pepvalue predictions are removed from the PSM.
```
lonelu: It seems like we have another ambiguous peptide removing mechanism in Protein parsimony. Although they are at different levels. Could we include Protein information in the model? It seems complicated for now.

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -135,63 +146,150 @@ public static string ComputePEPValuesForAllPSMsGeneric(List<PeptideSpectralMatch
...
         }
 
-        private static Dictionary<string, int> GetSequenceToPSMCount(List<PeptideSpectralMatch> psms)
+        /// <summary>
+        /// Assuming that we want to use different sets of training features for different search types, the search type needs to be determined.
+        /// In the future, we may have that specified in the GUI and this will be moot.
+        /// </summary>
+        /// <param name="psms"></param>
+        /// <returns></returns>
+        private static string DetermineSearchType(List<PeptideSpectralMatch> psms)
```
lonelu: I like the design here. This can help extend the function for crosslink search or others...

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -135,63 +146,150 @@ public static string ComputePEPValuesForAllPSMsGeneric(List<PeptideSpectralMatch
...
         /// <returns></returns>
-        public static Dictionary<string, int> GetAccessionCounts(List<PeptideSpectralMatch> psms)
+        private static string[] GetCustomFeatures(string searchType)
         {
-            Dictionary<string, int> accessionCountDictionary = new Dictionary<string, int>();
+            //non-specific and top-down searches don't used missedCleavages
+            switch (searchType)
+            {
+                case "topDown":
+                    return new string[] { "Intensity", "ScanPrecursorCharge", "DeltaScore", "Notch", "PsmCount", "ModsCount", "MissedCleavagesCount", "Ambiguity", "LongestFragmentIonSeries" };
```
rmmiller22: Comment above states that top-down searches don't use missedCleavages, but MissedCleavageCount is still included here. I was wondering why this is the case.

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -13,13 +14,33 @@ namespace EngineLayer
...
         public static string ComputePEPValuesForAllPSMsGeneric(List<PeptideSpectralMatch> psms)
         {
-            Dictionary<string, int> accessionAppearances = GetAccessionCounts(psms);
+            string searchType = DetermineSearchType(psms);
+
+            //These two dictionaries contain the average and standard deviations of hydrophobicitys measured in 1 minute increments accross each raw
+            //file separately. An individully measured hydrobophicty calculated for a specific PSM sequence is compared to these values by computing
+            //the z-score. That z-score is used in the the machine learning.
+            Dictionary<string, Dictionary<int, double>> RetentionTimeHydrophobicityDeviation = new Dictionary<string, Dictionary<int, double>>();
+            Dictionary<string, Dictionary<int, double>> RetentionTimeHydrophobicityAverage = new Dictionary<string, Dictionary<int, double>>();
```
rmmiller22: I think that a Dictionary<string, Dictionary<int, Tuple<double, double>>> may work to simplify things. Having item 1 as the deviation and item 2 as the average or vice versa. That way all of the information is stored in a single dictionary for unmodified and a single place for modified peptides. Then for calculation of Z-score you just have to pass in this single dictionary instead of both. Just a thought.

trishorts: switched to tuple now

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
trishorts: deleted missed cleavages

Test/RetentionTimeTest.cs
```diff
@@ -0,0 +1,49 @@
...
+                double expected = (double)_peptides300A[i, 1];
+                double actual = calc.ScoreSequence(peptide);
+
+                // Round the returned value to match the values presented
+                // in the supporting information of the SSRCalc 3 publication.
+                // First cast to float, since the added precision of the double
+                // caused the double representation of 12.805 to round to 12.80
+                // instead of 12.81.  When diffed with 12.81 the result was
+                // 0.005000000000002558.
+                double actualRound = Math.Round((float)actual, 2);
```
rmillikin: I don't really get this. I guess you're trying to get the same exact answer as the supp info in the paper? but their precision is only to two decimals. so you have to do some weird rounding thing. just assert that `actual-expected < 0.051` or something

Test/RetentionTimeTest.cs
```diff
@@ -0,0 +1,49 @@
...
+
+                // Extra conditional added to improve debugging of issues.
+                if (Math.Abs(expected - actual) > 0.005)
```
rmillikin: I don't see the purpose of this

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
rmillikin: i also think this should be removed, though it's fine to do this in a separate PR since this PR is a fairly high priority. when the code gets messy like this it's hard to follow and/or know what variables are dummy unit test variables and which are "real"

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -13,13 +14,29 @@ namespace EngineLayer
...
         public static string ComputePEPValuesForAllPSMsGeneric(List<PeptideSpectralMatch> psms)
         {
-            Dictionary<string, int> accessionAppearances = GetAccessionCounts(psms);
+            string searchType = DetermineSearchType(psms);
+
+            //These two dictionaries contain the average and standard deviations of hydrophobicitys measured in 1 minute increments accross each raw
+            //file separately. An individully measured hydrobophicty calculated for a specific PSM sequence is compared to these values by computing
+            //the z-score. That z-score is used in the the machine learning.
+            //Separate dictionaries are created for peptides with modifications becuase SSRcalc doesn't really do a good job predicting hyrophobicity
+            Dictionary<string, Dictionary<int, Tuple<double, double>>> fileSpecificTimeDependantHydrophobicityAverageAndDeviation_unmodified = new Dictionary<string, Dictionary<int, Tuple<double, double>>>();
```
rmillikin: a comment here describing what each element is would be nice. I think this is:
string - filename
int - rounded RT
double - ?? hydrophobicity?
double - std dev RT?

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -135,63 +142,146 @@ public static string ComputePEPValuesForAllPSMsGeneric(List<PeptideSpectralMatch
...
-        /// <param name="psms"></param>
-        /// <returns></returns>
-        public static Dictionary<string, int> GetAccessionCounts(List<PeptideSpectralMatch> psms)
-        {
-            Dictionary<string, int> accessionCountDictionary = new Dictionary<string, int>();
-
-            foreach (PeptideSpectralMatch psm in psms)
-            {
-                if (psm != null)
+                foreach (PeptideSpectralMatch psm in psms.Where(f => (f.FullFilePath == filename || f.FullFilePath == null) && f.FdrInfo.QValue <= 0.01))
```
rmillikin: ` || f.FullFilePath == null` FullFilePath should never be null, this should be removed. i'm assuming this is for unit tests. but the unit test should be better, instead of adding stuff like this to production code to compensate for poorly constructed unit tests. again, not immediately necessary, but fixing stuff like this should be the next PR.

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -135,63 +142,146 @@ public static string ComputePEPValuesForAllPSMsGeneric(List<PeptideSpectralMatch
...
+                            if (hydrobophobicites.ContainsKey(possibleKey))
+                            {
+                                hydrobophobicites[possibleKey].Add(predictedHydrophobicity);
+                            }
+                            else
+                            {
+                                hydrobophobicites.Add(possibleKey, new List<double>() { predictedHydrophobicity });
+                            }
+                        }
+                        else if (!pwsm.BaseSequence.Equals(pwsm.FullSequence) && modified)
```
rmillikin: you don't need to change this, but it would be faster to do `peptide.AllModsOneIsNterminus.Any()` instead of doing a string comparison

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -135,63 +142,146 @@ public static string ComputePEPValuesForAllPSMsGeneric(List<PeptideSpectralMatch
             //mlContext.Model.Save(trainedModel, trainingData.Schema, @"C:\Users\User\Downloads\TrainedModel.zip");
         }
 
-        private static Dictionary<string, int> GetSequenceToPSMCount(List<PeptideSpectralMatch> psms)
+        private static Dictionary<string, Dictionary<int, Tuple<double, double>>> ComputeHydrophobicityValues(List<PeptideSpectralMatch> psms, bool modified)
```
rmillikin: change `modified` to something more clear like `computingHydrophobicitiesForModifiedPeptides`, it confused me on line 163

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -135,63 +142,146 @@ public static string ComputePEPValuesForAllPSMsGeneric(List<PeptideSpectralMatch
...
-                                accessionCountDictionary.Add(accession, 1);
+                                hydrobophobicites.Add(possibleKey, new List<double>() { predictedHydrophobicity });
                             }
                         }
                     }
                 }
+
+                foreach (int key in hydrobophobicites.Keys)
+                {
+                    averagesCommaStandardDeviations.Add(key, new Tuple<double, double>(hydrobophobicites[key].Average(), hydrobophobicites[key].StandardDeviation()));
```
rmillikin: interquartile range is more robust to outliers than std dev but this doesn't need to be optimized now

rmillikin:
as a separate thought apart from the review, I'm curious how well this works for low-res data. it should be better than anything we've tried so far (morpheus score, xcorr, etc.) because it optimizes a score based on intensity, number of matched ions, etc. it would be nice to test this at some point

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
rmillikin: this is sort of nitpicky but it should probably be "The value Tuple is the average and standard deviation, respectively, of the predicted hydrophobicities of the observed peptides eluting at that rounded retention time"

Test/RetentionTimeTest.cs
```diff
@@ -0,0 +1,34 @@
...
+
+            SSRCalc3 calc = new SSRCalc3("SSRCalc 3.0 (300A)", SSRCalc3.Column.A300);
+
+            foreach (string peptideSequence in peptidesAndHyrophobicities.Keys)
+            {
+                var peptide = new PeptideWithSetModifications(peptideSequence, new Dictionary<string, Modification>());
+                double expected = peptidesAndHyrophobicities[peptideSequence];
+                double actual = calc.ScoreSequence(peptide);
+
+                Assert.That(expected, Is.EqualTo(actual).Within(0.01));
```
rmillikin: 👍

PR #1700, opened 2019-07-18
parallelize mass-diff localization; report progress during mass-diff …

rmillikin:
…localization engine; add +- 3 missed mono search type for easier/better top-down search

### DIFF
--- a/EngineLayer/Localization/LocalizationEngine.cs
+++ b/EngineLayer/Localization/LocalizationEngine.cs
@@ -1,8 +1,10 @@
 ﻿using MassSpectrometry;
 using Proteomics.Fragmentation;
 using Proteomics.ProteolyticDigestion;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 
 namespace EngineLayer.Localization
 {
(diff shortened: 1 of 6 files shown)


EngineLayer/Localization/LocalizationEngine.cs
```diff
@@ -26,36 +28,60 @@ public LocalizationEngine(IEnumerable<PeptideSpectralMatch> allResultingIdentifi
...
+                (range, loopState) =>
                 {
-                    // create new PeptideWithSetMods with unidentified mass difference at the given residue
-                    PeptideWithSetModifications peptideWithLocalizedMassDiff = peptide.Localize(r, massDifference);
+                    for (int i = range.Item1; i < range.Item2; i++)
+                    {
+                        PeptideSpectralMatch psm = unambiguousPsms[i];
+
+                        // Stop loop if canceled
+                        if (GlobalVariables.StopLoops) { break; }
```
rmillikin: this code is the same as before except parallelized, with search progress reported

GUI/SearchTaskWindow.xaml
```diff
@@ -815,6 +815,13 @@
                                                     </TextBlock>
                                                 </RadioButton.ToolTip>
                                             </RadioButton>
+                                            <RadioButton x:Name="massDiffAcceptPlusOrMinusThree" Content="+- 3 Missed Monoisotopic Peaks" ToolTipService.ShowDuration="999999" ToolTipService.InitialShowDelay="500">
```
rmillikin: this adds a new mass diff acceptor. plus or minus three daltons. it's good for top-down searches where the missed monos are in both directions

acesnik:
Looks good. I wonder how much more search time the +/- 3 Da setting adds for TD searches.

rmillikin:
not much extra time, I think. most of the time is spent reading/"digesting" the database (i.e., creating the proteoform search space) and localizing mass-differences, if an open-mass search is performed. fragmenting and searching for a theoretical proteoform is not too expensive

PR #1701, opened 2019-07-18
fix very slow parsimony

rmillikin:


rmillikin:
I tried this with ~100k peptides from one of @trishorts searches. Without any changes, parsimony took >5 min. I'm not sure of the exact length, I killed it after 5 min. This change decreases that time to about 30 seconds single-threaded. Probably closer to 5 sec multithreaded. I'd like to check the speed with a few million PSMs but I think it's ready to go.

rmillikin:
2,992,700 PSMs
299,270 peptides

creation of parsimony engine took 3.5sec
parsimony stage 0 took 129sec
parsimony stage 1 took 0.5 sec
parsimony stage 2 took 1.4 sec
parsimony stage 3 took 31 sec
parsimony stage 4 took 33 sec
parsimony stage 5 took 74 sec
constructing protein group objects took 1.2 sec

rmillikin:
I will note also this is an extreme case where a lot of peptide-protein connections had to be constructed. I deliberately created PSMs with only 1 protein connection so that it would need to create new peptide-protein connections. in "real life" it will probably take <1 second for stage 0.

trishorts:
i'm really excited about this improvement. Thanks Rob and reviewers!

EngineLayer/ProteinParsimony/ProteinParsimonyEngine.cs
```diff
@@ -27,7 +26,7 @@ public class ProteinParsimonyEngine : MetaMorpheusEngine
         public ProteinParsimonyEngine(List<PeptideSpectralMatch> allPsms, bool modPeptidesAreDifferent, CommonParameters commonParameters, List<string> nestedIds) : base(commonParameters, nestedIds)
```
trishorts: would be cool to add a comment that give the big picture overview of how parsimony is work in addition to the very nice comments that you have below explaining individual steps

EngineLayer/ProteinParsimony/ProteinParsimonyEngine.cs
```diff
@@ -126,38 +127,109 @@ private List<ProteinGroup> RunProteinParsimonyEngine()
...
+                                    {
+                                        continue;
+                                    }
+
+                                    foreach (var psm in baseSequence.Value)
+                                    {
+                                        // numProteinsForThisPsm is the number of proteins that this PSM's peptides are associated with
+                                        int numProteinsForThisPsm = psm.BestMatchingPeptides.Select(p => p.Peptide.Protein).Distinct().Count();
+
+                                        if (numProteinsForThisPsm != numProteinsForThisBaseSequence)
```
lonelu: These two numbers should generally be the same because they are from the same identifications. In what situation they will be different?

rmillikin: it's pretty rare. if two proteins have a peptide digest product with the same base sequence (a shared peptide), but only one of them is modified (a non-shared modification), and the modified version is observed in the data, this can affect parsimony. the user can disable whether to ignore mods or not for parsimony

PR #1703, opened 2019-07-23
Draw matched ion later so they will be on the top layer of the figure

lonelu:


zrolfs:
I'm unable to update this branch...

GUI/MetaDraw/PsmAnnotationViewModel.cs
```diff
@@ -122,6 +122,23 @@ public PsmAnnotationViewModel()
 
             LineSeries[] allIons = new LineSeries[spectrumMzs.Length];
 
+            // draw the remaining unmatched peaks
```
trishorts: please add the following comment:  //we draw the unmatched peaks after we draw the matched peaks so that the matched peaks are visually sitting on top of the unmatched peaks and are therefore readable by the user and on the printout.

PR #1705, opened 2019-07-30
SILAC turnover implementation, revamp SILAC multiplex code

zrolfs:


trishorts:
have to deal with metadraw changes. 
image

trishorts:
SILAC GUI interface seems clunky. I don't see a way how to search and quantify SILAC data that doesn't contain the natural amino acid (e.g. NeuCode SILAC). I think it's okay to search unmixed by SILAC data with a single amino acid specified. But, for quantitative SILAC with ratio's one should have to specify each label even if one is the naturally abundant item. I guess the interface does allow for more than two labels but I don't know how this is handled. Maybe there are just multiple columns in the quantitative output?

trishorts:
As for turnover (turn-on/turn-off), if we have one row for each label (including unlabeled), then we can just label before/after on the row. We don't have to do the turnover before or turnover after, which doesn't make lots of intuitive sense to me.

zrolfs:
> SILAC GUI interface seems clunky. I don't see a way how to search and quantify SILAC data that doesn't contain the natural amino acid (e.g. NeuCode SILAC). 

There's a check box above the grid that says "quantify unlabeled"

>I think it's okay to search unmixed by SILAC data with a single amino acid specified. But, for quantitative SILAC with ratio's one should have to specify each label even if one is the naturally abundant item. 

I don't understand. Can you clarify?

>I guess the interface does allow for more than two labels but I don't know how this is handled. Maybe there are just multiple columns in the quantitative output?

There is column output for each condition, which allows for (theoretically) infinite multiplexing. The interface only limits to two conditions if you're doing a turnover experiment.

trishorts:
So, is this a way to do NeuCode SILAC
image

trishorts:
Then this is the way to do NeuCode turnover?
image

zrolfs:
If, for some reason, you spent the time/money to make a completely heavy mouse using K+8.033 and then switched the mouse to another expensive label, K+8.050, then that's the way to do it.

trishorts:
so....for me, the confusion is that in this turnover configuration, the "end" state is implicitly the natural abundance. It is not listed as a line item. 
image

zrolfs:
I'll include another row that displays the unlabeled and automatically changes the LabelType if the a turnover is chosen.

trishorts:
Yes. I think if you click turnover, you should have to specify two conditions (start/end) and change which ever labels you wish in each. That would be a way that I could easily comprehend.

EngineLayer/Silac/SilacConversions.cs
```diff
@@ -332,267 +269,602 @@ public static SpectraFileInfo GetHeavyFileInfo(SpectraFileInfo originalFile, Sil
...
         //Create new files for each silac label and file so that "file 1" now becomes "file 1 (light)" and "file 1 (heavy)"
         //Change heavy residue into the light residue plus a string label ("PEPTIDEa" -> "PEPTIDEK(+8.014)")
         //This light to heavy conversion needs to happen for the flashLFQ peptides here, but can't for the psm peptides, which are constrained to the protein
         //i.e. pwsms currently don't have sequences; they have start/end residues and a protein sequence. We have to change the output sequences when they're created.
-        public static void SilacConversionsPostQuantification(List<SilacLabel> silacLabels, List<SpectraFileInfo> spectraFileInfo, List<ProteinGroup> ProteinGroups,
-            HashSet<DigestionParams> ListOfDigestionParams, Dictionary<string, List<string>> silacProteinGroupMatcher, FlashLfqResults FlashLfqResults,
-            List<PeptideSpectralMatch> allPsms, Dictionary<string, int> ModsToWriteSelection, bool Integrate)
+        public static void SilacConversionsPostQuantification(List<SilacLabel> allSilacLabels, SilacLabel startLabel, SilacLabel endLabel,
```
trishorts: this method is >500 lines. I see many lines being hit during unit tests, which is excellent. But, I'm not sure if the value returned from subsections are being evaluated. Maybe it's good enough if the final result is accurate. My personal preference would be for this section of code to be broken into isolated methods that could be tested. I don't think there is a policy. Like others to weigh in.

EngineLayer/Silac/SilacConversions.cs
```diff
@@ -332,267 +269,602 @@ public static SpectraFileInfo GetHeavyFileInfo(SpectraFileInfo originalFile, Sil
...
-                                    {
-                                        string labelString = "(" + label.OriginalAminoAcid + label.MassDifference;
-                                        int labelIndex = groupName.IndexOf(labelString);
-                                        if (labelIndex != -1) //labelIndex == 1 if a) 2+ peptides are required per protein or b) somebody broke parsimony
-                                        {
-                                            groupName = groupName.Substring(0, labelIndex);
-                                            updatedGroups.Add(new FlashLFQ.ProteinGroup(groupName, group.GeneName, group.Organism));
-                                        }
-                                    }
+                                    //LLLLL, LLLLH, LLLHH, LLHHH, LHHHH, HHHHHH have n combinations of 1, 5, 10, 10, 5, 1
```
trishorts: I wonder if there is a need for a restriction of protease to the amino acid label. I guess that most people will use trypsin and label on a residue that gets cleaved. But I imagine, that I could not do that in the program. one could do a lys C or arg C that might have multiple labels but not missed cleavages.

trishorts: this could just be a count of the possible label sites instead of missed cleavages

EngineLayer/Silac/SilacConversions.cs
```diff
@@ -332,267 +269,602 @@ public static SpectraFileInfo GetHeavyFileInfo(SpectraFileInfo originalFile, Sil
...
+                                    double LLLL = values[0] + values[1] / 5;
+                                    double LLLH = values[1] * 4 / 5 + values[2] * 4 / 10;
+                                    double LLHH = values[2] * 6 / 10 + values[3] * 6 / 10;
+                                    double LHHH = values[3] * 4 / 10 + values[4] * 4 / 5;
+                                    double HHHH = values[4] / 5 + values[5];
+                                    values = new List<double> { LLLL, LLLH, LLHH, LHHH, HHHH };
```
trishorts: i don't see any unit tests that hit mixed L and H. Only fully labeled is tested.

EngineLayer/Silac/SilacConversions.cs
```diff
@@ -332,267 +269,602 @@ public static SpectraFileInfo GetHeavyFileInfo(SpectraFileInfo originalFile, Sil
...
+                    //the probability a light amino acid is incorporated (Pl) is simply Pl = 1 - Ph
+                    //We can simplify the math if we assume that all protein started as completely light, @t=0: LL=100%, LH = 0%, and HH = 0%
+                    //N = 1 - Qll + Qlh * Qlh / (4 * Qhh), (where N is the fraction of all protein that has turned over and Qll is the Quantity(fraction) of LL at the current timepoint
+                    //Qhh = qhh -N*qhh+Ph*Ph*N, where Q is quantity(fraction) at the timepoint and q is the quantity(fraction) at the start point (time zero).
+                    //Ph = sqrt((Qhh-qhh+N*qhh)/N), and qhh = 0, so we can simplify as
+                    //Ph = sqrt(Qhh/N)
+                    Dictionary<SpectraFileInfo, double> fileToHeavyProbabilityDictionary = new Dictionary<SpectraFileInfo, double>();
+                    foreach (SpectraFileInfo info in spectraFileInfo)
+                    {
+                        //convert the absolute values of LL, LH, and HH to relative values
```
trishorts: nice to have a unit test to see if the probabilities used are correct.

EngineLayer/Silac/SilacConversions.cs
```diff
@@ -332,267 +269,602 @@ public static SpectraFileInfo GetHeavyFileInfo(SpectraFileInfo originalFile, Sil
...
-                List<string> pwsmKeys = lfqPwsms.Keys.ToList();
-                foreach (string key in pwsmKeys)
-                {
-                    FlashLFQ.Peptide currentPeptide = lfqPwsms[key];
-                    SilacLabel label = GetRelevantLabelFromFullSequence(currentPeptide.Sequence, silacLabels);
-                    if (label != null) //if it's a heavy peptide
+                    //with the summed LL, LH, and HH values for each file, we can calculate the probability that a heavy amino acid is incorporated (Ph)
+                    //the probability a light amino acid is incorporated (Pl) is simply Pl = 1 - Ph
+                    //We can simplify the math if we assume that all protein started as completely light, @t=0: LL=100%, LH = 0%, and HH = 0%
+                    //N = 1 - Qll + Qlh * Qlh / (4 * Qhh), (where N is the fraction of all protein that has turned over and Qll is the Quantity(fraction) of LL at the current timepoint
```
trishorts: is it defaulted to "light" or whatever the start condition is? don't you go from heavy to light in one of your experiments?

EngineLayer/ProteinParsimony/ProteinGroup.cs
```diff
@@ -86,6 +86,43 @@ public ProteinGroup(HashSet<Protein> proteins, HashSet<PeptideWithSetModificatio
...
+                }
+                else
+                {
+                    UniquePeptidesOutput = GlobalVariables.CheckLengthOfOutput(string.Join("|", UniquePeptides.Select(p => p.FullSequence).Distinct()));
+                    SharedPeptidesOutput = GlobalVariables.CheckLengthOfOutput(string.Join("|", SharedPeptides.Select(p => p.FullSequence).Distinct()));
+                }
+            }
+            else
+            {
+                if (!DisplayModsOnPeptides)
```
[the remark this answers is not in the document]
zrolfs: Clever!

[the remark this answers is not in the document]
zrolfs: Wait, no it does change the p.BaseSequence vs p.FullSequence

EngineLayer/PeptideSpectralMatch.cs
```diff
@@ -324,12 +320,22 @@ public PeptideSpectralMatch Clone(List<(int Notch, PeptideWithSetModifications P
 
         private PeptideSpectralMatch(PeptideSpectralMatch psm, List<(int Notch, PeptideWithSetModifications Peptide)> bestMatchingPeptides)
         {
-            _BestMatchingPeptides = bestMatchingPeptides ?? psm.BestMatchingPeptides.ToList();
+            if (bestMatchingPeptides == null)
+            {
+                //TODO why isn't this covered?
```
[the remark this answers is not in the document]
zrolfs: nope, it's a legacy check. Code has been deleted.

EngineLayer/ProteinParsimony/ProteinGroup.cs
```diff
@@ -483,14 +523,14 @@ public void MergeProteinGroupWith(ProteinGroup other)
 
         public ProteinGroup ConstructSubsetProteinGroup(string fullFilePath)
         {
-            var allPsmsForThisFile = new HashSet<PeptideSpectralMatch>(this.AllPsmsBelowOnePercentFDR.Where(p => p.FullFilePath.Equals(fullFilePath)));
+            var allPsmsForThisFile = new HashSet<PeptideSpectralMatch>(AllPsmsBelowOnePercentFDR.Where(p => p.FullFilePath.Equals(fullFilePath)));
```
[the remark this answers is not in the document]
zrolfs: The capitalization of the first letter signifies that it is a class variable. Local variables begin lower case.

EngineLayer/Silac/SilacConversions.cs
```diff
@@ -11,97 +11,99 @@ namespace EngineLayer
     public static class SilacConversions
     {
         private static readonly string LABEL_DELIMITER = " & ";
-        public static PeptideSpectralMatch GetSilacPsm(PeptideSpectralMatch psm, SilacLabel silacLabel, bool heavyToLight)
+
+        public static string GetLabeledBaseSequence(string unlabeledBaseSequence, SilacLabel label)
```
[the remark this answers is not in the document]
zrolfs: Not totally sure what's being asked here. You want a unit test that specifically calls GetLabeledBaseSequence?
97 is just the ASCII form of the char 'a'.

EngineLayer/Silac/SilacConversions.cs
zrolfs: The number of label sites is what is used. The missedCleavages nomenclature was meant to aid understanding (because that's what it usually mirrors), but perhaps it simply confuses. I'll change it.

EngineLayer/Silac/SilacConversions.cs
zrolfs: Added (I think?)

EngineLayer/Silac/SilacConversions.cs
zrolfs: Full task unit tests are testing that the probabilities were calculated and applied correctly.

EngineLayer/Silac/SilacConversions.cs
zrolfs: The comments use L/H for ease of understanding (because typical experiments move from L->H), but they're a proxy for start/end, which is what is actually used in the code.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -336,6 +281,132 @@ private void QuantificationAnalysis()
                 }
             }
 
+            //If SILAC (Pre-Quantification) preparation
+            //setup variables
+            List<SilacLabel> allSilacLabels = Parameters.SearchParameters.SilacLabels;
+            SilacLabel startLabel = Parameters.SearchParameters.StartTurnoverLabel;
+            SilacLabel endLabel = Parameters.SearchParameters.EndTurnoverLabel;
+            bool quantifyUnlabeledPeptides = Parameters.ListOfDigestionParams.Any(x => x.GeneratehUnlabeledProteinsForSilac);
+            if (Parameters.SearchParameters.SilacLabels != null)
```
trishorts: This is a place where many lines are "hit" in unit tests but it is not at all clear to me the the values computed are correct. Methods with more than 100 lines spook me.

TaskLayer/SearchTask/SearchTask.cs
```diff
@@ -90,35 +90,42 @@ protected override MyTaskResults RunSpecific(string OutputFolder, List<DbForTask
...
-                            heavyLabel++;
+                            var updatedLabel = SilacConversions.UpdateAminoAcidLabel(SearchParameters.StartTurnoverLabel, heavyLabel);
+                            heavyLabel = updatedLabel.nextHeavyLabel;
+                            SearchParameters.StartTurnoverLabel = updatedLabel.updatedLabel;
+                            SearchParameters.SilacLabels.Add(SearchParameters.StartTurnoverLabel);
                         }
-                        SilacLabel updatedLabel = SilacConversions.AssignValidHeavyCharacter(currentLabel, heavyLabel);
-                        heavyLabel++;
-                        if (currentLabel.AdditionalLabels != null)
+                        if (SearchParameters.EndTurnoverLabel != null)
```
trishorts: this can just be else since you have SearchParameters.StartTurnoverLabel != null || SearchParameters.EndTurnoverLabel != null to get inside.

Test/AnalysisEngineTest.cs
```diff
@@ -97,5 +97,13 @@ public static void TestAnalysisEngineTests()
 
             engine.Run();
         }
+
+        [Test]
+        public static void TestRunSpecificPostSearchAnalysis()
+        {
+            //code coverage unit test for an unused abstract method in post search analysis
```
trishorts: this test is doodoo

Test/SearchEngineTests.cs
```diff
@@ -1611,5 +1612,60 @@ public static void TestFileWithNoMs2Scans()
             var ms2Scans = MetaMorpheusTask.GetMs2Scans(fileWithNoMs2Scans, null, new CommonParameters()).OrderBy(b => b.PrecursorMass).ToArray();
             Assert.That(!ms2Scans.Any());
         }
+
+        //This is a code coverage unit test that mostly just checks there are no crashes in specific situations
+        [Test]
+        public static void TestUndefinedProteinGroup()
+        {
```
trishorts: are there supposed to be some Assertions here? like checking returned values match previously specified values?

Test/SilacTest.cs
```diff
@@ -310,6 +309,301 @@ public static void TestSilacWhenProteinIsMissing()
             string outputFolder = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestSilac");
             Directory.CreateDirectory(outputFolder);
             var theStringResult = task.RunTask(outputFolder, new List<DbForTask> { new DbForTask(xmlName, false) }, new List<string> { mzmlName }, "taskId1").ToString();
+            //No assertions, just checking it didn't crash
```
trishorts: i guess this is okay. but it kinda inflates code coverage. I'm not sure this is really what we are after.

zrolfs:
The unit tests you're requesting require a (realistic) psm object, which is difficult to manufacture outside of the search tasks. I think the comprehensive full tests that currently exist are sufficient to test the individual methods. If any of those methods get changed, one of the unit tests will fail.

TaskLayer/SearchTask/SearchTask.cs
zrolfs: if both start and end have values, then using else would skip this code.

Test/AnalysisEngineTest.cs
zrolfs: It's superficial. The method will never be used, because the post search task doesn't truly inherit everything from a MetaMorpheus task.

Test/SearchEngineTests.cs
zrolfs: No, just checking that the program keeps moving when things aren't functioning as expected.

Test/AnalysisEngineTest.cs
rmmiller22: Why does it exist then, if its not truly testing anything of substance?

EngineLayer/PeptideSpectralMatch.cs
```diff
@@ -277,10 +277,6 @@ public void ResolveHeavySilacLabel(List<SilacLabel> labels, IReadOnlyDictionary<
             //EssentialSequence
             EssentialSequence = PsmTsvWriter.Resolve(_BestMatchingPeptides.Select(b => b.Pwsm.EssentialSequence(modsToWritePruned))).ResolvedString; //string, not value
             EssentialSequence = SilacConversions.GetAmbiguousLightSequence(EssentialSequence, labels, false);
-
```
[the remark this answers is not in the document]
zrolfs: Silac proteins previously had different protein accessions, but no longer do after mzlib changes. (Recall this isn't the first SILAC pull request).

EngineLayer/ClassicSearch/ClassicSearchEngine.cs
```diff
@@ -16,12 +16,13 @@ public class ClassicSearchEngine : MetaMorpheusEngine
         private readonly List<Modification> FixedModifications;
         private readonly List<Modification> VariableModifications;
         private readonly List<SilacLabel> SilacLabels;
+        private readonly (SilacLabel StartLabel, SilacLabel EndLabel)? TurnoverLabels;
```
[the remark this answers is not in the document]
zrolfs: Yes. Tuples inherently can't be null, but the ? allows it to be.

EngineLayer/ProteinParsimony/ProteinGroup.cs
```diff
@@ -303,48 +353,36 @@ public void CalculateSequenceCoverage()
 
             foreach (var protein in ListOfProteinsOrderedByAccession)
             {
-                bool errorResult = false;
```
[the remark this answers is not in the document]
zrolfs: The error checking was a hold over from some weird old code/bug. Ask @rmillikin.

Test/AnalysisEngineTest.cs
zrolfs: The base method is used for most tasks. The "PostSearchTask" does not implement the abstract method, but still must have that method to inherit from the abstract class.

CMD/CMD.csproj
```diff
@@ -9,9 +9,9 @@
 
   <ItemGroup>
     <PackageReference Include="FluentCommandLineParser" Version="1.4.3" />
-    <PackageReference Include="mzLib" Version="1.0.431" />
     <PackageReference Include="Microsoft.ML" Version="1.1.0" />
     <PackageReference Include="Microsoft.ML.FastTree" Version="1.1.0" />
+    <PackageReference Include="mzLib" Version="1.0.444" />
```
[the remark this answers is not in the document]
zrolfs: I updated mzLib and the majority of unit tests are failing now. Standby.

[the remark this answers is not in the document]
rmmiller22: Yeah I am trying to fix that in my master where I updated mzLib. TestSilacMultipleModsPerCondition,TestSilacNoLightProtein and TestSilacQuantification are all failing for me. Maybe it is the same issue. We should chat tomorrow.

EngineLayer/Silac/SilacConversions.cs
```diff
@@ -11,97 +11,99 @@ namespace EngineLayer
...
+                pwsm.PeptideDescription,
+                pwsm.MissedCleavages,
+                pwsm.AllModsOneIsNterminus,
+                pwsm.NumFixedMods,
+                labeledBaseSequence);
+            return psm.Clone(new List<(int Notch, PeptideWithSetModifications Peptide)> { (notch, labeledPwsm) });
+
+        }
+
+        public static PeptideSpectralMatch GetSilacPsm(PeptideSpectralMatch psm, SilacLabel silacLabel)
```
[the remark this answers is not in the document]
zrolfs: Can you clarify? I don't see a labeled Psm. Is this a nomenclature comment?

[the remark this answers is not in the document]
rmmiller22: GetLabeledPSM Lines 35-48 is what I am referring to. It was more of a question as to as the purpose of that method, I was trying to get at the question how is getting a labeled psm different then getting the silac psm (silac psm being labeled). I believe that I get it now though.

zrolfs:
I kinda wanted to let codacy run on this one, since it was a larger change. Do you mind if I revert?

zrolfs:
Re-merged at #1722

PR #1710, opened 2019-08-15
metadraw reconfiguration

bintiibrahim:
- new treeview that categorizes psm into protein and peptide
- working search box and pdf export
- includes sequence coverage map visualization, highlights unique and shared peptides for each protein

Requires: AllProteinGroups.tsv file 

includes fix bug #1713


rmillikin:
@bintiibrahim please resolve merge conflicts

rmillikin:
I'm putting this PR on hiatus because we're going to revamp MetaDraw into a browser-based HTML system. We are going to implement protein sequence coverage maps, but I don't really want to do a whole bunch of back-and-forth to get this PR merged when we're going to turn around and re-implement it in HTML basically immediately. so, I'm closing this but this code is very useful and will be essentially translated to HTML.

PR #1717, opened 2019-08-27
no crosslink at cleavage site

lonelu:
no crosslink at cleavage site. user required.


EngineLayer/CrosslinkSearch/CrosslinkSearchEngine.cs
```diff
@@ -20,6 +20,7 @@ public class CrosslinkSearchEngine : ModernSearchEngine
 
         private readonly bool CrosslinkSearchTopN;
         private readonly int TopN;
+        private readonly bool NoCleaveAtCrosslinkSite;
```
zrolfs: This is a really petty thing, but a name like "CleaveAtCrosslinkSite" is a lot cleaner than "NoCleaveAtCrosslinkSite". I understand this would require you to invert the the conditionals, but it's just kind of a weird as is.

lonelu: Agreed.

MetaMorpheus.sln
```diff
@@ -25,8 +25,8 @@ Global
 		TestAndReleaseLinux|Any CPU = TestAndReleaseLinux|Any CPU
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
-		{57B2A726-CCA9-432E-B74A-2EFA962D1A74}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{57B2A726-CCA9-432E-B74A-2EFA962D1A74}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{57B2A726-CCA9-432E-B74A-2EFA962D1A74}.Debug|Any CPU.ActiveCfg = Debug|x64
```
trishorts: This should probably still be Any CPU unless there is a strong reason to change.

EngineLayer/CrosslinkSearch/CrosslinkSpectralMatch.cs
```diff
@@ -82,12 +82,27 @@ public void ResolveProteinPosAmbiguitiesForXl()
...
 
-            for (int r = 0; r < peptide.BaseSequence.Length; r++)
+            //Consider the possibility that the site is at Protein N terminal.  
+            if (peptide.OneBasedStartResidueInProtein == 1)
+            {
+                if (!peptide.AllModsOneIsNterminus.Keys.Contains(1) && crosslinkerModSites.Contains(peptide.BaseSequence[0]) && CrosslinkAtCleavageSite)
+                {
+                    possibleXlPositions.Add(1);
+                }
+            }
```
zrolfs: This is only checking the protein N-terminus, but not the C-terminus. Additionally, we're missing cases for variable methionine and endogenous cleavage products.

MetaMorpheus.sln
lonelu: changed back.

Issue #1714, opened 2019-08-23
Calibrating a Non-specific Digest Finds No Peptides


mobiusklein:
Running the Calibrate task with a non-specific digest failed to match any peptides, even after expanding its tolerance several times. I have insufficient information from the log to determine if there were matches but they failed to pass the FDR threshold. The search also runs unexpectedly quickly, but I do not know if the Calibrate task takes some shortcuts.

I know the sample has around 3000 good spectra and ~1250 reasonable peptide IDs from searching it with other tools.

rmillikin:
Thanks for the report - @zrolfs can you comment on this?

zrolfs:
Hi mobiusklein,
It sounds like the maximum number of missed cleavages wasn't changed to accomodate the non-specific digestion. For these digestions, each residue can be cleaved, so each if the max missed cleavage parameter is set to 2 (which is the default for fully-specific digestions), then the search will only include 3-mers and will run very fast. Let me know if this resolves your issue. 
![image](https://user-images.githubusercontent.com/16883585/63605219-1e41bd80-c593-11e9-9839-c20b766e3fde.png)

rmillikin:
@zrolfs  I thought there was some "reasonable" default value set here for the max missed cleavages upon selecting the non-specific search. If you select non-specific search is the max missed cleavages still set at 2?

zrolfs:
@rmillikin The functionality described is implemented in the search task, but not Calibration, GPTMD, or XL.

rmillikin:
sounds good. maybe @bintiibrahim can change this with your guidance

PR #1718, opened 2019-08-27
NonSpecific GUI Task AutoUpdate

zrolfs:
When the non-specific protease is selected in any task, the default max missed cleavages and max peptide length will be set to a reasonable default and track with eachother. #1714


zrolfs:
Still need reviews for this small PR.

Issue #1662, opened 2019-06-10
Move "All target peptides within 1% FDR" up in the text output


rmillikin:
![image (2)](https://user-images.githubusercontent.com/24705713/59212430-05229680-8b78-11e9-86a2-f9dd5efeee04.png)

This seems easy but is actually surprisingly difficult because all of the peptide-level analysis (q-values, etc) is performed after the PSM analysis..

PR #1719, opened 2019-08-27
edit results output text

bintiibrahim:
i used a simple solution (using two string builders) because i dont think it's worth disrupting the order of psm and peptide analysis. 

#1662


rmillikin:
"targetText" change to "PsmPeptideProteinSummaryText"
"niceText" change to "TaskSummaryText"

Issue #1127, opened 2018-04-19
warning if users check Mods, UniMod and UniProt in G-PTM-D


trishorts:
just got email from someone that tried this. I think we need a warning pop-up or something.

rmillikin:
I'm not even sure why we have sections for UniProt and Unimod.

UniProt to me means "mods that uniprot has annotated". It doesn't make sense to me to label mods discovered by GPTMD as UniProt.

Unimod has a variety of issues...

@blfrey is overhauling the mods list. I think this is probably a part of that?

blfrey:
My overhaul is only looking at the .txt files in the Mods folder that comes with the MM installation.   UniProt and UniMod classifications are not in that folder.   Does this mean that Stefan linked them somehow to UniProt and UniMod servers?   They are certainly very long generic lists of possible PTMs.

btw, this "UniProt" list is actually a list of all PTM types, not their specific locations on AA residues.  

Options:
1) warning message if someone chooses too many GPTMD options
2) get rid of UniProt and UniMod generic lists
3) do not let people check either of these entire long lists with one check box, and instead leave the lists there, but only so people can go in and find certain individual mods and turn those on.

Option #2 might be OK.  We have a pretty well curated set w/o those 2 lists, and people can go change .txt files if they need to add something else.  (But option #3 lets people choose a few mods of interest to them that are not in our curated sets w/o learning how to change .txt files).

Thoughts @trishorts and @rmillikin ?

rmillikin:
I'd like to have these lists only show what's in our curated text files (i.e., not UniProt and UniMod). We'll still search for UniProt-annotated mods in the databases, but it doesn't make sense to me to annotate mods discovered by GPTMD as "UniProt." They're different annotation sources and the current style makes it difficult to determine if a modification was already annotated in the UniProt database or if GPTMD discovered it with the "UniProt" category checked. So I guess I like option 2? It's essentially foolproof except for issue #819. We (I) can modify the code so that these mods are not displayed, so don't worry about that.

blfrey:
I am on board with #2--deleting these 2 long lists, especially if you add the  "Add custom mod" button

rmillikin:
Ok, I think that's reasonable. Our mods list is missing a lot of common chemical tagging reagents, etc. The problem with UniMod is that it's missing neutral loss information, etc which would likely be a problem anyway when searching with MM with those UniMod modifications enabled.

[...]

PR #1723, opened 2019-08-29
excluded uniprot option from mods

bintiibrahim:
#1127 removed uniprot category from mods


trishorts:
What is the purpose of this pull request? Why eliminate the uniprot option?

rmillikin:
from issue #1127 :

"We'll still search for UniProt-annotated mods in the databases, but it doesn't make sense to me to annotate mods discovered by GPTMD as "UniProt." They're different annotation sources and the current style makes it difficult to determine if a modification was already annotated in the UniProt database or if GPTMD discovered it with the "UniProt" category checked."

trishorts:
Prefer that we keep these mods and simply relabel them in the gptmd generated xml as "GPTMD" mods. That will eliminate the confusion and stop users from thinking certain discoverd mods were prefiously annotaed. Also they can see how many are from gptmd, which is a good thing.

PR #1725, opened 2019-08-30
5 percent PEP ambiguity cutoff

rmillikin:


rmillikin:
not really sure what this "should" be. it's subjective. i was thinking 1% or 5% or something

PR #1726, opened 2019-08-30
190830 silac edits

zrolfs:
1. Fixed GUI bug that wouldn't let you change back to multiplex if you selected turnover
    -Also moved some code into helper methods

2. Fixed bug in multiplex code that would result in a crash if a peptide was identified in multiple files
   -Also unit test updated


EngineLayer/Silac/SilacConversions.cs
```diff
@@ -409,15 +409,16 @@ public static HashSet<FlashLFQ.ProteinGroup> CleanPastProteinQuant(HashSet<Flash
                 }
                 else //multiplex
                 {
-                    //foreach original file
```
zrolfs: here's the bug fix: it's moving the loop order so that it's peptides first, then files. Previously, a peptide was being generated for each file, and then a crash downstream would happen because multiple entries existed for each file

GUI/SearchTaskWindow.xaml.cs
```diff
@@ -1017,6 +906,139 @@ private static Proteomics.SilacLabel ConvertSilacDataGridInfoToSilacLabel(SilacI
...
+                if (StaticSilacLabelsObservableCollection.All(x => x.LabelType == SilacModificationWindow.ExperimentType.Multiplex))
+                {
+                    List<Proteomics.SilacLabel> labelsToSave = new List<Proteomics.SilacLabel>();
+                    foreach (SilacInfoForDataGrid info in StaticSilacLabelsObservableCollection)
+                    {
+                        labelsToSave.Add(ConvertSilacDataGridInfoToSilacLabel(info));
+                    }
+                    TheTask.SearchParameters.SilacLabels = labelsToSave;
+                    //Clear existing start and end labels, if any
+                    TheTask.SearchParameters.StartTurnoverLabel = null;
```
zrolfs: Here's the GUI bug fix. Setting these to null if it's a multiplex to overwrite previous entries

Issue #1349, opened 2018-08-16
Allow customized defaults


zrolfs:
We're partial to our defaults, but it's likely that not every user agrees. Maybe some labs never use trypsin, so they have to switch their protease each time they open a new search task. The other option users currently have is to track down an old toml and drag it in, but this can be a pain.

Instead, it'd be nice if we made a button in each task window that said "save as default", which writes the toml to a data folder so that it can be used anytime a new task is made.

acesnik:
Brian, that's a great first step. Thanks for the suggestion. 

"Save as Default" would be a bit tougher to implement.

PR #1733, opened 2019-09-16
190916 default tasks

zrolfs:
Create functionality that allows a user to save their own set of custom defaults for routine analyses. #1349


GUI/CalibrateTaskWindow.xaml
```diff
@@ -376,6 +376,12 @@
         <StackPanel  Orientation="Horizontal" HorizontalAlignment="Center" Grid.Row="1" Margin="5">
             <Button x:Name="cancelButton" Content="Cancel" Click="CancelButton_Click" />
             <Button x:Name="saveButton" Content="Save Calibrate Task" Click="SaveButton_Click" FontSize="24" />
+            <Button x:Name="setDefaultbutton" Click="SaveAsDefault_Click" Width="100" >
+                <TextBlock Text="Save As Default" TextWrapping="Wrap" TextAlignment="Center"/>
+                <ToolTipService.ToolTip>
+                    <ToolTip Content="Save these parameters as the default. MetaMorpheus can return to original defaults by deleting the folder 'DefaultParameters'" ToolTipService.ShowDuration="999999" ToolTipService.InitialShowDelay="500" />
```
trishorts: rather have a button reset (which deletes the folder) than have the user have to track it down. i sometimes can't find these folders. same comment on similar items below

rmillikin:
how does this work when MM gets updated and new variables are added to, say, Common Params? .toml files aren't guaranteed to be backwards/forwards compatible

zrolfs:
> how does this work when MM gets updated and new variables are added to, say, Common Params? .toml files aren't guaranteed to be backwards/forwards compatible

There are a few different paths; I'll outline each one and describe how they're handled:
1. The user never sets a toml as default. If a toml doesn't exist in the default folder when a new task is created, then a toml is NOT created in that folder (i.e. the folder remains empty). The new task simply calls the method that creates a new task. This is the same functionality as we currently have in MM.
2. The user has previously set a toml as the default but it is no longer supported. A GUI warning message is displayed telling the user that it was unable to read the old toml and the MM original default settings are used.

rmillikin:
it seems that `Toml.ReadFile` is used to read the .toml and if there is an exception in that method, a warning is printed. 

1. does `Toml.ReadFile` throw an exception if there is a key/value pair in the .toml that does not correspond to a variable in the object it's creating?
2. if there is a new variable in CommonParams, does `Toml.ReadFile` throw an exception if there is no value in the .toml for that variable?

zrolfs:
1. Yes
2. No; the default value for the new variable is used.

rmillikin:
I guess that's acceptable behavior for now. It does seem like a bad idea to use the CommonParam default for any new settings without printing out a warning, but I think that's what we do when someone drags in an old .toml. I guess my issue with this PR is that dragging in an old .toml is deliberate, while using a default .toml can quickly get out of date without the user really thinking about potential problems there. 

Ultimately, it would be a good idea to create a method that, upon loading a .toml file, warnings are added to the notification area if variables are not in the .toml.

zrolfs:
Any variable that has a default of null is not written to a toml, so we would have lots of false-positive warnings.

PR #1734, opened 2019-09-20
disabled SILAC turnover. fixed binding reference names

rmillikin:


rmillikin:
#1718 changed a lot of variable names and apparently UI elements that bind to those variables were not renamed and did not throw an error message or warning. let's be careful about mass renames when it includes variables which UI elements bind to

acesnik:
Refactoring usually changes these ... maybe it was a manual change or a merge issue.

PR #1741, opened 2019-10-02
N terminal K is crosslinkable, C terminal K is not

lonelu:


lonelu:
Add code to solve Issue #1754

EngineLayer/CrosslinkSearch/CrosslinkSpectralMatch.cs
```diff
@@ -87,31 +87,12 @@ public static List<int> GetPossibleCrosslinkerModSites(char[] crosslinkerModSite
...
-
-            List<int> range = Enumerable.Range(1, peptide.BaseSequence.Length - 1).ToList();
+            List<int> range = Enumerable.Range(0, peptide.BaseSequence.Length).ToList();
             if (!CrosslinkAtCleavageSite && peptide.OneBasedEndResidueInProtein != peptide.Protein.Length 
                 && !peptide.Protein.ProteolysisProducts.Any(x => x.OneBasedEndPosition == peptide.OneBasedEndResidueInProtein))
             {
-                //The N terminal and C termial cannot be crosslinked and cleaved.
-                range = Enumerable.Range(1, peptide.BaseSequence.Length - 2).ToList();
+                //The C termial cannot be crosslinked and cleaved.
+                range = Enumerable.Range(0, peptide.BaseSequence.Length - 1).ToList();
```
zrolfs: Want a unit test the allows AspN to be crosslinked at the C-terminus, but not at the N-terminus. The current handling appears trypsin/LysC/ArgC-specific.

zrolfs: I think this is fine for right now, but it might crop up as an issue down the line. It depends on the protease mechanism, so I'm not sure we can predict this.

PR #1742, opened 2019-10-02
update mzLib and finalize Variant results summary

rmmiller22:
Update to the most current mzLib that contains all necessary adjustments for variant identifications.
Finalized changes to the variant result summary file and its generation. 

- Separate missense into SNV and MNV

- Identify both peptides and unique variants

- rephrase modification summaries to be more clear

- variant site modification code has been fixed to now function effectively


Test/VariantSearchTests.cs
```diff
@@ -213,106 +219,221 @@ public void VariantSpecificOutputFiles()
...
             Assert.AreEqual("Number of unqiuely identified variant peptides at 1% FDR: 1", checkResults[5]);
-            Assert.AreEqual("Number of SAV variant peptides at 1% FDR: 0", checkResults[6]);
-            Assert.AreEqual("Number of frameshift variant peptides at 1% FDR: 1", checkResults[7]);
-            Assert.AreEqual("Number of inframe insertion variant peptides at 1% FDR: 0", checkResults[8]);
-            Assert.AreEqual("Number of inframe deletion variant peptides at 1% FDR: 0", checkResults[9]);
-            Assert.AreEqual("Number of stop gain variant peptides at 1% FDR: 0", checkResults[10]);
-            Assert.AreEqual("Number of stop loss variant peptides at 1% FDR: 0", checkResults[11]);            
-            Assert.AreEqual("Number of modified variant peptides at 1% FDR: 1", checkResults[12]);
-            Assert.AreEqual("Number of modified variant sites at 1% FDR: 0", checkResults[13]);
+            Assert.AreEqual("Number of unique variants: 1", checkResults[6]);
```
acesnik: Nice test. 

Could consider making these a static list of strings in the postprocessing class, so that you don't have to change them more than once. Could loop through that list and a list of expected results here.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -1017,59 +1030,220 @@ private void WriteVariantResults()
...
+            variantResults[14] = "Number of unique inframe insertion variants: " + insertionSites;
+            variantResults[15] = "Number of inframe deletion variant containing peptides at 1% group FDR: " + deletionCount;
+            variantResults[16] = "Number of unique inframe deletion variants: " + deletionSites;
+            variantResults[17] = "Number of stop gain variant containing peptides at 1% group FDR: " + stopGainCount;
+            variantResults[18] = "Number of unique stop gain variants: " + stopGainSites;
+            variantResults[19] = "Number of stop loss variant containing peptides at 1% group FDR: " + stopLossCount;
+            variantResults[20] = "Number of unique stop loss variants: " + stopLossSites;            
+            variantResults[21] = "Number of variant peptides at 1% group FDR with unambiguous localized modifications: " + modifiedVariantPeptides.Count();
+            variantResults[22] = "Number of variant peptides at 1% group FDR with unambiguous localized modifications at the variant sites : " + modifiedVariantSitePeptides.Count();
```
acesnik: Check that you can't use  `.Count` property, not the `.Count()` method

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -1017,59 +1030,220 @@ private void WriteVariantResults()
...
-            variantResults[9] = "Number of inframe deletion variant peptides at 1% FDR: " + deletionCount;
-            variantResults[10] = "Number of stop gain variant peptides at 1% FDR: " + stopGainCount;
-            variantResults[11] = "Number of stop loss variant peptides at 1% FDR: " + stopLossCount;            
-            variantResults[12] = "Number of modified variant peptides at 1% FDR: " + modifiedVariantPeptides.Count();
-            variantResults[13] = "Number of modified variant sites at 1% FDR: " + modifiedVariantSitePeptides.Count();
+            variantResults[4] = "Number of potential variant containing peptides identified at 1% group FDR: " + variantPeptides.Where(p => p.IsDecoy ==false &&  p.IsContaminant ==false && p.FdrInfo.QValue <= 0.01 &&p.FdrInfo.QValueNotch <= 0.01).ToList().Count();
+            variantResults[5] = "Number of unqiuely identified variant peptides at 1% group FDR: " + confidentVariantPeps.Where(p => p.IsDecoy == false &&  p.IsContaminant == false && p.FdrInfo.QValue <= 0.01 && p.FdrInfo.QValueNotch <= 0.01).ToList().Count();
```
acesnik: `=> !p.IsDecoy && !p.IsContaminant` is a bit more typical

Check that you can't use  `.Count` property, not the `.Count()` method

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -1017,59 +1030,220 @@ private void WriteVariantResults()
...
 
                         }
                     }
                 
                 }
             }
+            int SNVmissenseSites = 0;
+            foreach (var entry in SNVmissenseVariants)
+            {
+                SNVmissenseSites += entry.Value.Count();
```
acesnik: Check that you can't use  `.Count` property, not the `.Count()` method

acesnik: And for the loops below. This leads to speedups especially in loops.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -1017,59 +1030,220 @@ private void WriteVariantResults()
             }
             foreach (var peptide in confidentVariantPeps)
             {
-                if (peptide.FdrInfo.QValue <= 0.01 && peptide.FdrInfo.QValueNotch <= 0.01 && peptide.IsDecoy == false)
+                if (peptide.FdrInfo.QValue <= 0.01 && peptide.FdrInfo.QValueNotch <= 0.01 && peptide.IsDecoy == false && peptide.IsContaminant == false)
```
acesnik: `=> !p.IsDecoy && !p.IsContaminant` is a bit more typical

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -966,46 +957,68 @@ private void WriteVariantResults()
...
             List<PeptideSpectralMatch> modifiedVariantSitePeptides = new List<PeptideSpectralMatch>();// modification is speciifcally on the variant residue within the peptide
             foreach (var entry in modifiedVariantPeptides)
-            {
-                var variants = entry.BestMatchingPeptides.SelectMany(p => p.Peptide.Protein.AppliedSequenceVariations.Where(v =>p.Peptide.IntersectsAndIdentifiesVariation(v).identifies == true).ToList()).ToList();
-                var mods = entry.ModsIdentified;
+            {                
+                var variantPWSM = entry.BestMatchingPeptides.FirstOrDefault().Peptide;
+                var peptideMods = variantPWSM.AllModsOneIsNterminus.Values.ToList();                
+                var variantProteinModifications = variantPWSM.Protein.OneBasedPossibleLocalizedModifications.Where(k => k.Key >= variantPWSM.OneBasedStartResidueInProtein && k.Key <= variantPWSM.OneBasedEndResidueInProtein).ToList();
+                var variants = entry.BestMatchingPeptides.SelectMany(p => p.Peptide.Protein.AppliedSequenceVariations.Where(v => p.Peptide.IntersectsAndIdentifiesVariation(v).identifies == true).ToList()).ToList();
```
acesnik: I don't think you need the internal `.ToList()`, and it might take more time to construct the list multiple times.

`.Where(v => p.Peptide.IntersectsAndIdentifiesVariation(v).identifies)` is a more typical pattern.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -966,46 +957,68 @@ private void WriteVariantResults()
...
+            Dictionary<Protein, HashSet<SequenceVariation>> SNVmissenseVariants = new Dictionary<Protein, HashSet<SequenceVariation>>();
+            Dictionary<Protein, HashSet<SequenceVariation>> insertionVariants = new Dictionary<Protein, HashSet<SequenceVariation>>();
+            Dictionary<Protein, HashSet<SequenceVariation>> deletionVariants = new Dictionary<Protein, HashSet<SequenceVariation>>();
+            Dictionary<Protein, HashSet<SequenceVariation>> frameshiftVariants = new Dictionary<Protein, HashSet<SequenceVariation>>();
+            Dictionary<Protein, HashSet<SequenceVariation>> stopGainVariants = new Dictionary<Protein, HashSet<SequenceVariation>>();
+            Dictionary<Protein, HashSet<SequenceVariation>> stopLossVariants = new Dictionary<Protein, HashSet<SequenceVariation>>();            
+
+            List<PeptideSpectralMatch> modifiedVariantPeptides = confidentVariantPeps.Where(p => p.ModsIdentified != null && p.ModsIdentified.Count() > 0 && p.FdrInfo.QValue <= 0.01 && p.FdrInfo.QValueNotch <= 0.01 && p.IsDecoy == false && p.IsContaminant ==false).ToList(); //modification can be on any AA in variant peptide
```
acesnik: Check whether you can use the `.Count` property rather than `.Count()`
Consider using `!p.IsDecoy && !p.IsContaminant`

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -966,46 +957,68 @@ private void WriteVariantResults()
...
 
             List<PeptideSpectralMatch> variantPeptides = possibleVariantPsms.GroupBy(b => b.FullSequence).Select(b => b.FirstOrDefault()).ToList();
             List<PeptideSpectralMatch> confidentVariantPeps = new List<PeptideSpectralMatch>();
-            WritePsmsToTsv(variantPeptides,variantPeptideFile, Parameters.SearchParameters.ModsToWriteSelection);            
+
+            new FdrAnalysisEngine(variantPeptides, Parameters.NumNotches, CommonParameters, new List<string> { Parameters.SearchTaskId }, "variant_Peptides").Run();
+
+            WritePsmsToTsv(variantPeptides,variantPeptideFile, Parameters.SearchParameters.ModsToWriteSelection);    
+            
             foreach (var entry in variantPeptides)
```
acesnik: Could you shorten this `foreach` with a linq operation using `.Any()`?

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -946,17 +945,9 @@ private void WriteVariantResults()
...
-            {
-                var variants = peptide.BestMatchingPeptides.SelectMany(p => p.Peptide.Protein.AppliedSequenceVariations.Where(v => p.Peptide.IntersectsAndIdentifiesVariation(v).identifies).ToList()).ToList();
-                if(variants.Count > 0)
-                {
-                        possibleVariantPsms.Add(peptide);                                        
-                }                
-            }
-            
+                && p.FdrInfo.QValueNotch <= CommonParameters.QValueOutputFilter && p.BaseSequence != null).ToList();             
+            var possibleVariantPsms = FDRPsms.Where(p => p.BestMatchingPeptides.Any(pep => pep.Peptide.IsVariantPeptide() == true)).ToList();
```
acesnik: Consider `.Any(pep => pep.Peptide.IsVariantPeptide()).ToList()`

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -1017,59 +1023,220 @@ private void WriteVariantResults()
             }
             foreach (var peptide in confidentVariantPeps)
             {
-                if (peptide.FdrInfo.QValue <= 0.01 && peptide.FdrInfo.QValueNotch <= 0.01 && peptide.IsDecoy == false)
+                if (peptide.FdrInfo.QValue <= 0.01 && peptide.FdrInfo.QValueNotch <= 0.01 && !peptide.IsDecoy && !peptide.IsContaminant)
```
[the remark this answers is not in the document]
rmmiller22: The first set of cutoffs at line 946 exist to filter the output that is written in the variant specific peptide output file. The cutoffs applied here are utilized for the variant result summary file in which we want to count the number of variants of each type at 1% group FDR which may be different from the CommonParameter settings. We also want to exclude any variant or contaminant peptides from these calculations so that the totals provided are only target peptide identifications.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -1017,59 +1023,220 @@ private void WriteVariantResults()
...
+                    bool MNVmissenseIdentified = false;
+                    bool insertionIdentified = false;
+                    bool deletionIdentified = false;
+                    bool frameshiftIdentified = false;
+                    bool stopGainIdentified = false;
+                    bool stopLossIdentifed = false;
+
                     foreach (var variant in variants)
                     {                       
                         if (variantPWSM.Peptide.IntersectsAndIdentifiesVariation(variant).identifies == true)
```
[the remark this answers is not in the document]
rmmiller22: This unit test does not read in a fake database made by spritz. Instead a protein database is generated internally with a fake protein and fake variant. Becasue of this, the variant description does not follow the same format as what is produced by spritz and doesn't contain all of the information such as the variant type. This is why the SNVmissenseCount is not incremented. This is not an error in the code, and the string that is passed in as the varaint description could be altered to contain "missense_variant"  if deemed necessary.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -1017,59 +1023,220 @@ private void WriteVariantResults()
...
+                                        MNVmissenseVariants.Add(variantPWSM.Peptide.Protein, new HashSet<SequenceVariation> { variant });
+                                    }
+                                }                                                                                         
+                                
                             }
+                            
                             else if (culture.CompareInfo.IndexOf(variant.Description.Description, "frameshift_variant", CompareOptions.IgnoreCase) >= 0)
                             {
-                                frameshiftCount++;
+                                if (frameshiftIdentified == false)
```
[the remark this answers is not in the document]
rmmiller22: The peptide identified in this unit test terminated 1 AA prior to the start of the frameshift variant. Therefore it does not intersect with the variant and cannot be used to identify the variant peptide, which is why the variant count doe snot increment. This test is behaving as is expected based on how it is written.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -1017,59 +1023,220 @@ private void WriteVariantResults()
...
+                    bool deletionIdentified = false;
+                    bool frameshiftIdentified = false;
+                    bool stopGainIdentified = false;
+                    bool stopLossIdentifed = false;
+
                     foreach (var variant in variants)
                     {                       
                         if (variantPWSM.Peptide.IntersectsAndIdentifiesVariation(variant).identifies == true)
                         {
                             if (culture.CompareInfo.IndexOf(variant.Description.Description, "missense_variant", CompareOptions.IgnoreCase) >= 0)
```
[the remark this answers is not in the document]
rmmiller22: I think I addressed this in a comment below. Some of the unit tests written before the results summary code was added do not have complete variant descriptions, which is why the tests I wrote in VariantSpecificOutputFiles() include all variant types with complete descriptions

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -1017,59 +1023,220 @@ private void WriteVariantResults()
...
+                                        SNVmissenseIdentified = true;
+                                    }
+                                    if (SNVmissenseVariants.ContainsKey(variantPWSM.Peptide.Protein))
+                                    {
+                                        SNVmissenseVariants[variantPWSM.Peptide.Protein].Add(variant);
+                                    }
+                                    else
+                                    {
+                                        SNVmissenseVariants.Add(variantPWSM.Peptide.Protein, new HashSet<SequenceVariation> { variant });
+                                    }
```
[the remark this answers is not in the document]
rmmiller22: else if statements are used for subsequent variant types achieving the same thing as "continue;"

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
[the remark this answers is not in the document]
trishorts: still feel that the cut-offs should be coincident. if user specifies 5% then 5% for all seems right. No issue with filtering decoys/contaminants. Doubtful that any contaminants have variants and so I don't think that's probably a major thing.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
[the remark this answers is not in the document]
trishorts: should we make it so that a defined variant peptides has defined/required fields that will provide the necessary data for correct analysis downstream?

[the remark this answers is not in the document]
rmmiller22: I do not think that is necessary. The SequenceVariation object already has defined/ required fields. The variant description would normally be extracted from the proteogenomic database and would always contain the variant type. But since a fake variant was made, a fake variant description was made that did not have the variant type in it which would only occur in the case of a unit test.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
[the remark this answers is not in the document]
rmmiller22: The default for the qValue output filter is 1.0 meaning that all peptides or psms up to 100% FDR is written into the psm file, not 1% FDR or (0.01). I do not think that we should have this value be set to the default variant result filter for the output text file. I will do some work to try and accommodate, if the value if no the default, to have the results be filtered to the user selected FDR

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -1021,58 +1030,221 @@ private void WriteVariantResults()
...
+            foreach (var entry in stopLossVariants)
+            {
+                stopLossSites += entry.Value.Count;
+            }
+
+            int totalVariantSites = SNVmissenseSites + MNVmissenseSites+ insertionSites + deletionSites + frameshiftSites + stopGainSites + stopLossSites;          
+
+            
+
+            string[] variantResults = new string[25];
```
rmillikin: this is probably better implemented as a `List<string>` instead of a fixed-length array but it's fine for now

Issue #1711, opened 2019-08-22
Hyperlinks in release notes text don't work


rmillikin:
Sometimes I put links in release notes. These work in GitHub, but not MetaMorpheus's autoupdater. Only the text is scraped off the internet, not the link itself. It would be nice if we had some way to get the link too.

rmillikin:
fixed in #1751

PR #1751, opened 2019-10-09
Fix issue #1711

rudderbucky:
Parsed the GitHub release notes from markdown into HTML and displayed it in a WebBrowser.


zrolfs:
#1711

Issue #1768, opened 2019-11-12
Bug in search task quant SILAC


zrolfs:
Save a task with a SILAC quantification label specified, open the task it, change it to no quantification, save it, reopen it, and it will say SILAC instead of no quant.

zrolfs:
fixed in #1769

PR #1769, opened 2019-11-12
191112 silac bug gui

zrolfs:
#1768


trishorts:
The gui bug seems fixed. But, in the process of checking, i noticed that none of the silac stuff is saved in the toml. if you do a silac search and then drag that toml into a new search, nothing about silac was saved. Is this the desired behavior? Seems like it would be handy to have the silac stuff saved so you could easily repeat the search.

zrolfs:
I'm unable to reproduce this behavior.

Issue #1767, opened 2019-11-12
Have Top-down protease selection automatically update recommended parameters


zrolfs:
Similar to what's already being done for non-specific searches

rmillikin:
done in #1770

PR #1770, opened 2019-11-12
191112 top down recommendations

zrolfs:
#1767


GUI/SearchTaskWindow.xaml.cs
```diff
@@ -729,7 +729,26 @@ private void NonSpecificUsingNonSpecific(object sender, RoutedEventArgs e)
             }
         }
 
-        private void NonSpecificUpdate(object sender, TextChangedEventArgs e)
+        //this one is used by the GUI
+        private void ProteaseSpecificUpdate(object sender, SelectionChangedEventArgs e)
+        {
+            const int maxLength = 25;
+            if (((Protease)ProteaseComboBox.SelectedItem).Name.Contains("non-specific"))
```
trishorts: should the proteases be enum type? in mzlib?

GUI/SearchTaskWindow.xaml.cs
```diff
@@ -729,7 +729,26 @@ private void NonSpecificUsingNonSpecific(object sender, RoutedEventArgs e)
...
+            {
+                MaxPeptideLengthTextBox.Text = maxLength.ToString();
+            }
+            else if (((Protease)ProteaseComboBox.SelectedItem).Name.Contains("top-down"))
+            {
+                UseProvidedPrecursor.IsChecked = false;
+                DeconvolutionMaxAssumedChargeStateTextBox.Text = "60";
```
trishorts: 60 is very high in my experience doing td searches in MM. maybe 2X too high. If there are no practical consequences (e.g. longer search times, larger search spaces, more false positives), then no big deal. But i suspect there is a cost and that this number should be lower

GUI/SearchTaskWindow.xaml.cs
zrolfs: No, they are complex objects that include digestion rules.

GUI/SearchTaskWindow.xaml.cs
zrolfs: 60 was pulled from the wiki's recommendations: https://github.com/smith-chem-wisc/MetaMorpheus/wiki/Top-Down

I have no opinion, but we should be consistent in what we recommend.

trishorts: LVS has seen high MW proteoforms up to c.s. of 50. So maybe 60 is fine.

trishorts:
Do we need defaults for other proteases? I like that when I select non-specific or top-down that the defaults automatically change. But, when I go back to trypsin, it keeps the ns or td settings. Then I have to manually change all of them or delete the task. So, if we could poplulate the defaults for each selected protease, that'd be great.

zrolfs:
My concern with adding defaults for other proteases is that the code becomes harder to maintain. There would be two "generic" defaults: one in the load params and one in the protease selection.
I propose we create a class property that stores the initial parameters (pre-top-down) and reverts to those when the protease is changed. It would only revert the handful of params that are changed in top-down/non-specific.

trishorts:
yes. that's what i wnat

rmillikin:
How about when you switch to top-down protease a popup is generated saying "we recommend the following settings, do you want to apply them?" or something like that. That way if someone misclicks the top-down protease their settings don't get messed up. And we probably don't need to store the "old" settings at that point, since the user is at least informed that there was a settings change (along with what changed).

rmillikin:
I do really like the "custom settings for each protease" idea though. Chymotrypsin settings shouldn't necessarily look like trypsin settings, etc.

zrolfs:
I like the idea of asking the user. I'll include a checkbox like "don't ask me again".
Are there any suggestions for recommended protease-specific parameters? I propose semi-specific digestions for chymotrypsin and elastase. Maybe a suggestion using trypsin when users select Arg-C?
Source: 
image

zrolfs:
#1774

GUI/Util/UpdateGUISettings.cs
```diff
@@ -0,0 +1,253 @@
...
+        }
+
+        public static bool UseTopDownRecommendedSettings()
+        {
+            bool useRecommendedSettings = false;
+            //check with the user to update params
+            if (Params.AskAboutTopDownParams )
+            {
+                var results = ProteaseSpecificMsgBox.Show("Use Top-Down Recommendations?",
+                    "We recommend using the following parameters for top-down searches:\n" +
```
rmillikin: I also recommend variable mods be disabled for top-down searches - want to add this?

EngineLayer/ProteolyticDigestion/proteases.tsv
```diff
@@ -4,7 +4,7 @@ Asp-N	|D			full	MS:1001304	Asp-N	(?=[BD])
 chymotrypsin (don't cleave before proline)	"F[P]|,W[P]|,Y[P]|"			full	MS:1001306	Chymotrypsin	(?<=[FYWL])(?!P)	
 chymotrypsin (cleave before proline)	"F|,W|,Y|"			full	MS:1001306	Chymotrypsin	(?<=[FYWL])	
 CNBr	M|			full	MS:1001307	CNBr	(?<=M)	
-Elastase	"A|,V|,S|,G|,L|,I|"			full		Elastase	(?<=[AVSGLI])	
+elastase	"A|,V|,S|,G|,L|,I|"			full		Elastase	(?<=[AVSGLI])
```
rmillikin: Great PR 👍

GUI/Util/UpdateGUISettings.cs
zrolfs: Agreed. If GPTMD is done and includes oxidation of M, then variable mods should not be done.

Without GPTMD, using a .fasta database (yeast):
1. Variable Ox
 PSMs: 509
 Time: 41 sec

2. No Variable Mods
 PSMs: 723
 Time: 22 minutes

So it's worth looking for oxidation, but there's a 30 fold increase in search time to do it using a variable search.

PR #1775, opened 2019-11-14
faster modern search

rmillikin:
I replaced some code in modern search that in my hands is about 6x faster than the current code. 

I've added some .tomls that I've been using to benchmark. Current MM takes ~12 min or so for the completely open search for a BU yeast file; PR code is ~2 min. An interval search from [-10,200] takes about 26 seconds with this PR. Note that I didn't change indexing, etc. so that part is still the same speed and can be further optimized.

[ModernSearchTest.zip](https://github.com/smith-chem-wisc/MetaMorpheus/files/3848901/ModernSearchTest.zip)

### DIFF
--- a/EngineLayer/ModernSearch/ModernSearchEngine.cs
+++ b/EngineLayer/ModernSearch/ModernSearchEngine.cs
@@ -60,7 +60,7 @@ namespace EngineLayer.ModernSearch
                         IndexScoreScan(scan, scoringTable, byteScoreCutoff, idsOfPeptidesPossiblyObserved, CommonParameters.DissociationType);
 
                         // take indexed-scored peptides and re-score them using the more accurate but slower scoring algorithm
-                        FineScorePeptides(idsOfPeptidesPossiblyObserved, scan, scanIndex, scoringTable);
+                        FineScorePeptides(idsOfPeptidesPossiblyObserved, scan, scanIndex, scoringTable, CommonParameters.DissociationType);
 
                         //report search progress
                         progress++;
@@ -79,7 +79,7 @@ namespace EngineLayer.ModernSearch
             {
                 psm.ResolveAllAmbiguities();
             }
-            
+
             return new MetaMorpheusEngineResults(this);
         }
 
@@ -347,7 +347,7 @@ namespace EngineLayer.ModernSearch
             return PeptideSpectralMatches[scanIndex];
         }
 
-        protected void FineScorePeptides(List<int> peptideIds, Ms2ScanWithSpecificMass scan, int scanIndex, byte[] scoringTable)
+        protected void FineScorePeptides(List<int> peptideIds, Ms2ScanWithSpecificMass scan, int scanIndex, byte[] scoringTable, DissociationType dissociationType)
         {
             // this method re-scores the top-scoring peptides until no peptide in the rough-scored list can out-score
             // the best-scoring peptide. this guarantees that peptides will be scored accurately, according to metamorpheus score,
@@ -358,7 +358,7 @@ namespace EngineLayer.ModernSearch
             
             foreach (int id in peptideIds.OrderByDescending(p => scoringTable[p]))
             {
-                if (scoringTable[id] < bestScore)
+                if (scoringTable[id] < bestScore && dissociationType != DissociationType.LowCID)
                 {
                     FineScorePeptide(id, scan, scanIndex);
 
--- a/EngineLayer/ModernSearch/ModernSearchEngine.cs
+++ b/EngineLayer/ModernSearch/ModernSearchEngine.cs
@@ -54,6 +54,12 @@ namespace EngineLayer.ModernSearch
 
                     for (int scanIndex = range.Item1; scanIndex < range.Item2; scanIndex++)
                     {
+                        // Stop loop if canceled
+                        if (GlobalVariables.StopLoops)
+                        {
+                            return;
+                        }
+
                         Ms2ScanWithSpecificMass scan = ListOfSortedMs2Scans[scanIndex];
 
                         // do a fast rough first-pass scoring for this scan
@@ -89,11 +95,15 @@ namespace EngineLayer.ModernSearch
         /// </summary>
         protected void IndexScoreScan(Ms2ScanWithSpecificMass scan, byte[] scoringTable, byte byteScoreCutoff, List<int> peptidesPossiblyObserved, DissociationType dissociationType)
         {
+            // get allowed theoretical masses from the known experimental mass
+            // note that this is the OPPOSITE of the classic search (which calculates experimental masses from theoretical values)	
+            // this is just PRELIMINARY precursor-mass filtering	
+            // additional checks are made later to ensure that the theoretical precursor mass is acceptable
             List<AllowedIntervalWithNotch> notches = MassDiffAcceptor.GetAllowedPrecursorMassIntervalsFromObservedMass(scan.PrecursorMass).ToList();
-
             double lowestMassPeptideToLookFor = notches.Min(p => p.AllowedInterval.Minimum);
             double highestMassPeptideToLookFor = notches.Max(p => p.AllowedInterval.Maximum);
 
+            // clear the scoring table to score the new scan (conserves memory compared to allocating a new array)
             Array.Clear(scoringTable, 0, scoringTable.Length);
             peptidesPossiblyObserved.Clear();
 
@@ -104,7 +114,7 @@ namespace EngineLayer.ModernSearch
 
                 for (int i = 0; i < masses.Length; i++)
                 {
-                    //convert to an int since we're in discreet 1.0005...
+                    //convert to an int since we're in discrete 1.0005...
                     int fragmentBin = (int)(Math.Round(masses[i].ToMass(1) / 1.0005079) * 1.0005079 * FragmentBinsPerDalton);
 
                     List<int> bin = FragmentIndex[fragmentBin];
@@ -355,7 +365,7 @@ namespace EngineLayer.ModernSearch
             // analysis trains on delta score, the modern search output is not guaranteed to be the same as classic search 
             // output, though it will be very close.
             byte bestScore = 0;
-            
+
             foreach (int id in peptideIds.OrderByDescending(p => scoringTable[p]))
             {
                 if (scoringTable[id] < bestScore && dissociationType != DissociationType.LowCID)
--- a/EngineLayer/ModernSearch/ModernSearchEngine.cs
+++ b/EngineLayer/ModernSearch/ModernSearchEngine.cs
@@ -44,7 +44,7 @@ namespace EngineLayer.ModernSearch
                 CurrentPartition + "/" + CommonParameters.TotalPartitions, NestedIds));
 
             byte byteScoreCutoff = (byte)CommonParameters.ScoreCutoff;
-            
+
             Parallel.ForEach(Partitioner.Create(0, ListOfSortedMs2Scans.Length),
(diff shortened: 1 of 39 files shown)


EngineLayer/ModernSearch/ModernSearchEngine.cs
```diff
@@ -39,84 +40,44 @@ protected override MetaMorpheusEngineResults RunSpecific()
...
+                        Ms2ScanWithSpecificMass scan = ListOfSortedMs2Scans[scanIndex];
 
-                        bool meetsScoreCutoff = thisScore >= CommonParameters.ScoreCutoff;
-                        bool scoreImprovement = PeptideSpectralMatches[i] == null || (thisScore - PeptideSpectralMatches[i].RunnerUpScore) > -PeptideSpectralMatch.ToleranceForScoreDifferentiation;
+                        // do a fast rough first-pass scoring for this scan
+                        IndexScoreScan(scan, scoringTable, byteScoreCutoff, idsOfPeptidesPossiblyObserved, CommonParameters.DissociationType);
 
-                        if (meetsScoreCutoff && scoreImprovement)
+                        // take top 10 indexed-scored peptides and re-score them using the more accurate but slower scoring algorithm
+                        // NOTE: this "Take 10" section means that modern search is NOT GUARANTEED to produce the same results as classic search
```
zrolfs: :(

rmillikin: there are ways around this, fortunately. we could include the distance from the bin edge with each fragment in the index - so instead of just being a peptide ID number, it would be a peptide ID number and a dalton "error" from the start or end of the bin, giving an exact dalton value rather than approximation. I'm pretty sure this would be slightly faster than the current implementation. it would of course cost more hard disk space/RAM.

EngineLayer/ModernSearch/ModernSearchEngine.cs
```diff
@@ -126,10 +87,19 @@ protected override MetaMorpheusEngineResults RunSpecific()
...
 
-        protected List<int> GetBinsToSearch(Ms2ScanWithSpecificMass scan, List<int>[] FragmentIndex, DissociationType dissociationType)
+        /// <summary>
+        /// This is a first-pass scoring method which is supposed to be *very fast* but may sometimes miscount the number of truly matched fragments. The number
+        /// of matched fragments should always be overestimated, never underestimated.
+        /// </summary>
+        protected void IndexScoreScan(Ms2ScanWithSpecificMass scan, byte[] scoringTable, byte byteScoreCutoff, List<int> peptidesPossiblyObserved, DissociationType dissociationType)
         {
-            int obsPreviousFragmentCeilingMz = 0;
-            List<int> binsToSearch = new List<int>();
```
rmillikin: this is just a guess, but I think line 132 of the old code may be why modern search is currently so slow. it's allocating a new list for every fragment in the data file. these lists are collected by the garbage collector, which stops all threads to perform garbage collection.

EngineLayer/ModernSearch/ModernSearchEngine.cs
lonelu: Not sure about how this could be faster.

lonelu: What Shortreed did was: find the score of the 10th id, then take all the ids >= the score.

lonelu: It is possible to keep all the 10 for PEP...

EngineLayer/ModernSearch/ModernSearchEngine.cs
```diff
@@ -141,9 +111,15 @@ protected List<int> GetBinsToSearch(Ms2ScanWithSpecificMass scan, List<int>[] Fr
                     //convert to an int since we're in discreet 1.0005...
                     int fragmentBin = (int)(Math.Round(masses[i].ToMass(1) / 1.0005079) * 1.0005079 * FragmentBinsPerDalton);
 
-                    if (FragmentIndex[fragmentBin] != null)
+                    List<int> bin = FragmentIndex[fragmentBin];
```
lonelu: If my understanding is right. This here is same as "List<int> binsToSearch = new List<int>();" in the old way.

EngineLayer/ModernSearch/ModernSearchEngine.cs
lonelu: It seems like to me the take(10) helps a lot. Not sure about this, you use "List<int> bin = FragmentIndex[fragmentBin];" in the new method.

rmillikin: I was wrong - the current code is allocating a new list for each scan, not for each fragment.

to test my hypothesis I inserted a `List<int> temp = new List<int>();` onto line 96 and added each valid bin to the list.

without new list: 127 sec
with new list, bins added to it: 127 sec

so that doesn't seem like it's the culprit.

I also tested how much time the "top 10" filter saved:
with top 10 filter: 127 sec
without filter: ~2625 sec (estimated based on 12% of the search completing in 315 sec)

So it's definitely the fine-scoring part that's slow. Based on my CPU usage (it's spikey instead of flat) I'm guessing it's because of frequent garbage collection. I'll see if I can optimize this part...

EngineLayer/ModernSearch/ModernSearchEngine.cs
rmillikin: `List binsToSearch = new List();` creates a new list, `List<int> bin = FragmentIndex[fragmentBin];` creates a new reference to an existing list

EngineLayer/ModernSearch/ModernSearchEngine.cs
lonelu: I think it is reasonable to only fine score top N. 
In Crosslink search, the alpha peptide always ranks within the top 50. You can output the rank to see if this is the truth.

rmillikin: I replaced the top N with a function which always accurately computes the best peptide score (it fine-scores all the peptides that could be the best-scoring), but the delta-score is only an approximation because it takes too long to score all of the peptides that could be second-best scoring. 

because the delta score is only approximated, and PEP analysis trains on delta score, the PEP values are different for classic and modern search. and because some ambiguity is removed based on PEP, some ambiguity is also slightly different. I think it's not an important degree of difference, though.

rmillikin: speed on my benchmark test is maintained at 127sec, btw, so this method is just as fast as re-scoring the top 10 and is more accurate.

rmillikin: building in release took ~60 seconds... 😵

rmillikin:
code coverage decreases because some methods are no longer used by modern search (hence, not hit by modern search unit tests). i left these in place because other search engines use them (nonspecific, XL)

EngineLayer/ModernSearch/ModernSearchEngine.cs
```diff
@@ -39,84 +40,40 @@ protected override MetaMorpheusEngineResults RunSpecific()
...
-                    double highestMassPeptideToLookFor = notches.Max(p => p.AllowedInterval.Maximum);
-
-                    // first-pass scoring
-                    IndexedScoring(FragmentIndex, allBinsToSearch, scoringTable, byteScoreCutoff, idsOfPeptidesPossiblyObserved, scan.PrecursorMass, lowestMassPeptideToLookFor, highestMassPeptideToLookFor, PeptideIndex, MassDiffAcceptor, MaxMassThatFragmentIonScoreIsDoubled, CommonParameters.DissociationType);
+                    byte[] scoringTable = new byte[PeptideIndex.Count];
+                    List<int> idsOfPeptidesPossiblyObserved = new List<int>(PeptideIndex.Count);
 
-                    // done with indexed scoring; refine scores and create PSMs
-                    foreach (int id in idsOfPeptidesPossiblyObserved)
+                    for (int scanIndex = range.Item1; scanIndex < range.Item2; scanIndex++)
```
zrolfs: We're losing a parallelization optimization by breaking up the msFile into continuous chunks rather than distributing scans every other. Is there a reason or is it an oversight? Previous code was ln53.

EngineLayer/ModernSearch/ModernSearchEngine.cs
```diff
@@ -39,84 +40,40 @@ protected override MetaMorpheusEngineResults RunSpecific()
...
-                byte[] scoringTable = new byte[PeptideIndex.Count];
-                List<int> idsOfPeptidesPossiblyObserved = new List<int>();
-
-                for (; i < ListOfSortedMs2Scans.Length; i += maxThreadsPerFile)
+            Parallel.ForEach(Partitioner.Create(0, ListOfSortedMs2Scans.Length),
+                new ParallelOptions { MaxDegreeOfParallelism = CommonParameters.MaxThreadsToUsePerFile },
+                (range, loopState) =>
                 {
-                    // Stop loop if canceled
-                    if (GlobalVariables.StopLoops)
```
zrolfs: Should keep this in?

EngineLayer/ModernSearch/ModernSearchEngine.cs
rmillikin: I did not see any meaningful change and this code is more interpretable to me but I can do a more thorough benchmark

EngineLayer/ModernSearch/ModernSearchEngine.cs
```diff
@@ -39,84 +40,40 @@ protected override MetaMorpheusEngineResults RunSpecific()
...
-
-                    // empty the scoring table to score the new scan (conserves memory compared to allocating a new array)
-                    Array.Clear(scoringTable, 0, scoringTable.Length);
-                    idsOfPeptidesPossiblyObserved.Clear();
-                    Ms2ScanWithSpecificMass scan = ListOfSortedMs2Scans[i];
-
-                    // get fragment bins for this scan
-                    List<int> allBinsToSearch = GetBinsToSearch(scan, FragmentIndex, CommonParameters.DissociationType);
-
-                    // get allowed theoretical masses from the known experimental mass
```
zrolfs: I still like this comment to help describe what we're doing for the indexed search.

zrolfs:
It seems to me that most of the code is conserved, just moved around into helper methods. 

I don't see anything obvious to attribute the increase in speed except that we are no longer fine scoring all peptides above the score threshold.

rmillikin:
there are some streamlined areas outside of the fine-scoring. the bins aren't added to a list, for example

EngineLayer/ModernSearch/ModernSearchEngine.cs
zrolfs: Did you get results of the benchmark?

rmillikin: yep - `Parallel.ForEach(threads, (i) =>` is about 10% faster than `Parallel.ForEach(Partitioner.Create(0, ListOfSortedMs2Scans.Length)...`. I'm hoping to further optimize this with clearer syntax in the future...

rmillikin: that's in release mode, btw

zrolfs:
There are some easy unit tests (lowCID, compIons) that would increase code coverage for this if you're willing to put them in.

rmillikin:
the code coverage is negatively effected only because the deprecated methods aren't hit by the modern search unit tests anymore. the new code is hit by the old existing tests. when the nonspecific and XL searches don't use the deprecated code then we can delete that old stuff and the code cov will increase again

PR #1780, opened 2019-11-26
Faster crosslink search

lonelu:
Faster crosslink search


EngineLayer/CrosslinkSearch/CrosslinkSearchEngine.cs
```diff
@@ -184,174 +181,186 @@ private void GenerateCrosslinkModifications(Crosslinker crosslinker)
...
 
-        /// <summary>
-        /// 
-        /// </summary>
-        private List<CrosslinkSpectralMatch> FindCrosslinkedPeptide(Ms2ScanWithSpecificMass theScan, List<BestPeptideScoreNotch> theScanBestPeptide, int scanIndex)
-        {
-            List<CrosslinkSpectralMatch> possibleMatches = new List<CrosslinkSpectralMatch>();
 
-            for (int alphaIndex = 0; alphaIndex < theScanBestPeptide.Count; alphaIndex++)
+        private void FindCrosslinkedPeptide(Ms2ScanWithSpecificMass scan, List<int> idsOfPeptidesPossiblyObserved, byte[] scoringTable, double[] massTable, byte byteScoreCutoff, int scanIndex, List<CrosslinkSpectralMatch> possibleMatches, HashSet<Tuple<int, int>> seenPair)
```
trishorts: add comments

EngineLayer/CrosslinkSearch/CrosslinkSpectralMatch.cs
```diff
@@ -34,7 +34,9 @@ public bool IsIntraCsm()
             
             if (this.ProteinAccession != null && this.BetaPeptide.ProteinAccession != null)
             {
-                if (this.ProteinAccession == this.BetaPeptide.ProteinAccession)
+                if (this.ProteinAccession == this.BetaPeptide.ProteinAccession || 
+                    this.ProteinAccession == "DECOY_"+ this.BetaPeptide.ProteinAccession ||
```
trishorts: does "this.ProteinAccession" really equal "DECOY_" and not something like "DECOY_P01234". Maybe this should just be looking for string to contain DECOY?

EngineLayer/CrosslinkSearch/CrosslinkSpectralMatch.cs
```diff
@@ -82,17 +85,18 @@ public void ResolveProteinPosAmbiguitiesForXl()
             }
         }
 
-        public static List<int> GetPossibleCrosslinkerModSites(char[] crosslinkerModSites, PeptideWithSetModifications peptide, InitiatorMethionineBehavior initiatorMethionineBehavior, bool CrosslinkAtCleavageSite)
+        public static void GetPossibleCrosslinkerModSites(char[] crosslinkerModSites, PeptideWithSetModifications peptide, InitiatorMethionineBehavior initiatorMethionineBehavior, bool CrosslinkAtCleavageSite, out List<int> possibleXlPositions)
```
trishorts: Why not public List<int> GetPossibleCrosslinkerModSites? Having "out" seems wrong and I think codacy will complain

GUI/XLSearchTaskWindow.xaml
```diff
@@ -81,7 +81,7 @@
                             <ComboBox x:Name="cbbXLprecusorMsTl" />
                         </StackPanel>
                         <StackPanel Orientation="Horizontal" Grid.Row="1" Grid.Column="1" Margin="5">
-                            <CheckBox x:Name="ckbXLTopNum" VerticalAlignment="Center" Content="Top N peptides per precursor" />
+                            <Label Content="Keep top N candidates" />
```
trishorts: Does this really need to be user settable?

TaskLayer/XLSearchTask/PostXLSearchAnalysisTask.cs
```diff
@@ -58,12 +58,6 @@ public MyTaskResults Run(string OutputFolder, List<DbForTask> dbFilenameList, Li
             }
             MyTaskResults.AddTaskSummaryText("Target inter-crosslinks within 1% FDR: " + interCsms.Count(p => p.FdrInfo.QValue <= 0.01 && !p.IsDecoy && !p.BetaPeptide.IsDecoy));
 
-            if (xlSearchParameters.WriteOutputForPercolator)
```
trishorts: I'd like to keep the percolator output.

TaskLayer/XLSearchTask/XLSearchTaskOutputMethod.cs
```diff
@@ -42,48 +42,6 @@ public static void WritePsmCrossToTsv(List<CrosslinkSpectralMatch> items, string
             }
         }
         
-        public static void WriteCrosslinkToTxtForPercolator(List<CrosslinkSpectralMatch> items, string outputFolder, string fileName, Crosslinker crosslinker)
```
trishorts: +1 for keeping percolator

Test/XLTest.cs
```diff
@@ -603,34 +613,6 @@ public static void TestTheoreticalFragmentsCleavableCrosslink()
             Assert.That(signatureIons.Select(v => (int)v.NeutralMass).SequenceEqual(new int[] { 814, 824 }));
         }
 
-        [Test]
```
trishorts: perc

EngineLayer/CrosslinkSearch/CrosslinkSpectralMatch.cs
lonelu: The idea is to create fewer lists in the search engine to free RAM. 
Not sure why codacy complain, I will check.

GUI/XLSearchTaskWindow.xaml
lonelu: I prefer it to be user settable now. This may vary with different datasets. I may able to get a good number to suit all the datasets.

lonelu: I prefer it to be user settable now. This may vary with different datasets. I may able to get a good number to suit all the datasets.

EngineLayer/CrosslinkSearch/CrosslinkSpectralMatch.cs
trishorts: I think you're still creating a list, just inside the method instead of outside. You can use out instead of tuple to return two values. But, you have a void method and are returning only the list. So, I don't see the point.

EngineLayer/CrosslinkSearch/CrosslinkSpectralMatch.cs
lonelu: It is "DECOY_ + BetaPeptide.ProteinAccession", which is treated as Intra-crosslink. Previously, this is treated as Inter-crosslink.

EngineLayer/CrosslinkSearch/CrosslinkSpectralMatch.cs
lonelu: removed the out method.

TaskLayer/XLSearchTask/PostXLSearchAnalysisTask.cs
trishorts: image

EngineLayer/CrosslinkSearch/CrosslinkSearchEngine.cs
```diff
@@ -625,5 +674,19 @@ public static bool DissociationTypeGenerateSameTypeOfIons(DissociationType d, Di
             }
             return false;
         }
+
+        public static int BinarySearchGetIndex(double[] massArray, double targetMass)
```
zrolfs: You can call the Binary search method from ModernSearchEngine. Remove this duplicate code.

EngineLayer/CrosslinkSearch/CrosslinkSpectralMatch.cs
```diff
@@ -49,14 +51,15 @@ public bool IsIntraCsm()
                 {
                     foreach (var beta in betaProteins)
                     {
-                        if (alpha == beta)
+                        if (alpha == beta || alpha == "DECOY_" + beta || beta == "DECOY_" + alpha)
```
zrolfs: this counts decoy-target matches as intra? Add a comment since that's kinda important.

TaskLayer/XLSearchTask/XLSearchParameters.cs
```diff
@@ -10,23 +10,21 @@ public XlSearchParameters()
         {
             DecoyType = DecoyType.Reverse;
             Crosslinker = GlobalVariables.Crosslinkers.First();
-            RestrictToTopNHits = true;
-            CrosslinkSearchTopNum = 300;
+            CrosslinkSearchTopNum = 50;
```
zrolfs: <3

TaskLayer/XLSearchTask/XLSearchTask.cs
```diff
@@ -19,11 +19,13 @@ public XLSearchTask() : base(MyTask.XLSearch)
         {
             //Default parameter setting which is different from SearchTask, can be overwriten
             var digestPara = new DigestionParams(
-                minPeptideLength: 5
+                minPeptideLength: 5,
+                maxPeptideLength: 60,
```
zrolfs: These are interesting choices. Would you mind adding a comment for why you chose these?

zrolfs:
This doesn't quite hit the code coverage. You could add a test that has a DeadEndNH2 and a test where either the alpha or beta peptides (but not both) has null for its protein.

EngineLayer/CrosslinkSearch/CrosslinkSearchEngine.cs
```diff
@@ -564,7 +614,7 @@ private List<MatchedFragmentIon> ScoreChildScan(Ms2ScanWithSpecificMass parentSc
         /// Localizes the loop to a begin and end residue
         /// </summary>
         private CrosslinkSpectralMatch LocalizeLoopSites(PeptideWithSetModifications originalPeptide, Ms2ScanWithSpecificMass theScan, CommonParameters commonParameters,
-            List<int> possiblePositions, Modification loopMod, int notch, int scanIndex, int peptideIndex)
+            List<int> possiblePositions, Modification loopMod, int notch, int scanIndex)
```
zrolfs: loopMod isn't used for anything?

Issue #1781, opened 2019-11-26
lowCID crashes during calibration


bhampton-som:
Hello,

I am trying to search 24 raw files collected with MS1 on the orbitrap and MS2 CID on the ion trap (MS2 CID/IT).  I selected the LowCID option, and set the window to 0.5 Da; other settings were default.  MM crashes at some point during the calibration step.

I have also searched these with the HCD setting just to test if there is something problematic with the raw files.  This search succeeds, so I conclude the raw files are not the problem.  I have other raw files with MS2 HCD/FT and they always successfully complete.  However, MM always crashes when I feed it raw files with MS2 CID/IT data

I've attached the results.txt file that was located in the CalibrateTask folder which shows some error messages for the MS2 CID/IT raw file search.

[results.txt](https://github.com/smith-chem-wisc/MetaMorpheus/files/3892361/results.txt)

Thanks for your help.

Brian Hampton

zrolfs:
Hi Brian,
Thank you for bringing this to our attention. The CID functionality is still relatively new, so we appreciate your help.
From the results.txt file, it looks like the crash is likely being caused by one of your MS2 scans not containing any peaks. We'll modify MetaMorpheus to handle such cases and the fix should be present in the next release.

Zach
Best,
Zach

zrolfs:
Hi again,
You actually have have a pretty interesting scan: it looks like it has peaks at one point, but then loses them?
Would you be willing to share your .toml and your raw files? If you know which file the crash occurs on, just that file would be great.

rmillikin:
fixed in #1782

PR #1782, opened 2019-11-26
Handle LowCID crash

zrolfs:
#1781 
There are two lists for experimental peaks... One was previously checked for values, but not the other. The crash is symptomatic of an underlying issue, but it's difficult to investigate without the raw data that caused the crash.


zrolfs:
@lonelu @rmillikin Please take a look at this PR that's been open for a while.

PR #1784, opened 2019-12-04
proteoform file naming for top-down searches

rmillikin:
Capture


rmmiller22:
Shoudl we also change PSMs to PrSMs?

rmillikin:
I thought about it but decided against it because I thought the typical user would be confused. I'm also not really sure what it adds, though "PrSM" is common in literature...

Issue #1693, opened 2019-07-09
Output options: Write psms checkbox


trishorts:
allow user to turn off writing individual file results and also global psms.tsv

Issue #1530, opened 2018-11-25
Compress individual file results


acesnik:
I rarely look at these files. If that's true for most of us, we should consider using compressed output. If one is double-clicked in the GUI, just decompress it before opening it.

For 18 fractions, these amount to 3.5 GB. After gzipping them, they're 0.8 GB. 

The remaining files (AllPSMs and AllPeptides mostly) are 3 GB.

PR #1789, opened 2019-12-05
Output Options

zrolfs:
#1693 : User can turn off individual results
#1530 : User can now automatically compress individual results

Turning off individual results does not prevent .mzID files from being written. These files can still be compressed.


rmillikin:
failing unit test:

```
Total tests: 298
     Passed: 296
     Failed: 1
    Skipped: 1
 Total time: 2.3193 Minutes
```

zrolfs:
Fixed failing unit test caused by change to percolator output type.

rmillikin:
looks good to me

I'd prefer if these 2 options were in the advanced settings, though if others think that they are OK where they are, it's fine w/ me. i just think the basic settings should be as minimal and obvious as humanly possible - the bread and butter stuff. i think someone starting with metamorpheus won't necessarily know what an individual file result is or if it should be enabled or disabled

PR #1790, opened 2019-12-09
[WIP] Aggregation Engine

zrolfs:
This engine looks at nearby MS1 scans to find shared MS1 peaks. If adjacent scans contain a similar MS1 peak, the median m/z is used as a more accurate measurement and replaces all original m/zs that were shared for that peak. This engine was implemented into the calibration task and requires ~1 min/raw file.

I searched-calibrated-searched the two vignette files (mouse, high-res, trypsin) with and without the aggregation engine.

**Results:
File 1 (Frac4) Suggested Tolerances (MS1/MS2)**
Original: 5.7 / 16.9
Updated: 2.6 / 12.2

**File 2 (Frac5) Suggested Tolerances (MS1/MS2)**
Original: 10.1 / 15.4
Updated: 2.9 / 12.3

**Search Type                                         PSMs/Peptides/Proteins/SearchTime**
Original Search:                                   29835/11037/3592/7:36
Original Post-Calibration Search:        29890/11032/3667/7:15
Updated Post-Calibration Search:       29900/11336/3636/5:51

There's a pretty small improvement in PSMs, but it's comforting having the number of unique peptides increase while the number of proteins decrease. Additionally, we save more time on the search than we spent aggregating.

I'll try a top-down file next. Let me know if there are any other datasets people are interested in.


zrolfs:
This is currently making things worse for top-down. I'll figure out what's going on...

zrolfs:
@trishorts

acesnik:
Cool idea, @zrolfs

PR #1792, opened 2019-12-13
Faster/RAM efficient precursor deconvolution for lowCID

rmillikin:
Bugfix and parallelization for LowCID precursor deconvolution:

Code | Data | "GetMs2Scans" Runtime | Max RAM Usage | PSMs
-- | -- | -- | -- | --
master | LowResHela | 93.0 sec | 7.8 GB | 7959
PR | LowResHela | 10.4 sec | 1.2 GB | 7959
  |   |   |   |  
master | HighResMouse | 10.9 sec | 1.9 GB | 14969
PR | HighResMouse | 11.0 sec | 1.9 GB | 14969


EngineLayer/MetaMorpheusEngine.cs
```diff
@@ -80,14 +80,14 @@ public static List<MatchedFragmentIon> MatchFragmentIons(Ms2ScanWithSpecificMass
...
 
-            // if the spectrum has no peaks
-            if (!scan.ExperimentalFragments.Any())
-            {
-                return matchedFragmentIons;
-            }
-
             if (scan.TheScan.MassSpectrum.XcorrProcessed)
             {
+                // if the spectrum has no peaks
```
zrolfs: We should keep this outside of the if statement, because it needs to be run regardless.

EngineLayer/Ms2ScanWithSpecificMass.cs
```diff
@@ -19,9 +19,12 @@ public Ms2ScanWithSpecificMass(MsDataScan mzLibScan, double precursorMonoisotopi
 
             TheScan = mzLibScan;
 
-            ExperimentalFragments = neutralExperimentalFragments ?? GetNeutralExperimentalFragments(mzLibScan, commonParam);
+            if (commonParam.DissociationType != DissociationType.LowCID)
```
rmillikin: this is the main thing i changed. the old code had a bug in it where the MS2 scan would get deconvoluted for lowCID where it should have skipped deconvolution

TaskLayer/MetaMorpheusTask.cs
```diff
@@ -258,25 +271,33 @@ public static IEnumerable<Ms2ScanWithSpecificMass> GetMs2Scans(MsDataFile myMSDa
...
             // resolutions, so this pre-processing must take place after the parent/child scans have been determined
-            foreach (var parentScan in parentScans)
-            {
-                if (commonParameters.DissociationType == DissociationType.LowCID && !parentScan.TheScan.MassSpectrum.XcorrProcessed)
-                {
-                    parentScan.TheScan.MassSpectrum.XCorrPrePreprocessing(0, 1969, parentScan.TheScan.IsolationMz.Value);
-                }
-
-                foreach (var childScan in parentScan.ChildScans)
+            Parallel.ForEach(Partitioner.Create(0, parentScans.Length), new ParallelOptions { MaxDegreeOfParallelism = commonParameters.MaxThreadsToUsePerFile },
```
rmillikin: other major thing i changed. parallelized xcorr processing

EngineLayer/MetaMorpheusEngine.cs
zrolfs: I think we need to keep it as checking scan.ExperimentalFragments, because the matching happens to deconvolutedMasses (private), which is related more with experimentalFragments than with raw xArray peaks.

EngineLayer/Ms2ScanWithSpecificMass.cs
zrolfs: I'm confused. Why is this not relevant for lowCID? This populates the experimental fragments and the deconvoluted masses.

TaskLayer/MetaMorpheusTask.cs
zrolfs: Move the "commonParameters.DissociationType == DissociationType.LowCID" and "commonParameters.ChildScanDissociationType == DissociationType.LowCID" outside of the loop so we can avoid this for non lowCID searches.

EngineLayer/Ms2ScanWithSpecificMass.cs
rmillikin: lowCID doesn't use the deconvoluted MS2 data. the xcorr score uses the peaks in the scan, not deconvoluted masses

EngineLayer/MetaMorpheusEngine.cs
rmillikin: there's a check after the lowCID stuff which checks if there are any deconvoluted masses

rmillikin: (line 111)

Issue #1793, opened 2019-12-13
Linux paths in command line generally don't work


rmillikin:
The nuget package we're using to parse command line arguments seems to not be able to parse the "/" character, which is not really a problem in Windows since file paths use the "\\" character. In Linux however it is a pretty major problem, since the file system uses "/" in file paths. It's possible to do a hacky workaround by putting files in the root directory or the desktop so that "/" is not needed in the file paths in the command line but it's not great.

I fixed this issue in FlashLFQ by switching nuget packages from FluentCommandLineParser to CommandLineParser.

rmillikin:
fixed in #1796

Issue #1592, opened 2019-03-13
Commandline tool to generate default TOMLs


acesnik:
Or just specify an output directory and no TOMLs for CMD.dll to generate some default ones.

acesnik:
As is, the commandline version isn't really independent because I have to generate the tomls in the GUI first.

rmillikin:
done in #1796

PR #1796, opened 2019-12-16
change command line parser package; remove meta-task stuff from comma…

rmillikin:
…nd-line; added command to generate default tomls


rmillikin:
changes in this PR:
1) changed command line parser packages from FluentCommandLineParser to CommandLineParser to address issue #1793 
2) added a command-line parameter to generate default tomls to address issue #1592 
3) added a microvignette to the command-line version of MetaMorpheus. this is a small yeast data file/database that gets calibration/gptmd/search run on it.

I attempted to make a unit test for the microvignette but appveyor would not build. For some reason, adding the CMD as a reference to the Test project makes the SolutionDir variable "undefined". I've left the unit test there, but commented out, in case we eventually figure out how to fix that.

PR #1804, opened 2020-01-03
Remove interfaces

zrolfs:


zrolfs:
@rmillikin, can you confirm if we still want these? I think they're just occupying space.

PR #1807, opened 2020-01-08
.NET Core 3

rmillikin:


rmillikin:
some things I've noticed that I still need to fix (probably in a separate near-future pull request):

- Some icons in the GUI don't show up
- Drag-and-drop doesn't seem to work on the installed version of MM?

TaskLayer/MetaMorpheusTask.cs
```diff
@@ -779,8 +791,39 @@ private static string GenerateOutputFolderForIndices(List<DbForTask> dbFilenameL
...
+                    {
+                        Status("Reading precursor index...", new List<string> { taskId });
+                        precursorIndex = ReadFragmentIndex(Path.Combine(pathToFolderWithIndices, PrecursorIndexFileName));
+                    }
+
+                    successfullyReadIndices = true;
+                }
+                catch
+                {
+                    // could put something here... this basically is just to prevent a crash if the index was unable to be read.
```
trishorts: we could provide an error message

PR #1831, opened 2020-02-06
Pruned Db edits

rmmiller22:
1. fix bug Leah found that only proteins, who have peptide evidence have modification properly filtered prior to pruned database generation
2. old code cleared all original modifications for each protein and replaced with modifications for the database. this is problematic, especially since the peptide results are calculated after database generation. Code was added to store the original modifications and then reset to these after the databases are created.


trishorts:
A unit test to hit these bulleted lines from the PostSearchAnalysisTask would be great. 
image

lschaffer2:
Rachel would it be hard to add a unit test just like that one but make sure an artifact isn't added if it shouldn't be?

rmmiller22:
Unit tests for the PostSearchAnalysisTask are pretty tricky but I will do my best to design ones to meet both of your requests

rmmiller22:
> Rachel would it be hard to add a unit test just like that one but make sure an artifact isn't added if it shouldn't be?

Hi Leah, Upon re-familiarizing myself with my previous unit test. We already do check that artifacts are not being added to the database when we do not want them to be!

PR #1832, opened 2020-02-07
[WIP] Protein pvalue

trishorts:
very first attempt at creating protein P-value from unique peptide PEP values.

Have a look. 

Referencing formula 3 from this paper. https://pubs.acs.org/doi/pdf/10.1021/ac0341261
Alexey I. Nesvizhskii,*,†Andrew Keller,*,†Eugene Kolker,‡and Ruedi Aebersold


rmillikin:
just a small note on nomenclature, i haven't looked at the code yet.
p-value is an informal term that typically describes the result of a frequentist test (i.e., from a t-test). PEP (posterior error probability) or PIP (posterior inclusion probability) is the terminology we should use for this http://varianceexplained.org/r/bayesian_fdr_baseball/ . i know that seems pedantic but it matters for communicating these concepts to those who are not in as deep

acesnik:
Is the goal of this PR changing? I'm curious about Rina's new commits.

Issue #1818, opened 2020-01-24
EngineLayer/GlobalVariables.cs writes into codebase, not allowed in Singularity or HPC 


jj-umn:
Files: CustomAminoAcids/CustomAminoAcids.txt and settings.toml are written into the code base when MetaMorpheus is first run.   

This is an issue when running MetaMorpheus in a Singularity container.   The codebase is in a read only file system for security reasons.  

This could also be an issue for an a HPC system in which multiple jobs could be run concurrently if  the jobs need to alter those files.   

Could there be command line option for a user DataDir that could override or modify defaults rather than writing back into the codebase?

trishorts:
we'll look into it jj

rmillikin:
We could very easily add a command-line option to change the data directory - this seems like the simplest path forward

rmillikin:
I've got this resolved in #1835 but needs testing + unit tests

rsdmse:
While our Docker container works fine on a personal laptop, we're experiencing probably the same issue with the converted Singularity container on a cluster. Here's the complete output:
```
Welcome to MetaMorpheus
Unhandled exception. System.TypeInitializationException: The type initializer for 'EngineLayer.GlobalVariables' threw an exception.
 ---> System.IO.IOException: Read-only file system
   at System.IO.FileSystem.CreateDirectory(String fullPath)
   at System.IO.Directory.CreateDirectory(String path)
   at EngineLayer.GlobalVariables.WriteAminoAcidsFile()
   at EngineLayer.GlobalVariables.RefreshAminoAcidDictionary()
   at EngineLayer.GlobalVariables..cctor()
   --- End of inner exception stack trace ---
   at MetaMorpheusCommandLine.Program.Run(CommandLineSettings settings)
   at MetaMorpheusCommandLine.Program.<>c__DisplayClass3_0.<Main>b__1(CommandLineSettings options)
   at CommandLine.ParserResultExtensions.WithParsed[T](ParserResult`1 result, Action`1 action)
   at MetaMorpheusCommandLine.Program.Main(String[] args)
CMD: line 2: 59159 Aborted 
```

rsdmse:
Our user replied that the dev container (converted into Singularity) mentioned in #1835 is working on our cluster. Thanks again for all your help!

rmillikin:
Great to hear! I'll make a new release soon (probably within a week or two). 

For those following this issue:
[...]

PR #1835, opened 2020-02-10
Optionally specify a data directory for command-line

rmillikin:
optionally specify a data directory, change a couple data dir things to prevent crashes on readonly filesystems

#1818


rsdmse:
@rmillikin Thanks a lot for fixing this issue. We have a group of users on our HPC cluster waiting to use this software. If the PR will take a while, can we use your forked repo?

rmillikin:
@rsdmse I think this should get merged quite soon, but if you want, you can pull the current smithchemwisc/metamorpheus:dev tag from dockerhub which includes this change. usually the dev tag is the current master branch, but I made an exception in this case and pushed this PR to the dev tag for testing in Singularity.

something like this (assuming you're using Singularity to pull from dockerhub):
singularity_mm

but in case that doesn't suit your purposes, yes you can definitely build this PR and use it as you wish.

rmillikin:
@rsdmse it's not from my forked repo though, it's from a branch in this repo. https://github.com/smith-chem-wisc/MetaMorpheus/tree/optionalDataDir

rsdmse:
Thanks so much for your prompt reply!

trishorts:
I was unable to specify and output folder with quotations (either "" or '') which is need for file paths with spaces. 

e.g. "C:\User\Bob The Builder\Metamorpheus"

trishorts (approved):

rmillikin:
@trishorts this might be a singularity issue but I can poke around for a solution.
https://github.com/hpcng/singularity/issues/5437
https://stackoverflow.com/questions/62963043/escaping-filename-spaces-with-singularity-files-directive

rmmiller22 (approved):

PR #1837, opened 2020-02-11
Fix 404 error from downloads badge

acesnik:
Link back to same page from downloads badge to not get 404 error


rmillikin:
we could link to the releases page or something but this is fine for now

PR #1841, opened 2020-02-18
Low res cid gptmd crash

trishorts:
i was processing some low res CID data from coon (fusion lumos stuff). During GPTMD, i experienced two separate crashes. The first was from scans where experimental fragments was null. The second was from closestexperimentalmass null. I add two null checks and the crashes disappeared. I do not know the source of the crash. Maybe an mzlib problem.


trishorts:
this bug is also still a bug but i gave up on the pr

zrolfs:
I thought we fixed this in mzlib and that those properties should never be null.

Issue #1852, opened 2020-03-22
Index search crashes when reading index if two proteins have the same accession


Dmorgen:
Hi,

please see the following crash report:
MetaMorpheus: version 0.0.304
Data files were processed on a VMware, Inc. VMware7,1 computer using Windows version 6.2.9200.0 with a 64-Bit processor and 32 cores operating at 2.1GHz and 128GB installed RAM.
.NET Framework Version: 4.7 or laterC:\Program Files\Thermo\MSFileReader\Fileio_x64.dll MISSING\nC:\Program Files\Thermo\MSFileReader\fregistry_x64.dll MISSING\nC:\Program Files\Thermo\MSFileReader\XRawfile2_x64.dll MISSING\n
e: MzLibUtil.MzLibException: A crash occured in FlashLFQ during the intensity normalization process:
Sequence contains no elements
   at FlashLFQ.FlashLfqEngine.Run()
   at TaskLayer.PostSearchAnalysisTask.QuantificationAnalysis() in C:\projects\metamorpheus\TaskLayer\SearchTask\PostSearchAnalysisTask.cs:line 455
   at TaskLayer.PostSearchAnalysisTask.Run() in C:\projects\metamorpheus\TaskLayer\SearchTask\PostSearchAnalysisTask.cs:line 66
   at TaskLayer.SearchTask.RunSpecific(String OutputFolder, List`1 dbFilenameList, List`1 currentRawFileList, String taskId, FileSpecificParameters[] fileSettingsList) in C:\projects\metamorpheus\TaskLayer\SearchTask\SearchTask.cs:line 381
   at TaskLayer.MetaMorpheusTask.RunTask(String output_folder, List`1 currentProteinDbFilenameList, List`1 currentRawDataFilepathList, String displayName) in C:\projects\metamorpheus\TaskLayer\MetaMorpheusTask.cs:line 430e.Message: A crash occured in FlashLFQ during the intensity normalization process:
Sequence contains no elementse.InnerException: e.Source: FlashLFQe.StackTrace:    at FlashLFQ.FlashLfqEngine.Run()
   at TaskLayer.PostSearchAnalysisTask.QuantificationAnalysis() in C:\projects\metamorpheus\TaskLayer\SearchTask\PostSearchAnalysisTask.cs:line 455
[...]

PR #1854, opened 2020-03-23
SearchTask: Option to Handle Contaminants

zrolfs:
#1852 
This looks intense, but it's not too bad.

I added options for how contaminants can be handled when they are also present in the target database. These are:
-Remove the target protein
-Remove the contaminant protein (default)
-Rename both the target and contaminant accessions
I left in the throw in MetaMorpheusTask.cs ln 698. Shouldn't be possible to hit anymore, so we could take it out, but you never know...

After our Slack conversations, I also:
-Removed the preferential treatment of contaminants over targets in parsimony. They are equal now.


zrolfs:
> Consider rephrasing the error message as:
> The protein DECOY_P02533 has multiple entries. Protein accessions must be unique. Contaminant protein DECOY_P02533 was ignored.

> Consider rephrasing the error message as:
> The protein DECOY_P04259 has multiple entries. Protein accessions must be unique. Target protein DECOY_P04259 was ignored.

> The protein Q04695 has multiple entries. Protein accessions must be unique. Target protein Q04695 was renamed Q04695_T1. Contaminant protein Q04695 was renamed Q04695_C1.

I don't think I can do this, because the decoys are not labeled as contaminants. This also prevents me from doing the same thing with the targets, because the decoys and targets must share their accession numbers for the top-picked fdr.

> The proteins file for this collision marks the column as contaminant, when it’s both. And, the underscore_number makes it look like a regular UniProt isoform, which it is not.

I could make it "PROTEIN_D1" for duplicate? 

> Protein.Equals() should compare also mods. But not in this PR. That can be done by someone else as an mzlib PR.
The index requires each accession to be unique, so it only made sense to compare by the accession. Unless we're talking about a different section?

Test/RobTest.cs
```diff
@@ -128,11 +128,10 @@ public static void TestParsimony()
             Assert.Contains("AB--------", parsimonyProteinSequences);
             Assert.Contains("--C-------", parsimonyProteinSequences);
             Assert.Contains("-B-D---HHH--", parsimonyProteinSequences);
-            Assert.Contains("----E----**", parsimonyProteinSequences);
             Assert.Contains("-B------I-", parsimonyProteinSequences);
             Assert.Contains("----EFG---", parsimonyProteinSequences);
             Assert.Contains("----EFG--J", parsimonyProteinSequences);
-            Assert.AreEqual(8, parsimonyProteinSequences.Count);
+            Assert.AreEqual(6, parsimonyProteinSequences.Count);
```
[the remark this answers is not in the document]
zrolfs: Where are you recommending using Protein.Equals?

EngineLayer/PsmTsv/PsmTsvWriter.cs
```diff
@@ -231,7 +231,7 @@ internal static void AddPeptideSequenceData(Dictionary<string, string> s, Peptid
 
             string theoreticalsSearched = " ";
             s[PsmTsvHeader.TheoreticalsSearched] = theoreticalsSearched;
-            s[PsmTsvHeader.DecoyContaminantTarget] = pepWithModsIsNull ? " " : psm.IsDecoy ? "D" : psm.IsContaminant ? "C" : "T";
+            s[PsmTsvHeader.DecoyContaminantTarget] = pepWithModsIsNull ? " " : Resolve(pepsWithMods.Select(b => b.Protein.IsDecoy ? "D" : b.Protein.IsContaminant ? "C" : "T")).ResolvedString;
```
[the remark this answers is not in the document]
zrolfs: What information would you like? This resolve method is used in dozens of places within this class.

TaskLayer/SearchTask/SearchTask.cs
```diff
@@ -453,5 +454,74 @@ private static MassDiffAcceptor ParseSearchMode(string text)
...
+        protected static void SanitizeProteinDatabase(List<Protein> proteins, TargetContaminantAmbiguity tcAmbiguity)
+        {
+            if (proteins.Count != 0) //if we have proteins
+            {
+                proteins.Sort((x, y) => x.Accession.CompareTo(y.Accession));
+                string previousAccession = proteins[0].Accession; //get first accession
+                for (int i = 1; i < proteins.Count; i++)
+                {
+                    string currentAccession = proteins[i].Accession; //get current accession
+                    if (currentAccession.Equals(previousAccession)) //if this accession was found in the previous protein (i.e. it's a duplicate)
```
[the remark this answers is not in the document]
zrolfs: We want to compare accessions and not protein objects. The index will crash if two identical accessions are present. The mods, sequences, variants, etc. are irrelevant.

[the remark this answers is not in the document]
trishorts: But, it does matter which one gets deleted. If you have xml and fasta, then removing xml could remove peptides with PTMs not present in the FASTA. That's a non-zero consequence.

trishorts:
> I could make it "PROTEIN_D1" for duplicate?
> 
I can live with this

EngineLayer/ProteinParsimony/ProteinParsimonyEngine.cs
```diff
@@ -45,11 +45,11 @@ public ProteinParsimonyEngine(List<PeptideSpectralMatch> allPsms, bool modPeptid
                 _fdrFilteredPsms = allPsms.Where(p => p.BaseSequence != null && p.FdrInfo.QValue <= FdrCutoffForParsimony && p.FdrInfo.QValueNotch <= FdrCutoffForParsimony).ToList();
             }
 
-            // peptides to use in parsimony = peptides observed in high-confidence PSMs
+            // peptides to use in parsimony = peptides observed in high-confidence PSMs (including decoys)
```
rmmiller22, line 48: Thanks Zach!

TaskLayer/SearchTask/SearchTask.cs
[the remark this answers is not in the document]
rmmiller22: Isn't the user determining which protein that they want to keep when they select whether they want to keep the contaminant or keep the decoy? Even if the contaminant version has mods not in the target database, if the user only wants to keep the target, then I am not sure it would be right for us to override that because we think one protein entry is better than the other.

zrolfs:
@rmmiller22 The xml only supersedes the fasta if they're both labeled as targets. If one is a contaminant and the other a decoy, then the user-specified choice is used.

TaskLayer/SearchTask/SearchTask.cs
```diff
@@ -453,5 +454,81 @@ private static MassDiffAcceptor ParseSearchMode(string text)
...
+
+        /// <summary>
+        /// Handle ambiguity when two theoretical proteins in the 
+        /// search space have the same accession number.
+        /// The accession must be unique for indexed searches
+        /// </summary>
+        /// <param name="proteins"></param>
+        /// <param name="tcAmbiguity"></param>
+        protected static void SanitizeProteinDatabase(List<Protein> proteins, TargetContaminantAmbiguity tcAmbiguity)
+        {
```
rmillikin, line 466: tbh this section of code seems complicated and error prone... i wrote a simpler version that i think is easier to follow, see what you think:

```
protected static void SanitizeProteinDatabase(List<Protein> proteins, TargetContaminantAmbiguity tcAmbiguity)
        {
            foreach (var accessionGroup in proteins.GroupBy(p => p.Accession))
            {
                int numProteinsWithSameAccession = accessionGroup.Count();

                if (numProteinsWithSameAccession == 1)
                {
                    continue;
                }

                List<Protein> proteinsWithThisAccession = accessionGroup.ToList();
                List<Protein> proteinsToRemove = new List<Protein>();

                if (tcAmbiguity == TargetContaminantAmbiguity.RemoveContaminant)
                {
                    // remove contaminants
                    proteinsToRemove = proteinsWithThisAccession.Where(p => p.IsContaminant).ToList();
                }
                else if (tcAmbiguity == TargetContaminantAmbiguity.RemoveTarget)
                {
                    // remove targets
                    proteinsToRemove = proteinsWithThisAccession.Where(p => !p.IsDecoy && !p.IsContaminant).ToList();
                }

                foreach (var protein in proteinsToRemove)
                {
                    proteins.RemoveAll(p => p == protein);
                    proteinsWithThisAccession.RemoveAll(p => p == protein);
                }

                proteinsToRemove.Clear();

                // if there are still duplicate proteins, rename them to avoid duplicate accessions
                if (proteinsWithThisAccession.Count > 1)
                {
                    int proteinNumber = 1;

                    foreach (var originalProtein in proteinsWithThisAccession)
                    {
                        var renamedProtein = new Protein(originalProtein.BaseSequence, originalProtein + "_" + proteinNumber.ToString(), originalProtein.Organism,
                            originalProtein.GeneNames.ToList(), originalProtein.OneBasedPossibleLocalizedModifications, originalProtein.ProteolysisProducts.ToList(), originalProtein.Name, originalProtein.FullName,
                            originalProtein.IsDecoy, originalProtein.IsContaminant, originalProtein.DatabaseReferences.ToList(), originalProtein.SequenceVariations.ToList(), originalProtein.AppliedSequenceVariations,
                            originalProtein.SampleNameForVariants, originalProtein.DisulfideBonds.ToList(), originalProtein.SpliceSites.ToList(), originalProtein.DatabaseFilePath);

                        proteins.Add(renamedProtein);
                        proteinsToRemove.Add(originalProtein);
                        proteinNumber++;
                    }
                }

                foreach (var protein in proteinsToRemove)
                {
                    proteins.RemoveAll(p => p == protein);
                    proteinsWithThisAccession.RemoveAll(p => p == protein);
                }
            }
        }
```

rmillikin: also does this belong in the search task? shouldn't this live in MetaMorpheusTask and then all the tasks would call this method upon loading in the protein DB?

zrolfs: That's a point I had considered and am open to doing. I ended up putting it in the search task because I didn't think it substantially influenced the other tasks. I could see a case for XLSearchTask, though. I can put it to the MetaMorpheusTask if you would prefer.

rmillikin: I think that's a good idea - as far as I know this mostly affects indexed searches (modern search, SNES, XL search, and probably glyco search). since those compose 3 different tasks it's probably a good idea for them all to reference this block of code

zrolfs: I like the revision. Is there a reason to use RemoveAll vs Remove? Shouldn't there only be one instance?

zrolfs: Ahh, it keeps decoys for both, though, which I don't like. Hm...

rmillikin: yes that is a rather strange quirk. the `RemoveAll(p => p == protein)` forces the use of reference equality and not value equality (`==` and not `.Equals()`). if you use `.Remove()`, then the equals method for the protein is used, which will remove the first protein in the list with the given accession and base sequence - the IsContaminant bool is not used to assess equality. We could fix that in mzLib but this is just a quick-and-dirty fix

Issue #1815, opened 2020-01-23
Crash when trying to write long file paths; error message not informative


rmillikin:
@blfrey got this error

When trying to write a filename, Windows has a maximum path length of 260 characters. If a filepath is longer than that, an exception is thrown with the message that "part of the path could not be found". We should check to see if the path length is >260 characters, catch the exception, perhaps write the file with a truncated name, and/or display a more meaningful error message

rmillikin:
fixed in #1868

rmillikin:
i've actually decided against fixing this. windows 10 and up no longer has this issue (paths >260 characters are allowed) and so it's difficult to figure out when this error is going to occur before it happens. the exception thrown is sort of a generic error message and i think is difficult to handle. and it's rare enough that i think it's not a major problem. could revisit it at some point though.

blfrey:
OK. that is fine by me.   But, just saying, that I'm pretty sure Artemis already had Windows 10 when this issue came up.

acesnik:
Does this file need to be updated somehow? https://github.com/smith-chem-wisc/MetaMorpheus/blob/master/GUI/app.manifest

That's how Stefan fixed this way back.

PR #1868, opened 2020-04-01
[WIP] issue 1815

rmillikin:
#1815


rmillikin:
Earlier versions of windows have a 260 character limit for paths. Apparently this limitation is not in effect for Windows 10. I originally planned to catch the exception and provide a better error message, and I may end up implementing that at some point, but it will be difficult to unit-test since this crash is machine-dependent. I've just added a warning to the GUI that displays if the path length looks like it might exceed 260 characters

Issue #1833, opened 2020-02-10
Files with "." in the name messes up Experimental Design if calibrated


rmillikin:
For example:
myfile.1.mzml calibrates fine, but the calibration task doesn't properly create the "myfile.1-calib.mzml" filename in the "ExperimentalDesign.tsv", causing a crash downstream in the search task if the experimental design is specified

rmillikin:
fixed in #1869

PR #1869, opened 2020-04-01
issue 1833

rmillikin:
#1833

### DIFF
--- a/TaskLayer/CalibrationTask/CalibrationTask.cs
+++ b/TaskLayer/CalibrationTask/CalibrationTask.cs
@@ -302,7 +302,7 @@ namespace TaskLayer
             for (int i = 1; i < lines.Length; i++)
             {
                 var split = lines[i].Split('\t');
-                string oldFileName = Path.GetFileNameWithoutExtension(split[0]);
+                string oldFileName = Path.GetFileName(split[0]);
                 string newFileName = oldFileName + CalibSuffix;
                 string newline;
 


Test/CalibrationTests.cs
```diff
@@ -9,34 +9,49 @@ namespace Test
...
     {
         [Test]
-        public static void ExperimentalDesignCalibrationTest()
+        [TestCase("filename1.mzML")]
+        [TestCase("filename1.1.mzML")]
+        public static void ExperimentalDesignCalibrationTest(string originalFileName)
         {
             CalibrationTask calibrationTask = new CalibrationTask();
             string outputFolder = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestCalibration");
-            string myFile = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\SmallCalibratible_Yeast.mzML");
```
zrolfs, line 16: Can this file be deleted, or do we need it for another unit test?

TaskLayer/CalibrationTask/CalibrationTask.cs
```diff
@@ -302,7 +302,7 @@ private static void WriteNewExperimentalDesignFile(string assumedPathToExperDesi
             for (int i = 1; i < lines.Length; i++)
             {
                 var split = lines[i].Split('\t');
-                string oldFileName = Path.GetFileNameWithoutExtension(split[0]);
+                string oldFileName = Path.GetFileName(split[0]);
```
zrolfs, line 305: Could you comment that the extension has already been pruned at this point?

Test/CalibrationTests.cs
zrolfs: jk, I see it further down

PR #1873, opened 2020-04-03
[WIP] output spectral library

Yuling1996:


lonelu:
In SpectralLibraryTest.SpectrumTest()
At line 50 
Assert.AreEqual(expectedResultHash, testResultHash);
Is passed in VS, but failed every time in the pull request. 
I checked the results and found that the spectralLibrary.msp is different from expected. I assume the   new SearchTask().RunTask(...) at line 47 have changes of parameters or so... please check it
@Yuling1996

Issue #1876, opened 2020-04-09
MS files don't always get added to the GUI.


zrolfs:
This is a super weird bug where when I drag and drop a folder, only one of the files from that folder is added in the GUI. If I use the "Add" button instead of drag and drop, I get a crash.

Works fine for 3/4 of my folders. Reproducibly has this problem for one of them. File path length is <260 char.

zrolfs:
This appears to occur because the (custom) protease specified in the file-specific toml is not present in the proteases.tsv file after updating. This is still a pretty strange behavior, though. I think we should throw a warning rather than just ignore it, and none of the files should be added if their tomls aren't valid?

rmillikin:
fixed in #1877

PR #1877, opened 2020-04-09
Catch "bad" proteases in file-specific params

zrolfs:
#1876 
Reproduce: 
-Create file-specific tomls with proteases that don't exist in the dictionary (e.g. "asdf").
-Try to load multiple files (drag-n-drop or Add button)
Current behavior: only one file will be added and no file-specific parameters. If using the Add button, MetaMorpheus will crash and close.

This PR changes the behavior to: 
-catch the exception from the missing key and provide a warning to the user. 
-Load all ms files and no file-specific parameters.
-MetaMorpheus will not crash.


trishorts:
The error message is better.

I think we need something better than a crash as a resolution to the problem. Loading the file while not loading the filespecific toml and then blocking the editing of that toml just doesn't work. if the protease is anything other than trypsin, which is likely given that there is an FSP doesn't make sense. if you block file load until the toml works, maybe that helps. You could, perhaps open a fsp gui window that allows the user to fix the problem but simply setting whatever parameters need fixing. That doesn't work for commandline. but i think crash in commandline with error message is more acceptable.

zrolfs:
@trishorts Can you clarify? The crashes are now handled by this PR and the user is provided with an error message. The toml will be loaded for any protease that is in the dictionary, not just trypsin. The issue is encountered when a toml is manually edited to a protease not found in proteases.tsv. Previously, this resulted in the program crashing (closing). This PR catches the crash and provides a warning to the user.
A warning is provided for commandline users as well.

trishorts:
maybe i don't have your latest commits in my test branch. i thought i did. when i try to edit bad toml in gui, mm just closes without warning.

trishorts:
i did not have your latest commits. have updated. will try again now.

zrolfs:
> no crash now.
> I would still like to see a better error message. you could add
> 
> "Proceeding with the search with use trypsin as the protease. If you wish to use a filespecific settings, you must clear the file from the mass spectra files window, edit the file specific parameters .toml in a text editor, then reload the file and the file specific paramters."

Trypsin isn't guaranteed to be used. It will be whatever protease was specified in the common params. I think "Problem parsing file-specific toml" is pretty clear that the toml will be ignored.

PR #1889, opened 2020-05-04
increased PEP training fraction to add more stability to small samples

trishorts:
increased percolator data to better match PEP training data BestMatchingPeptides

I felt like the training fraction (formerly 10% was low for single files). So i increased in to 0.25%. 
I also added a cap at 10^6 training values. Beyond that seems overkill.

Finally, I noticed that percolator results were incomplete. They did not have all of the training data that PEP had. So i increased the output to match PEP.


zrolfs:
> I felt like the training fraction (formerly 10% was low for single files). So i increased in to 0.25%.
> I also added a cap at 10^6 training values. Beyond that seems overkill.

You mean 25%?

PR #1890, opened 2020-05-05
enable writing of .mzid for searches with a single file. Disables .mz…

trishorts:
I discovered that it was hard/impossible to get the .mzid output for a single file search. Because of the placement of the write statement in postsearchanalysistask, protein parsimony is a pre-requisite. I think that the mzid file might need protein groups so okay fine. I changed the mzid checkbox default state to off and disabled it when parsimony was not selected. I guess i usually don't want to do either.... The other thing that was messing up the works were requirements for more than one file for individual file results. Which makes sense on the surface except that you can't get mzid.


GUI/TaskWindows/SearchTaskWindow.xaml
```diff
@@ -440,7 +440,11 @@
                     <GroupBox Header="Output Options" DockPanel.Dock="Top" ScrollViewer.CanContentScroll="True">
                         <!--<Expander x:Name="OutputExpander">-->
                         <StackPanel Orientation="Vertical" Margin="5">
-                            <CheckBox x:Name="CkbMzId" Content="Write .mzID" />
+                            <CheckBox x:Name="CkbMzId" Content="Write .mzID" IsEnabled="{Binding IsChecked, ElementName=CheckBoxParsimony}" ToolTipService.ShowOnDisabled="True">
+                                <ToolTipService.ToolTip>
+                                    Output of .mzID requires protein parsimony. Please check "Apply protein parsimony and construct protein groups".
```
zrolfs, line 445: Heads up that ToolTips don't display if the object is disabled, so the second part might not make sense (since they would only be able to read this if it was already checked).

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -607,7 +607,7 @@ private void WriteProteinResults()
                 // write all individual file results to subdirectory
                 // local protein fdr, global parsimony, global psm fdr
 
-                if ((Parameters.CurrentRawFileList.Count > 1 && Parameters.SearchParameters.WriteIndividualFiles ) 
+                if (Parameters.SearchParameters.WriteIndividualFiles
                     || Parameters.SearchParameters.WriteMzId || Parameters.SearchParameters.WritePepXml)
                 {
                     Directory.CreateDirectory(Parameters.IndividualResultsOutputFolder);
```
zrolfs, line 613: I don't think we should make this folder if we only have one file. The mzid should instead be put in the parent directory with all the other files.

Test/MyTaskTest.cs
```diff
@@ -443,6 +443,7 @@ public static void TestFileOutput()
 
             SearchTask task = Toml.ReadFile<SearchTask>(Path.Combine(TestContext.CurrentContext.TestDirectory, @"SlicedSearchTaskConfig.toml"), MetaMorpheusTask.tomlConfig);
             task.SearchParameters.DecoyType = DecoyType.None;
+            task.SearchParameters.WriteMzId = true;
```
zrolfs, line 446: Is there a test where this is false?

GUI/TaskWindows/SearchTaskWindow.xaml
trishorts: Line 443 ToolTipService.ShowOnDisabled="True" makes it show when disabled:
image

Test/MyTaskTest.cs
trishorts: Happy to make further updates but i need to understand better what you're looking for. Apparently i haven't had enough coffee yet.

GUI/TaskWindows/SearchTaskWindow.xaml
zrolfs: Nice! I was unfamiliar with that option. Thanks for pointing it out.

Test/MyTaskTest.cs
zrolfs: I think it's hit... feel free to ignore. I was trying to figure out why codecov had gone down, but I think it's just because of how it's handling the conditionals.

TaskLayer/SearchTask/SearchParameters.cs
```diff
@@ -25,7 +25,7 @@ public SearchParameters()
             KeepAllUniprotMods = true;
             MassDiffAcceptorType = MassDiffAcceptorType.OneMM;
             MaxFragmentSize = 30000.0;
-            WriteMzId = true;
+            WriteMzId = false;
```
rmillikin, line 28: why this change? it doesn't take much extra time to create the .mzID but it is really annoying if you wanted an mzID and forgot to check the box

trishorts: for me its not a time thing, its a space thing. both drives on my computer where i store data are alwasy full and i'm alwasy deleting stuff. my assumption is that a majority of users will never need mzid for anything. so its superfluous. they can tick the box when needed. I will change it back if this is really important to you.

rmillikin: it is very important to me.. right now mzid is the only "standard" output that we write (pepxml for xl task also I think) so if anybody wants to output our results into other programs (skyline) it's really important. the space constraints thing is something to consider but with @zrolfs change you can specify your own defaults, so you can disable mzid and save that as your default if you wish

trishorts: reverted

Issue #1593, opened 2019-03-13
Commandline improvements


acesnik:
- [x] Dockerize it
- [x] Make smithchemwisc Docker account
- [x] Automatic builds of docker image from the release branch on Docker Hub
- [x] Update readme for .NET Core vignette for mzMLs and using docker on Linux (or any) system
- [x] Add capability to generate default tomls from the commandline
- [x] Make a tutorial for MetaMorpheus Docker (see https://github.com/smith-chem-wisc/FlashLFQ/wiki/Docker-Image for similar docker tutorial template)

Any other things I'm missing?

acesnik:
Not sure how to deploy docker builds from GitHub organization repo https://forums.docker.com/t/cant-access-new-github-organization-for-automated-builds/1080/11

acesnik:
I made a smithlab account on Docker Hub, though, if anyone wants to use it for some reason.

acesnik:
Thanks for adding the default toml option. Just used it!

acesnik:
- [ ] note how to specify multiple files in the help menu (separated by space following the flag, e.g. -t toml1.toml toml2.toml -s spect1.mzML spect2.mzML)

rmillikin:
docker hub deployments are automatic as of v0.0.306. also all merged PRs create a dev build which is also auto uploaded to docker hub.

rmillikin:
the "latest" tag on dockerhub is also the latest release version

acesnik:
I checked off the "Update readme for .NET Core vignette for mzMLs and using docker on Linux (or any) system" -- I think that was before the .NETCore version could read RAW files.

acesnik:
@rmillikin You also ticked off the space-delimited issue. Feel free to close the issue 👍

rmillikin:
sure. i think we still need a docker tutorial though

rmillikin:
see https://github.com/smith-chem-wisc/FlashLFQ/wiki/Docker-Image for similar docker tutorial template

acesnik:
- [ ] Make a tutorial for MetaMorpheus Docker

acesnik:
Added a tutorial for Docker here, https://github.com/smith-chem-wisc/MetaMorpheus/wiki/Docker-Image

PR #1900, opened 2020-05-13
command-line error codes; verbosity settings

rmillikin:
- Add exit codes to the CMD version. If MM fails, a non-zero exit code is returned. This makes it easy to check via script if MM failed its run
- Add verbosity settings for CMD version. This reduces the amount of output written to the console
- The installer installs the CMD version along w/ the GUI version
- The installer installs the microvignette files, enabling the microvignette to be run by the installed version of MM. Currently, there is no option to run the microvignette in the GUI version, but this PR makes that very easy to set up now that the files are present after MM is installed
- Change -v flag to specify verbosity and --test to perform the microvignette

The general purpose of this PR is to make automatic testing of MM via AppVeyor easier; enabling testing of the installer, etc.

### DIFF
--- a/CMD/CommandLineSettings.cs
+++ b/CMD/CommandLineSettings.cs
@@ -30,10 +30,10 @@ namespace MetaMorpheusCommandLine
         [Option('g', HelpText = "Generate default task tomls")]
         public bool GenerateDefaultTomls { get; set; }
 
-        [Option('v', HelpText = "Runs a small test search using a database and yeast data file included with this MetaMorpheus installation")]
+        [Option("test", HelpText = "Runs a small test search using a database and yeast data file included with this MetaMorpheus installation")]
         public bool RunMicroVignette { get; set; }
 
-        [Option("verbosity", Default = VerbosityType.normal, HelpText = "Determines how much text is written. Options are no output ('none'), minimal output and errors  ('minimal'), or normal ('normal')")]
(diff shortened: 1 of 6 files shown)


CMD/CommandLineSettings.cs
```diff
@@ -32,7 +32,12 @@ public class CommandLineSettings
 
         [Option('v', HelpText = "Runs a small test search using a database and yeast data file included with this MetaMorpheus installation")]
         public bool RunMicroVignette { get; set; }
-        
+
+        [Option("verbosity", Default = VerbosityType.normal, HelpText = "Determines how much text is written. Options are no output ('none'), minimal output and errors  ('minimal'), or normal ('normal')")]
```
acesnik, line 36: Typically, this is `-v` in commandline programs. Consider changing the test flag to `--test`. Is the default showed in the help menu? If not, maybe add that to the description.

rmillikin: thx, i've noticed that too since implementing the microvignette. i've made -v refer to verbosity and --test runs the microvignette.

the default verbosity is displayed when --help is called

acesnik: Also, since we're here, I think it'd be helpful to note that the databases, spectra, and tomls are space-delimited in their descriptions (https://github.com/smith-chem-wisc/MetaMorpheus/issues/1593)

rmillikin: will do

rmillikin:
I hadn't thought about the calibration failure; I guess it is a partial failure, or maybe a warning. 

The --version info is automatically generated by the command line parser nuget package that we're using but I believe that method can be overridden. I'll look into it in a separate PR

PR #1903, opened 2020-05-18
Remove high q-value Targets from PEP training

trishorts:
Now that PEP is stabilized, I'm going to look at some small adjustments. In this first commit, I ran three separate large searches to compare test metrics of PEP between keeping high q-value Targets in the false training set or using only decoys. I found in all three cases that using only decoys for the false set produced superior recall results. Therefore, I am eliminating the high q-value targets from training. Anyway, there are a bazillion decoys to choose from.


rmillikin:
i agree w/ @zrolfs but will leave this choice up to @trishorts

PR #1905, opened 2020-05-20
GUI redesign

rmillikin:
this is kind of a learning experience for me regarding how to style xaml. so far so good though


rmillikin:
using this color palette btw https://flatuicolors.com/palette/gb

trishorts:
Absolutely love this opening page. You could add a link for .xml. I don't think many people understand the xml and its utility. I think it would help if we had and easy explainer on the concept.

also some bullet points to key MM features here would be great:

- calibration

- ptm discovery

- top-down

- crosslink

- glycopeptide

- non-specific (antigen)

image

trishorts:
image

The add buttons looked like tabs to me. I couldn't find "the button". Maybe these could look more like buttons some how? At the bottom of the page?

trishorts:
by accident, i dragged a raw file onto the databases page and it still went where I wanted it to go. that's cool. but a little scary. cuz you might drag something by accident and not know it.

trishorts:
image


I really like the bold text. Since you have lots of room on the right, you could have a separate button for each kind of task. that way people could see the whole menu without digging deeper. I don't know if new people understand the task system enough to know to look for search there.

trishorts:
Maybe "Quantification" should be its own task? It seems too buried in search to me. I'd like to showcase it more.

trishorts:
image

kinda feel like notifications box should be open by default?

rmillikin:
> Absolutely love this opening page. You could add a link for .xml. I don't think many people understand the xml and its utility. I think it would help if we had and easy explainer on the concept.
> 
> also some bullet points to key MM features here would be great:
> 
> * calibration
> * ptm discovery
> * top-down
> * crosslink
> * glycopeptide
> * non-specific (antigen)
> 
> image

we should probably add links to vignettes too

rmillikin:
> image
> 
> kinda feel like notifications box should be open by default?

the notifications pops up if/when a notification appears. it's just minimized by default because there are no notifications to display

trishorts:
I think each of these could be a separate line item and maybe boost the font size a point or two?

image

acesnik:
Wow, this new GUI is beautiful and nicely organized.

Could you add a link to Spritz (https://smith-chem-wisc.github.io/Spritz/) in the Help page with something like "Generate Sample-Specific Database for MetaMorpheus with Spritz" as the button?

rmmiller22:
On the about page, we should remove HPLC from "HPLC-MS/MS proteomics data" since MetaMorpheus can also analyze CZE data

rmmiller22:
> image
> 
> I really like the bold text. Since you have lots of room on the right, you could have a separate button for each kind of task. that way people could see the whole menu without digging deeper. I don't know if new people understand the task system enough to know to look for search there.

I agree that having separate buttons for each task would be nice

rmmiller22:
Should we update the task windows as well to go more with the new GUI. Nothing drastic, maybe just altering colors to have them look like they belong more with the new GUI than the old

rmmiller22:
I think that the order of the tasks in the drop down menu should go calibrate, GPTMD, search, crosslink and glyco instead of calibrate, search, GPTMD, Crosslink and glyco

rmillikin:
I added links to our papers along with a description of each task to the home page and boosted font size; I could not find good link to .xml but added a short description of the advantages of .xml in the database page; added link to spritz; removed HPLC; reordered tasks in menu.

I'd prefer to keep the design fairly minimal so I think one button for "add task" is preferable to me. It seems like the interface would be crowded to have a button for each task. Or I'd have to change the button size. I like the uniform size of the buttons currently, it's just a design thing. But I could change this somehow if people feel strongly about it.

Button styling - I'm using a "[flat UI](https://www.google.com/search?q=flat+ui&source=lnms&tbm=isch&sa=X&ved=2ahUKEwir0tT4zafrAhVSK80KHRZhBFQQ_AUoAXoECBMQAw&biw=2276&bih=1123&dpr=1.13)" style for the interface so I'd prefer to keep the button styling as-is but again I could change this if people feel very strongly about it.

Task windows - I do agree that these should be updated and I've laid some foundation for that in the app.xaml file (this controls app-wide styling). However it's more work than I want to do with this current PR.

acesnik:
Can you just have "+ADD TASK," and keep the dropdown open until the user clicks off somewhere else or hits escape? Then, it'll keep the GUI clean, but the user won't have to click it three times to add calibration/gptmd/search, for example.

acesnik:
Could change "+ADD TASK" to "+ADD TASKS" in that case.

acesnik:
And maybe add tooltip information for each of the tasks if it's not there already.

rmillikin:
separating the tasks into separate buttons seems to be very popular, let me see if I can make that look nice

rmillikin:
-added separate button for each "add task" operation
-added option for right click -> delete
-added bullet points for task descriptions in the about page
-added a column for the file name in addition to the file path for spectra + protein DB pages

PR #1910, opened 2020-05-21
show files in gui and rename psmtsv

trishorts:
in the current release, none of the glyco results files show up in the GUI

also, the file extension of them is incorrect (psm.tsv) which makes them invisible to MetaDraw without extension change.

This PR fixes those two problems


rmillikin:
any unit tests for this? seems important

rmillikin:
i guess for the CMD that doesn't require a unit test. better to test that via appveyor script if necessary.

the output files though, seem like they should be unit tested. there don't seem to be any unit tests that test the output of the glyco search otherwise they would have to be changed for this PR I believe. there should prob be at least one unit test for glyco search output.

PR #1913, opened 2020-05-24
Glycan add sugar

lonelu:
User can specify their own speical sugars in MM.
Note that current glycan database only contains limited number of glycans with specified sugars.


EngineLayer/GlycoSearch/Glycan.cs
```diff
@@ -5,9 +5,70 @@
 using System;
 using Proteomics;
 using MassSpectrometry;
-
+using System.Dynamic;
+using System.Reflection.Metadata;
+using System.Text;
 namespace EngineLayer
 {
+    public class Monosaccharide
```
trishorts, line 13: I think that you need to check if all of the elements are included (4). That names and Ids have not been reused. That nothing is of the wrong type (double instead of int). What happens if you get to a line that has been incorrectly entered? crash?

EngineLayer/GlycoSearch/GlycanBox.cs
```diff
@@ -103,7 +103,7 @@ public static IEnumerable<GlycanBox> BuildChildOGlycanBoxes(int maxNum, int[] gl
 
         public GlycanBox(int[] ids, bool targetDecoy = true):base(ids)
         {
-            byte[] kind = new byte[] { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
+            byte[] kind = new byte[Glycan.SugarLength];
```
trishorts, line 106: is length the number of sugars in the glycan, ignoring branching? maybe explain that with a comment somewhere

EngineLayer/GlycoSearch/GlycanDatabase.cs
```diff
@@ -123,7 +123,9 @@ public static List<GlycanIon> NGlycanCompositionFragments(byte[] kind)
             {
                 if (hexnac_count == 0)
                 {
-                    GlycanIon glycanIon = new GlycanIon(null, 8303819, new byte[] { 0, (byte)hexnac_count, 0, 0, 0, 0, 0, 0, 0, 0, 0 }, glycan_mass - 8303819);
+                    byte[] aIonKind = new byte[Glycan.SugarLength];
+                    aIonKind[1] = (byte)hexnac_count;
+                    GlycanIon glycanIon = new GlycanIon(null, 8303819, aIonKind, glycan_mass - 8303819);
```
trishorts, line 128: what is 8303819? comment?

Test/XLTestNGlyco.cs
```diff
@@ -25,7 +25,8 @@ public class XLTestNGlyco
         [Test]
         public static void GlyTest_GetKindString()
         {
-            byte[] kind = new byte[] {3, 4, 0, 0, 1, 0, 0, 0, 0, 0 };
+            byte[] kind = new byte[Glycan.SugarLength];
```
trishorts, line 28: here is where you can test file reading and making sure that errors in the file edited by the user are handled without crash.

PR #1915, opened 2020-05-25
generate glyco default toml via CMD

rmillikin:
**pull request type:**
bugfix

**problem:**
the command-line version of MM generates default .toml files for each task with the `-g` flag. the glyco search was not included in this, making it dependent on the GUI version to generate its .toml files.

**solution:**
added the glyco search task to the list of .toml files generated by the `-g` flag.

**repercussions of solution:**
none


trishorts:
this pr doesn't show up in my vs 
image

trishorts:
The commandline version doesn't have the "TaskDescriptor" line. don't know if that matters 
image

PR #1924, opened 2020-08-05
replace default contaminants DB with new one

rmillikin:


blfrey:
Thanks Rob!    And I like the name of the new default DB!

Issue #1761, opened 2019-10-26
Add lactylation mod


acesnik:
https://www.nature.com/articles/s41586-019-1678-1

acesnik:
Published 3 days ago 😉

PR #1927, opened 2020-08-05
Addition of Lactylation Mod

lawashburn:
Adding a lactylation mod as a form of PTM based on the initial issue [here](https://github.com/smith-chem-wisc/MetaMorpheus/issues/1761).


EngineLayer/Mods/Mods.txt
```diff
@@ -47,6 +47,12 @@ DI   HCD:C7 H11 N1 O1
 CF   H2 C2 O1
 DR   Unimod; 1.
 //
+ID   Lactylation
+TG   K
+PP   Anywhere.
+MT   Less Common
+CF   H4 C3 O2
```
acesnik, line 54: Should the number of hydrogens be 3, not 4? It looks like one is lost from nitrogen from lysine (with formal charge zero) with N-lactylation
image

e.g. methyl groups are -CH3, but we record CH2 in this file because one is lost.

lawashburn: Good point, thank you.

PR #1930, opened 2020-08-14
write indexed mzml

rmillikin:


rmillikin:
> Why do we want to do this?

it would be faster to load a dynamic .mzML file. if an index doesn't exist in an .mzML, then an index is built by the loading method, which takes time. it's better if we just have the index written already in the file

Issue #1958, opened 2020-09-24
GUI Update - details


trishorts:
1.   I miss having checkboxes by databases. I like to be able to mark which ones i want to use in a search, and leave them around so  i can use them later in a different search.

2.   I'd like to be able to delete a task. 
3.  I'd like to be able to re-order tasks.

rmillikin:
Let's use this issue to centralize to-do items for the new GUI.

- Double-click doesn't seem to open stuff for me anymore, need to fix that
- Need a right click -> Delete All option (equivalent to "Clear" in the old GUI)

rmillikin:
- Need to reset the output folder (to empty) when all spectra files are deleted from the GUI

rmillikin:
- Font size needs to be bigger

rmillikin:
- Add a resizing splitter thing between the spectra/proteinDB and task areas in the "Run" tab

PR #1959, opened 2020-09-24
bugfixes for the GUI

rmillikin:
#1958 

- Can select multiple DB/spectra files and delete them
- Delete key on the keyboard will delete tasks
- Plus or minus keys will move tasks up or down
- Double-clicking a file or task will open it
- If all spectra files are deleted, the output file path is cleared
- Can right click -> Delete All for spectra, protein DBs, tasks
- Can right click -> Move task up or Move task down
- Clicking the "Select" button in MetaDraw (to add a spectra file or results file) will correctly display .mgf and .tsv as options
- Fixed a crash loading XL Task results in MetaDraw
- Increased the font size of the tasks in the Run tab from 11 to 13
- Added grid splitters to the Run tab so that the DB/spectra/tasks areas can be resized

### DIFF
--- a/GUI/MetaDraw/MetaDraw.xaml.cs
+++ b/GUI/MetaDraw/MetaDraw.xaml.cs
@@ -178,9 +178,9 @@ namespace MetaMorpheusGUI
             try
             {
                 // TODO: print warnings
-                foreach (var psm in PsmTsvReader.ReadTsv(psmtsvFileName, out List<string> warnings))
+                foreach (PsmFromTsv psm in PsmTsvReader.ReadTsv(psmtsvFileName, out List<string> warnings))
                 {
-                    if (loadPsmsOfAllSpectraFiles || fileNamesWithExtension.Contains(psm.Filename) || fileNamesWithoutExtension.Contains(psm.Filename) || potentiallyCalibratedFileNames.Contains(psm.Filename)) // in case results are from a calibrated file
+                    if (loadPsmsOfAllSpectraFiles || fileNamesWithExtension.Contains(psm.FileNameWithoutExtension) || fileNamesWithoutExtension.Contains(psm.FileNameWithoutExtension) || potentiallyCalibratedFileNames.Contains(psm.FileNameWithoutExtension)) // in case results are from a calibrated file
                     {
                         allPsms.Add(psm);
                     }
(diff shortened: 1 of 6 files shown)


GUI/MetaDraw/MetaDraw.xaml.cs
```diff
@@ -48,6 +48,8 @@ public partial class MetaDraw : Window
         private SolidColorBrush modificationAnnotationColor;
         private Regex illegalInFileName = new Regex(@"[\\/:*?""<>|]");
         private ObservableCollection<string> plotTypes;
```
trishorts, line 50: i feel like we are going to run into case sensitivity problems. is that an issue? if yes,  I think it can be dealt with

rmillikin: the ".ToLower()" on line 129 (new code, line 127 in the old code) should take care of it

trishorts:
i can right click open a file but not double click to open.

trishorts:
neither delete nor backspace will delete a task thats already been run (after reset tasks has been hit)

trishorts:
+ and - works to change order

trishorts:
I think we really need the file check boxes. After calibration I have the following scenerio. and I don't want the un calibrated files hanging around
image

trishorts:
delete does work for files

rmillikin:
> i can right click open a file but not double click to open.
I wasn't able to reproduce this bug, can you give me more steps to help reproduce? The following works for me:

Drag-and-drop .raw, search task .toml, and .fasta DB into MM. Double-click opens all of them correctly. I click "Run MetaMorpheus" and let the run complete. Double-click opens files written by the task. Double-click will not open .raw or .fasta or the task while in progress. After the run completes, I click "Reset Tasks" and I'm able to double-click to open all 3 again.

> neither delete nor backspace will delete a task thats already been run (after reset tasks has been hit)

I wasn't able to reproduce this one either, can I get more steps? The following works for me:
Drag-and-drop .raw, search task .toml, and .fasta DB into MM. Click "Run MetaMorpheus", wait for run to complete, hit "Reset Tasks", hitting Del key deletes task.

We could screenshare too so I can get the exact steps and try to reproduce this on my machine.

trishorts:
> 
> 
> > i can right click open a file but not double click to open.
> 
> I wasn't able to reproduce this bug, can you give me more steps to help reproduce? The following works for me:
> 
The problem was on my end. I was reviewing PR but hadn't checked it out. After checking out and re-running, I was able to double click open the files. You can mark this as resolved. Sorry for the wild goose chase.

trishorts (approved):
You'll have to decide what you want to do about the check boxes for raw files and databases. You can merge this PR and open a separate issue or try to fix it in this PR.

lonelu (approved):

lonelu:
Just something not important. When I click Visualize, the MetaDraw pops out immediately. I prefer just to get into the page and click a 'MetaDraw' button. It is possible that we have other visualization tools in the Visualize page. For example, the peak extraction tool Rob developed.

PR #1970, opened 2020-10-25
Protein quant update

rmillikin:
- Better protein quant (median polish) via FlashLFQ/mzLib update
- Summarization of protein quant intensities for fractionated data into biorep-level intensities
- Fixed a bug where the experimental design did not get read in if normalization was not enabled
- Added more experimental design error checking to CMD and GUI pre-run
- Moved experimental design code around a bit to be more centralized/easily re-usable
- ExperimentalDesign.tsv file is copied to output folder in the Search task

Minor non-quant stuff:
- Added back the ability for the Calibration task to write indexed .mzML files, with a checkbox option to write non-indexed .mzML


trishorts (approved):
This is a good pr. I tested the GUI and a number of scenarios on the command line.

rmmiller22 (approved):

Issue #1974, opened 2020-11-09
MM_V0.0.312 Crashes when trying to visualize annotated MS2 spectra


kyleabrown20:
When I try to visualize the annotated MS2 results from a search, the program crashes when I click on a peptide spectra.

trishorts:
thanks kyle. i'll check this quick and see if i can reproduce the error.

trishorts:
i'm not getting an error.  Can you tell us a little more about your search? Maybe we could run a search with one of your files?

![image](https://user-images.githubusercontent.com/16841846/98590166-09367480-2294-11eb-8e80-4a13693338c2.png)

kyleabrown20:
Thanks for the help! Here is a screen shot of the error in case it helps.

[cid:6591fc0d-3ed2-4ab6-acab-1d8da7806a60]


-------------------------------------

Kyle A. Brown, PhD

Department of Surgery

University of Wisconsin-Madison


________________________________

trishorts:
somehow your screenshot didn't come through

rmillikin:
Just a note that this is resolved. It occurred because there was a period in the spectra file name. Fixed in #1976 .

PR #1976, opened 2020-11-09
custom file extension getter

rmillikin:
custom file extension getter that is more reliable than the microsoft one.
- can handle compressed files like .fasta.gz
- ignores periods in file names
- this change allows MetaDraw to load data from spectra files with periods in the name. related to #1974

### DIFF
--- a/EngineLayer/GlobalVariables.cs
+++ b/EngineLayer/GlobalVariables.cs
@@ -249,7 +249,7 @@ namespace EngineLayer
         }
 
         /// <summary>
-        /// Gets the file name with extension, keeping .gz appended for compressed files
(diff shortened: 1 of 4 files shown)


trishorts (approved):
seems fine. I tested that kyle's filename with the comma is properly handled. maybe it's a little inefficint but hardly every used so no big deal. looking at your code gave me an idea for an alternative.

you could reverse the whole file name
then do IndexOf(Char, Int32, Int32) for '.'
take the reverse of the string up to that point and see if it is gz
if gz then continue to the next '.' otherwise return reverse of what you already have. 

This way you don't have to build up string a char at a time until '.'

rmmiller22 (approved):

PR #1980, opened 2020-11-14
Metadraw reorganization, unit tests

rmillikin:
this PR moves a lot of metadraw code to the engine layer. the overall goal is to separate the MetaDraw logic and plotting from the GUI, and unit test the logic and plotting.


lonelu (approved):
All look good and It works on my local. 
As we allow multiple files to be load. Should we change the GUI to show multiple files as in Main GUI?

rmmiller22:
Overall looks good. a few minor comments:
1. For PSM visualization, the error message when only a subset of spectra files are loaded is great - but when there are a lot of files note loaded it can be a bit overwhelming. Maybe we could make it more general (X number of psms were not loaded because spectra files  1, 2, and 3 were not provided)

2. in spectral annotation the peptide sequence can overlap with the peptide information. Maybe we could put the info outside of the chart?
image

3. Similar issue in data visualization where key can be overrun by data, maybe have the key be outside of the graph.
image

I will approve but just some things to think about

rmmiller22 (approved):

rmillikin:
> Overall looks good. a few minor comments:
> 
> 1. For PSM visualization, the error message when only a subset of spectra files are loaded is great - but when there are a lot of files note loaded it can be a bit overwhelming. Maybe we could make it more general (X number of psms were not loaded because spectra files  1, 2, and 3 were not provided)
> 2. in spectral annotation the peptide sequence can overlap with the peptide information. Maybe we could put the info outside of the chart?
>    image
> 3. Similar issue in data visualization where key can be overrun by data, maybe have the key be outside of the graph.
>    image
> 
> I will approve but just some things to think about

Thx @rmmiller22 , I will edit the warning message to that in a separate PR. the latter issue - I did it this way in particular in an attempt to simulate what the output PDF looks like. so if your current window looks messed up, your export PDF will look equally messed up. By dragging the window around it helps you get a better-looking PDF without a lot of trial and error. I think we should do a better job of getting a good looking chart by default but that will prob be a separate PR.

Issue #1966, opened 2020-10-13
possible issue with editing save folder


trishorts:
Found a small bug in metamorpheous. When I try and select a filepath to save the results to on the run tab (from the below screenshot) the dialog window that pops up will not let me choose a new output folder 

![image](https://user-images.githubusercontent.com/16841846/95902835-f8f5ad00-0d5a-11eb-83aa-ee56cc102f41.png)

trishorts:
this was mailed to me by coner

rmillikin:
that button opens the folder that's currently in the "output folder" textbox, it's not a folder selection dialog. so this isn't really a bug, it's maybe more of an unclear button purpose. we could add a dialog selector button though.

PR #1983, opened 2020-11-16
fix output folder GUI bug

rmillikin:
moved `PreviewKeyDown="Window_PreviewKeyDown"` from a window-wide event to specific to certain GUI elements, fixing issue #1978


trishorts (approved):
seems fine. backspace works now, which i believe is the bug that is the target of the fix. 

Before the search runs, i do think this should be a folder dialog so that when you hit the open button, you can select a save folder. So there is still this issue: https://github.com/smith-chem-wisc/MetaMorpheus/issues/1966

After the search is run, then it should just pop open the folder with the saved data.

PR #1985, opened 2020-11-17
ability to add spectra file directory in CMD

### LINKED ISSUE #1901: Hitting script length limit in Powershell
I just encountered a Powershell issue with the commandline program. Running 400 spectra files or so, and I run into this:
image

Turns out there's a script length limit in Powershell.
http://systemcentersynergy.com/max-script-block-size-when-passing-to-powershell-exe-or-invoke-command/

Looks like the solution is to make the huge parameter list (spectra files) into a file. I think in this case we should allow the option of using a spectrum file list.
https://support.microsoft.com/en-gb/help/830473/command-prompt-cmd-exe-command-line-string-limitation

rmillikin:
can specify a directory of spectra files. fixes #1901 .

unfortunately, no unit test for the new CMD option as this seems to not play well w/ appveyor. but I did manually test it.

### DIFF
--- a/EngineLayer/GlobalVariables.cs
+++ b/EngineLayer/GlobalVariables.cs
@@ -226,7 +226,7 @@ namespace EngineLayer
                 FileName = path
             };
 
-            if (useNotepadToOpenToml && Path.GetExtension(path) == ".toml" && RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            if (useNotepadToOpenToml && Path.GetExtension(path).ToLowerInvariant() == ".toml" && RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
                 p.StartInfo.FileName = "notepad.exe";
                 p.StartInfo.Arguments = path;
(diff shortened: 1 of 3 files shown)


EngineLayer/GlobalVariables.cs
```diff
@@ -266,16 +266,44 @@ public static string GetFileExtension(string fileWithExtension)
                 {
                     extension = new string(sb.ToString().Reverse().ToArray());
 
-                    if (extension != ".gz" || !fileWithExtension.Substring(0, i).Contains('.'))
+                    if (!extension.ToLowerInvariant().EndsWith("gz") || extension.Count(p => p == '.') >= 2)
```
zrolfs, line 269: I don't understand this change.

zrolfs (approved):

EngineLayer/GlobalVariables.cs
rmillikin: I guess sometimes .bgz and similar extensions exist besides just .gz, so I made it `!  .EndsWith("gz")` instead of `!= ".gz"`. Second part was because sometimes files can contain periods ("my.file.mzML")

trishorts (approved):
works nice. like that we can use the whole folder. tried empty folder and wrong folder too and it dealt with that just fine.

PR #1996, opened 2020-11-24
[WIP] GPTMD task-friendly spectral libraries

rmillikin:
There are some issues w/ the current code when using spectral libraries and the GPTMD task. GPTMD writes new databases and then all the old databases are disabled. Since we're classifying spectral libraries as databases, these libraries mistakenly get disabled for subsequent tasks. I edited the code so that running the GPTMD task will not disable using a spectral library in a subsequent search task.


GUI/MainWindow.xaml.cs
```diff
@@ -166,7 +166,7 @@ private void AddNewProteinDatabaseFromGptmd(object sender, XmlForTaskListEventAr
             }
             else
             {
-                foreach (var db in ProteinDatabases)
+                foreach (var db in ProteinDatabases.Where(p => !GlobalVariables.AcceptedSpectralLibraryFormats.Contains(GlobalVariables.GetFileExtension(p.FilePath))))
```
zrolfs, line 169: unclear what this statement is needed for. A brief comment would be good.

Yuling1996 (approved):

zrolfs (approved):

PR #2000, opened 2020-12-03
added cross validaton to PEP calculation

trishorts:
This pull request adds cross-validation to the PEP calculation (see https://bmcbioinformatics.biomedcentral.com/articles/10.1186/1471-2105-13-S16-S3). Formerly, 25% of the data was used for training and a separate 25% of the data was used for testing. The metrics from the test were reported in the output. The trained model was then applied to all of the data including the training data, which could lead to overfitting.

In this PR, I added cross-validation. Now, all of the data is broken into four groups. Three groups are used for training and one is used for testing and application of the trained model. This process is repeated four times so that all PSMs are scored by a separately trained model. The metrics from the four models are averaged and presented as a single metric.


EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -552,7 +626,7 @@ private static float GetMobilityZScore(PeptideSpectralMatch psm, PeptideWithSetM
                                 label = false;
                                 newPsmData = CreateOnePsmDataEntry(searchType, fileSpecificParameters, psm, sequenceToPsmCount, timeDependantHydrophobicityAverageAndDeviation_unmodified, timeDependantHydrophobicityAverageAndDeviation_modified, fileSpecificMedianFragmentMassErrors, chargeStateMode, csm.BestMatchingPeptides.First().Peptide, 0, label);
                             }
-                            else if (!csm.IsDecoy && !csm.BetaPeptide.IsDecoy && psm.FdrInfo.QValue <= 0.01)
+                            else if (!csm.IsDecoy && !csm.BetaPeptide.IsDecoy && psm.FdrInfo.QValue <= 0.0005)
```
lonelu, line 629: Why change QValue to 0.0005?

trishorts: instead of training with 25%, i'm now training with 75%. So, i thought i could change the cutoff and train on better values. moving from 0.01 to 0.005 only eliminates a small fraction of PSMs, but it eliminates the worst of the PSMs below 1%.

lonelu (approved):

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -165,25 +231,33 @@ public static string ComputePEPValuesForAllPSMsGeneric(List<PeptideSpectralMatch
...
                     }
                 });
+            return ambiguousPeptidesResolved;
+        }
 
-            var predictions = trainedModel.Transform(testData);
-
-            CalibratedBinaryClassificationMetrics metrics;
-            try
+        public static List<int>[] Get_PSM_Group_Indices(int count, int numGroups)
```
zrolfs, line 240: "numPSMs" would be a more descriptive variable name than "count"

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -165,25 +231,33 @@ public static string ComputePEPValuesForAllPSMsGeneric(List<PeptideSpectralMatch
...
+                groupsOfIndicies[i] = new List<int>();
             }
-            catch
+
+            int myIndex = 0;
+            while (myIndex < count)
             {
-                return "";
+                int subIndex = 0;
+                while (subIndex < numGroups && myIndex < count)
```
zrolfs, line 252: Nice work separating them as 1,2,3,4,1,2,3,4 rather than 1,1,1,1,1,1,1,1 .... 4,4,4,4,4,4

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -57,84 +57,150 @@ public static string ComputePEPValuesForAllPSMsGeneric(List<PeptideSpectralMatch
...
 
-            //Parallel operation of the following code requires the method to be stored and then read, once for each thread
-            //if not output directory is specified, the model cannot be stored, and we must force single-threaded operation
-            if (outputFolder != null)
+                return AggregateMetricsForOutput(allMetrics, sumOfAllAmbiguousPeptidesResolved);
+            }
+            catch
             {
-                mlContext.Model.Save(trainedModel, trainingData.Schema, Path.Combine(outputFolder, "model.zip"));
+                return "Posterior error probability analyis failed. This can occur for small data sets when some sample groups are missing positive or negative training examples.";
```
zrolfs, line 108: "analysis"

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -57,84 +57,150 @@ public static string ComputePEPValuesForAllPSMsGeneric(List<PeptideSpectralMatch
...
             var trainer = mlContext.BinaryClassification.Trainers.FastTree(labelColumnName: "Label", featureColumnName: "Features", numberOfTrees: 400);
             var pipeline = mlContext.Transforms.Concatenate("Features", trainingVariables)
                 .Append(trainer);
-            var trainedModel = pipeline.Fit(trainingData); //this is NOT thread safe
 
-            int maxThreads = fileSpecificParameters.FirstOrDefault().fileSpecificParameters.MaxThreadsToUsePerFile;
-            object lockObject = new object();
+            List<CalibratedBinaryClassificationMetrics> allMetrics = new List<CalibratedBinaryClassificationMetrics>();
+            int sumOfAllAmbiguousPeptidesResolved = 0;
+            try
```
zrolfs, line 78: This is just a pet peeve, but I hate try-catch statements. They dramatically slow down the code.
If you can think of an if-else to handle it instead, that'd be preferred. No worries if not.

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -57,84 +57,150 @@ public static string ComputePEPValuesForAllPSMsGeneric(List<PeptideSpectralMatch
...
-            Parallel.ForEach(Partitioner.Create(0, psms.Count),
+        }
+
+        public static int Compute_PSM_PEP(List<PeptideSpectralMatch> psms, List<int> psmIndices, MLContext mLContext, TransformerChain<BinaryPredictionTransformer<Microsoft.ML.Calibrators.CalibratedModelParametersBase<Microsoft.ML.Trainers.FastTree.FastTreeBinaryModelParameters, Microsoft.ML.Calibrators.PlattCalibrator>>> trainedModel, string searchType, List<(string fileName, CommonParameters fileSpecificParameters)> fileSpecificParameters, Dictionary<string, int> sequenceToPsmCount, Dictionary<string, float> fileSpecificMedianFragmentMassErrors, int chargeStateMode, string outputFolder)
+        {
+            int maxThreads = fileSpecificParameters.FirstOrDefault().fileSpecificParameters.MaxThreadsToUsePerFile;
+            object lockObject = new object();
+            int ambiguousPeptidesResolved = 0;
+
+            if (String.IsNullOrEmpty(outputFolder))
```
zrolfs, line 174: This (and ln 187) feels hacky. Can you comment what's going on here?

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
trishorts: changed it to psmsCount, which is what is used in the function call "List<int>[] psmGroupIndices = Get_PSM_Group_Indices(psms.Count, numGroups);"

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
trishorts: thanks

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
trishorts: corrected

acesnik:
This is the two thousandth pull request 😮 👏

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
trishorts: i think it is. but i added a comment to explain. definitely up for alternatives.

zrolfs (approved):

rmillikin:
looks good to me but this is something I noticed that could happen on occasion:

this PR subdivides the PSMs into 4 groups and then if any of the groups doesn't have a true or false labeled PSM then the PEP training fails. it seems more robust to me to take the "trues", split those up randomly into 4 groups, then do the same for "falses". that way if there are at least 4 trues and 4 falses in the set of PSMs then you'll get valid subgroups

rmillikin (approved):

PR #2002, opened 2020-12-10
Pep cross validation

trishorts:
Training data for the binary decision tree is broken into four groups of PSMs. This is done consecutively by placing PSMs 1-4 into goups 1-4 and then repeating with the next group. By random chance, some groups can end up with only targets or decoys. This probably occurs for only targets and only with really small training sets like you'd get from a single raw file with not that many psms.

To rectify this problem, I first break up the psms into two groups, targets and decoys. I split all the targets between the four training groups. Then I split all the decoys into the four training groups. If there are at least four targets and four decoys, then every training group  gets at least one.


lonelu (approved):

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -251,23 +249,58 @@ public static int Compute_PSM_PEP(List<PeptideSpectralMatch> psms, List<int> psm
...
+                }
+                else
+                {
+                    targetPsmIndexes.Add(i);
+                }
+            }
+
             int myIndex = 0;
-            while (myIndex < psmsCount)
+            if (decoyPsmIndexes.Count > 0)
```
zrolfs, line 278: This check is unnecessary; the while loop will exit if this is false

zrolfs: Unclear how PEP trains if there are no decoys...

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -251,23 +249,58 @@ public static int Compute_PSM_PEP(List<PeptideSpectralMatch> psms, List<int> psm
...
+                    {
+                        groupsOfIndicies[subIndex].Add(decoyPsmIndexes[myIndex]);
+                        subIndex++;
+                        myIndex++;
+                    }
+                }
+            }
+
+            myIndex = 0;
+            if (targetPsmIndexes.Count > 0)
```
zrolfs, line 293: Same thing here; this conditional isn't needed.

zrolfs (approved):

EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
trishorts: I removed the unnecessary checks. 

PEP will not train if any of the four sets is missing both targets and decoys. The PEP columns will be empty. 

image

rmmiller22 (approved):

PR #2011, opened 2021-01-19
[WIP] install bash and procps in docker version for nextflow

rmillikin:
this PR installs bash and procps in the Linux docker image. bash is required to run metamorpheus in a nextflow pipeline, and procps lets nextflow monitor computational resources used by the image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,5 +4,11 @@ FROM mcr.microsoft.com/dotnet/core/runtime:3.1-alpine AS build
 ## Copies contents of the build folder into the Docker image
 ADD CMD/bin/Release/netcoreapp3.1/ /metamorpheus/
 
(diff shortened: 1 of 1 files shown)


rmillikin:
getting a weird failure message from docker. will investigate

zrolfs (approved):
No idea what's going on here, but I'll approve it for bureaucracy's sake

rmillikin:
seems to be a bug with Docker version 2.2.0.5, which is the version appveyor is running. https://github.com/microsoft/hcsshim/issues/813

rmmiller22 (approved):

trishorts (approved):
same as zach. seems innocuous enough and unlikely to break to MM. I can wait til the release happens and try to run the docker I guess.

Just asking, is their a way to download the docker from Appveyor and test it just like we download the installer?

Issue #2006, opened 2021-01-14
Bug with annotation in MetaDraw


Dmorgen:
Hi Guys,

got weird annotation in Metadraw whereby the b and y ions are all marked on the same exact position, even when it shouldn't.
[48474_FMEDMMPK.pdf](https://github.com/smith-chem-wisc/MetaMorpheus/files/5816531/48474_FMEDMMPK.pdf)

Cheers,
David.

trishorts:
David, 

Can you share the raw file and organism? I'd like to debug this and see what is going on.

Dmorgen:
please see your email.

trishorts:
I reproduced this bug. I'll try to figure out where the bug lies. 
![image](https://user-images.githubusercontent.com/16841846/105213452-58aae380-5b14-11eb-82a4-daf9242698bf.png)

rmmiller22:
This issue has been fixed, and this fix has been incorporated into the latest release made today (0.0.313)

PR #2014, opened 2021-01-21
Fix Complementary Ion Mz assignments

zrolfs:
This PR fixes the assignment of mz values to matched comp ions. Comp ions function as if they had created a comp ion for every observed fragment. However, we don't actually add these ions to the observed spectra. When a theoretical ion matched to an added comp ion, the mz of the original observed ion was recorded (because the m/z of the comp ion didn't exist). I changed the m/z to reflect the m/z of the comp ion as though it was present in the spectrum.


zrolfs:
Fixes #2006

trishorts (approved):
This is a good PR. Code changes are commented and covered by unit tests. Project builds and runs on my machine. Problem is successfully solved.

Yuling1996 (approved):

rmmiller22 (approved):

PR #2021, opened 2021-02-12
Download UniProt Database from GUI

trishorts:
This PR enables GUI users to download a proteome from UniProt directly from a pre-set list of available proteomes. At this time, the user cannot choose something that is not on the list. The proteome can be downloaded in either xml or fasta format, with and without unreviewed entries. The file can include protein isoforms when requested. And, it can either be compressed or uncompressed. The downloaded proteome is automatically placed in a proteomes folder. Downloaded the same proteome will overwrite the existing file. There is a button that allows the user to open the "proteomes" folder so that they can load them into the search gui. When downloads fail, the gui will not crash. Instead, no proteome will be loaded into the GUI. There is no error message.

Two methods were placed into a separate class outside of the GUI to enable complete unit test coverage.


trishorts:
Make sure that the File Creation Date can be accessed for the downloaded proteomes and add that date to the Prose.txt file.

GuiFunctions/Databases/DownloadUniProtDatabaseFunctions.cs
```diff
@@ -0,0 +1,162 @@
...
+{
+    /// <summary>
+    /// We place functions from the GUI in the class to enable testing.
+    /// </summary>
+    
+    public class DownloadUniProtDatabaseFunctions
+    {
+
+        public static string GetUniProtHtmlQueryString(string proteomeID, bool reviewed, bool isoforms, bool xmlFormat, bool compressed)
+        {
```
zrolfs, line 16: This could/should be refactored with a string builder, but it's not pressing.

zrolfs (approved):

GuiFunctions/Databases/DownloadUniProtDatabaseFunctions.cs
trishorts: Done

zrolfs: Sent a PM. I think we can do this method without nested conditionals.

zrolfs (approved):

rmillikin:
works well for me. I only have some aesthetic requests related to the buttons. I also think "databases" is clearer than "proteomes"  because to me a proteome is sort of an abstract concept and a database is a file, but if you really want "proteomes" that's fine w/ me.

In MainWindow.xaml:
```
<!--Buttons to add databases-->
<Button Name="DownloadUniProtDatabaseButton" Grid.Row="2" Content="DOWNLOAD DATABASE" 
        Click="DownloadUniProtDatabase_Click" Style="{StaticResource ImportantButtonStyle}" Margin="10,0,0,0" HorizontalAlignment="Left" />
                    
<Button Name="OpenProteomesFolder" Grid.Row="2" Content="OPEN DATABASE FOLDER" 
        Click="OpenProteomesFolder_Click" Style="{StaticResource ImportantButtonStyle}" Margin="211,0,0,0" HorizontalAlignment="Left"/>
                    
<Button Name="AddDatabaseButton" Grid.Row="2" Content="+ADD DATABASE" 
        Click="AddProteinDatabase_Click" Style="{StaticResource ImportantButtonStyle}" Margin="0,0,10,0" HorizontalAlignment="Right"/>

<Button Name="AddDefaultContaminantsButton" Grid.Row="2" Content="+ADD DEFAULT CONTAMINANTS" 
        Click="AddDefaultContaminantDatabase_Click" Style="{StaticResource ImportantButtonStyle}" Margin="0,0,211,0" HorizontalAlignment="Right"/>
```

and remove the columns from line 374.

In DownloadUniprotDatabaseWindow.xaml:
```
<Button Content="DOWNLOAD" Name="downloadProteomeButton" Click="DownloadProteomeButton_Click" Grid.Row="3" Width="200"
        Background="{StaticResource AccentColor}" Foreground="{StaticResource TextColor2}" Margin="10" 
        ToolTip="Downloads can take several minutes" ToolTipService.ShowOnDisabled="true" BorderThickness="0"/>
```

rmillikin (changes requested):
dummy comment. left real comments on PR

trishorts:
i used the columns to get the four buttons evenly spaced. that was easiest way I could find. 
image

rmillikin (approved):

rmmiller22 (approved):

kyp4 (approved):
Looks good - I was able to download databases and complete a search

PR #2022, opened 2021-02-12
Fix calibration/experimental design bug

rmillikin:
Fixed a bug where the Calibration task did not write the Experimental Design file correctly

### DIFF
--- a/EngineLayer/ExperimentalDesign.cs
+++ b/EngineLayer/ExperimentalDesign.cs
@@ -89,7 +89,7 @@ namespace EngineLayer
             return expDesign;
         }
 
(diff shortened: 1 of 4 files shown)


trishorts (approved):
i got no error this time when i ran hayes four files through calibration, search and quantification. and i got the correct number of intensities for proteins, peptides and peaks.

zrolfs (approved):
Looks good. Code coverage is down because there's no unit test with errors (ln 313), but that's superficial. Might want a test covering ln 336, though.

rmmiller22 (approved):

PR #2025, opened 2021-02-17
small experimental design GUI changes

rmillikin:
- Add ability to copy+paste tab- or comma-delimited cells into the Experimental Design window, e.g. from Excel (currently only w/ the ctrl+V keyboard shortcut)
- Fix an issue in the experimental design GUI where the buttons at the bottom of the window obscured data cells


zrolfs (approved):

trishorts (approved):
The file names are grey and the condition/biorep/fraction/tech rep are white. White boxes can be copied/pasted back and forth between MM and excel. After ED update, search ran to completion.

rmmiller22 (approved):

PR #2040, opened 2021-03-11
Fix small quant bugs

rmillikin:
- Fixed a bug where quantification only reported one column of intensity data if the experimental design was not defined
- Fixed a bug where the protein quantification intensity headers were not accurate if the samples were not fractionated but multiple technical replicates were present
- Made a unit test to check for these problems


EngineLayer/ProteinParsimony/ProteinGroup.cs
```diff
@@ -143,13 +144,16 @@ public string GetTabSeparatedHeader()
             if (FilesForQuantification != null)
             {
                 bool unfractionated = FilesForQuantification.Select(p => p.Fraction).Distinct().Count() == 1;
-                bool conditionsDefined = FilesForQuantification.All(p => p.Condition == "Default") || FilesForQuantification.All(p => string.IsNullOrWhiteSpace(p.Condition));
+                bool conditionsUndefined = FilesForQuantification.All(p => string.IsNullOrEmpty(p.Condition));
+
+                // this is a hacky way to test for SILAC-labeled data...
```
zrolfs, line 149: not sure why you'd need to test for SILAC-labeled data here, so the comment doesn't really make sense.

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -220,10 +220,12 @@ private void QuantificationAnalysis()
...
                 spectraFileInfo = new List<SpectraFileInfo>();
 
-                foreach (var file in Parameters.CurrentRawFileList)
+                for (int i = 0; i < Parameters.CurrentRawFileList.Count; i++)
                 {
+                    var file = Parameters.CurrentRawFileList[i];
+
                     // experimental design info passed in here for each spectra file
-                    spectraFileInfo.Add(new SpectraFileInfo(fullFilePathWithExtension: file, condition: "", biorep: 0, fraction: 0, techrep: 0));
+                    spectraFileInfo.Add(new SpectraFileInfo(fullFilePathWithExtension: file, condition: "", biorep: i, fraction: 0, techrep: 0));
```
zrolfs, line 228: can replace "file" with "Parameters.CurrentRawFileList[i]" so that you don't have to create an instance of "file" in memory.
We only use the variable once, so faster to look it up on the fly.

Test/SilacTest.cs
```diff
@@ -212,8 +212,11 @@ public static void TestSilacQuantification()
             //test proteins
             string[] output = File.ReadAllLines(TestContext.CurrentContext.TestDirectory + @"/TestSilac/AllQuantifiedProteinGroups.tsv");
             Assert.AreEqual(output.Length, 2);
-            Assert.IsTrue(output[0].Contains("Intensity_silac\tIntensity_silac(K+8.014)")); //test that two files were made
-            Assert.IsTrue(output[1].Contains("1237436.8670764582\t618718.4335382291")); //test the heavy intensity is half that of the light (per the raw file)
+            Assert.IsTrue(output[0].Contains("Intensity_silac\tIntensity_silacPart2\tIntensity_silac(K+8.014)\tIntensity_silacPart2(K+8.014)")); //test that two files were made
```
zrolfs, line 215: Why are there four headings now? What is Part2?

zrolfs (changes requested):

trishorts (approved):
This looks good to me. 

I made some edits to the Experimental Design and Quantification pages. Please spend five minutes on that and check that it matches with the current output expectations.

There is an open question about combining peptides intensities for the peptide output (only when protein inference is also selected). I sent Alexey an email for advice. 

The normalization algorithm should be mentioned somewhere explicitly.

EngineLayer/ProteinParsimony/ProteinGroup.cs
rmillikin: I wanted the SILAC reporting to be unaffected by the changes I made to biorep-level summarization. Currently SILAC will report 1 column per label per spectra file, and the SILAC code uses the "condition" field to organize this info, even if the experimental design is not defined by the user. So I needed a way to distinguish between experimental design being used in SILAC automatically vs. being defined by the user

TaskLayer/SearchTask/PostSearchAnalysisTask.cs
rmillikin: I prefer the clarity of the way I wrote it and I don't think the time or memory usage for this code block is meaningful

Test/SilacTest.cs
rmillikin: This part is a bit confusing to me, but hopefully you can clarify.. the unit test seems to me to be 2 spectra files which are identical, and 2 labels. So there should be 4 columns in the output right? 2 light and 2 heavy columns? 1 light/heavy for each spectra file?

zrolfs: Yep, you're right. Looks like I was lazy in my original coding for this test.

Test/SilacTest.cs
```diff
@@ -212,8 +212,11 @@ public static void TestSilacQuantification()
...
+
+            //TODO: the heavy peptide in "silacPart2" seems to be missing even though this spectra file is a copy of "silac"?
+            //Assert.IsTrue(output[1].Contains("875000\t875000\t437500\t437500")); //test the heavy intensity is half that of the light (per the raw file)
+            Assert.IsTrue(output[1].Contains("875000\t875000\t437500")); //test the heavy intensity is half that of the light (per the raw file)
```
zrolfs, line 219: Is there a reason the last intensity was removed?

EngineLayer/ProteinParsimony/ProteinGroup.cs
zrolfs: I think I understand. Could we name the boolean something that reflects that, if it's only being used for SILAC? Right now the connection isn't intuitive.

rmillikin: ya I will update the comment too

Test/SilacTest.cs
rmillikin: it appears to be a bug but I don't know how long it's been there. it seems to pre-date this change and the unit test never checked for 4 intensity values so it's unclear how long it's been like this. I added a todo item on line 217. I'll see if I can figure out what's going on in the next couple of days. The PSMs get passed correctly into FlashLFQ but one of the four peaks has a 0 intensity value, which is confusing because the peakfinding worked for an identical .mzML file.

zrolfs (approved):

Test/SilacTest.cs
rmillikin: fixed in 80d4d1a , a list variable was re-used but was modified in the called method

rmmiller22 (approved):

PR #2045, opened 2021-03-16
autodetect dissociation type from scan header in classic search

### LINKED ISSUE #2043: extra line breaks in mgf files cause a read crash (fix in mzLib).
Please see https://github.com/smith-chem-wisc/mzLib/issues/566

rmillikin:
this PR implements a new feature in which spectra files with mixed fragmentation techniques can be searched correctly via the classic search engine. the scan header will be used to determine the fragments to generate and search for, for each peptide/scan combo. this search mode is enabled by selecting "autodetect" as the dissociation type when creating a task.

also fixes #2043 via mzLib update.


trishorts (approved):
This is a good PR. Sufficient unit test coverage. I ran the whole vignette with unknown dissociation type. That was fine. I tested with an ethcd raw file and with a TMT file that had MS3. Both also fine. The only question i have is if we should change the word unknown to something else like unknown/variable or auto-detect.

kyp4 (approved):
Ran search with unknown fragmentation

rmmiller22:
I tested this and the performance on a normal file with Autodetect and its fragmentation type were identical. When I actually used a mixed fragmentation type raw file - the results were very poor. Will be investigating further to see if it is just a bad file, or if it is a larger issue.

rmmiller22 (approved):

PR #2047, opened 2021-03-31
Update to net5.0

acesnik:


acesnik:
ready for review accompanying https://github.com/smith-chem-wisc/mzLib/pull/569

acesnik:
Looks like this one will require some changes to the AppVeyor build. 😅

acesnik:
@rmmiller22 @rmillikin @trishorts 

Hi there! This is a PR that's ready for review. It aims to update MM to net5.0, which is now the framework for mzLib. The motivation here was that net5.0 is now the default `dotnet` that is downloaded from `homebrew` and `conda` repositories, and there is also pretty substantial speed increases for certain computer architectures with the switch from `netcoreapp3.1` to `net5.0`.

This PR also fixes a few issues with the installer, such as creating the `Proteomes` folder with the installer in the target programs folder. Before, the file was being installed to the `Mods` folder, which made a crash when trying the new `Download Proteomes` feature with an installed version of MM.

There was also an issue with `mzPlot.dll` being installed in the `CMD` that led to a crash on linux. This is fixed because net5.0 and net5.0-windows are now explicit, so the files needed for WPF are now explicitly only downloaded for net5.0-windows and not net5.0 that is used for the `CMD` project.

Thanks! -AC

acesnik:
NB: I changed `netcoreapp3.1` to `net5.0` in the AppVeyor environment variables `CommandLineBuildFolder` and `DotnetVersion` (a new variable, so that this is easier to reverse if needed) to get this PR to build.

zrolfs (approved):

rmillikin (approved):

rmillikin:
looks good to me, probably worth double-checking to make sure the installer works. installed version ran ok on appveyor though.

rmillikin:
I tested the installer on my machine (downloaded from appveyor via script) and the installer is 6x the size of the current installer (11 MB vs 65 MB) but it installs and MM runs ok

Issue #2039, opened 2021-03-08
Crosslink bug with database partition!!!


lonelu:
When database is partitioned, the current crosslink search is not able to find Beta peptide in another partition.

lonelu:
Related with #1962

PR #2057, opened 2021-05-11
Crosslink opt

lonelu:
Optimized RAM usage in Crosslink Search.
Optimized several other things, overall the results are better.
#2039


lonelu:
Similar new pull request is created. The branch will be kept for future usage but the pull request will be closed.

PR #2058, opened 2021-05-14
Add Ambiguity Level to Output

zrolfs:
I implemented the five-level classification for proteoform identifications, as described by [Smith et al](https://www.nature.com/articles/s41592-019-0573-x). I wrote this code in such a way that the five-level proteoform classification would be output even when not doing a top-down search. There's still ambiguity when we do bottom-up and I think this classification system could still be useful at the peptide level.


rmillikin:
I'd like to see at least one unit test; I know the method called is tested in mzLib but the PSM/PrSM output from MM should be checked that it contains correct classifications under at least one test case. And PsmFromTsv should be checked also

rmillikin (changes requested):
needs unit test

zrolfs:
unit test added

rmillikin (approved):

rmmiller22 (approved):

PR #2059, opened 2021-05-15
Decoy on the fly And Searching Match Ions with All Charges for decoy library generation 

Yuling1996:
There are 2 main parts which have been changed:
 1st one is decoy-on-the-fly method which generates decoy sequence by reversing target sequence. This is useful for decoy spectral  library generation from target spectral library.
 2nd one is looking for match ions with all charges. The original one finds a fragment with the theoretical mass, and if it accepts it and doesn't try to look for different charge states of that same fragment. This is not good for spectral library generation. 
 
I added a check box in Search task GUI and added the Boolean value "write spectral library" to classic search engine parameters. The new decoy and matched ions searching methods would only work under the condition that you check the write spectral library box.


EngineLayer/DecoyOnTheFly.cs
```diff
@@ -0,0 +1,168 @@
...
+{
+    public class DecoyOnTheFly
+    {
+        //This function maintains the amino acids associated with the protease motif and reverses all other amino acids.
+        //N-terminal modificatons are preserved. Other modifications travel with their respective amino acids. this results
+        //in a decoy peptide composed the same amino acids and modifications as the original. 
+        //Occasionally, this process results in peptide with exactly the same sequence. Therefore, there is a stop-gap measure
+        //the returns the mirror image of the original. N-terminal mods are preserved, but other mods are also reversed. 
+        //this should yield a unique decoy for each target sequence.
+        public static PeptideWithSetModifications GetReverseDecoyFromTarget(PeptideWithSetModifications pwsm, int[] revisedAminoAcidOrder)
```
zrolfs, line 18: I feel like this function should be in mzlib, rather than MetaMorpheus. We currently have all of our digestion/decoy code in mzLib, so there's a precedent.

EngineLayer/DecoyOnTheFly.cs
```diff
@@ -0,0 +1,168 @@
...
+        {
+            Dictionary<int, Modification> newModificationsDictionary = new Dictionary<int, Modification>();
+            //Copy N-terminal modifications from target dictionary to decoy dictionary.
+            if (pwsm.AllModsOneIsNterminus.ContainsKey(1))
+            {
```
zrolfs, line 23: Code coverage is a bit low. Adding unit tests for conditionals like this one would help.

EngineLayer/DecoyOnTheFly.cs
```diff
@@ -0,0 +1,168 @@
...
+                newModificationsDictionary.Add(1, pwsm.AllModsOneIsNterminus[1]);
+            }
+            char[] newBase = new char[pwsm.BaseSequence.Length];
+            ProteomicsExtenstionMethods.Fill(newBase, '0');
```
zrolfs, line 27: I don't think this line is needed. The default char value is '0'

EngineLayer/DecoyOnTheFly.cs
```diff
@@ -0,0 +1,168 @@
...
+            if (motifs != null && motifs.Count > 0)
+            {
+                foreach (var motif in motifs.Where(m => m.InducingCleavage != ""))//check the empty "" for topdown
+                {
+                    string cleavingMotif = motif.InducingCleavage;
+                    List<int> cleavageMotifLocations = new List<int>();
+
+                    for (int i = 0; i < pwsm.BaseSequence.Length; i++)
+                    {
+                        bool Fits;
```
zrolfs, line 39: keep local variables lowercase

EngineLayer/DecoyOnTheFly.cs
```diff
@@ -0,0 +1,168 @@
...
+                newModificationsDictionary.Add(1, pwsm.AllModsOneIsNterminus[1]);
+            }
+
+            //First step is to reverse the position of all modifications except the mod on the peptide N-terminus.
+            if (pwsm.AllModsOneIsNterminus.Any())
+            {
+                int newPosition = pwsm.BaseSequence.Length + 1;
+                for (int i = 2; i < pwsm.BaseSequence.Length + 2; i++)
+                {
+                    if (pwsm.AllModsOneIsNterminus.ContainsKey(i))
```
zrolfs, line 132: This is going to be slow, since you're looking through each amino acid position for a key. I'd recommend refactoring using a foreach(var kvp in pwsm.AllModsOneIsNterminus). The newPosition should be pwsm.baseSequence.Length -kvp.Key and the pwsm.AllModsOneIsNterminus[i] should be kvp.Value.

EngineLayer/MetaMorpheusEngine.cs
```diff
@@ -80,6 +80,55 @@ public static double CalculatePeptideScore(MsDataScan thisScan, List<MatchedFrag
             return score;
         }
 
+        public static double CalculateAllChargesPeptideScore(MsDataScan thisScan, List<MatchedFragmentIon> matchedFragmentIons)
```
zrolfs, line 83: This seems like a separate PR, but if we're trying to score all charges, shouldn't that happen in the matched fragment ion method? Those fragment ions with multiple charges could then be scored using the normal scoring method. Guess I'm not sure why this method is needed.

EngineLayer/MetaMorpheusEngine.cs
```diff
@@ -172,6 +221,50 @@ public static List<MatchedFragmentIon> MatchFragmentIons(Ms2ScanWithSpecificMass
             return matchedFragmentIons;
         }
 
+        public static List<MatchedFragmentIon> MatchFragmentIonsOfAllCharges(Ms2ScanWithSpecificMass scan, List<Product> theoreticalProducts, CommonParameters commonParameters)
+        {
+            var matchedFragmentIons = new List<MatchedFragmentIon>();
+            var Ions = new List<string>();
```
zrolfs, line 227: lower case

EngineLayer/MetaMorpheusEngine.cs
```diff
@@ -172,6 +221,50 @@ public static List<MatchedFragmentIon> MatchFragmentIons(Ms2ScanWithSpecificMass
...
+           
+            // if the spectrum has no peaks
+            if (scan.ExperimentalFragments != null && !scan.ExperimentalFragments.Any())
+            {
+                return matchedFragmentIons;
+            }
+
+            // search for ions in the spectrum
+            //foreach (Product product in theoreticalProducts)
```
zrolfs, line 236: I'd use the commented code instead. ("foreach" is better than "for" in this circumstance)

EngineLayer/MetaMorpheusEngine.cs
```diff
@@ -172,6 +221,50 @@ public static List<MatchedFragmentIon> MatchFragmentIons(Ms2ScanWithSpecificMass
...
+
+                var minMass = commonParameters.ProductMassTolerance.GetMinimumValue(product.NeutralMass);
+                var maxMass = commonParameters.ProductMassTolerance.GetMaximumValue(product.NeutralMass);
+                var closestExperimentalMassList = scan.GetClosestExperimentalIsotopicEnvelopeList(minMass, maxMass);
+                if (closestExperimentalMassList != null)
+                {
+                    foreach (var x in closestExperimentalMassList)
+
+                    {
+                        String Ion = product.ProductType.ToString() + product.FragmentNumber + "^" + x.Charge + "-" + product.NeutralLoss;
```
zrolfs, line 254: lower case

zrolfs: This method risks being super slow because of the long string lookups. Particularly in top-down where there are many deconvoluted experimental masses and many high scoring PrSMs. Might want to consider a way to refactor.

EngineLayer/ClassicSearch/ClassicSearchEngine.cs
```diff
@@ -22,11 +22,13 @@ public class ClassicSearchEngine : MetaMorpheusEngine
         private readonly PeptideSpectralMatch[] PeptideSpectralMatches;
         private readonly Ms2ScanWithSpecificMass[] ArrayOfSortedMS2Scans;
         private readonly double[] MyScanPrecursorMasses;
+        public Dictionary<PeptideWithSetModifications, PeptideWithSetModifications> decoyTargetPairs;
+        private readonly bool WriteSpetrcalLibrary;
```
zrolfs, line 26: Spectral

EngineLayer/ClassicSearch/ClassicSearchEngine.cs
```diff
@@ -22,11 +22,13 @@ public class ClassicSearchEngine : MetaMorpheusEngine
...
 
         public ClassicSearchEngine(PeptideSpectralMatch[] globalPsms, Ms2ScanWithSpecificMass[] arrayOfSortedMS2Scans,
             List<Modification> variableModifications, List<Modification> fixedModifications, List<SilacLabel> silacLabels, SilacLabel startLabel, SilacLabel endLabel,
             List<Protein> proteinList, MassDiffAcceptor searchMode, CommonParameters commonParameters, List<(string FileName, CommonParameters Parameters)> fileSpecificParameters,
-            SpectralLibrary spectralLibrary, List<string> nestedIds)
+            SpectralLibrary spectralLibrary, List<string> nestedIds, bool writeSpetrcalLibrary)
```
zrolfs, line 31: spectral

EngineLayer/ClassicSearch/ClassicSearchEngine.cs
```diff
@@ -62,101 +65,245 @@ protected override MetaMorpheusEngineResults RunSpecific()
 
             if (Proteins.Any())
             {
-                int maxThreadsPerFile = CommonParameters.MaxThreadsToUsePerFile;
-                int[] threads = Enumerable.Range(0, maxThreadsPerFile).ToArray();
-                Parallel.ForEach(threads, (i) =>
+                if (WriteSpetrcalLibrary == false)
```
zrolfs, line 68: There are a lot of changes in this file that are making it difficult to follow along. I'll take a look later...

zrolfs (changes requested):

EngineLayer/MetaMorpheusEngine.cs
acesnik: Can also use the $"{string1}{string2}" syntax. https://docs.microsoft.com/en-us/dotnet/csharp/how-to/concatenate-multiple-strings#string-interpolation

Yuling1996: Will a Tuple<Product , int > be better?
Codes are like: 
var ions = new List<Tuple<Product, int>>();
foreach (var x in closestExperimentalMassList)
 {
     Tuple<Product, int> ion = new Tuple<Product, int>(product, x.Charge);
     ....... 
}

EngineLayer/MetaMorpheusEngine.cs
Yuling1996: Normal search only looks for one match ion for one fragment, and if it accepts it  then it doesn't try to look for different charge states of that same fragment. The score will be the number of matched ions and plus some fraction calculated by intensity(matchedFragmentIons[i].Intensity / thisScan.TotalIonCurrent).
Like b1,b2,b3 will have score 3.xxx;
But when generating library, we need look for match ions with all charges. So we will have b1,b2,b3, b1^2, b2^3. If using the normal scoring function, the score will be 5.xxxx, which is not proper in my opinion. I think the score for b1 and b1^2 should also be 1 plus some some fraction calculated by intensity, because they are matching the same fragment ion just with different charges. So b1,b2,b3, b1^2, b2^3 should be also 3.xxx(but a little higher than b1,b2,b3 as the fraction part)  rather than 5.xxx.

I hope this answers your question. And please tell me if any of my thoughts are not proper.

zrolfs: That makes sense. Can you add a header to the method so that it's clear how this method is different?

EngineLayer/Ms2ScanWithSpecificMass.cs
```diff
@@ -114,6 +114,59 @@ public int GetClosestFragmentMass(double mass)
...
+                return index;
+            }
+            index = ~index;
+
+            return index;
+        }
+
+        //look for the closest index which is smaller than acceptable maximum mass
+        //if all the mass in the array are bigger than the maximum mass, then return -1, which means no mass is acceptable
+        public int GetClosestFragmentMassMaximum(double massMax)
```
rmillikin, line 159: this seems like a near-duplicate of `GetClosestFragmentMassMinimum`, and both of these are near-duplicates of `GetClosestFragmentMass`. is it possible to just use the existing method `GetClosestFragmentMass` to get the information you need?

EngineLayer/SpectralLibrarySearch/SpectralLibrarySearchFunction.cs
```diff
@@ -67,13 +67,13 @@ public static double CalculateNormalizedSpectralAngle(List<MatchedFragmentIon> t
                 double mz = scan.MassSpectrum.XArray[closestPeakIndex];
                 double experimentalIntensity = scan.MassSpectrum.YArray[closestPeakIndex];
 
-                // is the mass error acceptable?
-                if (commonParameters.ProductMassTolerance.Within(mz.ToMass(libraryIon.Charge), libraryIon.Mz.ToMass(libraryIon.Charge)))
-                {
-                    var test = new Product(libraryIon.NeutralTheoreticalProduct.ProductType, libraryIon.NeutralTheoreticalProduct.Terminus,
+                var test = new Product(libraryIon.NeutralTheoreticalProduct.ProductType, libraryIon.NeutralTheoreticalProduct.Terminus,
```
rmillikin, line 70: this will create a new Product for every library ion, even if the mass isn't within tolerance. it's more memory-efficient to leave it as the old way

Test/TestData/2021-06-08-19-48-32/Task Settings/Task1-SearchTaskconfig.toml
```diff
@@ -0,0 +1,88 @@
+TaskType = "Search"
```
rmillikin, line 1: i think this file should not be included?

Test/Test.csproj
```diff
@@ -18,7 +18,8 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.ML" Version="1.3.1" />
     <PackageReference Include="Microsoft.ML.FastTree" Version="1.3.1" />
-    <PackageReference Include="mzLib" Version="1.0.494" />
+    <PackageReference Include="Microsoft.NETCore.App" Version="2.2.8" />
```
rmillikin, line 21: pretty sure you can remove this line

Test/SpectralLibraryReaderTest.cs
```diff
@@ -7,6 +7,9 @@
 using TaskLayer;
 using System.Collections.Generic;
 using Proteomics.Fragmentation;
+using Proteomics.ProteolyticDigestion;
```
rmillikin, line 10: these 3 are probably unneeded

Test/TestData/2021-06-08-19-48-32/Task Settings/Task1-SearchTaskconfig.toml
rmillikin: same with the file "Test/TestData/2021-06-08-19-48-32/Task1-SearchTask/model.zip"

CMD/CMD.csproj
```diff
@@ -32,4 +32,8 @@
     <ProjectReference Include="..\TaskLayer\TaskLayer.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
rmillikin, line 35: pretty sure this is not needed

EngineLayer/ClassicSearch/ClassicSearchEngine.cs
```diff
@@ -22,11 +22,13 @@ public class ClassicSearchEngine : MetaMorpheusEngine
         private readonly PeptideSpectralMatch[] PeptideSpectralMatches;
         private readonly Ms2ScanWithSpecificMass[] ArrayOfSortedMS2Scans;
         private readonly double[] MyScanPrecursorMasses;
+        public Dictionary<PeptideWithSetModifications, PeptideWithSetModifications> decoyTargetPairs;
```
rmillikin, line 25: i think this dictionary is not used anywhere?

EngineLayer/ClassicSearch/ClassicSearchEngine.cs
```diff
@@ -203,5 +249,7 @@ private int GetFirstScanWithMassOverOrEqual(double minimum)
             // index of the first element that is larger than value
             return index;
         }
+
```
rmillikin, line 252: remove extra empty lines between brackets, i've noticed it a couple places elsewhere too

rmillikin (changes requested):

rmillikin (approved):

EngineLayer/ClassicSearch/ClassicSearchEngine.cs
```diff
@@ -39,9 +40,16 @@ public class ClassicSearchEngine : MetaMorpheusEngine
...
             SearchMode = searchMode;
             SpectralLibrary = spectralLibrary;
+            WriteSpectralLibrary = writeSpectralLibrary;
+
+            // if we're doing a spectral library search, we can skip the reverse protein decoys generated by metamorpheus.
+            // we will generate reverse peptide decoys w/ in the code just below this (and calculate spectral angles for them later).
+            // we have to generate the reverse peptides instead of the usual reverse proteins because we generate decoy spectral
+            // library spectra from their corresponding paired target peptides
+            Proteins = proteinList.Where(p => !p.IsDecoy || spectralLibrary == null).ToList();
```
zrolfs, line 52: It's a little odd to check if the spectral library is null for every protein.
More efficient code would be:
`Proteins = spectralLibrary == null ? proteinList : proteinList.Where(p=> !p.IsDecoy).ToList();`

EngineLayer/ClassicSearch/ClassicSearchEngine.cs
```diff
@@ -89,22 +100,28 @@ protected override MetaMorpheusEngineResults RunSpecific()
...
+                            {
+                                int[] newAAlocations = new int[peptide.BaseSequence.Length];
+                                reversedOnTheFlyDecoy = peptide.GetReverseDecoyFromTarget(newAAlocations);
+                            }
+
+                            // clear fragments from the last peptide
+                            foreach (var fragmentSet in targetFragmentsForEachDissociationType)
                             {
                                 fragmentSet.Value.Clear();
+                                decoyFragmentsForEachDissociationType[fragmentSet.Key].Clear();
```
zrolfs, line 115: I don't see when this is ever populated.

EngineLayer/ClassicSearch/ClassicSearchEngine.cs
```diff
@@ -66,19 +74,22 @@ protected override MetaMorpheusEngineResults RunSpecific()
                 int[] threads = Enumerable.Range(0, maxThreadsPerFile).ToArray();
                 Parallel.ForEach(threads, (i) =>
                 {
-                    var fragmentsForDissociationTypes = new Dictionary<DissociationType, List<Product>>();
+                    var targetFragmentsForEachDissociationType = new Dictionary<DissociationType, List<Product>>();
+                    var decoyFragmentsForEachDissociationType = new Dictionary<DissociationType, List<Product>>();
```
zrolfs, line 78: This is an unused variable?

EngineLayer/ClassicSearch/ClassicSearchEngine.cs
```diff
@@ -167,6 +184,34 @@ protected override MetaMorpheusEngineResults RunSpecific()
             return new MetaMorpheusEngineResults(this);
         }
 
+        private void AddPeptideCandidateToPsm(ScanWithIndexAndNotchInfo scan, object[] myLocks, double thisScore, PeptideWithSetModifications peptide, List<MatchedFragmentIon> matchedIons)
+        {
+            bool meetsScoreCutoff = thisScore >= CommonParameters.ScoreCutoff;
+
+            // this is thread-safe because even if the score improves from another thread writing to this PSM,
+            // the lock combined with AddOrReplace method will ensure thread safety
+            if (meetsScoreCutoff)
```
zrolfs, line 193: Can we keep this conditional outside of the method?

EngineLayer/ClassicSearch/ClassicSearchEngine.cs
```diff
@@ -116,32 +133,32 @@ protected override MetaMorpheusEngineResults RunSpecific()
...
+                                // calculate the peptide's score
+                                double thisScore = CalculatePeptideScore(scan.TheScan.TheScan, matchedIons, fragmentsCanHaveDifferentCharges: WriteSpectralLibrary);
 
-                                // this is thread-safe because even if the score improves from another thread writing to this PSM,
-                                // the lock combined with AddOrReplace method will ensure thread safety
-                                if (meetsScoreCutoff)
+                                AddPeptideCandidateToPsm(scan, myLocks, thisScore, peptide, matchedIons);
+
+                                if (SpectralLibrary != null)
                                 {
```
zrolfs, line 146: Put this logic in a helper method

EngineLayer/ClassicSearch/ClassicSearchEngine.cs
```diff
@@ -167,6 +184,34 @@ protected override MetaMorpheusEngineResults RunSpecific()
             return new MetaMorpheusEngineResults(this);
         }
 
+        private void AddPeptideCandidateToPsm(ScanWithIndexAndNotchInfo scan, object[] myLocks, double thisScore, PeptideWithSetModifications peptide, List<MatchedFragmentIon> matchedIons)
```
zrolfs, line 187: I'm wary of performance issues here because of the frequency with which this method is called. Can you benchmark the search time needed before and after this PR using the vignette?

EngineLayer/EngineLayer.csproj
```diff
@@ -20,7 +20,7 @@
     <PackageReference Include="Microsoft.ML" Version="1.3.1" />
     <PackageReference Include="Microsoft.ML.FastTree" Version="1.3.1" />
     <PackageReference Include="Microsoft.NETCore.App" Version="2.2.8" />
-    <PackageReference Include="mzLib" Version="1.0.494" />
+    <PackageReference Include="mzLib" Version="1.0.496" />
```
zrolfs, line 23: Please update <3

zrolfs (changes requested):

EngineLayer/ClassicSearch/ClassicSearchEngine.cs
Yuling1996: The decoyFragmentsForEachDissociationType? This is used in later part

EngineLayer/ClassicSearch/ClassicSearchEngine.cs
Yuling1996: We can do it but if keep it outside, then we will have to add one line to calculate the "meetsScoreCutOff" every time we call this method.

EngineLayer/ClassicSearch/ClassicSearchEngine.cs
Yuling1996: It is populated in line 86,92,155.

EngineLayer/EngineLayer.csproj
Yuling1996: Sorry I didn't get it. Could you explain more?

EngineLayer/ClassicSearch/ClassicSearchEngine.cs
Yuling1996: before this PR, the search time is 00:03:32.9719787
after this PR, the search time is 00:03:35.0665066.
The data I used is CAST_frac4&&frac5

zrolfs (approved):

rmmiller22 (approved):

PR #2081, opened 2021-07-07
Internal Fragment Ion Funtionality

zrolfs:
SEARCH TASK:
-Added internal fragments as a user-specified parameter.
-Users can choose the minimum allowed internal fragment length (MM uses "0" to ignore internal fragments).
-The default option is to ignore internal fragments.
-Internal fragments are not used in scoring, but can reduce proteoform ambiguity.
-At least two internal fragments are needed to disambiguate an identification
-Benchmarked on GELFrEE Jurkat dataset. Went from 34876 PrSMs at 1% FDR to 36608 PrSMs after inclusion of internal fragments, because T|D identifications were disambiguated as targets.
-Minor (~15%) increase in search time when looking for internal fragments. Search time went from 30:16 to 34:47

METADRAW:
-Added a sequence coverage map for visualizing terminal and internal fragments.
-Changed y-ions from purple to red. It provides more contrast with the blue b-ions. 
-Purple is now used for internal ions.


zrolfs:
TODO: 
-Unit tests
-MetaDraw spectrum bug

zrolfs:
Benchmarked on an MCF-7 GELFrEE dataset:
PrSMs: 11367 -> 12139
Time: 18:55 -> 21:42

trishorts:
did you figure out fdr? are there any issues?

zrolfs:
> did you figure out fdr? are there any issues?

image

zrolfs:
Internal fragment ions still have a pretty high FDR, which makes them useless for searching. I had tried to reduce FDR by filtering the internal fragments using "evidence" from the terminal ions (e.g. the ends of the internal ion must have been observed by terminal ions). This achieves lots of high quality internal fragments for proteoforms with lots of terminal ions, but doesn't do much for low scoring proteoforms. Low scoring proteoforms only have a few terminal ion match, so there's little evidence for internal ions, and consequently few or no internal ions match. These types of internal ions, although high-quality, are useless. They don't localize PTMs for high-scoring PrSMs, and they don't pull low-scoring PrSMs out of the mud.
If you try to use ppm tolerance as a filter instead of "evidence", you get worse results than if you removed all internal ions.

I took a different approach and searched for all internal fragments post-hoc to disambiguate proteoforms. There are still lots of incorrect internal ion matches, but the effect is more subdued because these ions don't contribute to the score. When requiring 2 or more internal fragments to disambiguate an unlocalized PTM, there's about a 30% chance MetaMorpheus will pick the wrong localization site (as evaluated by bottom-up). That's still a 70% hit-rate, which I think is pretty good.

trishorts:
its possible that you have multiple real localization sites that introduce some confusion. you may not measure them all in bottom up for various reasons.

rmillikin:
crash searching XL data. I was trying to test to see if the parent/child scan view worked given change on line 164 of MetaDrawLogic.xaml.cs. seems the code thinks M ions are internal ions. should also check D (diagnostic) ions

image

rmillikin (changes requested):
fix XL crash, test that parent/child scan works, check that visualizing glyco search still works

zrolfs:
@rmillikin Can you send me vignette files for testing the XL crash and glyco visualization?

rmillikin:
> @rmillikin Can you send me vignette files for testing the XL crash and glyco visualization?

i searched Test\XlTestData\ms2mixed_bsa_xlink.mzML with Test\XlTestData\BSA.fasta and Test\XlTestData\XLSearchTaskconfig_BSA_DSS_23747.toml but for the task i changed the MS2 child scan dissociation to ETD

zrolfs:
> fix XL crash, test that parent/child scan works, check that visualizing glyco search still works

Thanks for catching this @rmillikin . I patched the visualization crash and removed the sequence coverage map for things with child scans, since I'm not sure how to visualize those data. Somebody smarter than me will need to add that functionality if a user requests it.

I can now do both an XL and glyco search as well as visualize the data from both.

rmillikin (approved):

trishorts (approved):

rmmiller22 (approved):

PR #2098, opened 2021-10-31
Fix crash

zrolfs:


Test/CalibrationTests.cs
```diff
@@ -67,6 +67,32 @@ public static void ExperimentalDesignCalibrationTest(string nonCalibratedFile)
...
+            // set up original spectra file (input to calibration)
+            string nonCalibratedFilePath = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\subset20100611_Velos1_TaGe_SA_Hela_3.mzML");
+
+            // protein db
+            string myDatabase = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\FullHuman.fasta");
+
+            // set up original experimental design (input to calibration)
+            SpectraFileInfo fileInfo = new SpectraFileInfo(nonCalibratedFilePath, "condition", 0, 0, 0);
+            var experimentalDesignFilePath = ExperimentalDesign.WriteExperimentalDesignToFile(new List<SpectraFileInfo> { fileInfo });
+
```
acesnik, line 90: Could you add an assertion here that shows there is a zero-intensity peak?

EngineLayer/Calibration/LabeledDataPoint.cs
```diff
@@ -17,7 +17,7 @@ public LabeledDataPoint(double experimentalMz, int scanNumber, double logTotalIo
             this.ScanNumber = scanNumber;
             this.LogTotalIonCurrent = logTotalIonCurrent;
             this.LogInjectionTime = logInjectionTime;
-            this.LogIntensity = logIntensity;
+            this.LogIntensity = logIntensity > 0 ? logIntensity : 0; //peaks with zero intensity can cause downstream crash
```
acesnik, line 20: What is the value of `logIntensity` when the intensity is zero? NaN?

acesnik (changes requested):
This is a clear PR. I would just recommend adding an assertion to clarify that there is in fact a zero-intensity peak.

trishorts (approved):
This seems fine. Can we just eliminate the zero intensity points. I don't think they have any value (meaningless). Are are just going to slow us down. I know that changes the files more. But, honestly don't much care.

EngineLayer/Calibration/LabeledDataPoint.cs
zrolfs: Negative Infinity

zrolfs:
> This seems fine. Can we just eliminate the zero intensity points. I don't think they have any value (meaningless). Are are just going to slow us down. I know that changes the files more. But, honestly don't much care.

I agree, but I think the zero peaks should be filtered out upstream in mzlib, not in MM

Test/CalibrationTests.cs
zrolfs: done

trishorts:
> 
> 
> > This seems fine. Can we just eliminate the zero intensity points. I don't think they have any value (meaningless). Are are just going to slow us down. I know that changes the files more. But, honestly don't much care.
> 
> I agree, but I think the zero peaks should be filtered out upstream in mzlib, not in MM

I'll work on that.

trishorts (approved):

acesnik (approved):

rmmiller22 (approved):

PR #2102, opened 2021-11-11
mzlib update to 504 which removes zero intensity peaks

trishorts:
zero intensity peaks cause calibration task to crash.
I believe newer, fast mass spectrometers are producing files with extraneous zero intensity peaks
Zach fixed calibration in the prior commit to deal with zeros.
I updated mzlib so that when a file is read in any file format (raw, mgf, mzml) either statically or dynamically, zero intensity peaks are eliminated.
With this change, the unit test that zach created to test for how calibration handled zeros became obsolete. So, I eliminated the test. BUT, I retained his code fix, which should just remain as an active buffer. I don't believe that line of code can easily be tested w/o making a public function out of it.


trishorts:
I was able to calibrate an mzML with with known zero peaks without incident

acesnik:
It's also well tested in mzLib, now. Looks good!

acesnik (approved):

lonelu (approved):

rmmiller22 (approved):

PR #2107, opened 2021-12-06
Spectral library parallel

trishorts:


acesnik (approved):
I don't have detailed knowledge of this set of methods. My approval is for the locking and parallelization looking good, and the diff hit testing being about 90% is great to see.

Someone with more detailed knowledge should look more into the spectral library specific elements.

trishorts:
looks like this is already in place

PR #2114, opened 2022-01-19
Added in ability to use TMTpro-18 and corrected chemical formula for hydroxybutyrylation

nbollis:
Added in the ability to use TMT18 and two unit tests to go with it. 
Noticed the chemical formula for hydroxybutylation was incorrect and this request fixes it.


nbollis:
Removed extra label in TMT11 that was incorrect
Reordered the testcases to be in ascending mass order.

kyp4 (approved):
Looks good to me - all unit tests passed and found TMT18 in a test data set

EngineLayer/Mods/Mods.txt
```diff
@@ -155,7 +155,7 @@ TG   K
 PP   Anywhere.
 MT   Common Biological
 DI   HCD:169.110323533
```
acesnik, line 157: Good catch on this. C4H7O2 looks right.

Will this change the `DI` mass? Should that be updated? I'm not sure how that's calculated.

EngineLayer/Mods/tmt.txt
```diff
@@ -1,4 +1,89 @@
...
+#NL   HCD:C1 N1 H15 C{13}7
+DI   HCD:C1 N1 H15 C{13}7
+#NL   HCD:C1 H15 C{13}7 N{15}1
+DI   HCD:C1 H15 C{13}7 N{15}1
+#NL   HCD:N1 H15 C{13}8
+DI   HCD:N1 H15 C{13}8
+#NL   HCD:H15 C{13}8 N{15}1
+DI   HCD:H15 C{13}8 N{15}1
+//
+TMT Multiplex
```
acesnik, line 86: Should this line be deleted? Looks like the title was copy-pasted here

acesnik: I guess this is true below, too, so no worries

acesnik (approved):
This looks good. Nice work getting the mzML in there for testing actual data.

Just had one question below, but not sure it needs addressing.

trishorts (approved):
good work nic

rmmiller22 (approved):

PR #2116, opened 2022-01-26
Counting unique ambiguous sequences with fdr qvalue <= 0.01

nbollis:
The change implemented here was that instead of identical ambiguous psm's will now be counted and added the the PSMCount field. 

To do so without interfering with the rest of the program, a new field for ambiguous full sequence was added to the PeptideSpectralMatch object and initialized in each constructor and when resolving ambiguities. The CountPSM method was then updated to also count these on qualifying psm's.

This project was started when I began learning to program and thus has many commits that are not representative of the final code. The above description is accurate as of 3/9/22


acesnik:
Is this ready for review?

acesnik:
If you're interested in how to clean up the commit history, you can make a new branch the main upstream branch `git checkout -b newPSMCountBug upstream/master', and then squash merge this branch into it, `git merge --squash oldPSMCountBug`. Then, double check that all the changes are the same, `git diff oldPSMCountBug newPSMCountBug` and then you could make a new PR with a clean commit history (or less recommended, but would also work, would be force-pushing it here to erase the previous history).

PR #2117, opened 2022-02-01
top-down biomarker search

trishorts:
first stab at finding additional proteoforms. This PR enables detection of partially truncated proteoforms, both at the full protein level and at the annotated product (signal peptide, chain peptide) level. There is a new GUI functionality in advanced setttings to turn this on. Hidden for probably most users intentionally.


trishorts:
FYI, currently, biomarker search is only available as an option in search task (not in any other task). And, it's an advanced option that is off by default. It's marked as a file loading parameter b/c it happens during protein db load. It looks like this:
image

acesnik (approved):

nbollis (approved):

rmmiller22 (approved):

Issue #1846, opened 2020-02-26
Non-specific search mode not selected when switching protease


zrolfs:
Pop-up recommending non-specific settings says it will switch modes, but doesn't.

trishorts:
solved in PR 2120
https://github.com/smith-chem-wisc/MetaMorpheus/pull/2120

PR #2120, opened 2022-02-09
GUI Auto Settings Bug Fix

nbollis:
This is a GUI specific update. Selecting "non-specific" as the protease in the search task triggers a prompt suggesting default settings that we have found to be a useful start point for a non-specific search. This update changes one of the defaults, the mode of search from "classic" to "non-specific". This should give equivalent results but the search should be faster. The search mode can be found in "Advanced Options/Some Search Parameters/Search Mode".

This update solves the issue raised in #1846 
https://github.com/smith-chem-wisc/MetaMorpheus/issues/1846


trishorts (approved):
this is a good pull request. thanks for attending to it.

avcarr2 (approved):
Looks good. Good catch.

rmmiller22 (approved):

PR #2135, opened 2022-04-08
[WIP] Calibration bug fix

trishorts:
Failure to calibrate one file in a list of files leaves the tolerances of common params adjusted too wide for the subsequent scan. This PR changes the code to use a local copy of common params for each calibrated file that is created anew for each separate file.


trishorts:
Also using this opportunity to substitute explicit variables for var and to update code with discards, which showed up in C# 7
https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/discards

TaskLayer/CalibrationTask/CalibrationTask.cs
```diff
@@ -105,18 +105,18 @@ protected override MyTaskResults RunSpecific(string OutputFolder, List<DbForTask
 
                     if (i == 1) // failed round 1
                     {
-                        CommonParameters.PrecursorMassTolerance = new PpmTolerance(20);
```
acesnik, line 108: Is this the main change?

acesnik (approved):

TaskLayer/CalibrationTask/CalibrationTask.cs
acesnik: I just read the description, and it looks like the answer is yes. :)

trishorts: Yep. When calibration fails the tolerance doesn't get reset for the next file.

nbollis (approved):

PR #2140, opened 2022-05-10
Qvalue and PepQValue Output filtering  and PSM Count Changes

nbollis:
Recounted PSM's after pep calculation disambiguation
Added in ability for users to select if they would like to filter by qValue or pep qValue
Added options within the output settings of search task to select between qValue or pep qValue

### DIFF
--- a/CMD/CMD.csproj
+++ b/CMD/CMD.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
     <ApplicationManifest>app1.manifest</ApplicationManifest>
     <Configurations>Debug;Release</Configurations>
     <DebugType>full</DebugType>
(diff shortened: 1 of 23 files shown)


trishorts (approved):
looks good

avcarr2 (approved):

trishorts:
there is a bug now when adding the gptmd task. I think we need to have the defaults set so this question doesn't pop up.
image

trishorts:
also let's add a default value of 0.01 to q-value but leave it unchecked. 
image

rmmiller22:
I do not believe we should set a default value of 0.01. That would automatically filter all results to 1% FDR, which is not what we do currently. Previously that value was set to 1, so all IDs at any FDR would be provided in the output, and then if the user wanted truncated results they could change it. I personally like that model.

rmmiller22:
As long as setting it to a default of 0.01 in the GUI does not change the back end from not having a filter when unchecked then I am all good.

trishorts:
there is no effect if neither box is checked. it's just that the default value in grey text is already there. I don't want to filter anything automatically.

rmmiller22:
Perfect! I also do not want to filter anything automatically.

PR #2142, opened 2022-05-18
[WIP] single cell MBR with spectral library analysis

trishorts:
Currently, match between runs only looks for a precursor mass spec in MS1 at the predicted time. With this PR, we will look for any MS2 spectra that might have originated from the peptide in question withing a specified time window. If such an MS2 spectrum can be found, we will determine the MetaMorpheus score and also measure the spectral angle against a stored library spectrum of the peptide in question. Library spectra for all peptides with pep Q-value < 0.01 are stored in the first round of search before the quantitative analysis and mbr are performed. So each MBR peak is expected to have a corresponding library spectrum.


acesnik:
🤖 Closing this, though not because the idea went stale.

The branch cannot be brought forward. It predates the move of the projects into the `MetaMorpheus/` subdirectory, so `EngineLayer/`, `TaskLayer/` and `Test/` all sit at the repo root and every one of the fifteen files collides on a merge, including the ~42,500 lines of fixtures. It is 339 commits behind, the PEP change targets `PEPValueAnalysisGeneric.cs` which is now `PEPAnalysisEngine.cs`, and `MiniClassicSearchEngine` is a copy of a `ClassicSearchEngine` that has since been rewritten around `IBioPolymerWithSetMods` and `SpectralMatch` — `PeptideWithSetModifications`, `PeptideSpectralMatch[]` and `b.PrecursorMass` are all gone from it. Re-applying this is a rewrite rather than a rebase, which is the whole reason to close rather than fix.

Worth saying plainly: the idea in the description is still unimplemented. FlashLFQ grew a real MBR subsystem in the meantime — `MbrScorer`, a factory for it, and an MBR-specific PEP model — but its features are entirely MS1-derived, being ppm error, intensity, retention time, scan count and isotopic distribution. `MbrChromatographicPeak.cs:62` still carries "No Ms2 ID means no MS2 ID Retention time" as a given. So looking for an MS2 spectrum at an MBR peak and scoring it against a stored library spectrum is complementary to what was built rather than superseded by it. Tracked as #2786 so it does not disappear with this PR.

`mbr_spectralLibrary` stays on your fork. The fixtures are the part I would not want to lose — `K13_02ng` and `K13_20ng` at 0.2 ng and 2.0 ng in one experimental design, with a human FASTA slice and 261 reference IDs, is a single-cell-scale MBR test case that would be tedious to rebuild.

PR #2146, opened 2022-05-18
Edited PsmFromTsv to ignore enclosing quotes in fragment ion strings

### LINKED ISSUE #2144: MetaDraw can't read .psmtsv's edited in Excel
**Describing the issue**

Tried to open .psmtsv that I sorted in excel in MetaDraw. Excel appends quotation marks to the fragment ions, rendering them unreadable

**Reproducing the error**

Any edit to a .psmtsv in excel reproduces this.

Alexander-Sol:
Closes #2144.


avcarr2 (approved):
Looks good.

nbollis (approved):

trishorts:
Trim has to be applied to every cell that could take a comma

Alexander-Sol:
Changed so that the trim happens to each cell as soon as each line is read in. This should remove quotation marks inadvertently introduced into any cell/column. 

However, this will be problematic if we every introduced a field that's supposed to be enclosed in quotes.

EngineLayer/PsmTsv/PsmFromTsv.cs
```diff
@@ -85,7 +85,7 @@ public class PsmFromTsv
 
         public PsmFromTsv(string line, char[] split, Dictionary<string, int> parsedHeader)
         {
-            var spl = line.Split(split);
+            var spl = line.Split(split).Select(p => p.Trim('\"')).ToArray();
```
avcarr2, line 88: So this is some nightmare level regex stuff, but you should be able to deal with the edge case of a field enclosed by quotes using a conditional regex: https://docs.microsoft.com/en-us/dotnet/standard/base-types/alternation-constructs-in-regular-expressions#conditional-matching-based-on-a-valid-captured-group. Might need to combine the regex with a conditional that is triggered by a pattern match.

Alexander-Sol: Good to know it's possible, but let's hope it never comes to that

Test/PsvTsvTest.cs
```diff
@@ -25,6 +25,14 @@ public static void ReadOGlycoPsms()
             Assert.AreEqual(9, parsedPsms.Count);
         }
 
+        [Test]
+        public static void ReadExcelEditedPsms()
```
trishorts, line 29: you could add a test for one of the fields that has quotes and show that it was parsed correctly

trishorts (commented):
just a little more unit testing and we should be good.

Test/PsvTsvTest.cs
Alexander-Sol: New version of the test ensures that fragment ions (the field most often enclosed in quotes) are parsed correctly

rmmiller22 (approved):

PR #2152, opened 2022-05-31
MetaDraw Updates

nbollis:
Reorganized some of the files to be grouped more intuitively
     Several models were moved out of the GUI project so that they could be properly tested
Export type can be selected
Added a full sequence annotation below the sequence coverage map
Added a color selection options for ions and ptm's in the spectrum, sequence coverage, and new sequence annotation 
Sequence coverage and sequence annotation can be exported as well


acesnik (approved):
Code & tests look good. 👍🏼 

(I didn't manually test the GUI myself.)

avcarr2 (approved):
Tested on some data that I knew had a ton of internal fragment ion and would make the pre-MetaDraw update slow. Worked very well. Much faster and responsive. The little details, like the legend, are very nice. Good update.

rmmiller22 (approved):

PR #2155, opened 2022-06-22
update newtonsoftJson because of security concerns

trishorts:
Update EngineLayer and GUI nuget package Newtonsoft_Json to version 13.0.1, which is the current stable version.

https://github.com/smith-chem-wisc/MetaMorpheus/security/dependabot/1

Newtonsoft.Json prior to version 13.0.1 is vulnerable to Insecure Defaults due to improper handling of StackOverFlow exception (SOE) whenever nested expressions are being processed. Exploiting this vulnerability results in Denial Of Service (DoS), and it is exploitable when an attacker sends 5 requests that cause SOE in time frame of 5 minutes. This vulnerability affects Internet Information Services (IIS) Applications.


nbollis (approved):
10\10 would review again

Yuling1996 (approved):

PR #2162, opened 2022-07-25
Meta draw default settings bug fix and default mod colors

nbollis:
Fixed a bug where meta draw would crash if there was not a default param directory. Added default colors to some mods per Austin's suggestion in my previous pull request

These defaults in windows are added to \user\user\appdata\local\metamorpheus\defaultparameters

can be reset by deleting those files


Alexander-Sol (approved):
Looks good, no longer crashes when saving as defaults.

zdanaceau (commented):
Looks good overall

zdanaceau (approved):
Looks Good

trishorts (approved):

PR #2166, opened 2022-07-29
[WIP] Classic search DOTF implementation

zdanaceau:
Added functionality to classic search to allow for the use of decoy on the fly (DOTF). This should be entirely conservative of previous functionality with the exception of a change to how decoys are generated when a spectral library is in use. I have yet to implement unit testing for this.


EngineLayer/ClassicSearch/ClassicSearchEngine.cs
```diff
@@ -100,12 +103,29 @@ protected override MetaMorpheusEngineResults RunSpecific()
                         // digest each protein into peptides and search for each peptide in all spectra within precursor mass tolerance
                         foreach (PeptideWithSetModifications peptide in Proteins[i].Digest(CommonParameters.DigestionParams, FixedModifications, VariableModifications, SilacLabels, TurnoverLabels))
                         {
-                            PeptideWithSetModifications reversedOnTheFlyDecoy = null;
-
-                            if (SpectralLibrary != null)
+                            PeptideWithSetModifications generatedOnTheFlyDecoy = null;
+                            // Do rev check similarity, do scrambled, check sim, do mirrored
+                            if (SpectralLibrary != null || DecoyOnTheFly == true)
```
zdanaceau, line 108: This change causes decoys to be generated differently when using a spectral library, causing the TestReverseDecoyGenerationDuringSearch unit test to fail.

PR #2169, opened 2022-08-11
MetaDrawPolishing

nbollis:
Additional Features:
Color options for background peaks and internal ions
MetaDraw specific reset defaults button
Spectrum description elements PEP and PEP-Q only show three decimals
Sort list of psms by protein name, species of origin, q-value, pep, and MetaMorpheus Score
Filter psms by ambiguity level

Changes:
Data visualization charts now export to same folder as the rest of the images
Spectrum description 'spectral angle' field now off by default

Bug Fixes:
Resizing after psm or spectra files were reset caused a crash
Sequence annotation and stationary sequence ptm annotation positioning corrected
Export child scan as pdf now includes ptm color legend
Export sequence coverage no longer cuts off the sequence text
Sequence coverage now stays the correct size when the window is resolved
Psmtsv would use the ambiguous base sequence to set the positions of the matched ions instead of only one of the possible sequence

### DIFF
--- a/EngineLayer/PsmTsv/PsmFromTsv.cs
+++ b/EngineLayer/PsmTsv/PsmFromTsv.cs
@@ -125,7 +125,9 @@ namespace EngineLayer
             Score = double.Parse(spl[parsedHeader[PsmTsvHeader.Score]].Trim(), CultureInfo.InvariantCulture);
             DecoyContamTarget = spl[parsedHeader[PsmTsvHeader.DecoyContaminantTarget]].Trim();
             QValue = double.Parse(spl[parsedHeader[PsmTsvHeader.QValue]].Trim(), CultureInfo.InvariantCulture);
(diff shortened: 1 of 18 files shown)


Alexander-Sol (changes requested):
Y ions are displayed incorrectly for ambiguous peptides. Other than that, looks great!

nbollis:
> Y ions are displayed incorrectly for ambiguous peptides. Other than that, looks great!

Cannot seem to reproduce this error, could you attach the data that caused it?

Alexander-Sol:
> Cannot seem to reproduce this error, could you attach the data that caused it?

Files can be found here: https://drive.google.com/drive/folders/1QGfogd_mXcVSCRw-hzl-iwzxoex0SqIU?usp=sharing

The following peptide produces this error: EIVHIQAGQCGNQIGAK.

nbollis:
> > Cannot seem to reproduce this error, could you attach the data that caused it?
> 
> Files can be found here: https://drive.google.com/drive/folders/1QGfogd_mXcVSCRw-hzl-iwzxoex0SqIU?usp=sharing
> 
> The following peptide produces this error: EIVHIQAGQCGNQIGAK.

Bug was in the loading of the psmtsv.  Psmtsv would use the ambiguous base sequence to set the positions of the matched ions instead of only one of the possible sequences. It has been fixed

Alexander-Sol (approved):

avcarr2 (approved):
Tested it thoroughly. Works well

trishorts (approved):

PR #2171, opened 2022-08-19
 Added isotopic envelope information to PSMs for use in PEP calculations

Alexander-Sol:
Appveyor wasn't building, hoping re-opening the PR will help


EngineLayer/PeptideSpectralMatch.cs
```diff
@@ -56,6 +57,8 @@ public PeptideSpectralMatch(PeptideWithSetModifications peptide, int notch, doub
         public Dictionary<string, int> ModsIdentified { get; private set; } // these should never be null under normal circumstances
         public List<double> LocalizedScores { get; internal set; }
         public int ScanNumber { get; }
+        public IsotopicEnvelope  ScanPrecursorEnvelope {get; }
+        public float isotopeKullbackLeibler { get; set; }
```
nbollis, line 61: Maybe throw a summary comment on this field to briefly describe what the Kullback Liebler Divergence is and/or used for as it is not as readily apparent as the names of the other fields here

nbollis (dismissed):

EngineLayer/PeptideSpectralMatch.cs
Alexander-Sol: Added comment explaining KL, stating bound, and direction

PR #2173, opened 2022-08-24
WIP add water and ammonia loss fragment peak annotations and spectral library

trishorts:
water and ammonia loss peaks are prevalent in tandem mass spectra. However, including them during the search phase with primary ions leads to a loss in sensitivity. But, once the peptide ID is determined, we can go back and label fragments emanating from water and ammonia loss because these don't contribute to the peptide match score. These peaks can also be added to the library spectrum, which should improve the sensitivity of spectral similarity calculations.


nbollis (approved):

YulingDai (approved):

trishorts:
generically adding water and ammonia loss peaks to the spectrum library caused the best matching spectra to shift in the wrong direction. any benefit of including these peaks is in the noise. We could investigate this again using water and ammonia loss on specific ions. Tabling this for now.

PR #2178, opened 2022-09-09
Fragment coverage of proteins

weaversd:
Adds a column to AllProtein.tsv or AllQuantifiedProteins.tsv called Fragment Sequence Coverage. This column contains the amino acid sequence of the protein, where the one-letter code is capitalized if the amino acid is covered on both the c and the n terminal sides by some combination of b and y ions. 

- For each PSM, pulls out all n and c terminal ions.
- For each amino acid position within the PSM, checks if there are c and/or n ions covering each side of the amino acids. If there are, stores the amino acid position in a hash set, which is within the PeptideSpectralMatch class.
- For each protein, converts the covered positions within the PSM to the equivalent amino acid positions within the protein sequence, and stores these as a hashset
- Takes a lowercase string of the protein sequence, and converts AA positions to capital if they are found within the covered hashset for that protein.
- Print out this information in the appropriate tsv file.


EngineLayer/PeptideSpectralMatch.cs
```diff
@@ -393,5 +402,122 @@ private PeptideSpectralMatch(PeptideSpectralMatch psm, List<(int Notch, PeptideW
...
+                if (yIonPositionsUnique[i + 1] - yIonPositionsUnique[i] == 1)
+                {
+                    covered.Add(peptideLength - yIonPositionsUnique[i]);
+                }
+            }
+
+            // If the final y ion is present, the first amino acid is covered
+            if (yIonPositionsUnique.Count > 0)
+            {
+                if (yIonPositionsUnique[^1] == peptideLengthMinus1)
```
acesnik, line 500: Nice usage of this indexer!

acesnik (approved):
Logical & well commented code. 

I haven't tested the functionality, so hopefully someone does a run to double check things, too.

EngineLayer/PeptideSpectralMatch.cs
```diff
@@ -393,5 +402,122 @@ private PeptideSpectralMatch(PeptideSpectralMatch psm, List<(int Notch, PeptideW
...
+                {
+                    // put each ion position into a list, either b or y
+                    case Proteomics.Fragmentation.ProductType.b:
+                        bIonPositions.Add(ion.NeutralTheoreticalProduct.FragmentNumber);
+                        break;
+                    case Proteomics.Fragmentation.ProductType.y:
+                        yIonPositions.Add(ion.NeutralTheoreticalProduct.FragmentNumber);
+                        break;
+                }
+            }
```
nbollis, line 445: Do we want to account for internal ions as well?

trishorts: that point might be worth a discussion.

avcarr2: I think internal ion fragment coverage should be a separate pull request. Including internal ions in fragment coverage is a broader conversation in the field as a whole, so we should make a determination on that at a later time.

weaversd: My thought is that for the highest confidence of single amino acid coverage we would only want to use terminal b and y fragments. But I would defer to other opinions if people feel strongly.

avcarr2 (approved):
Code looks good. Tests are good. I ran real DDA data from my six protein standard mix and the sequence coverage looks good to me. It was a little non-intuitive at first, so I'd recommend adding something to the wiki to explain how to read the output and give a little example.

nbollis (approved):

EngineLayer/PeptideSpectralMatch.cs
```diff
@@ -268,6 +277,117 @@ public static int GetLongestIonSeriesBidirectional(Dictionary<PeptideWithSetModi
...
         }
 
+        /// <summary>
+        /// Determine the Fragment Coverage the PSM
+        /// Assigns fragment coverage indices for the PSM and the protein based on Amino Acid Position in Matched Ion Fragments
+        /// </summary>
+        public void GetAminoAcidCoverage()
+        {
+            if (this.BaseSequence != null && this.MatchedFragmentIons != null && this.MatchedFragmentIons.Any() &&
+                this.PeptideLength > 0)
```
trishorts, line 287: I think this can be shortened. suggest:
image

weaversd: I took out the check for peptide length, that one isn't necessary. I was still getting errors if I didn't have the MatchedFragmentIons null check, because it apparently can't do .Any() if the value is null. I also inverted it to reduce nesting.

EngineLayer/PeptideSpectralMatch.cs
```diff
@@ -268,6 +277,117 @@ public static int GetLongestIonSeriesBidirectional(Dictionary<PeptideWithSetModi
...
+                        p.NeutralTheoreticalProduct.Terminus == FragmentationTerminus.C)
+                    .Select(j => j.NeutralTheoreticalProduct.AminoAcidPosition).Distinct().ToList();
+
+
+                //Create a hashset to store the covered amino acid positions
+                HashSet<int> fragmentCoveredAminoAcids = new();
+
+
+                //Check N term frags first
+                if (nTermFragmentAAPositions.Count > 0)
```
trishorts: instead of count > 0 use .Any()

EngineLayer/PeptideSpectralMatch.cs
```diff
@@ -268,6 +277,117 @@ public static int GetLongestIonSeriesBidirectional(Dictionary<PeptideWithSetModi
...
+
+                    //Check all amino acids except for the last one in the list
+                    for (int i = 0; i < nTermFragmentAAPositions.Count - 1; i++)
+                    {
+                        //sequential AA, second one is covered
+                        if (nTermFragmentAAPositions[i + 1] - nTermFragmentAAPositions[i] == 1)
+                        {
+                            fragmentCoveredAminoAcids.Add(nTermFragmentAAPositions[i + 1]);
+                        }
+
```
trishorts, line 340: can these be as "elseif" clauses. I mean, if the position is covered, there is no need to see if its covered another way. just do additonal comparisons if not covered.

EngineLayer/PeptideSpectralMatch.cs
```diff
@@ -268,6 +277,117 @@ public static int GetLongestIonSeriesBidirectional(Dictionary<PeptideWithSetModi
...
+                        }
+                    }
+                }
+
+                //store in PSM
+                var fragmentCoveredAminoAcidsList = fragmentCoveredAminoAcids.ToList();
+                fragmentCoveredAminoAcidsList.Sort();
+                this.FragmentCoveragePositionInPSM = fragmentCoveredAminoAcidsList;
+
+                if (this.OneBasedStartResidueInProtein == null) return;
```
trishorts: why is this check here. i'm not sure this can happen. but if it can, you should do it in the very first step above and skip all the computation.

EngineLayer/PeptideSpectralMatch.cs
weaversd: Updated some tests that created PSMs with null matched ion fragments, so now we don't have to check if the matchedFragmentIons is null.

EngineLayer/PeptideSpectralMatch.cs
weaversd: Fixed

EngineLayer/PeptideSpectralMatch.cs
weaversd: These have to be several if statements, because we are actually checking two positions for complementary coverage. Either before or after the fragment currently being evaluated. They could both be covered, just one, or neither.

weaversd:
Now the calculation for fragment coverage is only done when reporting protein groups. This means that the proteins have already been assigned, and the coverage isn't calculated for proteins that don't pass the reporting thresholds.

EngineLayer/PeptideSpectralMatch.cs
```diff
@@ -80,6 +86,10 @@ public PeptideSpectralMatch(PeptideWithSetModifications peptide, int notch, doub
         public bool IsDecoy { get; private set; }
         public bool IsContaminant { get; private set; }
 
+        public List<int> FragmentCoveragePositionInPSM { get; private set; }
```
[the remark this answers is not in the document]
trishorts: also add a comment about the numbering. is the number in the peptide? is it one based? or is it numbered based on the complete protein sequence?

EngineLayer/ProteinParsimony/ProteinGroup.cs
```diff
@@ -349,11 +358,13 @@ public void CalculateSequenceCoverage()
...
+            var proteinsWithUnambigSeqPsmsPSM = new Dictionary<Protein, HashSet<PeptideSpectralMatch>>();
 
             foreach (var protein in Proteins)
             {
                 proteinsWithUnambigSeqPsms.Add(protein, new List<PeptideWithSetModifications>());
                 proteinsWithPsmsWithLocalizedMods.Add(protein, new List<PeptideWithSetModifications>());
+                proteinsWithUnambigSeqPsmsPSM.Add(protein, new HashSet<PeptideSpectralMatch>());
             }
 
             foreach (var psm in AllPsmsBelowOnePercentFDR)
```
trishorts, line 370: let's make sure that this is the field from ProteinGroup and not the complete set of psms from the search

variable should be protein.AllPsmsBelowOnePercentFDR

EngineLayer/PeptideSpectralMatch.cs
```diff
@@ -268,6 +278,117 @@ public static int GetLongestIonSeriesBidirectional(Dictionary<PeptideWithSetModi
...
+                }
+            }
+
+            //store in PSM
+            var fragmentCoveredAminoAcidsList = fragmentCoveredAminoAcids.ToList();
+            fragmentCoveredAminoAcidsList.Sort();
+            this.FragmentCoveragePositionInPSM = fragmentCoveredAminoAcidsList;
+        }
+
+        public void GetAminoAcidCoverageProtein()
```
trishorts, line 376: add a comment that this only concerns the fragment coverage of one peptide in the protein and not the coverage of the entire protein. the method name is a little misleading.

EngineLayer/ProteinParsimony/ProteinGroup.cs
```diff
@@ -379,6 +393,38 @@ public void CalculateSequenceCoverage()
...
 
+            foreach (var protein in ListOfProteinsOrderedByAccession)
+            {
+                HashSet<int> fragmentCoveredResidues = new HashSet<int>();
+
+                foreach (var psm in proteinsWithUnambigSeqPsmsPSM[protein])
+                {
+                    if (psm.FragmentCoveragePositionInProtein != null)
+                    {
+                        foreach (int i in psm.FragmentCoveragePositionInProtein)
```
trishorts, line 404: replace this for loop with 
fragmentCoveredResidues.AddRange(psm.FragmentCovereagePositionInProtein)

Test/SeqCoverageTest.cs
```diff
@@ -1,11 +1,15 @@
...
+using EngineLayer;
 using NUnit.Framework;
 using Proteomics;
 using Proteomics.Fragmentation;
 using Proteomics.ProteolyticDigestion;
 using System.Collections.Generic;
 using System.Linq;
 using MassSpectrometry;
+using FlashLFQ;
+using static iText.IO.Image.Jpeg2000ImageData;
```
trishorts, line 11: what the heck is that? looks like a virus

weaversd: Yikes... no idea but I removed it.

EngineLayer/ProteinParsimony/ProteinGroup.cs
weaversd: Replaced with UnionWith which is the AddRange equivalent for hash sets.

EngineLayer/PeptideSpectralMatch.cs
weaversd: This method is gone now

EngineLayer/ProteinParsimony/ProteinGroup.cs
weaversd: Confirmed that we're looping through the psms that correspond to the protein group. AllPsmsBelowOnePercentFDR

EngineLayer/PeptideSpectralMatch.cs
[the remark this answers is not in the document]
weaversd: Changed name, added comments.

weaversd:
Ran a search with 6 raw files, two sample types. Output looks good. Here is the comparison between 'normal' sequence coverage and fragment sequence coverage.
[FragmentCoverage vs SequenceCoverage.txt](https://github.com/smith-chem-wisc/MetaMorpheus/files/9842327/FragmentCoverage.vs.SequenceCoverage.txt)

PR #2182, opened 2022-09-13
File specific dissociation type

trishorts:
User's requested that we enable the use of file specific dissociation types. This PR accomplishes that task.


nbollis (approved):

Alexander-Sol (commented):
If you set file specific dissociation parameters, does it affect protein parsimony in any way? If so, should we test for that? If not, is that a future goal?

trishorts:
> If you set file specific dissociation parameters, does it affect protein parsimony in any way? If so, should we test for that? If not, is that a future goal?

It inadvertently effects inference. Proteins are inferred from the identified peptides. With this PR, peptides from different ionization sources can be identified in a single search. However, there is no direct effect that is testable (I think).

Alexander-Sol (approved):
Ran it on some data I had lying around, it worked well

weaversd (approved):

PR #2184, opened 2022-09-15
Mass histogram metadraw

weaversd:
- Moved PlotModelStats to GuiFunctions so that the methods in that file could be tested
- added three new plots 
     - Precursor mass histogram
     - Precursor m/z histogram
     - Fragment error (ppm) histogram

- For the final plot (fragment error in ppm), fixed the import of fragment error from psmtsv files. Previously, the fragment error was not imported, and so I added functionality within psmFromTsv to parse and import the Da error from the psmtsv file.
- Back-calculated theoretical fragment mass by subtracting the observed Da error from the observed fragment mass. The existing structure of MatchedFragmentIon then correctly calculates the ppm error for each fragment using the theoretical and observed masses.

- Added relevant tests for each plot type in PlotModelStats

### DIFF
--- a/GuiFunctions/MetaDraw/PlotModelStat.cs
+++ b/GuiFunctions/MetaDraw/PlotModelStat.cs
@@ -28,7 +28,7 @@ namespace GuiFunctions
             "Histogram of Precursor Masses",
             "Histogram of Precursor M/z",
             "Precursor PPM Error vs. RT",
-            "Fragment PPM Error vs. RT", 
+            //"Fragment PPM Error vs. RT",  //Math works, but takes a very long time to load.
             "Histogram of PTM Spectral Counts",
             "Predicted RT vs. Observed RT"
         };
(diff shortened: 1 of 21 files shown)


GuiFunctions/MetaDraw/PlotModelStat.cs
```diff
@@ -143,14 +146,16 @@ private void histogramPlot(int plotType)
...
                     break;
-                case 2: // Histogram of Fragment PPM Errors TODO not shown as an option in gui
-                    xAxisTitle = "Fragment error (ppm)";
-                    binSize = 0.1;
-                    foreach (string key in psmsBySourceFile.Keys)
-                    {
-                        numbersBySourceFile.Add(key, psmsBySourceFile[key].SelectMany(p => p.MatchedIons.Select(v => v.MassErrorPpm)));
-                    }
-                    break;
+                //case 2: // Histogram of Fragment PPM Errors TODO not shown as an option in gui
```
nbollis, line 149: If this is ready to be implemented, just not displayed as an option, would you be able to add it to MetaDraw?

weaversd: Added these two plots as an option. It is now possible to parse fragment ion error from psmtsv

nbollis (approved):
Looks good. Great work

avcarr2 (approved):
Looks good. Built and tested with my six protein standard mix DDA data.

trishorts (approved):
nice addition to the metadraw functionality

PR #2185, opened 2022-09-16
Fix mzID SILAC issue

zrolfs:
among others...


trishorts:
This PR appears to have solved a bug reported by a user where individual .mzid file results were not included in the output because of filename changes with silac labeling.

nbollis (approved):

acesnik (approved):

trishorts (approved):
This pr does generated the individual .mzid files. However, they cannot be read by skyline. I don't know if that is an us or them problem. The individual concerned is pushing the skyline people first. We'll see. Anyway, we're good to go for now.

PR #2187, opened 2022-09-21
option to use PEP as peptide filter for protein inference

trishorts:

Currently, we filter the peptides used for protein inference with qvalue < 0.01 and qvalue notch <0.01. This PR gives the user the option to use PEP instead, which is calculated in part using both qvalue and qvalue notch amongst several other parameters. We typically see more peptides at PEP qvalue < 0.01 than we see at qvalue <0.01. Thus we can expect to see an increase in the number of identified proteins at 1% FDR. The PEP filter is as follows: peptides are filtered for PEP <0.5 and PEP qvalue < 0.01. Methods for computing protein score and protein FDR are unchanged in this PR.


Test/ProteinParsimonyTest.cs
```diff
@@ -0,0 +1,157 @@
...
+using TaskLayer;
+using IO.MzML;
+using MassSpectrometry;
+using Org.BouncyCastle.Asn1.X509;
+using System.Text.RegularExpressions;
+
+namespace Test
+{
+    [TestFixture]
+    public class ProteinParsimonyTest
```
avcarr2, line 16: Tests take 25 seconds to run due to IO code that is duplicated and re-run for every test. Need to break out code shared between runs into [OneTimeSetup ](https://docs.nunit.org/articles/nunit/writing-tests/attributes/onetimesetup.html)and [OneTimeTearDown](https://docs.nunit.org/articles/nunit/writing-tests/attributes/onetimeteardown.html) methods to reduce the time it takes to run these tests.

avcarr2 (changes requested):
Everything looks good except for the amount of time it takes to run the tests. Duplicated code that is re-used between all tests needs to be broken out into OneTimeSetup and OneTimeTearDown methods.

nbollis (approved):

PR #2190, opened 2022-10-10
Hydrophobicity Histogram / Histogram Colors

weaversd:
- Added Histogram to display hydrophobicity scores for each psm
- Changed the default colors used for stacked histograms in oxyplot, so they are not all greens/blues. Makes it easier to see the differences between the samples. If there are suggestions for different colors I'm definitely open to discussion. Currently it is just a good variety, first with more primary colors and then starting to get into more nuanced ones.

Currently the hydrophobicity is calculated by the SSRCalc that was already built in to 'Proteomics' in MZLib. When DeepLC is implemented, it will be possible to use that to calculate predicted hydrophobicity or an equivalent.


nbollis (approved):
Looks good, tested on my machine and was able to export this image
image

trishorts (approved):

YulingDai (approved):

PR #2191, opened 2022-10-18
adjusted file specific parameters window

trishorts:
Adding the option for file specific dissociation type to file specific settings window caused the window to be undersized on some monitors. This PR adjusts the default window size and adds the ability to scroll the buttons if the user makes the window vertical size smaller than the height needed to accommodate all the buttons. Also, set button is pinned to the bottom and is always visible regardless of overall window height. Thanks Nic.


nbollis (approved):
10/10 would review again

YulingDai (approved):

weaversd (approved):

PR #2195, opened 2022-10-26
Visualization options

weaversd:
Added options to metaDraw visualization, as well as additional CZE plots

- New plot: Histogram of predicted precursor electrophoretic mobility
- New plot: Predicted Electrophoretic mobility vs observed migration time
- New option: Option to display the y-axis on a log scale instead of a linear scale.
- New option: Option to display the histogram bars next to eachother for different samples, as opposed to stacked. This makes sense especially for the log scale option.


weaversd:
Filtering by Sequence/PTM as an option:
fourPanelPlot mobility vs MT

PR #2197, opened 2022-11-03
update to peptide and psm count in results files

trishorts:
Currently, there is an error in the peptide count for the AllResults.tsv and Results.tsv because the filter was not set properly.
This PR fixes the bug.


Test/SearchEngineTests.cs
```diff
@@ -140,7 +140,7 @@ public static void TestClassicSearchXcorrWithToml()
 
             List<PsmFromTsv> parsedPsms = PsmTsvReader.ReadTsv(psmFile, out var warnings);
 
-            Assert.AreEqual(127, parsedPsms.Count);
+            Assert.AreEqual(37, parsedPsms.Count);
             Assert.AreEqual(0, warnings.Count);
         }
```
nbollis, line 145: Should you load back in the results.txt to check the lines where it states how many 1%  PSM's there are to test if it is the correct value?

avcarr2 (approved):

nbollis (approved):

acesnik (approved):

PR #2199, opened 2022-11-10
Updated Nuget packages that were updated in MzLib

nbollis:
nett is now version 0.15.0
Mathnet.Numerics is now version 5.0.0

One class, DataPointAcquisitionResults, had issues with the new version of Mathnet.Numerics and was changed to maintain the same behavior. The previous version of the code called the method MeanStandardDeviation whose definition in the Mathnet.Numerics code is below

image

The code that has been changed, simply implements this in our class. 

image


trishorts (approved):

Alexander-Sol (approved):
Good catch on the standard deviation stuff

YulingDai (approved):

PR #2204, opened 2022-11-17
TMT18 test rework and test file downsizing

nbollis:
TMT 18 test had a very large testing file. 

Significantly reduced the size of the testing file and simplified the test itself


trishorts (approved):

acesnik (approved):

Alexander-Sol (approved):
-330,000 lines of code. We should all strive for this level of efficiency

PR #2208, opened 2022-11-29
Creating averaging task for MM

nbollis:
Implementing spectral averaging as a task in MetaMorpheus

image

Averaging options will be set just like other tasks within MetaMorpheus and defaults are able to be saved

image

It operates on the same principles as the calibration task, where it will create a new mzML from the input file and use that for downstream tasks. 

It has been tested on-line by performing calibration -> averaging -> gptmd -> search
Results were identical to me calibrating, averaging that calibrated file off-line, then performing a GPTMD and search task. 

Output files for the above example appear as follows
image

Task Settings
image

Averaging Task Folder
image

results.txt
image

AutoGeneratedManuscriptProse.txt
Citation will be changed once averaging paper goes out
image


TaskLayer/TaskLayer.csproj
```diff
@@ -20,7 +20,7 @@
     <PackageReference Include="Microsoft.ML" Version="1.3.1" />
     <PackageReference Include="Microsoft.ML.FastTree" Version="1.3.1" />
     <PackageReference Include="Microsoft.NETCore.App" Version="2.2.8" />
-    <PackageReference Include="mzLib" Version="1.0.537" />
+    <PackageReference Include="mzLib" Version="5.0.542" />
```
avcarr2, line 23: This is still using your local version of mzLib.

avcarr2 (changes requested):
I've used this extensively in support of the spectral averaging paper. However, it's still dependent on a local version of mzLib, so that needs to be fixed before we continue.

PR #2212, opened 2023-01-08
Fix SILAC turnover crash

zrolfs:
Fixed bug that caused turnover searches to crash when data not in root directory. Bug introduced in #2185


trishorts (approved):

nbollis (approved):

Alexander-Sol (approved):
lgtm

PR #2226, opened 2023-01-23
[WIP] LFQ functionality for the glyco search

trishorts:
Users have requested LFQ functionality for the O-glyco search. I checked and FlashLFQ can read and quantify the output from an o-glyco search. Therefore, I believe that we can simply run FlashLFQ inside MetaMorpheus.


trishorts:
destroyed this when i pushed a big files.

PR #2241, opened 2023-02-10
I think there are configuration problems with the release so deleting ARM64 and x86

trishorts:


trishorts:
alex took care of this in his pr

Issue #2240, opened 2023-02-09
Cannot add Decimal Point in Product Mass Tolerance Box


conorjenkins:
**Describing the issue**
Attempting to add Product mass tolerance of 0.6 Da and the decimal point is not being accepted into the field

**Screenshots**
![image](https://user-images.githubusercontent.com/85502162/217726358-bdc7426a-bea8-44da-82f2-692cce38a83e.png).

trishorts:
Hi Conor. How lame. Who programmed this junk anyway....

This bug is confirmed. I'll get on it. Hope you are doing well.

trishorts:
please follow https://github.com/smith-chem-wisc/MetaMorpheus/pull/2248 for updates

conorjenkins:
Thanks! Love the update btw and maybe you should consider it a feature to force people to use ppm for their tolerances. Hope all is well with you and the team as well.

trishorts:
this is fixed. available in the next releae.

PR #2248, opened 2023-02-13
Allow decimal numbers for product mass tolerance in all tasks

trishorts:
user reported being unable to input 0.5 Da for product mass tolerance https://github.com/smith-chem-wisc/MetaMorpheus/issues/2240

Only integer input was allowed. 

This PR enables input of decimal values.

Addition of two utility methods to verify decimal and integer were also added to GUI.Utils

### DIFF
--- a/GUI/TaskWindows/CalibrateTaskWindow.xaml
+++ b/GUI/TaskWindows/CalibrateTaskWindow.xaml
@@ -71,7 +71,7 @@
                                 <StackPanel Grid.Column="1">
                                     <StackPanel Orientation="Horizontal" Margin="5">
                                         <Label x:Name="productMassToleranceLabel" Content="Product Mass Tolerance" />
-                                        <TextBox x:Name="ProductMassToleranceTextBox" HorizontalAlignment="Left" TextWrapping="Wrap" PreviewTextInput="CheckIfNumber" Width="45" ToolTipService.ShowDuration="999999" ToolTipService.InitialShowDelay="500">
+                                        <local:DoubleTexBoxControl x:Name="ProductMassToleranceTextBox" HorizontalAlignment="Left" TextWrapping="Wrap" Width="45" ToolTipService.ShowDuration="999999" ToolTipService.InitialShowDelay="500">
                                             <TextBox.ToolTip>
                                                 <TextBlock>
                                                     Accepted tolerance between experimental and theoretical MS2 peaks.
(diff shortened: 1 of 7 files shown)


elaboy (approved):

GUI/TaskWindows/CalibrateTaskWindow.xaml
```diff
@@ -56,30 +56,30 @@
                                 <StackPanel>
                                     <StackPanel Orientation="Horizontal" Margin="5">
                                         <Label x:Name="precursorMassToleranceLabel" Content="Precursor Mass Tolerance" />
-                                        <TextBox x:Name="PrecursorMassToleranceTextBox" HorizontalAlignment="Left" TextWrapping="Wrap" Width="45" PreviewTextInput="CheckIfNumber" ToolTipService.ShowDuration="999999" ToolTipService.InitialShowDelay="500">
+                                        <local:DoubleTexBoxControl x:Name="PrecursorMassToleranceTextBox" HorizontalAlignment="Left" TextWrapping="Wrap" Width="45" ToolTipService.ShowDuration="999999" ToolTipService.InitialShowDelay="500">
```
acesnik, line 59: Typo in class name. Should be `DoubleTextBoxControl`

GUI/Util/IntegerTextBoxControl.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace MetaMorpheusGUI
+{
+    /// <summary>
+    /// This text box requires input text to be integer only.
+    /// </summary>
+    public class IntegerTexBoxControl : TextBox
+    {
+        protected override void OnPreviewTextInput(TextCompositionEventArgs e)
+        {
+            foreach (var character in e.Text)
```
acesnik, line 13: Why not use `int.TryParse()`?

GUI/Util/DoubleTexBoxControl.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace MetaMorpheusGUI
+{
+    /// <summary>
+    /// This text box requires input text to be decimal only.
+    /// </summary>
+    public class DoubleTexBoxControl : TextBox
+    {
+        protected override void OnPreviewTextInput(TextCompositionEventArgs e)
+        {
+            foreach (var character in e.Text)
```
acesnik, line 13: why not use `double.TryParse()`?

acesnik (changes requested):

GUI/Util/DoubleTexBoxControl.cs
Alexander-Sol: With this approach, it prevents invalid input in real time. You can't type any invalid characters into the box. With the TryParse approach, you could type anything in, but it would only be rejected when you clicked away from the box/when focus is lost.

acesnik: got it!

acesnik (approved):

Alexander-Sol (approved):

PR #2252, opened 2023-02-16
Glyco write individual file results

trishorts:
Last step before I add label free quant to the glyco search task. This PR just splits the results into individual file folders for those users that want such a thing. It is similar to the individual file folder results of search task.


acesnik (approved):

acesnik:
The code changes look good. I haven't tested the GUI or functionality.

Issue #2253, opened 2023-02-19
Version 1.0.1 doesn't install runtimes on upgrade


CEBoos:
**Describing the issue**

MetaMorpheus v1.0.1 installer doesn't install the win-x64 runtime dlls when upgrading from an older version.

**Reproducing the error**

Upgrading from MM v1.0.0 to 1.0.1 on a computer without the  win-x64 runtime dlls should replicate

- Protein database

N/A
- Mass spectrometry file types

N/A
- Settings

N/A
- Example datafiles

N/A

**Expected behavior for GUI**

N/A

**Screenshots**

N/A
**Desktop information**

OS: Windows 10 Enterprise; 19044.2604
Intel(R) Xeon(R) CPU E5-1650 v4 @ 3.60GHz   3.60 GHz, 64-bit, x64 processor
MM v1.0.1

**Additional context**

Uninstalling then re-installing fixes the issue

PR #2254, opened 2023-02-20
Changed Target Architecture to x64

Alexander-Sol:
mzLib only targets x64 processor architecture. Targeting "Any CPU" introduced warning messages when building. 

This PR switches MetaMorpheus to only target x64 architecture. Warnings are no longer generated on build.

This may fix issue #2253, but I wouldn't hold my breath.


MetaMorpheusSetup/Product.wxs
```diff
@@ -1,698 +1,823 @@
...
-  <!--MetaMorpheus Program Menu Folder and Registry Key-->
-  <Fragment>
-    <Component Id="MetaMorpheusProgramMenuFolderComponent" Directory="MetaMorpheusProgramMenuFolder" Guid="c68bc3b0-cc4f-48ed-9ea0-25dbd7492292">
-      <RegistryValue Root="HKCU" Key="Software\SmithGroup\MetaMorpheus" Name="installed" Type="integer" Value="1" KeyPath="yes" />
-      <Shortcut Id="AppStartMenuSC" Name="MetaMorpheus" Description="Shortcut to App" Target="[#MetaMorpheusGUI.exe]" WorkingDirectory="MetaMorpheusInstallFolder" />
-      <Shortcut Id="UninstallProduct" Name="Uninstall MetaMorpheus" Description="Uninstalls My Application" Target="[System64Folder]msiexec.exe" Arguments="/x [ProductCode]" />
-      <RemoveFolder Id="MetaMorpheusProgramMenuFolder" On="both" />
-    </Component>
-  </Fragment>
+	<!--Shorthand way of getting the install directory-->
```
acesnik, line 2: Are these changes intentional?

TaskLayer/MbrAnalysis/MbrAnalysisResults.cs
```diff
@@ -15,12 +15,13 @@ public class MbrAnalysisResults
         public readonly ConcurrentDictionary<ChromatographicPeak, MbrSpectralMatch> BestMbrMatches;
         public readonly FlashLfqResults FlashLfqResults;
         private Dictionary<string, List<string>> PeptideScoreDict;
+        public bool MaxQuantAnalysis { get; }
```
acesnik, line 18: Are these changes intentional in this file, too?

acesnik (changes requested):
Looks good. I think keeping it to x64 is a more consistent with the build, as you mentioned.

I think there are maybe some unintentional changes to look into before merging.

acesnik:
🤖 I'm closing this PR, as the architecture question it raised has since been resolved in the other direction.

The reasoning was sound at the time: mzLib was x64-only, so aligning MetaMorpheus with it made sense. What has changed is that mzLib's x64 targeting was itself the problem. `PlatformTarget=x64` stamps AMD64 into the assembly's PE header, and CoreCLR on Apple Silicon refuses to load those assemblies — the root cause of #2503 and #2504. The two repositories are aligned now, but architecture-neutral rather than x64: smith-chem-wisc/mzLib#1127, released in 1.0.584, and #2701. MetaMorpheus therefore runs natively on Apple Silicon, `osx-arm64` conda-forge builds are now feasible, and #2709 can pursue the GUI on macOS. GUI, GuiFunctions and Test remain x64 as this PR intended, since they are Windows-only and feed the installer.

The review comments above were never addressed and the branch now conflicts with master, so there is no subset left to rebase. #2253 is a separate bootstrapper issue and remains open. Thanks @Alexander-Sol — the instinct that the two repositories needed to agree on architecture was correct.

PR #2258, opened 2023-02-27
[WIP] Unit Tests for LFQ in MetaMorpheus

trishorts:


acesnik:
:robot: Went digging through the old LFQ PRs and wanted to leave what I found, since it's easier than re-deriving it later.

The `PostSearchAnalysisTask` change here — swapping the PEP-vs-QValue branch for a flat `QValue`/`QValueNotch` filter on the PSMs going into FlashLFQ — looks superseded by #2387, which landed `EngineLayer/SpectrumMatch/FilteredPsms.cs` in August 2024. That gave the filter an explicit `FilterType` and threshold rather than hardcoding one choice, and it's now used at about eight sites in `PostSearchAnalysisTask`. Same concern, solved more generally.

There's a nice coda to that: the same refactor builds `peptideSequencesForQuantification` and hands it to FlashLFQ as `peptideSequencesToQuantify`, which is the `PeptideModifiedSequencesToQuantify` allow-list your mzLib#1151 is fixing right now.

On the `Test_LFQ` addition in #2258 — I don't think the central assertion can hold. It checks that summed peptide intensities equal summed protein intensities, but `FlashLfqEngine` computes protein quant via `CalculateProteinResultsMedianPolish`, which is a log-space median polish rather than a rollup sum (and `CalculateProteinResultsTop3` only sums the top three peptides). So the totals aren't expected to agree, which may be what your "some peptides have UNDEFINED in the protein group / this is problematic" note was circling.

Given that, I'd suggest closing both — the branches stay on your fork either way. Very happy to be wrong if there's something in here you were still planning to come back to.

acesnik (commented):
🤖 Went digging through the old LFQ PRs and wanted to leave what I found, since it's easier than re-deriving it later.

The `PostSearchAnalysisTask` change here — swapping the PEP-vs-QValue branch for a flat `QValue`/`QValueNotch` filter on the PSMs going into FlashLFQ — looks superseded by #2387, which landed `EngineLayer/SpectrumMatch/FilteredPsms.cs` in August 2024. That gave the filter an explicit `FilterType` and threshold rather than hardcoding one choice, and it is now used throughout `PostSearchAnalysisTask`. Same concern, solved more generally.

There's a nice coda to that: the same refactor builds `peptideSequencesForQuantification` and hands it to FlashLFQ as `peptideSequencesToQuantify`, which is the `PeptideModifiedSequencesToQuantify` allow-list your mzLib#1151 fixed — that one merged on 2026-08-24, so this part is settled.

On the `Test_LFQ` addition in #2258 — I don't think the central assertion can hold. It checks that summed peptide intensities equal summed protein intensities, but `FlashLfqEngine` computes protein quant via `CalculateProteinResultsMedianPolish`, which is a log-space median polish rather than a rollup sum (and `CalculateProteinResultsTop3` only sums the top three peptides). So the totals aren't expected to agree, which may be what your "some peptides have UNDEFINED in the protein group / this is problematic" note was circling.

Given that, I'd suggest closing both — the branches stay on your fork either way. Very happy to be wrong if there's something in here you were still planning to come back to.

acesnik:
🤖 Closing on the review above — reopen if there was something here you were still coming back to.

The `PostSearchAnalysisTask` edit can no longer be applied at all: it rewrites a `List<PeptideSpectralMatch> unambiguousPsmsBelowOnePercentFdr` block that is gone from master, down to the type name in its declaration. `FilteredPsms` (#2387, `EngineLayer/SpectrumMatch/FilteredPsms.cs`) replaced that PEP-vs-QValue branch with an explicit `FilterType` and threshold and is used twenty times in the file today, so the concern behind the edit got settled more generally than the edit would have settled it.

That leaves `Test_LFQ`, whose central assertion cannot pass for a reason independent of any of the above. `FlashLFQResults.CalculateProteinResultsMedianPolish` (`FlashLFQResults.cs:401`) fills a matrix with log2-transformed peptide intensities and median-polishes it, admitting only peptides that clear `UnambiguousPeptideQuant()`. Protein intensity is therefore neither a sum of peptide intensities nor computed over the same peptide set the test sums across, so `Is.EqualTo(proteinFileOneIntensitySum).Within(1.0)` is not a tolerance away from holding. I suspect that is what the `UNDEFINED in the protein group` note in the test was circling.

The coverage it was reaching for mostly exists too — `PostSearchAnalysisTaskTests.cs` and `SearchTaskTest.cs` both exercise `DoLabelFreeQuantification` and twelve test files assert on the quantified output tables, so `QuantificationTest.cs` holding only three tests makes the gap look larger than it is.

What would reopen this is a test pinning LFQ intensities to values computed outside MetaMorpheus, rather than reconciling the peptide and protein tables against each other. Nothing on master does that and it would be worth having. `LFQ_Testing` stays on the fork, and the two mzML fixtures are the expensive part of this PR — they remain available to whoever writes that test.

PR #2161, opened 2022-07-22
Mbr Spectral Library Matching

Alexander-Sol:
This PR is part of the larger single cell project. 

Added three classes: MbrAnalysisRunner, MbrSpectralMatch, and MiniClassicSearchEngine.

MbrAnalysisRunner is a static class that runs through every MBR Peak in the FlashLFQ results, then goes back to the original file and grabs all MS2 scans in a two minute window centered on the retention time of that peak. 

MiniClassicSearchEngine takes that array of MS2 scans, finds all the scans with a plausible precursor match, and scores them against a spectral library, saving scan with the highest spectral contrast angle. 

MbrAnalysisRunner saves the results of the MiniClassicSearch as a MbrSpectralMatch object. After all mbr peaks have been analyzed, FDR+PEP analysis takes place, and the results are written to a .psmtsv.

Issue #2218, opened 2023-01-17
UNDEFINED value in Protein Groups field in the AllQuantifiedPeptides file


MKatWIS:
I ran a non-specific search on peptidomics data (2 raw files, only the search task used, human swissprot xml for a DB) and in 3 of the identified and quantified peptides I got "UNDEFINED" in the protein group column. What does that mean? How were these peptides identified if not mapped to any protein in the DB?
![image](https://user-images.githubusercontent.com/52959056/212876143-4555936c-b530-40eb-ba83-bb563a83f8af.png)

acesnik:
🤖 `UNDEFINED` is an internal placeholder MetaMorpheus writes into the `Protein Groups` column of `AllQuantifiedPeptides.tsv` when the quantification step cannot find a protein group for a peptide. It does not mean the peptide matched nothing — those three peptides were identified normally and are counted in your 1% FDR peptide total. Their proteins are already in your output: look the peptides up in `AllPeptides.psmtsv` and read the `Protein Accession` and `Gene Name` columns.

It is a bug, and an ordering one. Protein inference is built only from peptides whose sequence is unambiguous, and it then uses the resulting protein list to resolve *other* peptides that were ambiguous. A peptide resolved at that second step becomes a normal identification, but the protein-group table quantification consults was built at the first step, so it has no entry and the placeholder is printed instead.

Reproduced on one of our own test files with default settings. There the ambiguity was an isoleucine/leucine swap between beta-tubulin paralogues — isobaric, so no spectrum can separate them — and inference resolved it by dropping the paralogue nothing else required. The peptide's group was in `AllQuantifiedProteinGroups.tsv` the whole time.

Fixed in #2781. Note it also changes protein intensities, because the intensity of such a peptide previously went into the discarded `UNDEFINED` group and will now count toward its actual group.

Sorry this sat for so long — and thank you for cross-comparing the two files, which is the only way this is visible.

PR #2259, opened 2023-02-27
[WIP] filter psms for flashlfq only by qvalue and qvalue notch

trishorts:
this code was changed in a prior commit https://github.com/smith-chem-wisc/MetaMorpheus/pull/2161
Files with values for PEP were filtered by PEP. But when assigning them to ProteinGroups, the filter was for qvalue and qvalue notch. Thus there was incosistency that produced UNDEFINED protein groups for some peptides. This messed up protein quantification.

### DIFF
--- a/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
+++ b/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
@@ -254,21 +254,11 @@ namespace TaskLayer
             }
 
             // get PSMs to pass to FlashLFQ
-            List<PeptideSpectralMatch> unambiguousPsmsBelowOnePercentFdr = new();
-            if (Parameters.AllPsms.Count > 100)//PEP is not computed when there are fewer than 100 psms
-            {
-                unambiguousPsmsBelowOnePercentFdr = Parameters.AllPsms.Where(p =>
-                    p.FdrInfo.PEP_QValue <= 0.01
-                    && !p.IsDecoy
-                    && p.FullSequence != null).ToList(); //if ambiguous, there's no full sequence
-            }
-            else
-            {
-                unambiguousPsmsBelowOnePercentFdr = Parameters.AllPsms.Where(p =>
+            List<PeptideSpectralMatch> unambiguousPsmsBelowOnePercentFdr  = Parameters.AllPsms.Where(p =>
                     p.FdrInfo.QValue <= 0.01
+                    && p.FdrInfo.QValueNotch <= 0.01
                     && !p.IsDecoy
                     && p.FullSequence != null).ToList(); //if ambiguous, there's no full sequence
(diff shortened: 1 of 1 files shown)


acesnik:
:robot: Went digging through the old LFQ PRs and wanted to leave what I found, since it's easier than re-deriving it later.

The `PostSearchAnalysisTask` change here — swapping the PEP-vs-QValue branch for a flat `QValue`/`QValueNotch` filter on the PSMs going into FlashLFQ — looks superseded by #2387, which landed `EngineLayer/SpectrumMatch/FilteredPsms.cs` in August 2024. That gave the filter an explicit `FilterType` and threshold rather than hardcoding one choice, and it's now used at about eight sites in `PostSearchAnalysisTask`. Same concern, solved more generally.

There's a nice coda to that: the same refactor builds `peptideSequencesForQuantification` and hands it to FlashLFQ as `peptideSequencesToQuantify`, which is the `PeptideModifiedSequencesToQuantify` allow-list your mzLib#1151 is fixing right now.

On the `Test_LFQ` addition in #2258 — I don't think the central assertion can hold. It checks that summed peptide intensities equal summed protein intensities, but `FlashLfqEngine` computes protein quant via `CalculateProteinResultsMedianPolish`, which is a log-space median polish rather than a rollup sum (and `CalculateProteinResultsTop3` only sums the top three peptides). So the totals aren't expected to agree, which may be what your "some peptides have UNDEFINED in the protein group / this is problematic" note was circling.

Given that, I'd suggest closing both — the branches stay on your fork either way. Very happy to be wrong if there's something in here you were still planning to come back to.

acesnik:
🤖 Closing this as superseded — but you were right about the symptom and right that it corrupted protein quantification, and it took a second cause to explain why it persisted.

The inconsistency you identified is gone. Both filters now resolve through `FilteredPsms`, and its `QValue` case checks `QValue` and `QValueNotch` together (`FilteredPsms.cs:40`), so the PEP-versus-QValue mismatch from #2161 that this PR targeted no longer exists — and the `QValueNotch` condition this PR added is in there too. The diff no longer applies either; it patches the `Parameters.AllPsms` / manual-branching version of `QuantificationAnalysis` that #2387 and #2604 replaced.

But #2218 still reproduces on current master, from a different cause: `ProteinParsimonyEngine` is built from PSMs filtered with `includeAmbiguous: false`, and `TrimProteinMatches` then re-resolves some of those PSMs, so a peptide can become a normal identification *after* the protein-group table quantification consults was built. On a shipped HeLa fixture with default settings, 1 of 23 quantified peptides comes out `UNDEFINED` for that reason, and fixing it moves one group's intensity from 27,774 to 380,500. Written up in #2781.

Reopen if you think the filter change here is still wanted on its own merits.

PR #2260, opened 2023-02-28
Uniprot api update

Jevs242:
I changed the url format in StringBuilder htmlQueryString and increased the client wait time because some xml files were taking too long and that was causing an error.


Alexander-Sol (changes requested):
Looks like this is failing some tests:

> Failed TestGetUniProtHtmlQueryString("UP000000625",True,False,False,True,"https://www.uniprot.org/uniprot/?query=proteome:UP000000625 &format=fasta&reviewed:yes&include:no&compress=yes") [51 ms]
  Error Message:
     Expected string length 110 but was 137. Strings differ at index 8.
  Expected: "https://www.uniprot.org/uniprot/?query=proteome:UP000000625 &..."
  But was:  "https://rest.uniprot.org/uniprotkb/stream?format=fasta&review..."

GUI/Views/DownloadUniProtDatabaseWindow.xaml.cs
```diff
@@ -55,9 +56,11 @@ private void DownloadProteomeWithProgress(string DownloadLink, string TargetPath
             {
                 WindowStartupLocation = WindowStartupLocation.CenterScreen
             };
-            progress.Show();
+            progress.Show();    
             Task.Factory.StartNew(() =>
             {
+                client.Timeout = TimeSpan.FromMinutes(30);
```
trishorts, line 62: this seems long. 30 minutes?

GUI/Views/DownloadUniProtDatabaseWindow.xaml.cs
```diff
@@ -73,7 +76,7 @@ private void DownloadProteomeWithProgress(string DownloadLink, string TargetPath
                     {
                         proteinDatabases.Add(new ProteinDbForDataGrid(TargetPath));
                     }
-                    File.SetCreationTime(TargetPath, DateTime.Now); //this is needed because overwriting an old file will not update the file creation time
+                    System.IO.File.SetCreationTime(TargetPath, DateTime.Now); //this is needed because overwriting an old file will not update the file creation time
```
trishorts, line 79: shouldn't need System.IO since it declared in the top using System.IO;

GUI/Views/DownloadUniProtDatabaseWindow.xaml.cs
```diff
@@ -92,7 +96,7 @@ private void DownloadProteomeButton_Click(object sender, RoutedEventArgs e)
 
         private static bool ProteomeDownloaded(string downloadedProteomeFullPath)
         {
-            return File.Exists(downloadedProteomeFullPath);
+            return System.IO.File.Exists(downloadedProteomeFullPath);
```
trishorts, line 99: same comment

GuiFunctions/Databases/DownloadUniProtDatabaseFunctions.cs
```diff
@@ -1,4 +1,6 @@
-﻿using System.Text;
+﻿using Proteomics.ProteolyticDigestion;
+using System.Text;
+using static Org.BouncyCastle.Bcpg.Attr.ImageAttrib;
```
trishorts, line 3: what is this bouncy castle thing?

trishorts (commented):
please address comments.

I tested this and it works nicely.

Thank you.

GuiFunctions/Databases/DownloadUniProtDatabaseFunctions.cs
Jevs242: It has probably been added automatically by visual studio, I see it is not used, I will delete it.

GUI/Views/DownloadUniProtDatabaseWindow.xaml.cs
```diff
@@ -55,9 +56,11 @@ private void DownloadProteomeWithProgress(string DownloadLink, string TargetPath
             {
                 WindowStartupLocation = WindowStartupLocation.CenterScreen
             };
-            progress.Show();
+            progress.Show();    
             Task.Factory.StartNew(() =>
             {
+                client.Timeout = TimeSpan.FromMinutes(30);
+                AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport", true);
                 HttpResponseMessage urlResponse = Task.Run(() => client.GetAsync(uri)).Result;
```
Alexander-Sol, line 64: I know you didn't write this, but it needs to be changed. If you use an Async method, it should include a call to either Wait or Await

Alexander-Sol: https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html

https://blog.stephencleary.com/2012/02/async-and-await.html

Based on those, I think we should use await so we don't lock up the UI thread.

Alexander-Sol (dismissed):
I think there may be a better way to use the HttpClient than what we implemented in the past. It would be nice to have a progress bar, a way to cancel the download, and a guarantee that the UI doesn't lock up during the download

The top answer in [this stack overflow question](https://stackoverflow.com/questions/20661652/progress-bar-with-httpclient) looks like it could be a better way to do it.

Jevs242:
>

trishorts:
Instead of a progress bar, maybe we can just report total bytes downloaded and update that frequently so the user knows there is some action.

`private void startDownload()
{
    Thread thread = new Thread(() => {
          WebClient client = new WebClient();
          client.DownloadProgressChanged += new DownloadProgressChangedEventHandler(client_DownloadProgressChanged);
          client.DownloadFileCompleted += new AsyncCompletedEventHandler(client_DownloadFileCompleted);
          client.DownloadFileAsync(new Uri("http://joshua-ferrara.com/luahelper/lua.syn"), @"C:\LUAHelper\Syntax Files\lua.syn");
    });
    thread.Start();
}
void client_DownloadProgressChanged(object sender, DownloadProgressChangedEventArgs e)
{
    this.BeginInvoke((MethodInvoker) delegate {
        double bytesIn = double.Parse(e.BytesReceived.ToString());
        double totalBytes = double.Parse(e.TotalBytesToReceive.ToString());
        double percentage = bytesIn / totalBytes * 100;
        label2.Text = "Downloaded " + e.BytesReceived + " of " + e.TotalBytesToReceive;
        progressBar1.Value = int.Parse(Math.Truncate(percentage).ToString());
    });
}
void client_DownloadFileCompleted(object sender, AsyncCompletedEventArgs e)
{
    this.BeginInvoke((MethodInvoker) delegate {
         label2.Text = "Completed";
    }); 
}`

trishorts:
maybe a cancel button too?

PR #2261, opened 2023-02-28
Reverted filtering method for psms passed to FlashLFQ

Alexander-Sol:
Psms were being filtered differently in ProteinScoringAndFdrEngine vs the QuantificationAnalysis method in PostSearchAnalysisTask, causing issues in protein quantification. This was caused by a switch to PEP based filtering in QuantificationAnalysis. This change was reverted, and should fix the issue.


nbollis (approved):

acesnik (approved):
Is PEP filtering implemented elsewhere?

Alexander-Sol:
> Is PEP filtering implemented elsewhere?

Currently, filtering happens in ~12 different places. Sometimes by PEP, sometimes by Q-Value. In the future, this should be refactored so that a common filtering method can be employed. For right now though, I'm only changing enough to stop protein quant from breaking

trishorts (approved):

PR #2274, opened 2023-04-28
Ethcd complementary ions

trishorts:
until now, we have not had support for complementary ion searches in MetaMorpheus for the dissociation type EtHcd. This is because there are two ion to add because there are two ion series b/y and c/z. This PR addresses that problem by adding a list of mass shifts rather than a single double.


nbollis (approved):

Alexander-Sol (approved):
**SPECIAL CASE**

Patch coverage is below threshold, but that's due to extremely minor changes to old code that wasn't covered previously. 

Modern Search and NonSpecific search both have **low** test coverage. That will be addressed at some point, but not today.

PR #2280, opened 2023-05-25
Spectral library refactor

Alexander-Sol:
Previously, spectral library writing and updating ran slowly. The bottleneck was grouping PSMs according to their sequence and charge state, which had an algorithmic complexity of O(n^2).

I refactored this section of the code using LINQ's IGrouping. It now runs in O(n) time. 

I also changed psmFiltering to be consistent between UpdateSpectralLibrary and GenerateSpectralLibrary


Alexander-Sol:
Merged into FilteringRefactor

PR #2281, opened 2023-05-25
Averaging task in MetaMorpheus

nbollis:
Wiki page is available here https://github.com/smith-chem-wisc/MetaMorpheus/wiki/Averaging-Task
The wiki page is also linked on the averaging task window and the main MM about tab
Implementing spectral averaging as a task in MetaMorpheus

image

Averaging options will be set just like other tasks within MetaMorpheus and defaults are able to be saved.
There are three preset averaging methods available just by clicking the buttons in the upper left
image
It operates on the same principles as the calibration task, where it will create a new mzML from the input file and use that for downstream tasks. 

It has been tested on-line by performing calibration -> averaging -> gptmd -> search
Results were identical to me calibrating, averaging that calibrated file off-line, then performing a GPTMD and search task. 

Output files for the above example appear as follows
image

Task Settings
image

Averaging Task Folder
image

results.txt
image


avcarr2 (approved):
Beautiful.

MetaMorpheus/GUI/Util/Converters/AveragingTaskConverters.cs
```diff
@@ -0,0 +1,253 @@
...
+        }
+    }
+
+    public class OutlierRejectionTypeToSigmaReadOnlyConverter : BaseValueConverter<OutlierRejectionTypeToSigmaReadOnlyConverter>
+    {
+        public override object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            if (value.GetType() == typeof(OutlierRejectionType))
+            {
+                OutlierRejectionType val = (OutlierRejectionType)value;
```
Alexander-Sol, line 124: Maybe do a switch statement here, like you did above?

MetaMorpheus/GUI/Util/Converters/AveragingTaskConverters.cs
```diff
@@ -0,0 +1,253 @@
...
+                        return 1;
+
+                    case SpectraFileAveragingType.AverageEverynScansWithOverlap:
+                        return 1;
+
+                    case SpectraFileAveragingType.AverageDdaScans:
+                        return 1;
+
+                    case SpectraFileAveragingType.AverageDdaScansWithOverlap:
+                        return 1;
```
Alexander-Sol, line 105: A different way to do this is just let them all fall through to the last switch case. That way, return 1 would only be written once

Alexander-Sol (approved):
A couple minor comments. Overall it looks great. The preset parameters are a really nice touch

elaboy (approved):

MetaMorpheus/GUI/Util/Converters/AveragingTaskConverters.cs
avcarr2: Even with the newer C# versions ostensibly allowing you to do it, it's still kind of a pain to implement

PR #2282, opened 2023-05-30
Database download Test

elaboy:
Made a test for making sure the Uniprot REST API is working.


MetaMorpheus/Test/GuiFunctionsTest.cs
```diff
@@ -1,35 +1,40 @@
...
-        [TestCase("UP000000625", false, true, true, false, "https://rest.uniprot.org/uniprotkb/stream?format=xml&query=%28%28proteome%3AUP000000625%29%29")]
-        [TestCase("UP000000625", true, false, false, false, "https://rest.uniprot.org/uniprotkb/stream?format=fasta&reviewed=true&query=%28%28proteome%3AUP000000625%29%29")]
-        [TestCase("UP000000625", false, false, true, false, "https://rest.uniprot.org/uniprotkb/stream?format=xml&query=%28%28proteome%3AUP000000625%29%29")]
-        [TestCase("UP000000625", false, false, false, false, "https://rest.uniprot.org/uniprotkb/stream?format=fasta&query=%28%28proteome%3AUP000000625%29%29")]
-        public static void TestGetUniProtHtmlQueryString(string proteomeID, bool reviewed, bool isoforms, bool xmlFormat, bool compressed, string expectedResult)
-        {
-            string queryString = DownloadUniProtDatabaseFunctions.GetUniProtHtmlQueryString(proteomeID, reviewed, isoforms, xmlFormat, compressed);
-            Assert.AreEqual(expectedResult, queryString);
-        }
+        //[Test]
```
avcarr2, line 14: Delete rather than comment out

MetaMorpheus/Test/GuiFunctionsTest.cs
```diff
@@ -48,5 +53,66 @@ public static void TestGetUniprotFilename(string proteomeID, bool reviewed, bool
...
+                File.Delete(filePath);
+
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine(ex);
+                Console.WriteLine(proteomeURL);
+                Assert.Fail();
+            }
+
```
avcarr2, line 113: Need to check that result is correct before asserting pass.

avcarr2 (changes requested):

MetaMorpheus/GuiFunctions/Databases/DownloadUniProtDatabaseFunctions.cs
```diff
@@ -11,32 +12,19 @@ public class DownloadUniProtDatabaseFunctions
...
+            queryArray[0] = compressed ? "compressed=true" : "";
+            queryArray[1] = xmlFormat ? "format=xml" : "format=fasta";
+            queryArray[2] = isoforms && !xmlFormat ? "includeIsoform=true" : "";
+            queryArray[3] = reviewed ? $"query=reviewed:true&proteome:{proteomeID}" : $"query=proteome:{proteomeID}";
 
-            if (reviewed)
-            {
-                htmlQueryString.Append("&reviewed=true");
-            }
+            string[] queryArrayReduced = queryArray.Where(x => x != "").ToArray() ;
```
Alexander-Sol, line 24: You don't need to convert back to an array here, you can leave it as an IEnumerable.

Also, you could set unneeded fields as null instead of "". I think a null check is slightly faster than string equality comparison, but it's a small difference.

MetaMorpheus/Test/GuiFunctionsTest.cs
```diff
@@ -48,5 +31,72 @@ public static void TestGetUniprotFilename(string proteomeID, bool reviewed, bool
...
+        {
+            var proteomeURL = DownloadUniProtDatabaseFunctions.GetUniProtHtmlQueryString(proteomeID, reviewed,
+                isoforms, xmlFormat, compressed);
+
+            var filePath = Path.Combine(TestContext.CurrentContext.TestDirectory, $@"DatabaseTests\{testName}");
+
+            List<Protein> reader = new List<Protein>();
+
+
+            try
```
Alexander-Sol, line 59: I don't think you need a try catch block - any unhandled assumption that's thrown will fail the test

Alexander-Sol (changes requested):

avcarr2 (approved):

nbollis (approved):

Alexander-Sol (approved):

Issue #2289, opened 2023-06-28
XL toml causes crash


Alexander-Sol:
Reading in an old XLSearchTask toml (found in Lei's data, [here](https://drive.google.com/drive/folders/1TrgwX0weDCIgiizsBHW0gdNjOTXYTRwe)) results in MetaMorpheus crashing when you try and open the XL Search Task Window

PR #2290, opened 2023-06-28
Added null checks to enable display of XL results

Alexander-Sol:
MetaDraw was crashing when you attempted to view XL results. I added a couple of null checks that resolved the issue (#2289.)

MetaDraw was and continues to crash when attempting to read in the Deadends.tsv output from XLSearchTask. This is due to dead ends being recorded as an unrecognized modification. Should maybe be addressed at some point, but it's outside the scope of this PR.


Alexander-Sol:
This PR can handle visualization of top-down and conventional bottom-up data w/in MetaDraw.

elaboy (approved):

trishorts (approved):

nbollis (approved):

PR #2295, opened 2023-07-19
Write spectrum library from PsmTsv in MetaDraw

trishorts:
I'd like to be able to create library spectra in MetaDraw. For this to happen, we need to be able to use the info from PsmTsv to generate the spectrum. This PR adds a new method to PsmFromTsv called ToLibrarySpectrum() that produces the library spectrum object. Once you have a library spectrum object, you simply need to call the .ToString() method to get the .msp format.

I copied the pdf/png export code in MetaDraw and added a button in the gui that lets the user export any number of selected PSMs for immediate export into a spectrum library.

The output was tested both in metadraw and in search. The written libraries load into metadraw and produce spectral angles in search.

I did need to update psmtsv read to correctly accomodate child scans (MS3) for both crosslink and glyco. There was previously an error reading the scan intesnities.


Alexander-Sol:
I feel like ToLibrarySpectrum should spit out a LibrarySpectrum object. Then you could call to string or whatever on that. If we ever change how spectral libraries are written, I don't want to have to edit multiple different ToLibrarySpectrum methods

Alexander-Sol (changes requested):
I feel like ToLibrarySpectrum should spit out a LibrarySpectrum object. Then you could call to string or whatever on that. If we ever change how spectral libraries are written, I don't want to have to edit multiple different ToLibrarySpectrum methods

trishorts:
> I feel like ToLibrarySpectrum should spit out a LibrarySpectrum object. Then you could call to string or whatever on that. If we ever change how spectral libraries are written, I don't want to have to edit multiple different ToLibrarySpectrum methods

that makes sense. It is done now.

nbollis (approved):

Alexander-Sol (approved):

trishorts:
i broke this writing glyco spectrum library

nbollis (approved):

PR #2297, opened 2023-08-02
Changed the main window to own the subsequent windows

nbollis:
This should make the task windows (and MetaDraw) pop up in front of the main window


trishorts (approved):
brilliant

Alexander-Sol (approved):
Works as intended when installed

elaboy (approved):

PR #2298, opened 2023-08-02
Add Tailor per-spectrum score calibration as an opt-in PSM column

trishorts:


acesnik:
🤖 Pushed an implementation onto this branch, since the calculation here could not have produced a Tailor score. Current master is merged in and the three conflicting files are resolved in favour of the new code, so the diff is now 9 files, +373/-11. Your nine commits are still in the history — this was a fast-forward, nothing was rewritten.

**The measurement is the main result, and it is negative.** The code is correct and cheap; on real data the calibrated score is a *worse* discriminator than the raw Morpheus score.

### Why the old calculation returned the raw score

`ComputeTailorDenominator` sorted ascending and averaged the **bottom** 1% — Tailor uses the 99th percentile, the other end — then floored it with `Math.Max(1, ...)`. The lowest-scoring candidates for a spectrum match no fragment ions and score ~0, so the floor was reached every time: the denominator was exactly `1`, and the Tailor score was identical to `Score` for every PSM. Simulated at n = 219/345/508/816 candidates (the range actually observed below), the denominator is `1.000` in all four cases. The scores were also collected only from decoys, and there was a hardcoded `myTailor.tsv` write to an absolute `E:` path.

### MetaMorpheus does not give Tailor the candidate set it assumes

Over 66,987 scans and 13,958,782 candidate evaluations, one mouse file at the default ±5 ppm:

| candidates/scan | min | 25th | median | 75th | 90th | 99th | max |
|---|---|---|---|---|---|---|---|
| | 0 | 43 | **219** | 345 | 407 | 508 | **816** |

Comet and Crux, where Tailor was developed, use a wide precursor window and see thousands to tens of thousands. At N = 219 the 99th percentile is the *third-largest* score for that spectrum, so `Score / q99` compresses to a median of 1.95 with a 5th percentile of 0.99, against a raw-score median of 8.16. Only 1.2% of scans reach 500 candidates.

### Calibration loses identifications, in both regimes

Target/decoy competition inside the calibrated subset, evaluated block-wise so tied scores are all-in-or-all-out — the raw score is 85% tied at the wide setting and Tailor 62%, so a prefix scan would flatter the less-tied metric:

| precursor window | column populated | subset | raw Score | Tailor | change |
|---|---|---|---|---|---|
| ±5 ppm (default) | 2.3% | 461 T / 102 D | 308 | 242 | **−66 (−21.4%)** |
| ±100 ppm | 98.1% | 26,528 T / 9,226 D | 13,950 | 12,663 | **−1,287 (−9.2%)** |

Widening to ±100 ppm populates the column for 98.1% of PSMs, so the sparsity really is the window — but it does not rescue the discrimination. Nor is the loss specific to the 99th percentile: dividing by the 50th, 75th, 90th and 95th instead costs 244, 2,073, 241 and 470 targets over all 24,291 PSMs. Every variant tested is worse than the raw score.

I'd read that as the Morpheus score already being better calibrated across spectra than XCorr is, which is what Tailor was built to fix. That is an inference, not something I measured.

### What the code does

`ScoreDistribution` is a fixed-width histogram rather than a `List<double>`. The classic loop is protein-major and parallel, so one scan's scores arrive from every thread across the whole run: a list needs a lock per candidate and grows without bound, while a histogram bin takes an `Interlocked.Increment` and costs a fixed ~1 kB per MS2 scan, allocated only when the option is on. Scores are recorded before the score cutoff, and the on-the-fly library decoys are included so the distribution is not halved in library search.

Two cases refuse to emit a number rather than guess one: a percentile whose rank lands in the clamped top bin, and a spectrum with fewer than 500 candidates — where the divisor approaches the best score and the ratio is pinned near 1, the same degeneracy as before but from the opposite end.

Off by default. With `ComputeTailorScore = false` the column is absent and `AllPSMs.psmtsv` is **byte-identical** to unmodified master on the same search, as are `AllPeptides.psmtsv` and `AllProteinGroups.tsv`. The score is output only — not an FDR or PEP input.

### Not done, and not verified

- **Classic search only.** Modern and non-specific leave the column blank rather than emitting zero.
- The published score adds a constant to numerator and denominator for negative XCorr values. The Morpheus score is non-negative so this is plain `Score / q99`; **I did not check the constant against the paper.**
- One raw file, one organism, one enzyme. The direction held across five percentiles and two windows, but this is not a survey.
- The column name is a local constant in `PsmTsvWriter`, not in mzLib's `SpectrumMatchFromTsvHeader`, to keep this to one repo. It belongs there at the next mzLib release.
- Tests are 11/11 green via a throwaway net10.0 harness; `Test.csproj` is `net*.0-windows`, so CI is the first Windows run.

Two ways to go, and I don't have a stake in the first: merge it as a documented negative so the column is there for anyone wanting to re-test on top-down or a genuinely open search, where the calibration has more room to matter — or close it and keep the measurement. The branch is real either way now.

PR #2304, opened 2023-09-01
Updated to MzLib 1.0.541 and updated one unit test

nbollis:
Updated to mzlib with reduced complexity scan accessing, this will reduce time to run calibration. 

Updated one unit test to no longer run in parallel as collisions were occurring with file accessing. 
Updated the same unit test to reflect a change in the total number of proteins present in the database as changed by uniprot. 

This version of mzlib will also correctly load fasts.gz files


elaboy (approved):

trishorts (approved):
this looks good to me. run time looks good. results look good.

Alexander-Sol (approved):

PR #2310, opened 2023-09-12
Added SpectralAveraging.dll to project.wxs

nbollis:
SpectralAveraging.dll was not being packaged properly with the Wix Installer. This PR adds the lines to the wix install script to package the new dll within mzlib correctly. 

Future work should dig into why this is not happening automatically.


elaboy (approved):

trishorts (approved):

Alexander-Sol (approved):
Downloaded, installed, and tested on vignette data. Averaging works and results in +20 Protein Groups (w/o GPTMD).

PR #2327, opened 2023-12-26
Invert qvalue calculation

trishorts:
We have never used the correct calculation for q-value, which is (decoy count + 1)/(target count). This is because, the highest scoring PSM would have q-value = 0.5. Bill Noble alerted me to an alternate strategy for computing q-value that would eliminate this problem and allow us to use the correct formula. This is accomplished by computing q-value from lowest scoring PSM to highest scoring PSM subject to the following rule. Whenever the current q-value is greater than the previous q-value, we keep the previous q-value. Now, the highest scoring PSM will have a q-value equal to the q-value of the highest scoring decoy PSM. It will never be zero. c est la vie.

overall we expect to see a higher yield of both peptides and PSMs with q-value < 0.01. 

An unfortunate by product of this update is that some unit tests with low psm count fail to have sufficent psms with low q-value. In certain instances, this prevents PEP from being calculated. This problem was solved by adding another mzML that could boost the psm count to a sufficiently high level to allow all unit tests to complete successfully while testing what they were intended to test. Please note, some test assertions had to be changed to accommodate the new results.


elaboy (approved):

nbollis (commented):
How do Target/Decoy curves change based upon this implementation?

Alexander-Sol (approved):

PR #2329, opened 2023-12-28
Andromeda score

trishorts:

Adding a popular peptide spectrum match scoring option.

image


acesnik (commented):
🤖 I worked through the arithmetic on this rather than reading it, because a score is either right or it isn't. The short version: the idea is defensible but the implementation does not compute the Andromeda score, and the error is of a kind that cannot be recalibrated away.

## The binomial coefficient is missing

Andromeda (Cox et al. 2011) is `-10 log10 P` where `P` is the binomial tail — the probability of matching at least *k* of *n* theoretical fragments by chance:

```
P = Σ(j=k..n) C(n,j) · p^j · (1-p)^(n-j)
```

`AndromedaScore` computes the sum without `C(n,j)`:

```csharp
localSum += Math.Pow(p, j) * Math.Pow(1 - p, n - j);
```

So `localSum` is not a probability. It is too small by the omitted coefficients, which for a mid-range *j* is five to sixteen orders of magnitude, and the resulting score is inflated:

```
   n    k    q |  Andromeda   this PR |  P wrong by
  20   10   10 |       51.5     104.1 |     182,299x
  40   20   10 |       97.3     208.6 |  1.4e+11x
  60   30   50 |        2.6     165.7 |  2.1e+16x
```

The reason this matters more than a scale factor: **the inflation is a function of n**, so it is a peptide-length bias, not an offset.

```
  n=10  published  27.9   this PR  51.8   inflation  23.9
  n=30             74.5           156.4              81.9
  n=60            142.5           313.2             170.7
```

Two PSMs of different length are inflated by different amounts, so the *ranking* between them differs from Andromeda's. That is why no post-hoc correction fixes it — the ordering is already wrong.

## p can exceed 1, and the score goes negative

The call site derives the peak density from every peak in the scan:

```csharp
numberOfPeaksPer100Thomsons = scan.TheScan.NumPeaks / ((max - min) / 100.0);
```

For 2,000 peaks over a 300-1800 Th window that is 133.3 peaks per 100 Th, so `p = 1.333`. Then `(1-p)` is negative, `localSum` comes out greater than 1, and the score is **negative** — I get -24.0 for n=20, k=10. A `-10 log10(probability)` score cannot be negative, so any such row in the psmtsv is a tell.

Even below that threshold the density is implausible as a chance-match probability: 500 peaks over 200-2000 Th gives `p = 0.278`, i.e. a 28% chance that any given theoretical fragment matches at random.

That is the third divergence: Andromeda does not use all peaks. It retains the top *q* per 100 Th window and **maximises the score over q**, which is what keeps p small and makes the statistic meaningful. Here `q` is a single derived density, used once.

## Two things that limit the damage, and two that limit the usefulness

Reassuringly, this is an output column only — `AndromedaScore` appears nowhere in `FdrAnalysis/`, so it does not reach FDR or PEP and cannot move an identification. Given what the numbers are, that is a good thing.

But the column is also only populated for Classic search: `ModernSearchEngine` and `NonSpecificEnzymeSearchEngine` both pass a hardcoded `0`, so two of three search modes emit `0.000` for every PSM. And in Classic, the on-the-fly decoy is passed `0` too, so targets and decoys are not comparable — the same trap #2650 handled deliberately by scoring both sources on one scale.

Also `using Newtonsoft.Json.Linq;` is added to `MetaMorpheusEngine.cs` and there are zero JSON types used in the file — an IDE auto-import that should not land in the core engine.

## Is it still potentially useful?

The *purpose* is, if the purpose is cross-tool comparison — being able to put a MetaMorpheus PSM next to a MaxQuant one on a comparable axis is genuinely handy. But this is a rewrite rather than a patch: the coefficient, the peak-depth filtering, the optimisation over *q*, and the per-engine wiring are four separate pieces, and at that point it is worth asking first whether we want to emit another engine's score at all, versus documenting how to map our own.

If it is wanted, the thing that would make it checkable is a handful of PSMs scored by both MaxQuant and this, agreeing to within rounding. That is a much better acceptance test than any amount of review.

## Not verified

Everything above is arithmetic and reading — `Test/Test.csproj` is `net*.0-windows` and the branch is `dirty`, so I ran no MetaMorpheus code. I have not compared against MaxQuant output, and I am taking the published formula from Cox et al. 2011 as I understand it; if the intended target is a later Andromeda revision with a different form, that would change the first section and is worth saying so.

PR #2339, opened 2024-02-09
#2- Deconvolution Generalization

nbollis:
Enables the generalization of deconvolution to other deconvolution types in MetaMorpheus. 
This is necessary for if/when Decharge is added to MzLib. 

Created separate deconvolution parameters for product and precursor scans. 
Removed deocn tolerance and intensity ratio from toml. 
Added negative mode deconvolution tests. 

Updated custom fragment ion window to only display Protein products with ability to do RNA once implemented in MzLib. This eliminates a crash


Alexander-Sol (changes requested):
Fix the two statements on one line and we're good to go

elaboy (approved):

elaboy (approved):

YulingDai (approved):

Alexander-Sol (approved):

PR #2340, opened 2024-02-09
#3 -Spectral Match Generalization

nbollis:
Generalized the PeptideSpectralMatch class to a generic SpectralMatch class. This enables OligoSpectralMatch to use all of the same Writing classes and a vast majority of the FDR calculations. 
Also added some tests to the Bin class for test coverage


trishorts:
wonder about switching from psm to sm for all things. would be a big change.

trishorts (approved):

YulingDai (approved):

Alexander-Sol (approved):

PR #2341, opened 2024-02-09
#1 - Digestion params generalization

nbollis:
Generalizes digestion parameters in common parameters to enable rna digestion and protein digestion. 
Adds rnase.tsv to the installed MetaMorpheus. 

In the task windows, I had to move all things dependent upon the digestion agent being a protease into an if clause that casts the IDigestionParams.DigestionAgent into a protease.


trishorts:
wherein nic embraces the var side

Alexander-Sol (approved):

elaboy (approved):

acesnik (approved):

RayMSMS (approved):

MetaMorpheus/EngineLayer/CrosslinkSearch/CrosslinkSearchEngine.cs
```diff
@@ -320,9 +320,10 @@ private void FindCrosslinkedPeptide(Ms2ScanWithSpecificMass scan, List<(double,
                 possibleMatches = new List<CrosslinkSpectralMatch>();
             }
 
+            var initiatorMethionine = ((DigestionParams)CommonParameters.DigestionParams).InitiatorMethionineBehavior;
             foreach (var id in idsOfPeptidesPossiblyObserved)
```
Alexander-Sol, line 324: Should there be some kind of null check here? What happens if CommonParameters.DigestionParams is an RnaDigestionParams object

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -444,24 +457,47 @@ public static CommonParameters SetAllFileSpecificCommonParams(CommonParameters c
...
-                maxModsForPeptides: maxModsForPeptide,
+            int maxModsForPeptide = fileSpecificParams.MaxModsForPeptide ?? commonParams.DigestionParams.MaxMods;
 
-                //NEED THESE OR THEY'LL BE OVERWRITTEN
-                maxModificationIsoforms: commonParams.DigestionParams.MaxModificationIsoforms,
-                initiatorMethionineBehavior: commonParams.DigestionParams.InitiatorMethionineBehavior,
-                fragmentationTerminus: commonParams.DigestionParams.FragmentationTerminus,
-                searchModeType: commonParams.DigestionParams.SearchModeType
+            IDigestionParams fileSpecificDigestionParams;
+            if (commonParams.DigestionParams is DigestionParams digestionParams)
```
Alexander-Sol, line 466: This would be easier to understand and maintain if it was a switch statement on typeof(commonParams.DigestionParams). The current implementation assumes that there will only ever be two classes that implement IDigestionParams

Alexander-Sol (changes requested):
Need to check for bad casts in CrosslinkSearch. 
IDigestionParams handling in MetaMorpheusTask would be better as a switch statement

MetaMorpheus/EngineLayer/CrosslinkSearch/CrosslinkSearchEngine.cs
nbollis: I am assuming that we will not use the crosslink search engine for RNA without serious refactoring that would catch this assumption.

Alexander-Sol: Would it require a serious refactor? Why can't you just use a
`if (CommonParameters.DigestionParams is not DigestionParams digestionParams)`
check and throw an exception if false?

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
nbollis: Good point. I converted this to a switch and added more FileSpecificParameter weirdness handling.

MetaMorpheus/EngineLayer/CrosslinkSearch/CrosslinkSearchEngine.cs
nbollis: > Why can't you just use a `if (CommonParameters.DigestionParams is not DigestionParams digestionParams)` check and throw an exception if false?

That is the affect of the current implementation. I can add more explicit checks if you'd like.

RayMSMS (approved):

trishorts (approved):

PR #2353, opened 2024-02-23
Shared peptides tic box for lfq

trishorts:
Important back and forth on twitter today about the inclusion of shared peptides for label free quant. General sentiment is no, which has been the default parameter that we use in MetaMorpheus for flashlfq. However, the flashlfq app gives the user the option. Therefore I added a checkbox to the quantification section of the search task that allows the user to choose. The default is still to not include them.


Alexander-Sol (changes requested):
The tick box should go next to the other boxes, probably below "Normalize quantification results." Also needs a tooltip

I made these changes myself in #2354

PR #2355, opened 2024-03-08
Invert q-value

trishorts:
We have never used the correct calculation for q-value, which is (decoy count + 1)/(target count). This is because, the highest scoring PSM would have q-value = 0.5. Bill Noble alerted me to an alternate strategy for computing q-value that would eliminate this problem and allow us to use the correct formula. This is accomplished by computing q-value from lowest scoring PSM to highest scoring PSM subject to the following rule. Whenever the current q-value is greater than the previous q-value, we keep the previous q-value. Now, the highest scoring PSM will have a q-value equal to the q-value of the highest scoring decoy PSM. It will never be zero. c est la vie.

overall we expect to see a higher yield of both peptides and PSMs with q-value < 0.01.

An unfortunate by product of this update is that some unit tests with low psm count fail to have sufficent psms with low q-value. In certain instances, this prevents PEP from being calculated. This problem was solved by adding another mzML that could boost the psm count to a sufficiently high level to allow all unit tests to complete successfully while testing what they were intended to test. Please note, some test assertions had to be changed to accommodate the new results.

### DIFF
--- a/MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
+++ b/MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
@@ -189,7 +189,7 @@ namespace EngineLayer.FdrAnalysis
             }
         }
         /// <summary>
-        /// This method gets the PSM count for each psm with q-value < 0.01 and adds that information to the individual psm.
+        /// This method gets the count of PSMs with the same full sequence (with q-value < 0.01) to include in the psmtsv output
         /// </summary>
         public void CountPsm(List<SpectralMatch> proteasePsms)
         {
--- a/MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
+++ b/MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using EngineLayer;
+using EngineLayer.FdrAnalysis;
 
 namespace EngineLayer.FdrAnalysis
 {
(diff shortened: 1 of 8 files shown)


MetaMorpheus/EngineLayer/PsmTsv/PsmTsvWriter.cs
```diff
@@ -161,6 +162,7 @@ internal static (string ResolvedString, string ResolvedValue) Resolve(IEnumerabl
             }
         }
 
+        //The null-coalescing operator ?? returns the value of its left-hand operand if it isn't null; otherwise, it evaluates the right-hand operand and returns its result.
```
Alexander-Sol, line 165: there aren't any null-coalescing operators nearby

MetaMorpheus/EngineLayer/PsmTsv/PsmTsvWriter.cs
```diff
@@ -177,6 +179,7 @@ internal static void AddBasicMatchData(Dictionary<string, string> s, SpectralMat
             s[PsmTsvHeader.Notch] = psm == null ? " " : Resolve(psm.BestMatchingBioPolymersWithSetMods.Select(p => p.Notch)).ResolvedString;
         }
 
+        //The null-coalescing operator ?? returns the value of its left-hand operand if it isn't null; otherwise, it evaluates the right-hand operand and returns its result.
```
Alexander-Sol, line 182: there aren't any null-coalescing operators nearby

MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
```diff
@@ -223,35 +188,31 @@ public static void Compute_PEPValue_Based_QValue(List<SpectralMatch> psms)
                 psms[psmsArrayIndicies[i]].FdrInfo.PEP_QValue = Math.Round(qValue, 6);
             }
         }
-
-        public void CountPsm()
+        /// <summary>
+        /// This method gets the PSM count for each psm with q-value < 0.01 and adds that information to the individual psm.
```
Alexander-Sol, line 192: I don't understand this summary comment. What is the PSM count for an individual psm? Is it always 1 or 0?

trishorts: comment updated w/ new explanation

image

MetaMorpheus/EngineLayer/PsmTsv/PsmTsvWriter.cs
trishorts: deleted

MetaMorpheus/EngineLayer/PsmTsv/PsmTsvWriter.cs
trishorts: comment deleted

nbollis (approved):

elaboy (approved):

PR #2360, opened 2024-04-17
added hyperscore calculation to PEP

trishorts:


MetaMorpheus/EngineLayer/FdrAnalysis/PEPValueAnalysisGeneric.cs
```diff
@@ -606,6 +607,29 @@ private static float GetMobilityZScore(SpectralMatch psm, IBioPolymerWithSetMods
...
 
+        private static float GetFraggerHyperscore(SpectralMatch psm, IBioPolymerWithSetMods selectedPeptide)
+        {
+            var peptideFragmentIons = psm.BioPolymersWithSetModsToMatchingFragments[selectedPeptide];
+            var nIons = peptideFragmentIons.Where(f => f.NeutralTheoreticalProduct.Terminus == FragmentationTerminus.N).ToList();
+            var cIons = peptideFragmentIons.Where(f => f.NeutralTheoreticalProduct.Terminus == FragmentationTerminus.C).ToList();
+            ulong nIonFactorial = GetFactorial((ulong)nIons.Count);
+            nIonFactorial = nIonFactorial < 0 ? ulong.MaxValue : nIonFactorial; // this is a hack to prevent overflow
+            ulong cIonFactorial = GetFactorial((ulong)cIons.Count);
+            cIonFactorial = cIonFactorial < 0 ? ulong.MaxValue : cIonFactorial; // this is a hack to prevent overflow
```
nbollis, line 618: Ulong will overflow after 20 fragment ions. This check will not prevent overflow due to it being unsigned. We should discuss what max value is required. BigInteger does not appear to have a practical limit to its value. 
image

trishorts:
I think it goes a bit higher. Nonetheless it is not meant to prevent overflow. When overflow happens a negative number is returned. This check will substitute ulong max for that negative number   The answer won't be correct but it will be large. In fact it will be much larger than a decoy.

PR #2361, opened 2024-04-29
MetaDraw Fragmentation Reanalysis

nbollis:
Enables the reanalysis of a fragmentation spectrum with additional fragment ion types. 

The plotting methods use the MatchedIons of the psm that gets passed in to determine what to plot. To enable this functionality, I had to temporarily replace the fragment ions in the psm with the new matched ions, then put the original ions back in once the plotting has finished. 

To use: have your normal MetaDraw Display
image

Select the additional ions you want to search for
image

His the search button. The persist check box will enable this research on all new PSMs you plot

image


trishorts:
pretty cool. i only ever see yAmmoniaLoss. maybe we need unit tests for each of the ion types? sounds like lots of work...... I wonder if we paste a couple of example spectra into proteinprospector if it sees any other fragments.

nbollis:
> pretty cool. i only ever see yAmmoniaLoss. maybe we need unit tests for each of the ion types? sounds like lots of work...... I wonder if we paste a couple of example spectra into proteinprospector if it sees any other fragments.
That is a good idea and maybe a good starter project. Our current tests only truly test b, y, and c ions

Alexander-Sol (approved):
I pulled it and it worked great

avcarr2 (approved):

trishorts:
is the top supposed to look like this? 
image

trishorts:
i think that export spectrum library should maybe contain the newly annotated fragments. 
image

nbollis:
> is the top supposed to look like this? image

This has been fixed

trishorts (approved):

PR #2374, opened 2024-06-13
Pep bug fix

Alexander-Sol:
Fixed an issue where PEP was duplicating PSMs when creating training data.

Fixed SpectralRecoveryTests (Those tests depended on duplicating one decoy PSM 4 times in order for PEP to run. Once PSMs were no longer duplicated, PEP broke. Fixed the test by adding in three additional decoys)

Also, edited the PostSearchAnalysisTaskTest. We now see fewer PSMs with PEP-Q < 0.01 and more peptides with PEP-Q < 0.01. The test has been edited to reflect that.


trishorts (approved):

avcarr2 (approved):
Looks good

PR #2375, opened 2024-06-18
Only quantify peptides that clear the q-value threshold

Alexander-Sol:
Previously, when we wrote the AllQuantifiedPeptides output, we filtered based off of PSM q-value. So you would have some peptides with high q-values reported in the quant results. This fixes that!

In order to fix that, I had to reconfigure FDR calculation at the PSM and Peptide level. I did this by adding two new fields to SpectralMatch: PsmFdrInfo and PeptideFdrInfo. The names are self explanatory. There are a variety of toggles that were added to specify when Psm level FDR info should be used and peptide level info should be used.


MetaMorpheus/Test/PostSearchAnalysisTaskTests.cs
```diff
@@ -101,6 +101,16 @@ public static void AllResultsAndResultsTxtContainsCorrectValues_PepQValue_Bottom
             Assert.AreEqual("TaGe_SA_A549_3_snip_2 - Target protein groups within 1 % FDR: 140", results[15]);
         }
 
+        [Test]
+        public static void TestPeptideQuantificationResults()
```
nbollis, line 105: This test seems subject to break in the future due to the hard coded number. Could you instead read in the peptides file, count the results at threshold, then use that in your assert?

MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -501,16 +501,32 @@ private void QuantificationAnalysis()
...
+            {
+                peptideSequencesForQuantification = FilteredPsms.Filter(Parameters.AllPsms,
+                                CommonParameters,
+                                includeDecoys: false,
+                                includeContaminants: true,
+                                includeAmbiguous: false,
+                                includeAmbiguousMods: false,
+                                includeHighQValuePsms: false,
+                                filterAtPeptideLevel: true)
+                    .Select(pep => pep.FullSequence).Distinct().ToList();
```
nbollis, line 516: We are excluding ambiguous now, has that always been the case?

nbollis (changes requested):

MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
trishorts: yes

trishorts (approved):

PR #2376, opened 2024-06-18
new style computation of pep q-value

trishorts:
Currently, the pep q-value is computed by summing the pep and dividing by row number for psms or peptides sorted in descending order by MetaMorpheus score.

In this PR, pep is used to re-score the psms/peptides. PSMs are sorted in ascending order (good to bad) and then pep q-value is computed from bottom to top like we recently did for q-value using the formula (cumulative decoy count + 1) / cumulative target count).


MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -126,105 +125,81 @@ protected override MyTaskResults RunSpecific(string OutputFolder, List<DbForTask
...
             {
-                fileSpecificPsmsOrPeptides.RemoveAll(b => b.IsDecoy);
+
+                _filteredPsms = _filterType.Equals("q-value")
+                    ? psms.Where(p => p.PsmFdrInfo.QValue <= _filterThreshold && p.PsmFdrInfo.QValueNotch <= _filterThreshold).ToList() 
+                    : psms.Where(p => p.PsmFdrInfo.PEP_QValue <= _filterThreshold)
+                        .ToList();
+                _filteredPsms.RemoveAll(p => (p.IsDecoy && includeDecoys));
+                _filteredPsms.RemoveAll(p => (p.IsContaminant && includeContaminants));
+                _filteredPsms.RemoveAll(p => (p.BaseSequence.IsNullOrEmpty() && removeAmbiguous));
```
Alexander-Sol, line 164: You've got your booleans mixed up. It should be (p.IsDecoy && !includeDecoys)

PR #2379, opened 2024-06-25
Meta Draw Quality of Life Updates

nbollis:
Added additional customization options to the meta draw spectral match plot. 
MetaDrawSettings is now more robust to prevent crashes due to saved parameter reading/writing.
Meta Draw settings window was reorganized into logical sections.

New maximum annotation font size is 36 (up from 18). 
Annotated peak thickness is now customizable. 
Unannotated peak thickness is now customizable. 
Axis text size is now customizable. 
Axis title text size is now customizable.


Alexander-Sol (approved):
I really love this PR, it's running well for me.

My one gripe is that it's difficult to find where all the new options are. Previously the "Annotation text size" would pop up as an editable option without having to expand the "Advanced Ion Annotations."

I think it would be more user friendly if "Annotation text size" and maybe "Annotated peak thickness" were displayed by default, and all the other options were kept where they are.

trishorts (approved):

elaboy (approved):

PR #2380, opened 2024-06-25
DIA_XIC

YulingDai:


acesnik:
🤖 Closing this — DIA and XIC construction were built out on master while this sat, and the shape they took is different enough that rebasing is not the right move.

Two PRs cover it. `#2546` "MetaMorpheusDIA" (merged 2025-10-08) created `EngineLayer/DIA/`, including `XicConstruction/XicConstructor.cs` with the `MzPeakXicConstructor` and `NeutralMassXicConstructor` implementations. `#2573` "XicGrouping part2" (merged 2026-01-22) added `PrecursorFragmentGrouping/XicGroupingEngine.cs` alongside `PfGroupingEngine` and `UmpirePfGroupingEngine`.

Mapping this branch onto that: the precursor-and-fragment grouping added here as `GroupPrecursorPeaksAndFragmentIonsForDIA` and `GroupPrecursorPeakAndFragmentIons` is what the three grouping engines do, `GetPeakCurve` and `GetPeakCurves` are what the XIC constructors do, and the peak-curve type itself now lives a layer down in mzLib as `MassSpectrometry/PeakIndexing/ExtractedIonChromatogram.cs` with the splines in `PeakIndexing/PeakSpline/`. The data model is `PrecursorFragmentPair.cs` and `PrecursorFragmentGroup.cs`.

The structural difference is the reason this is a close rather than a rebase. Those five methods were added to `TaskLayer/MetaMorpheusTask.cs`, the abstract base every task derives from, as 364 lines of statics; master has them in engine classes behind `DIAEngine` and `DIAparameters`, which is where a search mode belongs and what makes it configurable per task. Porting the branch forward would mean moving the same logic into those classes, at which point it is a reimplementation of code that already exists rather than a rebase of code that does not.

Deleting the `DIA_XIC` branch along with it — the commits are not lost, since GitHub keeps the pull-request head ref: `git fetch origin refs/pull/2380/head` still retrieves `efc61995` after the branch is gone. This branch was never mirrored to a fork, so that ref is the copy.

**Reopening signal**: if there is behaviour in these 925 lines that `#2546` and `#2573` do not cover — a grouping rule, an XIC boundary condition, a parameter — reopen and name it, and it can go in as a change against the current engines. The concrete way to check is to diff this branch's grouping logic against `PfGroupingEngine` and `XicGroupingEngine` rather than against master as a whole.

PR #2426, opened 2024-10-15
Q value threshold adjustment and bug fix

trishorts:
This PR changes the threshold used for using the more stringent and correct q-value calculation (D+1)/T to a total of 1000 PSMs. Below that level we use D/T. This will help with top-down files that contain fewer PTMs. It is hoped that in the future, we can eliminate this threshold and use one threshold for all. Some PRs that might help would be paired target-decoy and decoy on the fly using 10X decoys. 

An threshold override was also added to FdrAnalysisEngine that enables it to compute PEP when desired for lists of PSMs <1000 used "Reflection"

For reference on the equation please see:

J Proteome Res. 2022 Oct 7;21(10):2412-2420. doi: 10.1021/acs.jproteome.2c00282. Epub 2022 Sep 27. 
https://pubmed.ncbi.nlm.nih.gov/36166314/

Traditional Q-Value (D/T) for PSM Count < 1000
image
image
image


Alexander-Sol (approved):

Alexander-Sol (changes requested):
the QvalueThresholdOverride property and associated field are a little wonky, I suggest changing it to a property with public get; private set;, then changing it in the relevant tests using reflection. Also, add the non-parallelizable tags

nbollis (dismissed):

elaboy (dismissed):

nbollis (approved):

elaboy (approved):

Alexander-Sol (approved):
Looks good

PR #2430, opened 2024-10-28
Updated FilteredPsms to only iterate through PSMs once

Alexander-Sol:
Previously, when filtering, we iterated through the list of PSMs 6 times. Now, we only iterate through once in a single LINQ statement. This should give a modest performance increase.


nbollis (approved):
LGTM

RayMSMS (approved):

PR #2431, opened 2024-10-28
Peptide Level PEP-Q Value changes

Alexander-Sol:
This PR changes the way PSMs are rolled up to peptides for the purpose of PEP-Q Value calculations.

When calculating PEPQ Values at the peptide level, you first have to select the a PSM to represent each full sequence (peptide/proteoform). Previously, PSMs were selected based on their MetaMorpheus score. Now, PSMs will be selected based on PEP (lowest PEP PSM represents the full sequence, MetaMorpheus score is used as a tie-breaker). 

The AllPeptides/AllProteoforms outputs will change as a result of this PR. The PSM with the lowest PEP will be used to represent each peptides. The output will still be ordered by MetaMorpheus score. All entries will have a PEP-Q value <= 1.


MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
```diff
@@ -83,7 +83,7 @@ private void DoFalseDiscoveryRateAnalysis(FdrAnalysisResults myAnalysisResults)
 
                         // peptides are ordered by MM score from good to bad in order to select the best PSM for each peptide
                         peptides = psms
-                            .OrderByDescending(p => p)
+                            .OrderBy(p => p.FdrInfo.PEP)
                             .GroupBy(p => p.FullSequence)
```
nbollis, line 87: should we add a thenby in case two results get the same pep value?

MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
```diff
@@ -110,7 +110,7 @@ private void DoFalseDiscoveryRateAnalysis(FdrAnalysisResults myAnalysisResults)
                     // really, in this case, we only need to run one or the other (i.e., only peptides or psms are passed in)
                     // but there's no mechanism to pass that info to the FDR analysis engine, so we'll do this for now
                     peptides = psms
-                            .OrderByDescending(p => p)
+                            .OrderByDescending(p => p.FdrInfo.PEP)
```
nbollis, line 113: Why are we using linq statements here and above instead of the filtering methods you wrote?

MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
Alexander-Sol: You're right, changed it

nbollis (approved):

trishorts (approved):
surprised that no unit tests needed an update

PR #2438, opened 2024-11-08
eliminate nunit legacy code

trishorts:
Some time in the last few months we updated the NUNIT nuget package, which eliminated some assertions that we had used frequently, such as Assert.AreEqual. To get around the problem of changing all those lines of code in our unit tests we used a "Legacy" option that allowed us to retain the old terms for a while. This PR eliminates all of the legacy assertions bringing us up to date with the current NUNIT nuget package


Alexander-Sol (approved):
good change

nbollis (approved):

RayMSMS (approved):

PR #2439, opened 2024-11-14
Fragger and xcorr modules

trishorts:

Here I add two candidate modules for use in the PEP analysis engine. These should be investigated for efficacy in discrimination between correct and incorrect identifications.

MSFragger computes a hyperscore similar to that of X!:
hyperscore=log(Nb!Ny!∑i=1NbIb,i∑i=1NyIy,i)

where Nb is the number of matched b-ions, Ny is the number of matched y-ions, Ib,i are the intensities of matched b-ions, and Iy,i are the intensities of matched b-ions. While the theoretical fragment index can be adapted to include other fragment ion types, only b and y ions are included and used for scoring at this time. Expectation calculation is also performed in a similar manner as X! Tandem through linear regression of the survival function to estimate the expectation of a given hyperscore [51](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5409104/#R51).

[Nat Methods.](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5409104/#) Author manuscript; available in PMC 2017 Oct 10.
Published in final edited form as:
Nat Methods. 2017 May; 14(5): 513–520.
Published online 2017 Apr 10. doi: [10.1038/nmeth.4256](https://doi.org/10.1038%2Fnmeth.4256)
PMCID: PMC5409104
NIHMSID: NIHMS861287
PMID: [28394336](https://pubmed.ncbi.nlm.nih.gov/28394336)
MSFragger: ultrafast and comprehensive peptide identification in shotgun proteomics

XCorr - 
        /// <summary>
        /// https://willfondrie.com/2019/02/an-intuitive-look-at-the-xcorr-score-function-in-proteomics/
        /// 
        /// A mass spectrum can be preprocessed by subtracting the mean intensities at all of the offsets. 
        /// Then a single dot product between the preprocessed mass spectrum and the theoretical peptide 
        /// mass spectrum yields the xcorr score, which is made possible because of the distributive 
        /// property of the dot product.
        /// 
        /// Since we have already chosen the match for this scan, we can use the matched ions to calculate the
        /// xcorr and skip the dot product step.


acesnik (commented):
🤖 Read this alongside #2329 since they are the same idea — alternative scoring functions — but this one is in much better shape, mostly because it is honest about being a module rather than a feature.

**Blast radius is zero, and that is deliberate.** `GetFraggerHyperScore` and `Xcorr` are `public static` and called from nothing but `PepAnalysisEngineTests`. Nothing in `PEPAnalysisEngine`'s feature construction references them, so this is +105/-0 that cannot move an identification. Staging a scoring function with tests before wiring it up is the right order, and the title says so.

`GetLog10Factorial` is also properly tested — negative input returns null, a large input is checked with a stated tolerance, and the nullable return is guarded at both call sites rather than `.Value`-ed blindly.

## The one thing worth settling before anything consumes this

The hyperscore's intensity term uses the **product** of the two per-terminus sums:

```csharp
double log10IntensitySum = (nSum > 0 && cSum > 0) ? Math.Log10(nSum * cSum) : 0.1;
return log10Fact(nCount) + log10Fact(cCount) + (float)log10IntensitySum;
```

X!Tandem's classic hyperscore, which MSFragger describes itself as following, is `Nb! · Ny! · ΣI` where `ΣI` is a single sum over all matched fragment intensities — which in log space is `log10(Σb + Σy)`, not `log10(Σb) + log10(Σy)`. **I could not verify which form the MSFragger paper specifies**, and I am not going to assert an error I have not confirmed. But the two differ materially and not by a constant:

```
  Σb      Σy    | log10(sum)  log10(product)  diff
  1e3     1e3   |      3.30            6.00   2.70
  1e6     1e6   |      6.30           12.00   5.70
  5e4     2e4   |      4.85            9.00   4.15
```

The gap grows with intensity, so as with #2329's length bias it changes ranking rather than shifting a scale.

**The reason this needs checking against the paper rather than against the test:** the test's expected value is the implementation's own output. `7.112605` is exactly `log10(3!) + log10(3!) + log10(600 × 600)` for the b1/b2/b3 and y1/y2/y3 fixture at intensities 100/200/300. The sum form gives `4.635484`. So the assertion pins the arithmetic — usefully — but it cannot detect a wrong formula, because it was derived from the formula it is testing.

That makes it cheap to settle: one look at the paper's equation tells you whether the expected value should be 7.112605 or 4.635484, and whichever it is, the test then means something it currently cannot.

## Smaller

The `: 0.1` fallback when either terminus has no matched intensity is a magic constant. It is there to avoid `log10(0)`, which is right, but `0.1` gives a PSM with only b-ions a small positive intensity contribution rather than a defined one — and a one-sided match is exactly the case where a hyperscore should be low. Zero, or the single-sided sum, would both be more defensible than 0.1.

The `Xcorr` remarks cite a blog post. It is a good explanation and I have no objection to linking it, but the primary source (Eng et al., the original SEQUEST cross-correlation) is what a reader will want to check the implementation against, and the comment's shortcut — "since we have already chosen the match for this scan, we can use the matched ions and skip the dot product" — is a real claim about equivalence that deserves the citation to lean on.

## Is it still potentially useful?

Yes, more so than #2329, and for a specific reason: these are the two scores most often asked for when comparing against MSFragger and Comet output, and having them as tested static functions costs nothing while nothing calls them.

One flag for whenever they do get wired up: `PEPAnalysisEngine` is where PEP features live, and a new populated feature changes q-values for every search — that is precisely the finding I left on #2640 today. So the moment either of these becomes a `PsmData` field, it needs a parameter defaulting off and a before/after identification count, not just a correctness check.

## Not verified

`Test/Test.csproj` is `net*.0-windows` and the branch is `behind`, so I ran nothing; the arithmetic above is hand-computed from the fixture in the test. I have not read the MSFragger or SEQUEST papers' equations directly, which is exactly why the formula question is phrased as a question.

trishorts:
🤖 Note to self before picking this back up — where this stands as of master `92dfdd6c`.

## The red CI is stale, not real

The failing checks shown here are from run `17212667324` and still include `continuous-integration/appveyor/pr`, so they pre-date the move to Actions. Nothing has re-run since; this branch is ~65 commits behind master.

## Merging master is clean, but it does not compile

`git merge upstream/master` produces **no textual conflicts**. It then fails to build with 4 errors, because `SpectralMatch` has changed underneath the Fragger/xcorr code.

Three fixes, all verified locally — after them the **solution builds with 0 errors and 14 of 16 tests pass**:

1. **The peptide→ions dictionary is gone.** `SpectralMatch.BioPolymersWithSetModsToMatchingFragments` no longer exists; master keeps a list of `SpectralMatchHypothesis` (`EngineLayer/SpectrumMatch/SpectralMatchHypothesis.cs`), each carrying its own `MatchedIons`. In `PEPAnalysisEngine.cs` (two call sites, `GetFraggerHyperScore` and `Xcorr`):

   ```csharp
   // was: psm.BioPolymersWithSetModsToMatchingFragments[selectedPeptide]
   psm.BestMatchingBioPolymersWithSetMods
      .First(h => h.SpecificBioPolymer.Equals(selectedPeptide)).MatchedIons
   ```

   `First` rather than `FirstOrDefault` on purpose: the old dictionary indexer threw on a missing key, so this keeps that contract instead of silently scoring against an empty ion list.

2. **`Test/PepAnalysisEngineTests.cs`** — `BestMatchingBioPolymersWithSetMods.First().Peptide` → `.SpecificBioPolymer` (3 sites).

3. **`TestDataFile`** now takes `List<IBioPolymerWithSetMods>`, not `List<PeptideWithSetModifications>` — so the old call silently bound to the `TestDataFile(double)` overload. Fix the list type and add `using Omics;`.

## The actual blocker: `Xcorr` has no scan to read

`Xcorr` did `psm.MsDataScan.MassSpectrum.XArray`. That property is gone. `psm.Ms2Scan` compiles, **but it is not the same thing** — on master it is opt-in, populated only by `GPTMDTask.cs:171` calling `SetMs2Scan`. In a search flow it is null, so xcorr throws:

```
System.NullReferenceException
   at EngineLayer.PepAnalysisEngine.Xcorr(...) PEPAnalysisEngine.cs:1105
```

That accounts for the 2 remaining failures (`Xcorr_ValidInput_ReturnsExpectedResult`, `Xcorr_EmptyFragments_ReturnsZero`). The opt-in-ness looks deliberate — holding an `MsDataScan` on every `SpectralMatch` is a real memory cost on large runs — so this needs a decision rather than a rename:

1. **Pass the scan into `Xcorr` / `GetFraggerHyperScore`.** The caller in `PEPAnalysisEngine` already has it. No memory change, smallest blast radius. Preferred unless there is a reason not to.
2. **Have the search engines call `SetMs2Scan`.** Makes every PSM retain its scan — needs a look at peak memory on a big top-down run before committing.
3. **Pass the `XArray`/`YArray` in directly.** Narrowest signature, but pushes spectrum plumbing onto every caller.

## Nothing has been pushed

The branch is untouched — head is still `9d397bbb`, matching this PR. The three fixes above exist only in a local worktree, so they need re-applying (or a push) when this is picked up. Ordering: merge master → apply fixes 1–3 → decide the `Xcorr` scan question → re-run the suite, which will be the first real CI this branch has had in months.

PR #2442, opened 2024-11-21
IsoDec incorporation

nbollis:
Added Isodec deconvolution to MetaMorpheus. 
Allowed customization of MS2 deconvoltuion parameters. 

Isodec Decon can be switched out for classic decon in the GUI without the Search Task needing to know which decon type it is using. 

#### PR Summary
Enhances deconvolution parameter handling by making them serializable, adding new UI controls, and improving view model management.
- `CommonParameters.cs`: Removed `[TomlIgnore]` attribute from `DeconvolutionParameters` properties to ensure decon params get saved with each task ran
- Added `IsoDecDeconParamControl` user control and updated `HostDeconParamControl.xaml` for new UI elements.
- `DeconHostViewModel.cs` and `DeconParamsViewModel.cs`: Updated to manage `IsoDecDeconvolutionParameters`.
- Updated test files to include tests for `IsoDecDeconvolutionParameters` and related view models.


trishorts:
I'm wondering if there are different default parameters for bottom up and top down. I ran default isodec for bottom up and the results are less than classic. probably that makes sense. but i wonder if i make adjustments if isodec bottom up results can be improved. do we need explainers to help users?

trishorts (approved):

nbollis:
> I'm wondering if there are different default parameters for bottom up and top down. I ran default isodec for bottom up and the results are less than classic. probably that makes sense. but i wonder if i make adjustments if isodec bottom up results can be improved. do we need explainers to help users?
> 
> top down results are similar but not the same which make sense. not sure what to think about tha.
> 

Default parameters for top down and bottom up are found in DeconHostViewModel. Your results are similar to what I get for top-down. I simply copied the default bottom-up parameters from classic deconvolution, so there is likely a more optimal default.

trishorts (approved):

Alexander-Sol (dismissed):

Alexander-Sol (approved):

trishorts (approved):

PR #2448, opened 2024-12-09
Speedier gptmd again

trishorts:


MetaMorpheus/EngineLayer/Gptmd/GptmdEngine.cs
```diff
@@ -76,6 +75,7 @@ protected override MetaMorpheusEngineResults RunSpecific()
             }
             Parallel.ForEach(Partitioner.Create(0, psms.Count), new ParallelOptions() { MaxDegreeOfParallelism = maxThreadsPerFile }, (range) =>
             {
+                var modDict = new ConcurrentDictionary<string, ConcurrentBag<Tuple<int, Modification>>>();
```
Alexander-Sol, line 78: This doesn't need to be concurrent, as it is created inside the parallel loop, i.e., one per thread

Alexander-Sol (changes requested):

PR #2455, opened 2025-01-07
Stability Update

Alexander-Sol:
This PR is intended to address longstanding instability in MetaMorpheus results. Specifically, Calibration can produce different outputs with identical inputs. 

Unfortunately, even with this PR, variable results from  Calibration were observed

### Changes
- Fixed locks in DataPointAcquisitionEngine
    - I believe this is the main issue. Previously, improper locking led to race conditions when adding to the lists of LabeledDataPoint

- Changed MsDataFileManager to use ConcurrentDictionary to eliminate warnings about race conditions caused by asynchronous loading 
    - (To be clear, I don't think race conditions could have arisen with how FileManagers are currently used, but better safe than sorry)


trishorts (dismissed):

nbollis (approved):

pcruzparri (dismissed):

trishorts (approved):

MetaMorpheus/TaskLayer/CalibrationTask/CalibrationTask.cs
```diff
@@ -253,12 +253,12 @@ private DataPointAquisitionResults GetDataAcquisitionResults(MsDataFile myMsData
 
             //get the deconvoluted ms2scans for the good identifications
             List<Ms2ScanWithSpecificMass> goodScans = new List<Ms2ScanWithSpecificMass>();
-            List<SpectralMatch> unfilteredPsms = allPsmsArray.ToList();
-            foreach (SpectralMatch psm in goodIdentifications)
+            HashSet<int> goodScanIndices = new HashSet<int>(goodIdentifications.Select(psm => Array.IndexOf(allPsmsArray, psm)));
```
nbollis, line 256: why make this change?

nbollis (commented):
Why do we need a new resolve and changes to crosslink?

Alexander-Sol:
> Why do we need a new resolve and changes to crosslink?

> CrosslinkSpectralMatch keeps the legacy version of ResolveAllAmbiguities because I don't understand how XL search works and the new method broke things

nbollis:
> > Why do we need a new resolve and changes to crosslink?
> 
> > CrosslinkSpectralMatch keeps the legacy version of ResolveAllAmbiguities because I don't understand how XL search works and the new method broke things

Let me rephrase. Why do we need a new list based resolve method? (Which in turn forces you to do that implementation of cross-link)

Alexander-Sol:
> > > Why do we need a new resolve and changes to crosslink?
> > 
> > 
> > > CrosslinkSpectralMatch keeps the legacy version of ResolveAllAmbiguities because I don't understand how XL search works and the new method broke things
> 
> Let me rephrase. Why do we need a new list based resolve method? (Which in turn forces you to do that implementation of cross-link)

Calling ResolveAllAmbiguities sets the MatchedFragmentIons property by getting the first element of the BioPolymersWithSetModsToMatchingFragments dictionary. Dictionaries aren't intrinsically ordered. Writing the resolve method guarantees consistent behavior when setting MatchedFragmentIons

trishorts (approved):

PR #2459, opened 2025-01-28
Added Decon Parameters to Calibration Window

nbollis:
Added the option to set custom deconvolution parameters in calibration. 

Added MetaMorpheusTask.DetermineAnalyteType(TheTask.CommonParameters) 
at the beginning of the UpdateFieldsFromTask method in task windows. This ensures analyte type is determined and set
before updating fields from the task.

### DIFF
--- a/MetaMorpheus/GUI/TaskWindows/CalibrateTaskWindow.xaml
+++ b/MetaMorpheus/GUI/TaskWindows/CalibrateTaskWindow.xaml
@@ -49,36 +49,6 @@
                     <Expander x:Name="FileLoadExpander" >
                         <StackPanel>
                             <local:HostDeconParamControl x:Name="DeisotopingControl" DataContext="{Binding}"/>
-                            <GroupBox Header="Peak Trimming">
-                                <StackPanel>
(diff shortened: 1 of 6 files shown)


Alexander-Sol (changes requested):
I think the default parameters in Calibration should match those in search, i.e.:

- Trim MS1 Peaks = False
- Trim MS2 Peak = True
- Top N Peaks = 200
- Minimum Intensity Ratio = 0.01

Additionally, there are options in Search that aren't shown in Calibration. Nominal window width, Number of windows, and Normalize peaks. I'm fine with those being excluded from Calibration if that is intended, just curious

nbollis:
> I think the default parameters in Calibration should match those in search, i.e.:
> 
>     * Trim MS1 Peaks = False
> 
>     * Trim MS2 Peak = True
> 
>     * Top N Peaks = 200
> 
>     * Minimum Intensity Ratio = 0.01
> 
> 
> Additionally, there are options in Search that aren't shown in Calibration. Nominal window width, Number of windows, and Normalize peaks. I'm fine with those being excluded from Calibration if that is intended, just curious

That was not the intention. I copied too much of the xaml code from other windows. I have removed the accidental peak filtering addition to the calibration window.

Alexander-Sol (approved):

trishorts (approved):

PR #2462, opened 2025-01-30
Add timsTOF support to MetaMorpeus

Alexander-Sol:
Calibration and quant aren't currently supported, and will be quietly disabled or skipped if the user tries to quantify/calibrate a timsTOF data file. 

Search and GPTMD will work and are supported

SpectrumMatchFromTsvHeader class is duplicated from mzLib. This will be deleted as soon as a new version of mzLib is released. Should be ignored for now


MetaMorpheus/TaskLayer/CalibrationTask/CalibrationTask.cs
```diff
@@ -83,10 +83,22 @@ protected override MyTaskResults RunSpecific(string OutputFolder, List<DbForTask
 
                 // get filename stuff
                 string originalUncalibratedFilePath = currentRawFileList[spectraFileIndex];
+                var fileExtension = Path.GetExtension(originalUncalibratedFilePath);
                 string originalUncalibratedFilenameWithoutExtension = Path.GetFileNameWithoutExtension(originalUncalibratedFilePath);
+                bool calibrated = false;
+                if (fileExtension.Equals(".mgf", StringComparison.OrdinalIgnoreCase) || fileExtension.Equals(".d", StringComparison.OrdinalIgnoreCase))
```
nbollis, line 89: Did we calibrate mgf in the past? What would happen?

nbollis (commented):
Why do we need a project reference to sqlite in CMD, EngineLayer, and GUI but not TaskLayer and Test?

MetaMorpheus/TaskLayer/CalibrationTask/CalibrationTask.cs
Alexander-Sol: Previously, there was a warning that popped up that .mgfs can't be calibrated, but there were no checks to prevent it.

Alexander-Sol:
Sql dependencies are required in CMD, EngineLayer, TaskLayer, and GUI for installed versions of MM to work properly

Test project isn't installed, so they don't need to be bundled

nbollis (dismissed):

trishorts (approved):

RayMSMS (approved):

PR #2463, opened 2025-01-31
Model lock

Alexander-Sol:
In the PEP analysis engine, multiple threads need to read in the stored model. Currently, no lock is activated during reading, which can result in multiple threads reading the model simultaneously, resulting in a crash.

I'm not sure how to write unit tests for lock objects, so I was hoping we could let this one get through with low patch coverage

### DIFF
--- a/MetaMorpheus/EngineLayer/FdrAnalysis/PEPAnalysisEngine.cs
+++ b/MetaMorpheus/EngineLayer/FdrAnalysis/PEPAnalysisEngine.cs
@@ -17,6 +17,7 @@ using System.Threading.Tasks;
 using Omics.Modifications;
 using Omics;
 using Easy.Common.Extensions;
+using System.Threading;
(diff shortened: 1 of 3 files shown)


MetaMorpheus/EngineLayer/FdrAnalysis/PEPAnalysisEngine.cs
```diff
@@ -455,7 +460,7 @@ public int Compute_PSM_PEP(List<SpectralMatchGroup> peptideGroups,
                         }
                     }
 
-                    lock (lockObject)
+                    lock (_peptideCountLock)
```
nbollis, line 463: Interlocked.Add will perform this action without needing to lock.

RayMSMS (approved):

nbollis (approved):

PR #2465, opened 2025-02-05
Resolve ions

Alexander-Sol:
### Changes to how PeptidesWithSetMods are ordered and resolved within SpectralMatch

- Previously, BestMatchingBioPolymerWithSetMods was set by sorting PWSMs based on FullSequence and Parent Accession
    - Now, PWSMs are sorted based on Notch, the number of matched ions, the number of mods, then full sequence and accession

- SpectralMatch.MatchedFragmentIons is now set based on the best PWSM. The ions corresponding to the top ranked PWSM are written to that property
    - Previously, the first element of the BioPolymerWithSetModsToMatchingFragments dictionary was used. As dictionaries aren't ordered, this could lead to stochastic results
    - CrosslinkSpectralMatch.MatchedFragmentIons keeps the legacy behaviour because I don't understand how XL search works and the new method broke things


MetaMorpheus/EngineLayer/SpectralMatch.cs
```diff
@@ -225,9 +235,13 @@ public void ResolveAllAmbiguities()
...
             }
 
-            // TODO: technically, different peptide options for this PSM can have different matched ions
-            // we can write a Resolve method for this if we want...
-            MatchedFragmentIons = BioPolymersWithSetModsToMatchingFragments.First().Value;
+            // Technically, different peptide options for this PSM can have different matched ions
+            // However, writing out all the matched ions for all the peptide options would break excel
+            // Instead, we set MatchedFragmentIons as the ions matched to the best peptide option
+            if (this is CrosslinkSpectralMatch)
+                MatchedFragmentIons = BioPolymersWithSetModsToMatchingFragments.Values.First();
```
trishorts, line 242: can we add a comment here that in classic search this kind of selection made for problematic results and should be evaluated carefully at som etimein the future.

nbollis (approved):

Issue #955, opened 2018-02-20
Note in the readme which of the growing number of releases to choose for first-time users


acesnik:

acesnik:
- Mostly referring to the first line: Download the current version at https://github.com/smith-chem-wisc/MetaMorpheus/releases.
- Something like this line would probably be more helpful: Download the latest MetaMorpheusInstaller.exe release, and install MetaMorpheus.

rmillikin:
Done

PR #2475, opened 2025-02-17
Ptm stoich

pcruzparri:
### Framing
Master wrote a single `Modification Info List` column: count-based occupancy per site, pooled across
every spectra file, computed at the tail of `CalculateSequenceCoverage()`. This replaces it with
per-sample-group occupancy columns, adds an intensity-weighted variant, and moves the calculation into
mzLib's `BioPolymerGroup` so digestion-product groups can reuse it later.

**Where to start reading.** Occupancy is computed in mzLib by
`BioPolymerGroup.PopulateSampleGroupResults()`, from the spectra files and FlashLFQ intensities handed
to each group by `PostSearchAnalysisTask.QuantificationAnalysis()` and its glyco counterpart.
`ProteinGroup.GetTabSeparatedHeader()` and `ToString()` render the columns. The FlashLFQ area split
that weights intensity occupancy is `PostSearchAnalysisTask.DistributeQuantifiedIntensities()`.

### Output change (breaking)

`Modification Info List` is gone. In its place, per sample group:

| column | meaning |
|---|---|
| `SpectralCount_<label>` | PSMs from that sample group |
| `Intensity_<label>` | FlashLFQ intensity, when quantification ran |
| `CountOccupancy_<label>` | modified / total PSM count, per site |
| `IntensityOccupancy_<label>` | modified / total FlashLFQ area, per site |

The per-site cell format also changed, from `#aa71[Oxidation on S,info:occupancy=0.50(1/2)]` to
`pos71[Oxidation on S,info:fraction=0.50(1/2)]`. Denominators are now per sample group rather than
pooled over the whole run, so values are not comparable to previous releases even where a site appears
in both.

**The `#aa` → `pos` rename is not cosmetic — it marks a numbering change.** Terminal modifications used
to be reported on residues 1 and *length* alongside side-chain mods; they are now `pos0` and
`pos length+1`, distinct from the 1..length side-chain range. Renaming the prefix means a parser written
against the old format fails outright rather than silently reading new numbers under the old convention.

**The column set is dynamic.** Anything parsing this file by column index will break — resolve by
header name. `Number of PSMs` and everything after it shift by the number of sample groups.

### The parts worth stopping on

**Counts are per spectra file, never per channel.**
A TMT reporter channel and a SILAC channel are inferred from one spectrum, not observed separately, so
neither has a spectral count of its own. Multiplex runs get one count column per file. SILAC quantifies
by inventing a spectra file per label (`sample(K+8.014).mzML`); no PSM belongs to those, so they report
an intensity and nothing PSM-derived — the count and occupancy columns are omitted rather than written
as zeros. The **acquired** files are carried alongside the invented ones so counts and count occupancy
still describe what was actually measured, which is what master reported unconditionally. Intensity
columns are unaffected throughout, because those come from a dictionary rather than from a path join,
and that asymmetry is behind most of the edge cases here.

**Intensity occupancy is weighted by FlashLFQ peak areas.**
Each quantified peptidoform's area is split across the PSMs that identified it in that file, and each
PSM carries its share. A peptidoform identified but never quantified contributes nothing rather than
entering as a zero. The divisor is the filtered PSM set FlashLFQ was given, not every PSM in the run —
dividing by the unfiltered count credited part of each area to spectra that occupancy never sums over.
Measured on a local 18-file HeLa run, which is not in the repo and so is not reproducible from this PR:
the shortfall reached 57% of sites, with the old denominator a median of 86% of the corrected value. It
largely cancels in the ratio, since both arms lose signal together — 3.3% of sites move by more than
0.05 occupancy, 1.3% by more than 0.10.

**Occupancy survives every path that skips quantification.**
Count-based columns are populated before the multiplex return, the label-free return, the
experimental-design failure return, and before anything that can throw; the `catch` re-populates them so
a mid-quantification failure yields one uniform count-only schema instead of a mix. Glyco gets the same
treatment, including its individual-file protein reports, which are written before quantification would
ever run.

**Occupancy is computed twice on a label-free run.**
Once before the early returns, then again after quantification propagates, because assigning the
quantification state invalidates the first result. That is the cost of having the columns survive every
early return and any throw. Not profiled.

**`SpectralMatch` equality is reference equality.**
Matching the inherited `GetHashCode`, and matching master, which declared neither `Equals` nor
`GetHashCode`. The `HashSet<ISpectralMatch>` in `ProteinGroup` therefore behaves exactly as before.

### Also in this PR

- `ProteinGroup` extends mzLib's `BioPolymerGroup`, delegating quantification, sample grouping and
  occupancy; alias properties keep existing consumers working.
- Scan metadata moves into an immutable `ScanMetadata` record on `Ms2ScanWithSpecificMass` and
  `SpectralMatch`, so heavyweight scan objects can be released after scoring. `Num Experimental Peaks`
  is re-read after XCorr pre-processing rewrites the spectrum, including for every chimeric precursor
  sharing it.
- `SpectralMatch` implements `ISpectralMatch`.
- `IntensitiesByFile` and `FilesForQuantification` became copy-on-read adapters over the base
  properties, so callers moved from mutating the getter to build-locally-then-assign, hoisted to one
  read per call on the hot path.
- The copy constructor now carries `PeptideFdrInfo` and `IsobaricMassTagReporterIonIntensities`, which
  were being dropped by every clone including SILAC's.
- Smaller fixes: a discarded `possibleVariantPsms` ordering result, which changes row order in
  `AllVariantPSMs.psmtsv`, and a stale `ExperimentalDesign.tsv` leaking between tests via shared
  `TestData/`.

### What was run

Full suite green at **1783 passed / 0 failed / 1 skipped**, against stock mzLib **1.0.585**. No mzLib
change is required to build or test this.

Each occupancy behaviour has a test verified in both directions — reverted the production change,
confirmed the test fails, restored it, confirmed it passes:

- `PostSearchCountOnlyMultiFileWritesOneSchema` — two files, disjoint proteins, quantification off
- `GetMs2ScansRefreshesPeakCountForEveryChimericPrecursor` — chimeric low-CID precursors
- `GlycoCountBasedOccupancySurvivesQuantificationOff` — both glyco gaps, combined and per-file
- `TestTmtProteinGroupsHaveCountColumnsButNoIntensityColumns` — one count column per file, none per channel
- `TestSilacQuantification` — invented channels blank, real files keep their counts

The area-splitting change is not reachable by any test in the suite — no peptidoform in the test data
mixes accepted and rejected PSMs — so it was validated instead by re-running the full 18-file HeLa
search and diffing the occupancy columns against the pre-fix output. Both controls (`CountOccupancy` and
the FlashLFQ `Intensity` columns) came back bit-identical across all 5,244 sites, so the change is
isolated. Sites where every covering form is modified are exactly invariant, as the algebra requires.

### Deliberately out of scope

- **Label-based runs report no intensity occupancy at all**, by decision rather than omission. Splitting
  a peak area from the acquired file would hand light and heavy identifications a share of the light
  channel alone, since each channel's area is filed under an invented file and `ResolveHeavySilacLabel`
  rewrites heavy sequences to light before distribution. A wrong number is worse than an absent one, so
  SILAC reports spectral counts and count occupancy for its acquired files, intensities for every
  channel, and no `IntensityOccupancy_`. This PR is scoped to label-free occupancy; label-based
  occupancy is better served by a three-state estimator than by this one, and that is its own change.
- **The acquired file also carries a zero intensity purely so it has a key for
  `ConstructSubsetProteinGroup`**, which costs a blank `Intensity_` column per acquired file in a SILAC
  run. Temporary: once mzLib gates columns from a shared schema, the file can be carried without it.
- **A site with no measured intensity prints `0.0000(0/0)` rather than an empty cell**, which reads as a
  measured zero. The fix is a per-arm evidence filter in mzLib's `FormatOccupancy`, which is on a branch
  rather than in 1.0.585 or on master, so it is blocked on that landing first.
- **`IntensityOccupancy_` columns are emitted whenever FlashLFQ produced results at all**, rather than
  when a share was actually assigned, so a run that quantified nothing still advertises the columns.

### Follow-ups

Builds and tests against the released mzLib 1.0.585, so nothing has to land first.

Two follow-ups do need an mzLib release before they can land: peptide-group (digestion-product)
occupancy reporting, and MBR features feeding into occupancy. Both will prompt a small refactor of the
occupancy pipeline introduced here.


trishorts (commented):
Automated code review (local pipeline). Critical/Major findings are posted as inline comments on the relevant lines.

MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -1308,9 +1322,9 @@ private void WriteVariantResults()
 
                 foreach (var variant in variants)
                 {
-                    if (variantPWSM.IntersectsAndIdentifiesVariation(variant).identifies == true)
+                    if (variantPWSM.IntersectsAndIdentifiesVariation(variant).identifies == true && variant.Description.IsNotNullOrEmpty())
                     {
-                        if (culture.CompareInfo.IndexOf(variant.VariantCallFormatDataString.Description, "missense_variant", CompareOptions.IgnoreCase) >= 0)
+                        if (culture.CompareInfo.IndexOf(variant.Description, "missense_variant", CompareOptions.IgnoreCase) >= 0)
                         {
                             if (variant.VariantCallFormatDataString.ReferenceAlleleString.Length == 1 && variant.VariantCallFormatDataString.AlternateAlleleString.Length == 1)
```
trishorts, line 1329: **Major** - Variant null-safety fix is incomplete — VariantCallFormatDataString still dereferenced

**Problem:**
This PR changed the variant classification guard/lookup from `variant.VariantCallFormatDataString.Description` to `variant.Description` and added a `variant.Description.IsNotNullOrEmpty()` null guard, which signals that `VariantCallFormatDataString` (or its `Description`) can be null. However, inside the `missense_variant` branch the code still dereferences `variant.VariantCallFormatDataString.ReferenceAlleleString.Length` and `variant.VariantCallFormatDataString.AlternateAlleleString.Length` with no null check. If a variant has a non-empty `Description` but a null `VariantCallFormatDataString`, this throws a `NullReferenceException`. `WriteVariantResults()` is not wrapped in a try/catch, so the exception propagates and aborts the post-search analysis.

**Impact:**
Any database containing applied sequence variations that carry a missense `Description` but lack `VariantCallFormatDataString` data will crash the entire search at the variant-writing step, losing all results. The new guard gives a false sense of safety because the very field it was meant to protect against is still dereferenced unconditionally a few lines later.

**Suggested fix:**
Either gate the whole branch on `variant.VariantCallFormatDataString != null`, or read the allele strings through the same null-safe accessor the `Description` change relies on, falling back to the MNV path (or skipping) when the VCF data is unavailable. Ensure `Description` and the allele-length sub-classification draw from a consistent, non-null source.

**Affected test scenarios:**
- WriteVariantResults - missense variant with non-empty Description but null VariantCallFormatDataString - completes without NRE
- WriteVariantResults - SNV vs MNV classification - still distinguished correctly when VCF allele data is present
- WriteVariantResults - variant with null/empty Description - skipped without throwing

trishorts (commented):
Automated code review - 4 finding(s) at severity >= Major, posted as inline comments.

trishorts (commented):
batch test

MetaMorpheus/EngineLayer/ProteinParsimony/ProteinGroup.cs
```diff
@@ -182,7 +183,15 @@ public void GetIdentifiedPeptidesOutput(List<SilacLabel> labels)
...
+        #region TSV Output (MetaMorpheus column names, base quantification format)
+
+        /// <summary>
+        /// MetaMorpheus TSV header with "Protein" column names and BestPeptidePEP.
+        /// Quantification/occupancy columns use the base BioPolymerGroup SampleGroupResult format.
+        /// </summary>
+        public new string GetTabSeparatedHeader()
         {
             var sb = new StringBuilder();
             sb.Append("Protein Accession" + '\t');
```
trishorts, line 205: **Major** - Per-row quant columns derived from each group's own SampleGroupResults can desync from the single header

**Problem:**
`GetTabSeparatedHeader()` emits the dynamic quant/occupancy columns by iterating `SampleGroupResults` and, per group, checking `group.HasIntensityData`. But the header is produced from one representative `ProteinGroup`, whereas `ToString()` re-derives the same columns from *each* `ProteinGroup`'s own `SampleGroupResults` and per-group `HasIntensityData`. The old base code derived columns from the globally-shared `FilesForQuantification.GroupBy(p => p.Condition)`, which was identical for every protein group and therefore could not desync. The new consumption pattern assumes `SampleGroupResults` count, ordering, and each group's `HasIntensityData` flag are uniform across all protein groups, but nothing in this file enforces that.

**Impact:**
If any protein group's `SampleGroupResults` differs in count/ordering, or a group has `HasIntensityData == false` while the header group had it `true` (e.g., a protein with no measured intensity in one sample), that row emits fewer/extra tab-separated cells than the header, silently shifting every downstream column and corrupting the TSV for parsers and MetaDraw.

**Suggested fix:**
Drive both header and rows from a single shared schema (e.g., the global sample-group list and a per-sample-group `HasIntensityData` decision computed once), rather than letting each `ProteinGroup` independently decide its own column set. Ensure the column shape is invariant across all rows and the header.

**Affected test scenarios:**
- ProteinGroupTest - two protein groups where one sample group has intensity for PG-A but not PG-B - header and both rows have identical column counts
- QuantificationTest - mixed-observation experiment - every protein row column count equals header column count

MetaMorpheus/EngineLayer/ProteinScoringAndFdr/ProteinScoringAndFdrEngine.cs
```diff
@@ -165,8 +165,8 @@ private List<ProteinGroup> DoProteinFdr(List<ProteinGroup> proteinGroups)
                 }
 
                 pg.BestPeptideScore = pg.AllPsmsBelowOnePercentFDR.Max(psm => psm.Score);
-                pg.BestPeptideQValue = pg.AllPsmsBelowOnePercentFDR.Min(psm => psm.FdrInfo.QValueNotch);
-                pg.BestPeptidePEP = pg.AllPsmsBelowOnePercentFDR.Min(psm => psm.FdrInfo.PEP);
+                pg.BestPeptideQValue = pg.AllPsmsBelowOnePercentFDR.OfType<SpectralMatch>().Min(psm => psm.FdrInfo.QValueNotch);
+                pg.BestPeptidePEP = pg.AllPsmsBelowOnePercentFDR.OfType<SpectralMatch>().Min(psm => psm.FdrInfo.PEP);
             }
 
             // pick the best for each paired accession based on filter type
```
trishorts, line 186: **Major** - `.OfType<SpectralMatch>()` silently drops PSMs from best-QValue/PEP computation

**Problem:**
The only functional change in this file adds `.OfType<SpectralMatch>()` before `.Min(...)` on the `BestPeptideQValue` and `BestPeptidePEP` assignments, while `BestPeptideScore` on the line above still aggregates over the full `AllPsmsBelowOnePercentFDR` set. Because `AllPsmsBelowOnePercentFDR` is now exposed through the `BioPolymerGroup`/`ISpectralMatch` refactor, the cast was presumably added to reach `FdrInfo` (not on `ISpectralMatch`). But `OfType` *filters* rather than *casts*: any element that is not exactly a `SpectralMatch` is silently excluded from the QValue/PEP minimum, yet still included in `BestPeptideScore.Max`. The three "best peptide" metrics are then computed over inconsistent subsets of PSMs.

**Impact:**
If the set ever contains an `ISpectralMatch` that is not a `SpectralMatch`, `BestPeptideQValue`/`BestPeptidePEP` are computed from a different (smaller) population than `BestPeptideScore`, producing inconsistent sort keys in `SortProteinGroupsByFilterType` and incorrect target/decoy "picked" selection and protein-level FDR. If the filter removes *all* elements from a non-empty set, `.Min()` throws `InvalidOperationException` (sequence contains no elements), aborting the whole scoring engine.

**Suggested fix:**
Use `.Cast<SpectralMatch>()` (which throws loudly on an unexpected type, surfacing the real bug) instead of `.OfType<SpectralMatch>()`, or downcast a single shared filtered list once and reuse it for all three metrics so Score, QValue, and PEP are computed over the same PSM population. If non-`SpectralMatch` elements are genuinely possible, handle that case explicitly rather than letting it skew aggregates silently.

**Affected test scenarios:**
- DoProteinFdr - protein group whose PSM set mixes `SpectralMatch` with another `ISpectralMatch` impl - QValue/PEP/Score computed over identical population
- DoProteinFdr - protein group with non-empty PSM set but no `SpectralMatch` after filter - does not throw; metrics well-defined
- ScoreProteinGroups/DoProteinFdr - all-`SpectralMatch` set (normal case) - best metrics unchanged from pre-refactor behavior

MetaMorpheus/GuiFunctions/Util/ThreadSafeObservableCollection.cs
```diff
@@ -0,0 +1,78 @@
...
+        return ((IEnumerable<T>)snapshot).GetEnumerator();
+    }
+
+    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+
+    // ── UI-thread marshalling ─────────────────────────────────────────────────
+
+    protected override void OnCollectionChanged(NotifyCollectionChangedEventArgs e)
+    {
+        Dispatcher dispatcher = Application.Current?.Dispatcher;
```
trishorts: **Critical** - Synchronous dispatcher.Invoke raised while holding _lock can deadlock

**Problem:**
Every mutation (`InsertItem`, `RemoveItem`, `SetItem`, `MoveItem`, `ClearItems`) acquires `_lock` and then calls the `base.*` method, which synchronously raises `OnCollectionChanged`. The overridden `OnCollectionChanged` calls `dispatcher.Invoke(() => base.OnCollectionChanged(e))`, a *blocking* marshal to the UI thread — and it runs while `_lock` is still held. If the UI thread is concurrently blocked trying to acquire `_lock` (e.g., enumerating via the new `GetEnumerator`, or performing its own mutation), the background thread cannot complete its `Invoke` because the UI thread is parked on the lock, and the UI thread cannot release because it is waiting for the lock the background thread holds — a classic two-party deadlock.

**Impact:**
The application hangs (frozen UI) under exactly the concurrent read/write scenario this class was created to make safe. Because the class is intended for background-thread mutation with UI-thread binding, this is a high-probability hang in production.

**Suggested fix:**
Do not hold `_lock` across the change-notification raise. Capture/perform the mutation under the lock, then raise `OnCollectionChanged` outside the lock; or marshal notifications asynchronously (`BeginInvoke`) rather than synchronously, accepting that the dispatched delegate must not assume the collection is unchanged. Re-evaluate whether the lock must wrap the entire `base.InsertItem` call given that call also raises events.

**Affected test scenarios:**
- ThreadSafeObservableCollection - background thread Adds while UI thread enumerates under an active Dispatcher - completes without deadlock
- ThreadSafeObservableCollection - background thread Add while UI thread Insert - both complete, no hang

MetaMorpheus/GuiFunctions/Util/ThreadSafeObservableCollection.cs
```diff
@@ -0,0 +1,78 @@
...
+        lock (_lock)
+            snapshot = Items.ToArray(); // Items is the protected List<T> from Collection<T>
+        return ((IEnumerable<T>)snapshot).GetEnumerator();
+    }
+
+    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+
+    // ── UI-thread marshalling ─────────────────────────────────────────────────
+
+    protected override void OnCollectionChanged(NotifyCollectionChangedEventArgs e)
```
trishorts: **Major** - PropertyChanged notifications not marshalled to UI dispatcher

**Problem:**
The class overrides only `OnCollectionChanged` to marshal to the dispatcher. `ObservableCollection<T>` also raises `PropertyChanged` for `"Count"` and `"Item[]"` via `OnPropertyChanged`, which is left unmarshalled. When a mutation occurs on a background thread, the `PropertyChanged` event fires on that background thread directly. The class summary claims "CollectionChanged notifications are automatically marshalled to the UI dispatcher," but the equally cross-thread-sensitive `PropertyChanged` path is not handled.

**Impact:**
WPF bindings that listen to `Count` (or any property-change subscriber expecting UI-thread delivery) can receive notifications on a background thread, raising cross-thread `InvalidOperationException` or causing undefined binding behavior — defeating the thread-safety guarantee for those consumers.

**Suggested fix:**
Override `OnPropertyChanged` with the same dispatcher-marshalling logic used in `OnCollectionChanged`, or centralize the marshalling so both notification paths are covered consistently.

**Affected test scenarios:**
- ThreadSafeObservableCollection - background thread Add with a Count-bound subscriber under an active Dispatcher - PropertyChanged delivered on UI thread
- ThreadSafeObservableCollection - verify both CollectionChanged and PropertyChanged are marshalled identically

trishorts (commented):
Automated code review - 4 finding(s) at severity >= Major, posted as inline comments.

MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
pcruzparri: Good catch. That variant.Description change was a temporary patch I'd added while working through some mzLib/MetaMorpheus integration — it isn't actually needed here, since this branch is on the same mzLib version as master. I reverted it back to the original VariantCallFormatDataString.Description code, so the incomplete guard (and the NRE risk it introduced) is gone.

MetaMorpheus/EngineLayer/ProteinParsimony/ProteinGroup.cs
pcruzparri: You're right that the header (built from the first group) and the per-group rows could drift apart in principle. In the normal pipeline every group gets IntensitiesByFile assigned (with zeros for unmeasured proteins), so the column schema is uniform in practice today. I agree it's still fragile. For now, I've added a guard test that builds multiple protein groups and asserts every row's column count matches the header. The actual header refactor (driving header + rows from one shared schema) I'll do in a separate PR since it'd be best to address that at the BioPolymerGroup (mzlib) level.

MetaMorpheus/EngineLayer/ProteinScoringAndFdr/ProteinScoringAndFdrEngine.cs
pcruzparri: Agreed. Switched to a single Cast<SpectralMatch>().ToList() that's reused for Score, QValueNotch, and PEP, so all three are computed over the same PSM set and an unexpected type throws loudly instead of being silently dropped. Parsimony/FDR tests still pass.

MetaMorpheus/GuiFunctions/Util/ThreadSafeObservableCollection.cs
pcruzparri: This whole ThreadSafeObservableCollection was out of scope for this PR — it was an attempt to fix a test flake that only showed up in my local environment, not in CI. I've reverted it out of the branch entirely, so the deadlock risk no longer exists here. If we decide we want a thread-safe collection later, I'll do it in its own PR and handle the locking/marshalling properly.

MetaMorpheus/GuiFunctions/Util/ThreadSafeObservableCollection.cs
pcruzparri: Same as the deadlock comment — the ThreadSafeObservableCollection has been removed from this PR, so this one no longer applies.

MetaMorpheus/EngineLayer/ProteinParsimony/ProteinGroup.cs
```diff
@@ -1,145 +1,146 @@
...
 
-        public HashSet<IBioPolymer> Proteins { get; set; }
+        /// <summary>Maps to <see cref="BioPolymerGroup.BioPolymerGroupName"/>.</summary>
+        public string ProteinGroupName => BioPolymerGroupName;
 
-        public string ProteinGroupName { get; private set; }
+        /// <summary>Maps to <see cref="BioPolymerGroup.BioPolymerGroupScore"/>.</summary>
+        public double ProteinGroupScore
+        {
+            get => BioPolymerGroupScore;
```
nbollis, line 63: [review-backend, review-generic, review-oop] Alias collection getters silently changed from mutable views to detached copies

`FilesForQuantification` and `IntensitiesByFile` now materialize a fresh collection on every getter call. That keeps the new base-property adapters working internally, but it also changes the public `ProteinGroup` API from reference semantics to copy semantics: any existing caller that mutates the getter result now silently mutates a throwaway copy. The PR already had to rewrite internal callers around this, so the break is real; it is just quiet for any remaining external consumers.

Suggestion: Either preserve the old mutable semantics for these alias properties, or make the break explicit by returning read-only views / renaming the APIs so callers cannot mistake them for live collections.

MetaMorpheus/EngineLayer/ProteinParsimony/ProteinGroup.cs
```diff
@@ -199,36 +208,21 @@ public string GetTabSeparatedHeader()
...
+            // Quantification and occupancy columns from base SampleGroupResult system.
+            // Dynamic columns appear only when SampleGroupResults has been explicitly populated
+            // upstream (e.g., LFQ-success path). Workflows that return early from quantification
+            // leave it null/empty and emit only the static columns.
+            if (SampleGroupResults != null)
             {
-                bool unfractionated = FilesForQuantification.Select(p => p.Fraction).Distinct().Count() == 1;
-                bool conditionsUndefined = FilesForQuantification.All(p => string.IsNullOrEmpty(p.Condition));
-
-                // this is a hacky way to test for SILAC-labeled data...
```
nbollis, line 208: [review-oop] ProteinGroup coverage no longer satisfies the inherited BioPolymerGroup coverage contract

`ProteinGroup.CalculateSequenceCoverage()` is now a hidden `new` method that fills only the legacy flat coverage lists. It never updates mzLib's cached `CoverageResult`, even though `ProteinGroup` now inherits `BioPolymerGroup` and therefore exposes the inherited `CoverageResult` property. That property lazily falls back to the base `CalculateSequenceCoverage()` implementation, but `SpectralMatch` does not implement `IHasSequenceCoverageFromFragments`, so callers using the inherited contract can observe different or emptier coverage than the `ProteinGroup` TSV output.

Suggestion: Populate `BioPolymerGroup.CoverageResult` from the MetaMorpheus-specific calculation, or move `SpectralMatch` onto `IHasSequenceCoverageFromFragments` and let the base coverage pipeline own the calculation end to end.

nbollis (commented):
## Review Summary

Approved findings from the dashboard were revalidated against the current PR head after a freshness sync. The head commit changed since initial analysis, but both approved findings still resolved to valid inline anchors on the updated patch.

### Major (2)
- `MetaMorpheus/EngineLayer/ProteinParsimony/ProteinGroup.cs:99` Alias collection getters now return detached copies instead of live mutable collections, creating a silent behavioral break for callers that mutate getter results.
- `MetaMorpheus/EngineLayer/ProteinParsimony/ProteinGroup.cs:415` `ProteinGroup`'s custom coverage calculation no longer keeps the inherited `BioPolymerGroup.CoverageResult` contract in sync, so inherited callers can observe coverage that diverges from MetaMorpheus output.

Grouping: 2 approved findings -> 2 inline comments | 0 summary-only.

MetaMorpheus/EngineLayer/ProteinParsimony/ProteinGroup.cs
```diff
@@ -182,7 +183,15 @@ public void GetIdentifiedPeptidesOutput(List<SilacLabel> labels)
             }
         }
```
trishorts, line 192: **Major** - Per-group HasIntensityData gating can desync TSV header and row columns

**Problem:**
Both `GetTabSeparatedHeader()` and `ToString()` emit the `Intensity_`/`IntensityOccupancy_` columns conditionally on `group.HasIntensityData`, evaluated per `SampleGroupResult` of the *individual* protein group. The header is produced once from a single representative `ProteinGroup`, while each data row is produced by a different `ProteinGroup` instance via its own `SampleGroupResults`. If `HasIntensityData` (or the count/identity of `SampleGroupResults`) for a given sample-group label differs between the header-source group and any row group, the row will emit a different number of tab-delimited fields than the header, misaligning the entire TSV from that column onward. The same hazard applies to the outer `if (SampleGroupResults != null)` guard: a row whose `SampleGroupResults` is null while the header group's was non-null (or vice-versa) silently drops/adds all quantification columns for that row.

**Impact:**
Column-misaligned protein TSV output: intensity, occupancy, PSM count, decoy flag, and QValue values shift into the wrong columns for affected rows, corrupting downstream parsing and any occupancy analysis. The failure is data-dependent (a protein with no intensity in one sample group), so it can pass CI yet break on real datasets.

**Suggested fix:**
Make the column shape a single property of the file/run rather than per-protein-group: decide the set of emitted columns (and each label's HasIntensityData) once from the union of all groups, and have every row honor that same shape (emitting empty fields where a particular group lacks intensity). Ensure the header is derived from the same authoritative shape used by the rows.

**Affected test scenarios:**
- WriteProteinGroupsTsv - one protein group has intensity in sample group A, another has none - all rows have identical column count equal to header
- WriteProteinGroupsTsv - header-source group's SampleGroupResults null but a later group non-null - column counts stay consistent

trishorts (commented):
Automated code review - 1 finding(s) at severity >= Major, posted as inline comments.

MetaMorpheus/TaskLayer/GlycoSearchTask/PostGlycoSearchAnalysisTask.cs
```diff
@@ -538,25 +545,30 @@ private void QuantificationAnalysis()
...
+                                    ? flashLfqProteinGroup.GetIntensity(spectraFile)
+                                    : 0);
                         }
+                        proteinGroup.IntensitiesByFile = intensities;
                     }
+
+                    // Populate SampleGroupResults from the shared spectraFileInfo so
+                    // every PG carries the same dynamic-column schema. Without this, the writer
+                    // would have no way to produce uniform headers/rows.
+                    proteinGroup.PopulateSampleGroupResults();
```
trishorts, line 571: **Major** — PopulateSampleGroupResults runs even when FlashLfqResults is null, on unset intensities

`proteinGroup.PopulateSampleGroupResults()` is now called unconditionally inside `if (ProteinGroups != null)`, *outside* the `if (Parameters.FlashLfqResults != null)` guard that assigns `IntensitiesByFile`. When `DoQuantification` is true but `flashLFQIdentifications.Any()` is false (no IDs below 1% FDR), `Parameters.FlashLfqResults` stays null, so the intensity dictionary is never built — yet `PopulateSampleGroupResults()` still executes and must read `IntensitiesByFile` to produce the `Intensity_<sample>` / `IntensityOccupancy_<sample>` columns. Since this PR changed the `IntensitiesByFile` getter to return `backingField.ToDictionary(...)`, a never-assigned backing field yields either a `NullReferenceException` or all-zero intensity columns. The base code skipped this path entirely (`ProteinGroups != null && Parameters.FlashLfqResults != null`).

**Impact:** A glyco run with quantification enabled but zero quantifiable identifications can throw during protein-group result writing, or emit misleading all-zero intensity/occupancy columns that did not previously exist.

**Suggested fix:** Only call `PopulateSampleGroupResults()` after intensity state is known valid, or initialize `IntensitiesByFile` to an empty dictionary before the call when `FlashLfqResults` is null. Mirror whatever guard the `PostSearchAnalysisTask` path uses so the two quantification implementations behave identically.

MetaMorpheus/EngineLayer/ProteinParsimony/ProteinGroup.cs
pcruzparri: Good call. Switched both to IReadOnly (list/dictionary) and noted in the doc that the getter hands back a snapshot, so it's clear you can't mutate it in place. Nothing internal was relying on mutating the result anymore anyway.

MetaMorpheus/EngineLayer/ProteinParsimony/ProteinGroup.cs
pcruzparri: Agreed. Nothing actually reads the inherited CoverageResult yet, so I'd rather not balloon this stoich-focused PR. A follow up would further the migration from the ProteinGroup class to the BioPolymerGroup. That being said, to just address the Coverage result concern, it can be addressed by either filling CoverageResult from our calc or having SpectralMatch implement IHasSequenceCoverageFromFragments. Recording here as a future note.

MetaMorpheus/EngineLayer/ProteinParsimony/ProteinGroup.cs
pcruzparri: Same thing as the earlier finding. Added a test that checks every row's column count matches the header across multiple groups. The actual header refactor should go in a separate PR. It is currently fragile but fully working.

MetaMorpheus/TaskLayer/GlycoSearchTask/PostGlycoSearchAnalysisTask.cs
pcruzparri: The NRE can't really happen here — the getter is null-safe and PopulateSampleGroupResults is fine with a null intensity dict (just gives count-only columns). It was inconsistent with the search path though, so I made glyco always set IntensitiesByFile (zeros when there's no LFQ) to match. Tests still pass.

trishorts (dismissed):
Approving. I re-reviewed against the current head (`ae97ad2`) — not just the reply threads — and every Critical/Major finding from this round is resolved or acknowledged-and-deferred with a test:

- **Variant NRE** — reverted to master's `VariantCallFormatDataString.Description`; the incomplete guard is gone. ✅
- **`.OfType<SpectralMatch>()` dropping PSMs** — now a single `Cast<SpectralMatch>().ToList()` reused for Score/QValue/PEP, so all three are computed over the same population and an unexpected type throws loudly. ✅
- **ThreadSafeObservableCollection deadlock (Critical) + PropertyChanged marshalling** — whole (out-of-scope) class reverted out of the PR. ✅
- **Alias getters silently detached** — now `IReadOnly` with the snapshot semantics documented. ✅
- **Glyco `PopulateSampleGroupResults` on null LFQ** — glyco now always assigns `IntensitiesByFile` (zeros when no LFQ) and populates uniformly, matching the search path. ✅

Nice, responsive turnaround — the two genuinely risky/out-of-scope items (the deadlock and the variant patch) were cleanly reverted rather than papered over.

## Follow-up notes (non-blocking — for the deferred BioPolymerGroup migration PR)

Two of the findings were reasonably deferred with tests. Recording the residual debt so it isn't lost in the reply threads:

1. **Method-hiding is a trap for the very migration this defers to.** `GetTabSeparatedHeader()` and `CalculateSequenceCoverage()` are now `new`-hidden on a type that *is-a* `BioPolymerGroup`. It's safe **today** only because every caller uses the concrete type — I confirmed `WriteProteinGroupsToTsv` (`PostSearchAnalysisTask.cs:1554`) holds `List<ProteinGroup>` and calls `.First().GetTabSeparatedHeader()`, so the MM override wins. But the follow-up that moves logic onto `BioPolymerGroup` must resolve this **together with** the `CoverageResult` contract finding (#`3443992211`) — the first base-typed caller would otherwise silently get empty columns / diverging coverage.

2. **Header/row column uniformity is enforced by producer discipline, not structure.** The header is built from one representative group while each row re-derives its own columns from its own `SampleGroupResults`/`HasIntensityData`. It stays aligned only because the search and glyco paths now both assign `IntensitiesByFile` for every group (zeros for unmeasured), and `TestMultipleProteinGroupsHeaderAndRowsHaveSameColumnCount` locks that case. It remains latent TSV-corruption risk if any *future* path populates `SampleGroupResults` non-uniformly. The single-shared-schema fix (drive header + rows from one authoritative column shape) is the real resolution and should be the deferred PR — please track it as an issue, not just a reply.

3. **Trivial perf, optional:** the `IntensitiesByFile` getter allocates a fresh dictionary on every call, and around `ProteinGroup.cs:650` it's called twice to read a single value (null-check then `GetValueOrDefault`). Cache in a local. Not worth holding anything for.

None of these block the merge.

MetaMorpheus/EngineLayer/SpectralMatch.cs
```diff
@@ -80,26 +60,82 @@ protected SpectralMatch(IBioPolymerWithSetMods peptide, int notch, double score,
...
+            (PrecursorScanIntensity > 0 ? new[] { PrecursorScanIntensity } : null);
+
+        /// <summary>
+        /// Returns the identified biopolymers (peptides/proteoforms) for ISpectralMatch compatibility.
+        /// Unwraps SpectralMatchHypothesis to the underlying IBioPolymerWithSetMods.
+        /// </summary>
+        public IEnumerable<IBioPolymerWithSetMods> GetIdentifiedBioPolymersWithSetMods() =>
+            BestMatchingBioPolymersWithSetMods.Select(h => h.SpecificBioPolymer);
+
+        public int CompareTo(ISpectralMatch? other)
```
nbollis, line 136: Careful. We already have a carefully orchestrated compare method for spectral matches, we should unify them

MetaMorpheus/TaskLayer/GlycoSearchTask/PostGlycoSearchAnalysisTask.cs
```diff
@@ -451,7 +456,8 @@ private void QuantificationAnalysis()
                         string.Join("|", proteinsOrderedByAccession.Select(p => p.GeneNames.Select(x => x.Item2).FirstOrDefault())),
                         string.Join("|", proteinsOrderedByAccession.Select(p => p.Organism).Distinct()));
 
-                    foreach (var psm in proteinGroup.AllPsmsBelowOnePercentFDR.Where(v => v.FullSequence != null))
+                    foreach (var psm in proteinGroup.AllPsmsBelowOnePercentFDR.OfType<SpectralMatch>()
```
nbollis: There may be issues here with the spectral matches being of the type GlycoSpectralMatch. A cast or something may be more appropriate here, unsure, but please verify this.

pcruzparri: Good catch, forgot to address this on the glyco path. Fixed now.

MetaMorpheus/EngineLayer/SpectralMatch.cs
pcruzparri: This one was a bit of a rabbithole but unified it. CompareTo(ISpectralMatch) now does score-with-tolerance, then scan number reversed, instead of the raw score comparison. Note that in instances where we are comparing by SpectralMatch type, this interface method is not called, and is just implemented for ISpectralMatch contract. So, comparing SpectralMatch objects will go through comparing score->runner up score->precursor mass error ppm->scan number. Calling a comparison on a generic ISpectralMatch (currently no references in MM) will go through this interface method and only compare by score->Scan number, since the other two fields aren't in the interface.

trishorts (commented):
Re-reviewed against the current head. My 2026-07-09 approval was dismissed by later pushes, so this supersedes it.

Context that shaped the review: the occupancy backend is already merged and shipped upstream — mzLib **#955**, itself titled "Ptm stoichiometry", landed `PositionFrequencyAnalysis`, `BioPolymerGroup.PopulateSampleGroupResults()`, `SampleGroupResult`, `ModificationOccupancyCalculator` and `ScanMetadata` in one commit, first tagged **1.0.577**. MM master pins **1.0.585**, so it is on the classpath today and no bump is needed. This PR is the consuming half of that pair, and it consumes it correctly: no local `ScanMetadata` is redefined, and `IntensitiesByFile` is a genuine adapter over the base's `IntensitiesBySample`. The direction is right. The findings below are all about fidelity at that seam, not about duplication.

Nine inline comments: six I'd consider blocking, three worth a decision.

**One thing in the description doesn't match the diff.** Item 5 describes `ThreadSafeObservableCollection` improvements "with comprehensive test coverage" — that identifier appears **zero** times in the diff, and exists in neither MetaMorpheus nor mzLib. What the diff actually does to concurrency is weaken an existing race regression test (see the comment on `FragmentReanalysisRaceConditionTest.cs`). Worth correcting the description either way.

Also of note: MM **#2187** edits the same `ProteinParsimony/ProteinGroup.cs` and will conflict with this refactor.

MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -616,6 +621,16 @@ private void QuantificationAnalysis()
...
                         Parameters.FlashLfqResults, Parameters.AllSpectralMatches.Cast<PeptideSpectralMatch>().ToList(), Parameters.SearchParameters.ModsToWriteSelection, quantifyUnlabeledPeptides);
                 }
+
+                // Populate SampleGroupResults AFTER all quant-state mutation (including SILAC
+                // re-labeling) so every PG carries the same dynamic-column schema for the writer.
+                if (ProteinGroups != null)
+                {
+                    foreach (var proteinGroup in ProteinGroups)
+                    {
+                        proteinGroup.PopulateSampleGroupResults();
```
trishorts, line 631: **Major** — multiplex and non-LFQ searches now emit no PTM occupancy at all

This is the only site (with `:1053` and `PostGlycoSearchAnalysisTask.cs:571`) that calls `PopulateSampleGroupResults()`. `ProteinGroup.cs` never calls it itself — the `new`-hidden `GetTabSeparatedHeader()` bypasses the base class's lazy auto-population, and its own comment says so: *"Dynamic columns appear only when SampleGroupResults has been explicitly populated"* (`ProteinGroup.cs:221`).

But this call sits at the **end** of `QuantificationAnalysis()`, after two early returns:

```csharp
if (Parameters.SearchParameters.DoMultiplexQuantification) { ...; return; }   // :256-263
if (!Parameters.SearchParameters.DoLabelFreeQuantification) { return; }        // :266-269
```

So for TMT/iTRAQ runs and any search with LFQ off, `SampleGroupResults` stays null, and both `GetTabSeparatedHeader()` (`ProteinGroup.cs:224`) and `ToString()` (`:344`) skip every occupancy column.

Meanwhile this PR deletes the `"Modification Info List"` column and the occupancy block at the tail of `CalculateSequenceCoverage()` — which master wrote **unconditionally**, independent of quantification. Net effect: those runs go from always having per-residue occupancy to having none.

To be fair to the stated scope: the PR promises count-based columns survive when the *experimental design* is unavailable, and that path does reach this line. It's the other two early returns that aren't covered. Options are to populate count-based results before the early returns, or to keep writing the old `ModsInfo` column when `SampleGroupResults` is null.

Note the whole method is wrapped in `try/catch → EngineCrashed`, so any exception mid-quantification also skips this call and silently drops the columns.

MetaMorpheus/EngineLayer/SpectralMatch.cs
```diff
@@ -116,9 +108,55 @@ protected SpectralMatch(IBioPolymerWithSetMods peptide, int notch, double score,
...
+            // Delta score and precursor mass error aren't on ISpectralMatch, so those were dropped from this
+            // CompareTo(SpectralMatch) method. Still, same ordering that mzLib's BaseSpectralMatch uses.
+            if (Math.Abs(this.Score - other.Score) > ToleranceForScoreDifferentiation)
+            {
+                return this.Score.CompareTo(other.Score);
+            }
+            return other.OneBasedScanNumber.CompareTo(this.ScanNumber); //reverse the comparision so that the lower scan number comes first.
+        }
+
+        public bool Equals(ISpectralMatch? other)
```
trishorts: **Major** — value equality declared with no matching `GetHashCode`

This adds a value-based definition (`FullFilePath`, `ScanNumber`, `FullSequence`), but `grep -c GetHashCode` over this file returns **0** — `SpectralMatch` still inherits `object.GetHashCode()` (reference identity) and does not override `Equals(object)`.

That matters because PSMs are stored in a hash-based set keyed by the interface:

```csharp
// ProteinGroup.cs:40
AllPsmsBelowOnePercentFDR = new HashSet<ISpectralMatch>();
```

`ISpectralMatch : IEquatable<ISpectralMatch>`, so `EqualityComparer<ISpectralMatch>.Default` resolves to the generic comparer and *does* call this new `Equals` — while hashing by reference. Two PSMs equal by the definition just declared land in different buckets, so `Contains`/`UnionWith` never honour it. The equality contract is broken in the way that's hardest to see: the code compiles, the tests pass, and the set silently stops deduping by the rule it now advertises.

mzLib's sibling pairs the same definition with `GetHashCode() => HashCode.Combine(FullFilePath, OneBasedScanNumber, FullSequence)` **and** an `Equals(object)` override — worth mirroring.

There's also an asymmetry: `BaseSpectralMatch.Equals(ISpectralMatch other)` is `if (other is BaseSpectralMatch sm) return Equals(sm); return false;`. `SpectralMatch` is not a `BaseSpectralMatch`, so `mmPsm.Equals(baseSm)` can be true while `baseSm.Equals(mmPsm)` is always false.

MetaMorpheus/EngineLayer/SpectralMatch.cs
```diff
@@ -116,9 +108,55 @@ protected SpectralMatch(IBioPolymerWithSetMods peptide, int notch, double score,
...
+
+        /// <summary>Maps to <see cref="ScanNumber"/> for ISpectralMatch compatibility.</summary>
+        int ISpectralMatch.OneBasedScanNumber => ScanNumber;
+
+        /// <summary>
+        /// Consolidates quantification intensities for ISpectralMatch compatibility.
+        /// Returns ReporterIonIntensities if available (isobaric), a singleton array of
+        /// PrecursorScanIntensity for LFQ, or null if neither is populated.
+        /// </summary>
+        double[]? ISpectralMatch.Intensities =>
```
trishorts, line 133: **Major** — `IntensityOccupancy_*` is computed from precursor intensity, not FlashLFQ

```csharp
double[]? ISpectralMatch.Intensities =>
    IsobaricMassTagReporterIonIntensities ??
    (PrecursorScanIntensity > 0 ? new[] { PrecursorScanIntensity } : null);
```

mzLib's consumer treats a single-element array as *the* quantification intensity for that PSM:

```csharp
// ModificationOccupancyCalculator.cs:83-84 (also :125, :170, :202)
if (psm.Intensities is { Length: 1 })
    totals.totalIntensity += psm.Intensities[0];
```

and that value is what feeds `IntensityBasedStoichiometry` → the `IntensityOccupancy_<sample>` column. So the "intensity-based stoichiometry from FlashLFQ peptide intensities" the description advertises is actually derived from each PSM's raw MS1 precursor intensity — a **different data source** from the adjacent `Intensity_<sample>` column, which really does come from `IntensitiesBySample`/FlashLFQ. Two neighbouring columns in the same TSV, drawn from two different measurements.

Worse, `Ms2ScanWithSpecificMass.cs:21` does `PrecursorIntensity = precursorIntensity ?? 1`. When the deconvoluter supplied no precursor intensity, the `> 0` guard still passes and **every PSM contributes exactly 1.0** — so `IntensityOccupancy` silently degenerates into a copy of `CountOccupancy` while still being labelled as intensity-based. There's no way for a user to tell the two cases apart from the output.

If the intent is FlashLFQ intensities, this needs to read from the quantification results rather than the scan. If the precursor-intensity fallback is deliberate, it should at least distinguish "no intensity recorded" from a real 1.0.

MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -551,18 +550,19 @@ private void QuantificationAnalysis()
                 var flashLFQIdentifications = new List<Identification>();
                 foreach (var spectraFile in psmsGroupedByFile)
                 {
-                    var rawfileinfo = spectraFileInfo.First(p => p.FullFilePathWithExtension.Equals(spectraFile.Key));
+                    var rawfileinfo = spectraFileInfo.FirstOrDefault(p => p.FullFilePathWithExtension.Equals(spectraFile.Key));
```
trishorts: **Major** — a file/path mismatch now silently drops every PSM from that file

`.First(...)` became `.FirstOrDefault(...)` plus a bare `continue`. Previously the mismatch threw, the outer `catch` called `EngineCrashed("Quantification", e)`, and the user got a visible warning plus a `Quantification_crash.txt`. Now every PSM from the unmatched file is dropped from `flashLFQIdentifications` with no `Warn(...)`, no status message, and no crash file.

A partial mismatch is the dangerous case: protein and peptide intensities get computed from a subset of the data and the run looks completely normal. A total mismatch leaves `Parameters.FlashLfqResults` null while the output still advertises `Intensity_*` and `IntensityOccupancy_*` columns — quantified-looking output with every intensity zero.

If skipping is genuinely the right behaviour here, it needs a `Warn($"...")` naming the file so it appears in the run log; if it isn't, the `.First()` should stay.

MetaMorpheus/EngineLayer/SpectralMatch.cs
```diff
@@ -83,28 +63,40 @@ protected SpectralMatch(IBioPolymerWithSetMods peptide, int notch, double score,
...
+        /// from Ms2ScanWithSpecificMass during construction.
+        /// </summary>
+        public ScanMetadata ScanMetadata { get; }
+
+        // Pass-through properties for backwards compatibility.
+        // Callers can be migrated to ScanMetadata.* over time.
+        public int ScanNumber => ScanMetadata.OneBasedScanNumber;
+        public int? PrecursorScanNumber => ScanMetadata.OneBasedPrecursorScanNumber;
+        public double ScanRetentionTime => ScanMetadata.RetentionTime;
+        public int ScanExperimentalPeaks => ScanMetadata.NumPeaks;
```
trishorts, line 81: **Major** — "Num Experimental Peaks" changes value for XCorr-preprocessed data

This now reads a snapshot taken in the `Ms2ScanWithSpecificMass` constructor. Master read it live at PSM-construction time (`SpectralMatch.cs:30`, `ScanExperimentalPeaks = scan.NumPeaks`).

The ordering is what breaks it. `Ms2ScanWithSpecificMass` objects are built inside `_GetMs2Scans` (`MetaMorpheusTask.cs:350`, `:367`, `:403`), but `XCorrPrePreprocessing` mutates the spectrum **in place, afterwards**:

```csharp
// MetaMorpheusTask.cs:466 (and :481 for child scans) — after _GetMs2Scans returns at :434
parentScan.TheScan.MassSpectrum.XCorrPrePreprocessing(0, 1969, parentScan.TheScan.IsolationMz.Value);
```

So the snapshot records the **pre**-processing peak count where the old code recorded the **post**-processing count. For the affected dissociation path the column in `AllPSMs.psmtsv` silently changes value against previous releases, and no test asserts it.

Capturing `NumPeaks` lazily, or re-snapshotting after preprocessing, would preserve the old value.

MetaMorpheus/Test/ProteinGroupTest.cs
```diff
@@ -244,12 +413,82 @@ public static void TestModificationInfoListInProteinGroupsOutput()
...
+                new HashSet<IBioPolymer> { protein },
+                new HashSet<IBioPolymerWithSetMods> { pwsm1, pwsm2 },
+                new HashSet<IBioPolymerWithSetMods> { pwsm1, pwsm2 });
+
+            // Branch 1: labels == null, DisplayModsOnPeptides == false → BaseSequence
+            pg.DisplayModsOnPeptides = false;
+            pg.GetIdentifiedPeptidesOutput(null);
+            var tsv1 = pg.ToString();
+            Assert.That(tsv1, Does.Contain(pwsm1.BaseSequence));
+            Assert.That(tsv1.Split('\t')[4], Does.Not.Contain("[")); // unique-peptides column has no mod notation
```
trishorts: **Major (test)** — this asserts on the mass column, not the unique-peptides column

`GetTabSeparatedHeader()` orders columns as Accession(0), Gene(1), Organism(2), Protein Full Name(3), **Protein Unmodified Mass(4)**, Number of Proteins in Group(5), **Unique Peptides(6)**.

Index `[4]` is the unmodified-mass list, so `Does.Not.Contain("[")` is being applied to a `"|"`-joined list of doubles and is true no matter what `GetIdentifiedPeptidesOutput` produced. The two `Is.Not.Empty` checks in branches 3 and 4 have the same problem.

Three of the four branch assertions in this new test can therefore never fail — the SILAC label branches end up covered only by `Assert.DoesNotThrow`. Index `[6]` is the column the comment describes.

Given the column set is now dynamic, indexing by name (find the header offset for `"Unique Peptides"`) would be more durable than a literal index.

MetaMorpheus/Test/MetaDraw/FragmentReanalysisRaceConditionTest.cs
```diff
@@ -47,39 +45,53 @@ public static void MatchIonsWithNewTypes_ConcurrentCalls_DoNotThrow()
...
-                Assert.That(completed, Is.True, "Concurrent rematching did not complete within the timeout.");
-                Assert.That(exceptions, Is.Empty,
-                    $"Concurrent rematching threw: {string.Join(Environment.NewLine, exceptions.Select(p => p.ToString()))}");
-            }
-            finally
-            {
-                Directory.Delete(outputFolder, true);
-            }
+            // Wait for matcher to finish (or cancel)
+            matcherTask.Wait(TimeSpan.FromSeconds(10));
```
trishorts: **Please reconsider** — this weakens a race regression test with no accompanying production change

The rewrite drops `[NonParallelizable]`, collapses `Enumerable.Range(0, Environment.ProcessorCount)` concurrent callers down to a single matcher plus one mutator, narrows `catch (Exception)` to `catch (InvalidOperationException)`, and removes the `try/finally` that cleaned `outputFolder`.

The deciding line is this one:

```csharp
matcherTask.Wait(TimeSpan.FromSeconds(10));   // return value discarded
```

The previous version asserted on it — `Assert.That(completed, Is.True, "Concurrent rematching did not complete within the timeout.")`. `Task.Wait(TimeSpan)` returns `false` on timeout rather than throwing, so a hang or deadlock now passes silently. Combined with dropping the `ProcessorCount` fan-out, the test is unlikely to reproduce the race it was written for.

`FragmentationReanalysisViewModel.cs` isn't in this PR, so nothing in the production code motivates the change. If the old test was flaky, that's worth fixing on its own terms — but as it stands this removes coverage without replacing it. At minimum, restore the `Is.True` assertion on the `Wait` result.

MetaMorpheus/EngineLayer/SpectralMatch.cs
```diff
@@ -462,22 +506,12 @@ protected SpectralMatch(SpectralMatch psm, List<SpectralMatchHypothesis> bestMat
...
-            ScanPrecursorMass = psm.ScanPrecursorMass;
             ScanPrecursorMostAbundantMass = psm.ScanPrecursorMostAbundantMass;
             IsMostAbundantMode = psm.IsMostAbundantMode;
             PrecursorAveragine = psm.PrecursorAveragine;
-            ScanOneOverK0 = psm.ScanOneOverK0;
             CollisionalEnergy = psm.CollisionalEnergy;
-            FullFilePath = psm.FullFilePath;
-            ScanIndex = psm.ScanIndex;
             FdrInfo = psm.FdrInfo;
+            PeptideFdrInfo = psm.PeptideFdrInfo;
```
trishorts, line 518: **Question** — this line is new, and unrelated to the stated purpose

The copy-constructor previously copied `FdrInfo` but not `PeptideFdrInfo`. Adding it changes peptide-level FDR propagation through every `CrosslinkSpectralMatch` / `GlycoSpectralMatch` clone, and through the SILAC clone path.

That may well be a genuine bug fix — but it isn't mentioned in the description, no test covers it, and FDR propagation is not somewhere a silent change is comfortable. Could you confirm it's deliberate, and say what it fixes? If it is a fix, it deserves its own test; if it's incidental, it should probably come out of this PR.

MetaMorpheus/EngineLayer/ProteinParsimony/ProteinGroup.cs
```diff
@@ -736,23 +641,32 @@ public ProteinGroup ConstructSubsetProteinGroup(string fullFilePath, List<SilacL
...
 
-                subsetPg.FilesForQuantification = new List<SpectraFileInfo> { spectraFileInfo };
+                if (spectraFileInfo != null)
+                {
+                    subsetPg.FilesForQuantification = new List<SpectraFileInfo> { spectraFileInfo };
+                }
             }
 
-            if (IntensitiesByFile == null)
+            if (IntensitiesByFile == null || spectraFileInfo == null)
```
trishorts: **Performance** — the copying getters are re-evaluated several times per call here

`FilesForQuantification` and `IntensitiesByFile` are no longer fields; each read allocates a fresh collection:

```csharp
get => SamplesForQuantification?.OfType<SpectraFileInfo>().ToList();          // :105
get => IntensitiesBySample?.ToDictionary(kvp => (SpectraFileInfo)kvp.Key, ...); // :116
```

Within `ConstructSubsetProteinGroup`, `FilesForQuantification` is read four times (the null check, the main lookup, and both SILAC fake-path lookups) and `IntensitiesByFile` twice — this null check and the `GetValueOrDefault` just below. Each is a full O(files) copy.

The method is called once per protein group per spectra file (`PostSearchAnalysisTask.cs:1038`), so total copying is on the order of `files² × proteinGroups` — a path that previously allocated nothing, because these were plain fields.

Hoisting each getter into a local at the top of the method fixes the hot path cheaply. Worth noting more broadly: eager copy-on-read getters are new to this codebase — the nearest precedent (`SpectralMatch.cs:230`) is lazy and still carries the author's own comment worrying about its per-call cost.

trishorts: **Performance** — the copying getters are re-evaluated several times per call here

`FilesForQuantification` and `IntensitiesByFile` are no longer fields; each read allocates a fresh collection:

```csharp
get => SamplesForQuantification?.OfType<SpectraFileInfo>().ToList();          // :105
get => IntensitiesBySample?.ToDictionary(kvp => (SpectraFileInfo)kvp.Key, ...); // :116
```

Within `ConstructSubsetProteinGroup`, `FilesForQuantification` is read four times (the null check, the main lookup, and both SILAC fake-path lookups) and `IntensitiesByFile` twice — this null check and the `GetValueOrDefault` just below. Each is a full O(files) copy.

The method is called once per protein group per spectra file (`PostSearchAnalysisTask.cs:1038`), so total copying is on the order of `files² × proteinGroups` — a path that previously allocated nothing, because these were plain fields.

Hoisting each getter into a local at the top of the method fixes the hot path cheaply. Worth noting more broadly: eager copy-on-read getters are new to this codebase — the nearest precedent (`SpectralMatch.cs:230`) is lazy and still carries the author's own comment worrying about its per-call cost.

MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
pcruzparri: You're right, and it was worse than the two early returns.

`PopulateCountBasedOccupancy()` now runs at the top of `QuantificationAnalysis()`, ahead of every return and ahead of anything that can throw, so TMT/iTRAQ and LFQ-off searches get `SpectralCount_` and `CountOccupancy_` back. The columns are per spectra file rather than per reporter channel — a channel carries no spectral count of its own, but the file it was measured in does. Intensity columns stay absent for multiplex: FlashLFQ never runs, and the occupancy calculator only accepts a single per-file intensity, not a reporter array.

Glyco had two separate gaps behind the one call site you cited, and both are closed: the `DoQuantification` early return, and the individual-file protein reports, which are written before quantification would run at all.

Chasing this also turned up something adjacent. In the count-only path each protein group was deriving its own columns from its own PSMs, so with more than one spectra file a group absent from one file wrote a short row against the header taken from the first group — sliding `Number of PSMs` and every column after it two positions left. Every group is now handed the same file list.

Tests: `PostSearchCountOnlyMultiFileWritesOneSchema` (two files, disjoint proteins), `GlycoCountBasedOccupancySurvivesQuantificationOff` (both glyco gaps, combined and per-file), and `TestTmtProteinGroupsHaveNoQuantColumns` inverted to `TestTmtProteinGroupsHaveCountColumnsButNoIntensityColumns` — it had been asserting exactly the behaviour you flagged.

**Your closing paragraph is handled as well.** Populating first means an exception mid-quantification no longer drops the columns outright, but that wasn't sufficient on its own: the `FilesForQuantification` and `IntensitiesByFile` setters null `SampleGroupResults` on assignment, so a throw partway through the propagation loop left some groups count-based and some null under a single header. The `catch` now restores count-based occupancy before `EngineCrashed` runs, so a failure anywhere in `QuantificationAnalysis` — the SILAC block, the distribution step, `PopulateSampleGroupResults` itself — ends at one uniform count-only schema instead of a mixed one.

MetaMorpheus/EngineLayer/SpectralMatch.cs
pcruzparri: Fixed by going the other way. Rather than add `GetHashCode`, `Equals(ISpectralMatch)` is now `ReferenceEquals`, which is what the inherited hash code already implements — so the `HashSet<ISpectralMatch>` in `ProteinGroup` behaves exactly as it did before this PR.

Worth noting this restores master exactly: master's `SpectralMatch` declared neither `Equals` nor `GetHashCode`, so it already had reference semantics. This PR introduced the mismatch and now removes it, rather than adding a new contract.

The asymmetry you named is fixed unconditionally, and without depending on what mzLib does: `mmPsm.Equals(baseSm)` is now `ReferenceEquals` across two different types, which is always false. So the "can be true one way, always false the other" case is gone regardless of `BaseSpectralMatch`'s side.

For what it's worth it doesn't arise here anyway — MetaMorpheus never constructs a `BaseSpectralMatch`; both `HashSet<ISpectralMatch>` sites in `ProteinGroup` only ever hold `SpectralMatch` instances, and the second is a filtering copy where identity dedup is what you want.

What's left is a semantic difference rather than a broken contract — mzLib's sibling has value semantics, this has identity semantics, behind one interface. Worth knowing about, but the contract itself holds.

MetaMorpheus/EngineLayer/SpectralMatch.cs
pcruzparri: Correct, including the `?? 1` degeneration. `Intensities` no longer reads the scan:

```csharp
double[]? ISpectralMatch.Intensities =>
    IsobaricMassTagReporterIonIntensities ??
    (QuantifiedIntensityShare is > 0 ? new[] { QuantifiedIntensityShare.Value } : null);
```

`QuantifiedIntensityShare` is the PSM's share of its peptidoform's FlashLFQ peak area, so the column now comes from the same measurement as the neighbouring `Intensity_<sample>` instead of a different one. A PSM whose peptidoform was never quantified contributes nothing rather than a fabricated 1.0.

**Your second point is not fully answered, though, and I'd rather say so than let it look closed.** You asked that the output distinguish "no intensity recorded" from a real value. The fabricated 1.0 is gone, but the same ambiguity reappears one layer out: a site whose covering PSMs were all unquantified prints `fraction=0.0000(0/0)`, where the `(0/0)` is the only thing separating absent data from a measured zero. And `DistributeQuantifiedIntensities` returns true whenever `FlashLfqResults` is non-null rather than when it actually assigned a share, so the `IntensityOccupancy_` column can be emitted for a run where nothing was quantified at all — every site reading `0.0000(0/0)`.

The fix for both is a per-arm evidence filter that emits blank instead of a zero-denominator fraction. It already exists on the mzLib side; the question is which release MetaMorpheus picks it up in, not whether it gets written.

**One more thing in the same code, which you didn't flag but which fell out of looking at it:** the area was divided by the count of *all* that peptidoform's PSMs while occupancy sums only over the ones below 1% FDR, so a form with rejected PSMs contributed less than its full area. Now fixed — the divisor is the same filtered set FlashLFQ was given.

This was common on a real dataset: 15.9% of quantified (peptidoform, file) pairs were affected, and those contributed a median of half their area. I re-ran a full 18-file search before and after the change and diffed the occupancy columns — `CountOccupancy_` and the FlashLFQ `Intensity_` columns came back bit-identical across all sites, so the delta is isolated to this fix, and sites where every covering form is modified are exactly invariant as the algebra requires. Most sites move very little; the tail does not, so per-site values change even though the distribution barely does.

The arithmetic itself is now unit-tested rather than resting on that run: `SplitAreaAcrossPsms` is a static helper with a test that n PSMs over area A each get A/n and sum back to A, and that a non-positive area leaves the shares unset.

MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
pcruzparri: Done — and chasing it turned up that your two paragraphs have different answers.

You gave two options for the first half: add a `Warn`, or put `.First()` back. It's `.First()` — reverted to master, so `git diff master` on that line is empty again.

That's because the branch turned out to be unreachable. `ExperimentalDesign.ReadExperimentalDesign` selects each `SpectraFileInfo`'s path *out of* the raw file list it's handed (`:57`, `:65`), and errors if any raw file is missing from the design (`:70-76`), with the caller returning before quantification. PSMs carry that same raw-file-list string, so every PSM group has a matching entry by construction. A `Warn` that can't fire is just diff, so it's gone.

The same reasoning answers "quantified-looking output with every intensity zero": that state is real, but not via the experimental design — a design that reads without errors cannot produce a sample group that fails the path join. A raw file that yielded no PSMs reports `SpectralCount = 0`, which is true rather than contradictory.

Where it was real is SILAC. FlashLFQ can't resolve two channels within one file, so quantification invents one file per label — `sample(K+8.014).mzML` — and re-files the quantified peptides into them. PSMs keep the file they were acquired in; `FullFilePath` is scan metadata and get-only. So the invented files matched no PSM and printed `SpectralCount_ = 0` beside a real `Intensity_`.

The fix isn't a path mapping, because there is nothing to map to: a SILAC channel is inferred from one spectrum's isotope envelope, not observed separately, so it has no spectral count of its own — the same reason a TMT reporter channel doesn't. Those channels now report an intensity and nothing PSM-derived, with the columns omitted rather than emitted empty.

The acquired files are carried alongside the invented ones so counts and count occupancy still describe what was actually measured. That mattered more than it first looked: in turnover, and in multiplex without an unlabeled condition, *every* file in the list is invented, so those runs would otherwise have reported no counts at all — less than master, which wrote its occupancy column unconditionally.

Checking that also turned up something worse next door, so intensity occupancy is now switched off for label-based runs entirely. `ResolveHeavySilacLabel` rewrites heavy sequences to light before areas are distributed, so light and heavy identifications collapse into one group keyed on the real file and split the *light* channel's area — the heavy area sits under the invented file and reaches nothing. That predates this PR, but it means the column was wrong rather than merely thin, and a wrong occupancy is worse than an absent one. SILAC keeps counts, count occupancy and per-channel intensities.

This PR is scoped to label-free occupancy. Label-based occupancy wants a three-state estimator rather than this one, which is a separate piece of work. `SilacTest` covers the column set for the multiplex, no-light and turnover cases.

MetaMorpheus/EngineLayer/SpectralMatch.cs
pcruzparri: Fixed, and your suggested remedy needed one more step than it looked.

The count is re-read after preprocessing. But re-snapshotting alone doesn't restore the old value for chimeric precursors: one `MsDataScan` gets one wrapper per deconvoluted precursor, all sharing the spectrum, and the refresh was sitting inside the `!XcorrProcessed` guard — so only the wrapper that happened to do the preprocessing got a fresh count, and the rest kept the pre-XCorr one. The guard now covers the preprocessing only; the refresh runs for every wrapper, inside the lock. Same shape for child scans.

Test: `GetMs2ScansRefreshesPeakCountForEveryChimericPrecursor`. It pins the thread count to 1 on purpose — with only two scans the partitioner splits them across threads, both take the preprocessing branch, and the regression hides.

MetaMorpheus/Test/ProteinGroupTest.cs
pcruzparri: Right on both counts, including that three of the four branch assertions could never fail. Resolved by header name rather than by a corrected literal index, as you suggested — the column set is dynamic now, so a fixed offset would only break again the next time a column moves.

MetaMorpheus/Test/MetaDraw/FragmentReanalysisRaceConditionTest.cs
pcruzparri: Reverted to master's version, which restores everything you named: `[NonParallelizable]`, the `ProcessorCount` fan-out, the broad catch, the `try/finally` cleanup, and the `Assert.That(completed, Is.True, ...)` on the `Wait` result. `git diff master` on this file is empty.

You were right that nothing in the production code motivated the change.

MetaMorpheus/EngineLayer/SpectralMatch.cs
pcruzparri: Deliberate, and it is a fix. The copy constructor was dropping both `PeptideFdrInfo` and `IsobaricMassTagReporterIonIntensities`, so every `CrosslinkSpectralMatch` / `GlycoSpectralMatch` clone and the SILAC clone path lost peptide-level FDR and reporter intensities. It surfaced through SILAC, where the cloned PSMs are the ones that reach quantification.

It is covered — `SilacTest.TestSilacClonePreservesQuantAndFdrData`. Fair point that it wasn't called out; it went in under a commit message about the copy constructor rather than in the PR description.

MetaMorpheus/EngineLayer/ProteinParsimony/ProteinGroup.cs
pcruzparri: Fixed. Both getters are hoisted into locals at the top of `ConstructSubsetProteinGroup`, so the four `FilesForQuantification` reads and two `IntensitiesByFile` reads become one copy each per call instead of six.

Taking the broader point too — these getters are adapters over the mzLib base properties, and the copy-on-read shape isn't something to spread further.

MetaMorpheus/EngineLayer/SpectralMatch.cs
trishorts: Reopening this one. The first half I'm satisfied with: `Intensities` reads `QuantifiedIntensityShare`, so this column and `Intensity_<sample>` come from the same measurement and the fabricated 1.0 is gone. Nice catch on the divisor bug that fell out of it, too.

It's the residual you flagged yourself. On `0.0000(0/0)` vs a measured zero you wrote that the per-arm evidence filter "already exists on the mzLib side; the question is which release MetaMorpheus picks it up in." I can't find it. mzLib master is past 1.0.586 and neither `ModificationOccupancyCalculator` nor `SiteSpecificModificationOccupancy.ToModInfoString` filters on evidence — the latter formats `ModifiedIntensity:G4/TotalIntensity:G4` unconditionally. There's also no open mzLib PR from you (last is #1073, merged 2026-07-09), and this branch pins mzLib 1.0.585.

Could you either link the mzLib PR, or open an issue there and link it here? I don't object to this shipping with the ambiguity — I object to the thread closing as though the fix were scheduled when nothing schedules it. Everything else in the 2026-08-19 pass I've verified against the branch and it's addressed.

trishorts (approved):

Alexander-Sol (approved):

MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -254,6 +253,12 @@ private void QuantificationAnalysis()
         {
             try
             {
+                // Spectral counts and count-based occupancy need no quantification, so fill them in
+                // before every return below and before anything that can throw. Without an experimental
+                // design the columns are labelled per spectra file; the label-free path repopulates
+                // them with intensities once FlashLFQ has run.
+                PopulateCountBasedOccupancy();
```
nbollis, line 260: Should count based occupancy only run if we are doing quantification? Or should count based occupancy be run regardless of if we are doing quantification?

MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -616,13 +615,170 @@ private void QuantificationAnalysis()
...
         }
 
+        /// <summary>
+        /// Populates spectral counts and count-based occupancy from the PSMs alone, so those columns
+        /// survive searches where quantification is switched off, is skipped because the experimental
+        /// design could not be read, or fails partway through. The columns are per spectra file rather
+        /// than per reporter channel, so multiplex runs get them too: a channel carries no spectral
+        /// count of its own, but the file it was measured in does.
+        /// </summary>
+        private void PopulateCountBasedOccupancy()
```
nbollis, line 657: You should put all these new helpers in a region called something like #region Occupancy 

This will prevent postsearchanalysistask from getting cluttered.

nbollis (approved):
Approved with a comment and a question

MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
pcruzparri: count-based occupancy should always run regardless of whether quantification is asked for or not, because 1) no new information is needed from what's already produced (the PSM counts), 2) can't think of a harm, especially when the previous Modification Info List column was always produced even though it was arguably misleading. So, regardless of quant, the marginal improvement is the spectral count and occupancy (previously the Modification Info List) columns are per file searched or spec'ed to the experimental design, whereas before the number of PSMs and the Mod Info List were aggregated across all files and condition to just produced two columns.

MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
pcruzparri: PR was merged, but noted for the follow up refactor after cleaning up some of the biopolymergroup machinery for the peptide group reporting extension of this PR.

PR #2486, opened 2025-03-06
Hotfix: IsoDec and other Missing dlls :(

nbollis:
Isodec dlls were not added to product.wxs on release. 

Oops

This PR adds them and all other missing references


Alexander-Sol (approved):

trishorts (approved):
I downloaded the installer and ran it w/ isodec. No crash. DK if that's the proper way to test but that is what I did.

PR #2491, opened 2025-03-07
Refactored GPTMD to use scan array

Alexander-Sol:
The GPTMD engine needs to access MsDataScans for each PSM to do its thing. Previously, the scans were obtained from a property in each SpectralMatch object. Now, the GPTMD engine takes in an array of Ms2ScanWithSpecificMass and gets the scans from there.

This refactor shouldn't change results from the GPTMD engine or improve performance, but it will enable us to remove the MsDataScan property from SpectralMatch in a future PR


trishorts (approved):

MetaMorpheus/EngineLayer/Gptmd/GptmdEngine.cs
```diff
@@ -175,13 +190,27 @@ protected override MetaMorpheusEngineResults RunSpecific()
...
-                kvp => new HashSet<Tuple<int, Modification>>(kvp.Value)
-            );
-            return new GptmdResults(this, finalModDictionary, modsAdded);
+            //UpdateModDictionary(modDict);
+            foreach(var kvp in modDict)
+                ModDictionary.MergeOrCreate(kvp.Key, new HashSet<Tuple<int, Modification>>(kvp.Value));
+            return new GptmdResults(this, ModDictionary, modsAdded);
+        }
+
+        private void UpdateModDictionary(ConcurrentDictionary<string, ConcurrentBag<Tuple<int, Modification>>> concurrentDict)
```
nbollis, line 199: This can be removed

nbollis (changes requested):

nbollis (approved):

RayMSMS (approved):

PR #2506, opened 2025-05-01
Generalize MetaDraw to use SpectrumMatchFromTsv

nbollis:
Removed protein specific fields from MetaDraw and generalized for oligos as well.


MetaMorpheus/GuiFunctions/MzLibExtensions.cs
nbollis, line 1: These conditional branches will be uncommented once Oligos are searchable.

trishorts (approved):

PR #2526, opened 2025-06-24
Added SQLite.Interop.dll to Product.wxs

Alexander-Sol:
Fixed issue where a dependency (SQLite.Interop.dll) was not installed, causing MetaMorpheus to crash when users attempted to search bruker timsTOF files.


trishorts (approved):
installed from appveyor. worked as expected.

nbollis (approved):

PR #2529, opened 2025-06-30
Meta draw settings to singleton

nbollis:
MetaDrawSettings view models take a bit of time to load as it must process all of the color information stored in the MetaDrawSettings static class into visualizable objects. 

This PR ensure this time-consuming operation only ever occurs once. This is done by converting MetaDrawSettings into a singleton, or a class which only ever has one instance at run-time. 

This PR also breaks out the settings window button into a distinct control which can be dropped anywhere else in the MetaDraw GUI

### DIFF
--- a/MetaMorpheus/GUI/MetaDraw/MetaDraw.xaml
+++ b/MetaMorpheus/GUI/MetaDraw/MetaDraw.xaml
@@ -152,15 +152,8 @@
 
                                 <Grid Grid.Row="0" Background="Gray" Opacity=".65"></Grid>
                                 <!--Settings menu-->
-                                <Button x:Name="Settings" Grid.Row="0" Grid.Column="1" Click="settings_Click" Background="White" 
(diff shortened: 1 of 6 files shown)


Alexander-Sol:
Don't know if it's related to this PR, but after accessing the Chimeric Scan View tab in MetaDraw, a significant lag occurs. Decreased performance continues if other tabs are accessed before the chimeric scan view tab loads

nbollis:
> Don't know if it's related to this PR, but after accessing the Chimeric Scan View tab in MetaDraw, a significant lag occurs. Decreased performance continues if other tabs are accessed before the chimeric scan view tab loads

This is because I do not load the chimera groups until they are needed. No reason to spend the time parsing out that data until a user decides to actually visualize it. This will be accelerated in an upcoming PR (after these PRs) that creates a separate big picture tab for chimeric annotations.

Alexander-Sol (approved):

trishorts (approved):

RayMSMS (approved):

PR #2536, opened 2025-07-17
Spectral library reading for MS2PIP outputs

zhuoxinshi:
ReadLibrarySpectrum in SpectralLibrary is reading in molecular weight of peptides as precursor MZ (which is fine for some types of spectral library but does not work with MS2PIP output). This PR changed the reader by skipping the line with molecular weight which is not used for constructing library spectra. Also all empty lines are removed when reading in library spectra.


MetaMorpheus/EngineLayer/SpectralLibrarySearch/SpectralLibrary.cs
```diff
@@ -189,6 +189,10 @@ private LibrarySpectrum ReadLibrarySpectrum(StreamReader reader, bool onlyReadHe
             while (reader.Peek() > 0)
             {
                 string line = reader.ReadLine();
+                if (!line.IsNotNullOrEmpty())
```
nbollis, line 192: You should change this to line.IsNullOrEmpty() to make it easier to read

MetaMorpheus/EngineLayer/SpectralLibrarySearch/SpectralLibrary.cs
```diff
@@ -213,10 +217,7 @@ private LibrarySpectrum ReadLibrarySpectrum(StreamReader reader, bool onlyReadHe
                 }
                 else if (line.StartsWith("MW", StringComparison.InvariantCultureIgnoreCase))
                 {
-                    split = line.Split(mwSplit);
-
-                    // get precursor m/z
-                    precursorMz = double.Parse(split[1].Trim(), CultureInfo.InvariantCulture);
+                    continue;
```
nbollis, line 220: So we do use precursor mz to construct the library spectrum below. With this change, precursor mz will always be set to zero. Is this the desired approach?

nbollis (changes requested):

PR #2537, opened 2025-07-18
Deconvolution Negative mode in GUI

nbollis:
This PR changes a little bit of how the deconvolution control works. We can now set the polarity directly.  One other thing was done, and that is enable custom fragmentation window to do RNA once it is prepared.


MetaMorpheus/GUI/Util/IntegerTextBoxControl.cs
```diff
@@ -25,6 +25,14 @@ protected override void OnPreviewTextInput(TextCompositionEventArgs e)
             {
                 if (!char.IsDigit(character))
                 {
+                    if (character == '-')
```
Alexander-Sol: This control is used a whole bunch, and most of the time it assumes/requires a positive integer. Can we make a new control instead?

nbollis: Good catch. I made it so you have to explicitly set AllowNegative="True" in the xaml for the integer control to allow it.

Alexander-Sol (approved):

RayMSMS (approved):

PR #2542, opened 2025-07-24
GPTMD Revisions

nbollis:
This PR makes 2 foundational changes

## 1. We only add the mod to the new database if it is the best scoring mod and localization for a given delta mass
This changes from previous behavior where we would add one mod to database per best localization of each mod matching that delta mass. 

## 2. Custom filters allow the user to define the stringency with which to add a modification with built filters including
Previous behavior: Add best scoring per mod without consideration for unmodified score. 
- Score Improvement
- Unidirectional coverage - one fragment contains mod
- Bidirectional coverage - one fragment from either terminus contains mod
- Flanking - one fragment immediately to the left and right of modified residue

### DIFF
--- a/.github/workflows/BuildReleaseArtifact.yml
+++ b/.github/workflows/BuildReleaseArtifact.yml
@@ -1,0 +1,105 @@
+name: Release — Patch Version & Build MSI
+
+on:
+  release:
+    types: [released, edited]
+
+jobs:
+  patch-build-upload:
+    runs-on: windows-latest
+    timeout-minutes: 60
+
+    steps:
+      - name: Checkout release tag
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.release.tag_name }}
+
+      - name: Set up .NET
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: 8.0.x
+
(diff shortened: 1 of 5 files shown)


trishorts (dismissed):
please consider copilot suggestions

MetaMorpheus/GUI/TaskWindows/GPTMDTaskWindow.xaml.cs
```diff
@@ -240,6 +248,28 @@ private void PopulateChoices()
...
+            FilterOptions.Add(new GptmdFilterViewModel(
+                new UniDirectionalIonCoverageFilter(),
+                "Uni-Directional Ion Coverage",
+                "Requires the mod site to be covered by at least one N-terminal or one C-terminal ion. That is, ions from one direction must include the mod, even if not flanking it."
+            ));
+            FilterOptions.Add(new GptmdFilterViewModel(
+                new FlankingIonCoverageFilter(),
+                "Flanking Ion Coverage",
+                "Requires flanking ions — a fragment from *before* and one from *after* the mod site, regardless of fragmentation direction."
+            ));
```
Alexander-Sol, line 272: The tool-tip should specify that this option won't add any terminal mods. That way we don't forget to adjust that filtering method

PR #2545, opened 2025-07-29
MetaDraw auto loaded with search results 

nbollis:


trishorts (dismissed):
this pr is awesome. i took it for the test drive. nice work. 

do we need some way to let the old users know that their results are ready and waiting?

RayMSMS (dismissed):

PR #2550, opened 2025-08-02
MetaDraw: Data Visualization Tab Updates

nbollis:
This PR overhauls the data visualization tab in `MetaDraw`

## Adding 4 new plots
- Histogram of Missed Cleavages
- Histogram of Fragment Ion Types by Count
- Histogram of Fragment Ion Types by Intensity
- Histogram of Ids by Retention Time

## Enabled more customization
- Only plot those that pass filter (previously we plotted all values)
- Normalize within file to have the y axis be total instead of counts
- Allow user to customize the colors in the histograms
- Allow user to reorder results in the histgram
- Created new tab in `MetaDrawSettingsWindow` for data visualization settings 

## Additional changes made to support this PR
- Generalized many of the color settings view models to share a base class
- Added many of the filtering properties `MetaDrawSettingsViewModel` instead of setting them from the code behind
- Added Upper and Lower bound dependency properties to `DoubleTextBoxControl` so that the QvalueFilter is bounded between 0 and 1. 

## Other miscellaneous changes
- Added a try catch in constructing chimera groups to prevent silly crashes when decon results do not align with search results

### DIFF
--- a/MetaMorpheus/EngineLayer/Util/PathSafety.cs
+++ b/MetaMorpheus/EngineLayer/Util/PathSafety.cs
@@ -1,0 +1,152 @@
+﻿using System;
+using System.IO;
+using System.Linq;
+
+namespace EngineLayer.Util
+{
+    public static class PathSafety
+    {
+        // Conservative defaults for broad Windows tooling compatibility.
+        // Increase if your environment supports long paths (\\?\).
+        public const int DefaultMaxPath = 260;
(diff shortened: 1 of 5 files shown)


trishorts (approved):

trishorts (dismissed):

Alexander-Sol (approved):
Great PR, very slick UI. 10/10

PR #2553, opened 2025-08-12
Add Precursor and PrecursorSet classes for m/z tolerance–based precursor management

Alexander-Sol:
**Summary**
This PR introduces two new classes to the `EngineLayer` namespace that provide a standardized way to represent precursor features and manage collections of unique precursors. These additions will be used in upcoming workflows that require deduplication of precursor lists (e.g., external decon pairing, MRM selected precursor handling).

**Key Changes**

* **`Precursor` class**

  * Immutable representation of a precursor feature, storing:

    * `MonoisotopicPeakMz`
    * `Mass`
    * `Charge`
    * `Intensity`
    * `EnvelopePeakCount`
    * Optional `FractionalIntensity`
  * Constructor enforces initialization of all required fields.

* **`PrecursorSet` class**

  * Maintains a collection of non-overlapping precursors, defined by a user-specified m/z tolerance.
  * Supports adding new precursors only if no equivalent precursor (within m/z tolerance) already exists in the set.
  * Implements `IEnumerable<Precursor>` for easy iteration.
  * Provides `Add` and `ContainsEquivalent` methods for controlled set membership.

**Motivation**
Currently, there is no unified way in MetaMorpheus to store and deduplicate precursors with m/z-based uniqueness rules. These classes encapsulate that logic in a reusable, testable form, enabling future features like:

* Efficient merging of precursor lists from multiple sources.
* Tolerance-aware filtering for external deconvolution results.
* MRM precursor table handling without duplicate entries.

**Next Steps**

* Integrate `PrecursorSet` into the `GetMs2Scans` workflow for per-scan precursor list construction.
* Use `Precursor` and `PrecursorSet` in the new MRM and external decon features currently in development.


MetaMorpheus/EngineLayer/Util/Precursor.cs
```diff
@@ -0,0 +1,102 @@
...
+/// Each precursor refers to one isotopic envelope observed in a single MS1 scan. 
+/// It is meant to hold experimentally observed data for a peptide/proteoform precursor.
+/// </summary>
+public record Precursor
+{
+    /// <summary>
+    /// Each precursor refers to one isotopic envelope observed in a single MS1 scan. 
+    /// It is meant to hold experimentally observed data for a peptide/proteoform precursor.
+    /// </summary>
+    /// <param name="MonoisotopicPeakMz"></param>
```
trishorts, line 20: good pr. some clarification of definitions in the comments. Since this is experimental, does MonoisotopicPeakMz refer to the lowest observed above threshold? or the maximum isotope? or some imputed isotope based on an envelope that is matched to it during deconvolution

MetaMorpheus/EngineLayer/Util/Precursor.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// <param name="Charge"></param>
+    /// <param name="Mass"></param>
```
trishorts: assume NeutralMass not mass with charges? this requires a deconvolution I guess since this is experimental data.

MetaMorpheus/EngineLayer/Util/Precursor.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// <param name="Intensity">Either the most abundant isotope intensity, or the summed intensity from all isotopes. For MM decon, this is determined by CommonParameters </param>
```
trishorts, line 23: but only one charge state? I could see wanting to know which intensity is being referred to. not sure future folks can tell.

MetaMorpheus/EngineLayer/Util/Precursor.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// <param name="EnvelopePeakCount"> The number of peaks observed in the Isotopic Envelope. 
+    /// The minimum intensity for a peak to be considered is specified by the user in decon parameters, or in arguments passed to external software</param>
+    /// <param name="FractionalIntensity"> The fraction of the intensity in the MS1 scan that is accounted for by this precursor</param>
```
trishorts, line 26: I see this is limited to the isolation window. but what about this precursor in other charge states. can charge state 3 have a fractional intensity of 0.2 and charge state 4 have a fractional intensity of 0.5? or is it 0.7 (summed value) for both charge states i guess it is not fraction in the window. also, does the whole envelope have to be inside the window? certainly some chimeras are on the envelope boundary.

MetaMorpheus/EngineLayer/Util/Precursor.cs
```diff
@@ -0,0 +1,102 @@
...
+{
+    /// <summary>
+    /// Each precursor refers to one isotopic envelope observed in a single MS1 scan. 
+    /// It is meant to hold experimentally observed data for a peptide/proteoform precursor.
+    /// </summary>
+    /// <param name="MonoisotopicPeakMz"></param>
+    /// <param name="Charge"></param>
+    /// <param name="Mass"></param>
+    /// <param name="Intensity">Either the most abundant isotope intensity, or the summed intensity from all isotopes. For MM decon, this is determined by CommonParameters </param>
+    /// <param name="EnvelopePeakCount"> The number of peaks observed in the Isotopic Envelope.
```
trishorts: do these isotopes have to be consecutive. if we see five isotopes but one in the middle is corrupted by some noise or another ion? what's the count

MetaMorpheus/EngineLayer/Util/PrecursorSet.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// The m/z tolerance for determining if two precursors are considered equivalent.
+        /// Must be positive.
+        /// </param>
+        public PrecursorSet(Tolerance tolerance)
+        {
+            Tolerance = tolerance;
+            PrecursorDictionary = new Dictionary<int, List<Precursor>>();
+        }
+
+        public bool IsDirty { get; private set; } = false;
```
trishorts, line 47: ?

MetaMorpheus/EngineLayer/Util/Precursor.cs
nbollis: It is simply the count of the peaks that are in the `IsotopicEnvelope` class. Ideally, it would not spit out any corrupted peaks

MetaMorpheus/EngineLayer/Util/Precursor.cs
nbollis: It is based upon the envelope that is matched during deconvolution

MetaMorpheus/EngineLayer/Util/Precursor.cs
nbollis: This is the neutral monoisotopic mass, will name to reflect that.

nbollis (dismissed):

PR #2563, opened 2025-08-27
Adjust notch reporting

nbollis:
Notch FDR has been updated. They will be written out as doubles to the spectral match results. 

In the case where we have an ambiguous PSM which is ambiguous between multiple notches, we will calculate a q-value notch for each hypothesis. We can then use this information to disambiguate results during post-search analysis. 

This was done by creating nullable double fields that will only be set when we have a notch ambiguous PSM. FDRAnalysisEngine was updated to make many of its function static, allowing other engines to set q-values. 

Future work should move other sources of disambiguation to this new engine. 

Also. We now delete database index directories at the end of test runs.

### DIFF
--- a/MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
+++ b/MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
@@ -240,7 +240,6 @@ namespace EngineLayer.FdrAnalysis
                 //the QValueThreshodOverride condition here can be problematic in unit tests. 
                 if (psms.Count < PsmCountThresholdForInvertedQvalue && !QvalueThresholdOverride)
                 {
-
                    QValueTraditional(psms, peptideLevelAnalysis: peptideLevelCalculation);
                 }
                 else
--- a/MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
+++ b/MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
@@ -158,8 +158,8 @@ namespace EngineLayer.FdrAnalysis
             double cumulativeDecoy = 0;
 
             //set up arrays for local FDRs
-            double[] cumulativeTargetPerNotch = new double[MassDiffAcceptorNumNotches + 1];
-            double[] cumulativeDecoyPerNotch = new double[MassDiffAcceptorNumNotches + 1];
+            Dictionary<int, double> cumulativeTargetPerNotch = psms.Select(p => p.Notch).Distinct().ToDictionary(p => p!.Value, _ => 0.0);
+            Dictionary<int, double> cumulativeDecoyPerNotch = psms.Select(p => p.Notch).Distinct().ToDictionary(p => p!.Value, _ => 0.0);
 
             //Assign FDR values to PSMs
             foreach (var psm in psms)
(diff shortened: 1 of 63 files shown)


pcruzparri (approved):

RayMSMS (approved):

MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
```diff
@@ -152,54 +155,82 @@ private void DoFalseDiscoveryRateAnalysis(FdrAnalysisResults myAnalysisResults)
...
-                    cumulativeTargetPerNotch[notch]++;
+                    int notch = group.Key;
+                    double groupTargetHits = group.Count(h => !h.IsDecoy);
+                    double groupDecoyHits = group.Count(h => h.IsDecoy);
+
+                    totalTargetHits += groupTargetHits;
+                    totalDecoyHits += groupDecoyHits;
+
+                    cumulativeTargetPerNotch[notch] += groupTargetHits / totalHits;
+                    cumulativeDecoyPerNotch[notch] += groupDecoyHits / totalHits;
```
Alexander-Sol, line 198: Should the denominator here be groupHits?

MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
```diff
@@ -254,20 +321,45 @@ private void QValueInverted(List<SpectralMatch> psms, bool peptideLevelAnalysis)
...
+                            }
+                        }
+                    }
+                }
+                else
+                {
+                    int notch = psms[i].Notch.Value;
+                    qValueNotch[notch] = Math.Min(qValueNotch[notch], (psms[i].GetFdrInfo(peptideLevelAnalysis).CumulativeDecoyNotch + 1) / Math.Max(psms[i].GetFdrInfo(peptideLevelAnalysis).CumulativeTargetNotch, 1));
+                    psms[i].GetFdrInfo(peptideLevelAnalysis).QValueNotch = Math.Min(qValueNotch[notch], 1);
+                }
```
Alexander-Sol, line 362: Does this duplicate lines 269-308? If so, should we break this out into a function?

MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
nbollis: No. Total hits refers to all hits for this given PSM. For each notch, we are adding:
 
cumulativeTargetNotch += Targets in Notch / Ambiguous Result Count for PSM
cumulativeDecoyNotch += Decoys in Notch / Ambiguous Result Count for PSM

MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
nbollis: No. One is reversed from the other. Notice the Math.Min in the section you annotated versus the Max in the section you quoted. 

` int notch = psms[i].Notch.Value; `

` qValueNotch[notch] = Math.Max(qValueNotch[notch], psms[i].GetFdrInfo(peptideLevelAnalysis).CumulativeDecoyNotch / Math.Max(psms[i].GetFdrInfo(peptideLevelAnalysis).CumulativeTargetNotch, 1)); `

` psms[i].GetFdrInfo(peptideLevelAnalysis).QValueNotch = Math.Min(qValueNotch[notch], 1); `

Alexander-Sol (approved):

Issue #1098, opened 2018-04-13
Peptide Output: Reclaim PSM information.


zrolfs:
We currently show the highest scoring scan for a unique peptide. Have a separate column that includes all scans the peptide was identified in (not just the highest scoring) as well as how many PSMs it was found in.

rmillikin:
isn't this in the PSMs file?

zrolfs:
It is present in the PSMs file as separate row entries. It would be convenient to have it all in one place when looking at the peptide list, however.

rmillikin:
see #752

PR #1084, opened 2018-04-07
Toml null fix

zrolfs:

Issue #969, opened 2018-02-22
Nightly automated testing


rmillikin:
Jenkins, our software for running automated tests, should build and run the MetaMorpheus master branch nightly for a battery of tests. This way we will catch bugs early so we won't be bogged down when it comes time to check to see if a version is OK to release.

I would also like to get a daily morning email about the status of the Jenkins test after it has completed or not, with any issues, number of PSMs before/after calibration, etc.

rmillikin:
Done!

PR #2572, opened 2025-10-10
New variant stuff

trishorts:


acesnik (commented):
🤖 The naming this fixes is worth fixing — `MaxHeterozygousVariants` describes a genotype when what the parameter actually bounds is combinatorial expansion, and `MinVariantDepth` says variant where it means allele. Splitting the max into "variants per isoform" and "how many isoforms" is also the right distinction to draw, because those are two different limits that the single old parameter was conflating.

The problem is what it depends on, and I think that has to be settled before the code is worth reading line by line.

## The mzLib API this targets was abandoned

The branch pins `mzLib 9.9.998`, a local build, and needs three parameters where mzLib master offers two: `VariantApplication.GetVariantBioPolymers(int maxAllowedVariantsForCombinatorics = 4, int minAlleleDepth = 1)`. There is no `MaxSequenceVariantsPerIsoform` and no separate isoform count anywhere in mzLib production code.

That is not a matter of waiting for a release. mzLib #967 was Part 4 of "Update Handling of Sequence Variants" and merged in February, but **Parts 5 through 8 — #968, #969, #970, #972 — and #974 were all closed unmerged on 2026-08-27**. So the API surface this PR was written against is not coming from that series. What did land instead took a different route: the variant test corpus (#1084, #1098) and the one-decoy-per-target fix (#1154).

**Half of it landed anyway, which is the good news.** mzLib's own parameter is already `minAlleleDepth`, so renaming MetaMorpheus's `MinVariantDepth` to match upstream is *better* justified now than when this opened — that half is a straight alignment against a name that exists. It is the max-parameter split that has nothing to bind to.

## The defaults change, and that changes variant-search output

Worth calling out prominently because it is easy to read as a rename: `MaxHeterozygousVariants = 4` becomes `MaxSequenceVariantsPerIsoform = 0`, and `MinVariantDepth = 1` becomes `MinAlleleDepth = 0`, with a new `MaxSequenceVariantIsoforms = 1`. mzLib's defaults are 4 and 1. So every variant search would get different bounds than today, and 0 for a combinatorics limit reads ambiguously — unlimited, or none? Whichever it means, it wants saying in the description, and it is the kind of change where a before/after count on a real variant database would settle more than any amount of review.

## Two things I checked that are *not* problems

Both looked alarming in the diff and both are staleness rather than loss, so I want to say so rather than have you chase them.

`Test/DIATests/TestXicGrouping.cs` appears to delete `TestCorrelationCalculationWithSpline`. It does not — master moved that test to `TestPfPairAndPfGroup.cs`, and this branch still has it in the old file. Present in both.

`PopulateCountBasedOccupancy` is on master and absent here, which reads like a revert of the count columns. It is not: the merge base (2025-10-10) had zero references, and master gained it via #2475 merged 2026-08-27, ten months after this branched. A rebase restores it, as it does the `VariantCallFormatDataString` migration from #2620 that #2587 was separately proposing.

## What I would do

Decide whether the three-parameter model is still the intended direction now that Parts 5-8 are closed. If it is, the mzLib side needs reopening or rewriting first and this rebases onto it. If it is not, the `MinAlleleDepth` rename plus the tooltip and GUI label changes are independently worth having and would make a small, mergeable PR against the current two-parameter API — and the 806 lines added to `GPTMDengineTest` are probably worth keeping either way.

Commenting rather than requesting changes: the branch cannot restore at `9.9.998`, so it cannot merge regardless, and the open question is a direction decision rather than a defect in the code.

## Not verified

I ran nothing — the pin makes that impossible, and `Test/Test.csproj` is `net*.0-windows` in any case. I have not read the `+108/-161` rewrite of the variant-reporting block in `PostSearchAnalysisTask` or the `+129/-34` in `MetaMorpheusTask` line by line, deliberately: both are written against the abandoned API, so their shape may change entirely depending on the answer above.

PR #2576, opened 2025-10-10
Database Loading Engine 

nbollis:
## 90% of the files changed are due to moving DbForTask from TaskLayer to EngineLayer

Create a new engine that handles database loading.
All methods called are public and static so that we can use each piece individually if needed. 

This will report values in the results.txt and ensure we have better crash reports.
image

Added Decoy Identifier to `GlobalVariables` and allow user to set their own custom decoy identifier in the global params text file, or for a specific search. 
image
image


Alexander-Sol (changes requested):
image

Gui crashes when underscore character ('_') is entered into the "Decoy Identifier" field

nbollis:
> image
> 
> Gui crashes when underscore character ('_') is entered into the "Decoy Identifier" field

Good Catch. This has been fixed.

RayMSMS (dismissed):

Issue #968, opened 2018-02-22
warning for incorrect version of msfilereader when loading .raw files


trishorts:
We could warn people that some users experience instability/crashes when loading thermo raw files with msfilereader versions other than 3.0 sp2

rmillikin:
This is has been a consistent problem over the entire lifespan of MetaMorpheus and I'd really like to get a robust solution in place. I think it should be one of our top priorities.

This issue already exists though (#876), so I'm going to close this. Sheeran will be working to solve this problem but I suspect he will need help from all of us to ensure this crash is captured from all angles

PR #2587, opened 2025-11-12
Accomodate variant call format description. This PR is a complement to mzLib PR 967 Add VariantCallFormat to SequenceVariation

trishorts:


acesnik (commented):
🤖 This is already on master, so I would close it.

The change here is the seven `variant.Description.X` → `variant.VariantCallFormatDataString.X` call sites in `PostSearchAnalysisTask`, adapting to mzLib #967. That landed as part of `b0172ac6`, "Update mzLib to 573 (#2620)" — master's `PostSearchAnalysisTask.cs` now uses `VariantCallFormatDataString` at exactly seven sites and `variant.Description` at none, so the adaptation arrived bundled with the pin bump that made it necessary, which is the sensible place for it.

The remaining difference between this branch's copy of that file and master's is all master being ahead — `using TaskLayer.MbrAnalysis`, `SpectralRecoveryResults`, and some formatting drift. There is nothing here master lacks.

The pin is also `mzLib 9.9.999`, a local build, so the branch cannot restore as it stands.

**Reopening signal**, so this is not a guess: if any `variant.Description` use survives on master where it should have migrated, reopen and name the file — a grep for `variant\.Description` across `TaskLayer` and `EngineLayer` is the quickest check, and it comes back empty for me today.

**Leaving the branch in place**, unlike the other stale PRs closed today — `accomodateVariantCallFormatDescription` stays on the fork, so if this turns out to be needed after all there is nothing to restore first, just reopen.

Worth noting for context rather than as a criticism of this PR: mzLib #967 was Part 4 of the "Update Handling of Sequence Variants" series, and Parts 5 through 8 (mzLib #968, #969, #970, #972) plus #974 were all closed unmerged on 2026-08-27. So Part 4 is the only piece of that series in the library, and this was its MetaMorpheus counterpart. That bears directly on #2572, which I have commented on separately.

Issue #1212, opened 2018-06-08
add spectrum annotation viewer


trishorts:

trishorts:
maybe one of you java people can have a look at this as a non-integrated solution. https://github.com/PRIDE-Toolsuite/pride-inspector

rmillikin:
my preference would be to format our mzID (and in the future pepXML) for import into someone else's viewer rather than making our own

Issue #1210, opened 2018-06-07
new scoring modes to access low-res data


trishorts:
let's try andromeda 1.0 and 2.0 scoring to see if we can get those low res scans done right. we can review the asms talk by Petra Gutenbrunner to see the actual procedure.

rmillikin:
see #300 
I suspect weighting intensity more will mean more IDs for low-res data while probably decreasing very slightly high-res data PSMs

andromeda 2.0 seems very similar to andromeda 1.0 except with fragment ion intensity prediction, which seems to (currently) only work well for unmodified trypsin-digest peptides. I'm not a fan of limiting our scoring algorithm to that subset of peptides, or having two separate scoring algorithms, etc, though I can see the merits of it

PR #1209, opened 2018-06-06
Rearrange results order

zrolfs:
It was looking messy with the PSMs, MS2s, and fragmented species all together instead of separated like peptides and proteins.

PR #2596, opened 2025-11-21
Tmt post processing

trishorts:
Used Alex TMT PR as a start.
Adds new TMT Post Processing Engine to process TMT data


trishorts:
Closing on architectural grounds rather than staleness — the staleness is a symptom.

`TmtPostProcessingEngine` is mzLib's `Quantification` project rebuilt inside `EngineLayer`. Its five private methods map one-to-one onto work that has since shipped in mzLib:

| this PR | mzLib |
|---|---|
| `NormalizeToReference` | `ReferenceChannelNormalization` |
| `SumPsmsForPeptideQuant` / `SumPeptidesForProteinQuant` | `SumRollUp` / `AggregatingRollUp` (mzLib #1211) |
| `SelectPsmForProteinQuant` | the unambiguous-match filter (mzLib #1210), `GetAllPeptideToProteinMap` (mzLib #1209) |
| `NormalizeLoading` | the normalization strategy set in `mzLib/Quantification/Strategies` |

That makes it a second quantification engine competing with the one the `Quantification` project exists to establish — the same situation as mzLib #984, which was closed for the same reason. MetaMorpheus #2755 does the intended thing instead: it calls mzLib's `QuantificationEngine` rather than reimplementing it.

Two other reasons not to revive this branch as it stands. `EngineLayer/Util/IsobaricMassTag.cs` appears here as a **new file** — that type reached master by another route and has evolved since, so this PR would take it backwards. And the five methods are `private static void` taking no parameters, i.e. a sketch rather than working code.

### Recorded before closing, because it is not in the code

The design choices are worth more than the implementation, and they should be checked against mzLib's strategies rather than lost:

- **"PAW Pipeline sums PSMs for peptide quantification"** and **"sums peptides for protein quantification"**. Naming a published method is the useful part — sum rather than mean at both roll-up levels is a deliberate choice, and mzLib now has `SumRollUp` alongside `MedianRollUp` and `Top3` (#1211, #1212), so the question "which is right for isobaric data, and why" is live and answerable there.
- **Reference-channel normalization ordering.** This engine ran loading normalization *before* reference normalization. Whether mzLib's strategy composition preserves that order is worth confirming.
- **`TmtProcessingParams`** encoded the option surface. mzLib #2598's `Sample Type` column (`study sample` / `reference` / `bridge` / `carrier`) and mzLib's `QuantificationParameters` should between them cover it; anything in `TmtProcessingParams` that neither expresses is a real gap and should become an issue rather than disappearing with this branch.

Reopen if any of the above turns out not to be covered.

PR #2598, opened 2025-11-24
tmt experiment design

trishorts:
This pull request adds support for handling TMT (Tandem Mass Tag) experimental design files in addition to the existing classic experimental design files. It introduces a new file format (`TmtDesign.txt`) for TMT experiments, ensures mutual exclusivity between classic and TMT design files, and provides robust validation and user feedback for both design types. The GUI is also updated to allow users to set TMT experimental design files. Below are the most important changes grouped by theme:

**Support for TMT Experimental Design Files:**

* Added a new static class `TmtExperimentalDesign` with methods for reading, validating, and writing TMT design files (`TmtDesign.txt`). This includes validation for file consistency, uniqueness of sample/replicate/fraction/tech combinations, and annotation handling. (`MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs`)
* Introduced a new global variable `TmtExperimentalDesignFileName` with the default value `"TmtDesign.txt"` to `GlobalVariables`. (`MetaMorpheus/EngineLayer/GlobalVariables.cs`) [[1]](diffhunk://#diff-2f5beae5761bd49a2e2c6a46f4aa84be6d23fad67b36e5816cb9804439e379cdR66) [[2]](diffhunk://#diff-2f5beae5761bd49a2e2c6a46f4aa84be6d23fad67b36e5816cb9804439e379cdR79)

**Command-line and Engine Logic:**

* Updated the main command-line logic to detect and handle both classic and TMT design files, enforce that only one design file type is present, and provide user feedback and error codes for various scenarios (missing files, both files present, parse errors, etc.). (`MetaMorpheus/CMD/Program.cs`) [[1]](diffhunk://#diff-c4ac059cbc68195f58233f80debf71b5c668f35491788b07e5c85a6d7411e206L252-R288) [[2]](diffhunk://#diff-c4ac059cbc68195f58233f80debf71b5c668f35491788b07e5c85a6d7411e206R306-R312) [[3]](diffhunk://#diff-c4ac059cbc68195f58233f80debf71b5c668f35491788b07e5c85a6d7411e206L309) [[4]](diffhunk://#diff-c4ac059cbc68195f58233f80debf71b5c668f35491788b07e5c85a6d7411e206R335-R379)

**GUI Enhancements:**

* Added a new button `SetTmtExperimentalDesignButton` to the main window for setting TMT experimental design files. (`MetaMorpheus/GUI/MainWindow.xaml`)
* On adding new spectra files, the GUI now attempts to seed the TMT design state from any design files found next to the raw files. (`MetaMorpheus/GUI/MainWindow.xaml.cs`)

**Minor UI Adjustments:**

* Adjusted the width of the "File-Specific Parameters" column in the spectra files data grid for improved layout. (`MetaMorpheus/GUI/MainWindow.xaml`)


trishorts:
Moved to draft, and pushed two additions that turn this from a file MetaMorpheus can read into a design mzLib's `QuantificationEngine` can run on.

## `Sample Type` — a ninth column

Says what each channel is for: `study sample`, `reference`, `bridge`, `carrier`, `empty`. Without it there was no way to mark a pooled reference channel, which left mzLib's `ReferenceChannelNormalization` — the only isobaric-aware normalization it ships — with nothing to normalize against.

The spellings come from SDRF-Proteomics' `characteristics[sample type]` rather than being invented here, so a design read from an SDRF file later maps through the same switch. That is also why it is an enum rather than the single bool mzLib consumes: `IsobaricQuantSampleInfo.IsReferenceChannel` only needs to know whether a channel normalizes the others, but the file is worth keeping more faithful than that.

**The column is optional on read.** Requiring it would invalidate every `TmtDesign.txt` written before now, and a file without it is unambiguous — every channel is a study sample. The row-width check now measures against the required columns rather than the full header, which is what a pre-existing row looks like.

## `ToMzLibDesign` — the projection

```csharp
public static IExperimentalDesign ToMzLibDesign(
    IEnumerable<TmtFileInfo> files, IsobaricMassTag tag, out List<string> errors)
```

Two contracts here are load-bearing, and getting either wrong mislabels every channel silently rather than failing:

- **Order.** The `ISampleInfo[]` is emitted in ascending reporter m/z — the order `IsobaricMassTag.GetReporterIonIntensities` fills `ISpectralMatch.Intensities` — *not* the order rows appeared in the design file. mzLib maps the two positionally.
- **Length.** Every channel of the tag gets an entry whether or not the design annotates it, because an intensity is reported for every channel. Skipping an unannotated one would shift every later channel by one; it becomes an `Empty` channel instead.

Plex is free text in the file and an `int` on `IsobaricQuantSampleInfo`, so ids are assigned by sorted plex name and are stable run to run. A channel that is not part of the tag is an error and the whole design is refused, rather than returning something half usable.

## Drive-by fix worth a look

`AnnotatePlexWindow` held a **second copy of the channel-label table**, and the two had drifted:

| Tag | AnnotatePlexWindow | IsobaricMassTag |
|---|---|---|
| TMT10 | … `130C`, **`131`** | … `130C`, **`131N`** |
| iTRAQ8 | … `119`, **`121`** | … `119`, **`120`** |
| diLeu12 | `115`–`126` | `115a`–`118d` |

So a label chosen in that window matched no channel of the tag, and a design written through the GUI could not be projected at all. It now calls the one table, and a test asserts every tag's label count matches its reporter ion count.

## Tests

`TmtSampleTypeAndMzLibDesignTests`, 31 tests — parsing and round-tripping every Sample Type spelling, a legacy design file with no Sample Type column still loading, channel ordering against tag order rather than file order, unannotated channels becoming `Empty` rather than being skipped, off-tag channels refused, stable plex ids, and keying by file name with extension. 150 tests pass across this file, `TmtExperimentalDesignTests`, `ResolveExperimentalDesignTests` and the multiplex suite.

## Still open before this leaves draft

- The GUI collects the design but nothing calls `TmtExperimentalDesign.Write`, so the file is never saved from the window.
- Nothing calls `ToMzLibDesign` yet — that is the `PostSearchAnalysisTask` wiring, and it needs mzLib #1193 (which makes the engine deliver its results) to land first.
- Worth deciding whether `TmtDesign.txt` should exist long-term at all: mzLib #1138/#1174 landed a full SDRF-Proteomics reader/writer/validator, and SDRF covers 7 of these 8 columns directly plus the reference-channel concept. Notes in the project write-up; the short version is that the format is a carrier and `IExperimentalDesign` is the interface, so a second carrier is a second `ToMzLibDesign`, not a rewrite.

acesnik (commented):
🤖 Read through this focusing on the projection onto mzLib and the design parser. The code is careful and the doc comments on `ToMzLibDesign` — the order and length contracts in particular — are the kind that stop someone breaking it later. Four things, one of which I think is a real bug.

## The iTRAQ 8-plex channel is labelled `120`, and the reagent's channel is `121`

`IsobaricMassTag.GetReporterIonLabels(iTRAQ8)` ends `..., "119", "120"`. iTRAQ 8-plex has no 120 channel — it is skipped because of the phenylalanine immonium ion at 120.081, and the eighth reagent is 121.

The m/z is right; only the name is wrong. Computing the eighth diagnostic ion from `Mods/tmt.txt`, `C{13}6 N{15}2 H12`, and adding a proton as `ToMz(1)` does:

```
computed 121.1215   published iTRAQ 8-plex 121.1220
```

So MetaMorpheus extracts the correct ion and calls it by a name that does not exist on the kit.

That matters for this PR specifically, because the drive-by consolidation went the wrong way for this tag. `AnnotatePlexWindow` had `121`, which was right, and now calls `GetReporterIonLabels`, which says `120`. A user annotating an 8-plex design will look for the channel their kit calls 121, not find it, and either pick the wrong row or be stuck. Aligning the two tables was the right instinct; this one wants `IsobaricMassTag` corrected to `121` rather than the window changed to match it. The `# 120` comment on that `DI` line in `Mods/tmt.txt` looks like where the name originally came from.

TMT10 `131` → `131N` and diLeu12 `115`–`126` → `115a`–`118d` both look like genuine improvements.

## The label-to-m/z correspondence is not tested, only the count

`EveryTagsLabelsMatchItsReporterIonCount` asserts `labels.Count == ReporterIonMzs.Length`. A count cannot catch the above, and the count is the weaker half of the invariant: `ToMzLibDesign` pairs `channelLabels[i]` with `ReporterIonMzs[i]`, so what has to hold is that label *i* names the channel whose m/z is at *i*.

That is assertable, since every label starts with its nominal mass:

```csharp
// label "127N" -> 127 must equal round(reporter m/z)
int nominal = int.Parse(new string(label.TakeWhile(char.IsDigit).ToArray()));
Assert.AreEqual(nominal, (int)Math.Round(tag.ReporterIonMzs[i]));
```

I ran that by hand over the tags defined in `Mods/tmt.txt`: iTRAQ-4plex, DiLeu-4plex and DiLeu-12plex all pass, and iTRAQ-8plex is the only failure — so it catches the bug above and does not fire spuriously. I did not check the TMT tags that way, because their `DI` blocks are laid out differently in that file, so worth confirming they pass before committing to it.

## Nothing calls `Write` or `ToMzLibDesign` outside tests

Both are still test-only, which matches what you listed as open when this went to draft. Since it has come back out of draft, is the intent to land the format and validation now and wire them up when mzLib #1193 is in, or was the wiring meant to be here? Either is reasonable; it changes what "TMT support" means for whoever reads the release notes, since as merged the design file is read and validated but does not yet reach quantification.

## A design whose paths do not resolve reads as success

In `Read`, a row is kept only if its `File` cell resolves into `fullFilePathsWithExtension`, and rows that do not match are skipped without comment. `Path.GetFullPath` resolves relative to the working directory rather than to the design file, so a `TmtDesign.txt` written with bare file names and sitting beside the raw files matches only when the process happens to be running in that directory. When it does not match, `Read` returns no errors and an empty annotation set, and the command line prints "Read TmtDesign.txt successfully".

Given how carefully the rest of this refuses ambiguity — refusing both design files, refusing off-tag channels, refusing duplicate file names — a design that matched no file at all seems worth at least a warning.

## Things I checked that are fine

- The order and length contracts in `ToMzLibDesign` do what the comment says: tag order rather than file order, and every tag channel present so an unannotated one becomes `Empty` rather than shifting the rest.
- The optional `Sample Type` column is handled correctly on read — `minimumCells` measures the required columns, so a pre-existing file with eight cells still parses, and column order is free because indices are resolved by name.
- Deleting the design file on a parse error at verbosity `none` reads alarmingly, but master already does exactly that; this preserves it faithfully. Worth knowing it now applies to `TmtDesign.txt` as well.
- `codecov/patch` is red at 85.86%, all of it in error branches.

trishorts:
@acesnik — all four addressed in `2a8ff48`, plus tests for the coverage gap.

## The iTRAQ 8-plex label — you were right, and this PR had made it worse

Corrected `IsobaricMassTag` to `121` rather than changing the window back, as you suggested. Your arithmetic checks out: `H12 C{13}6 N{15}2` is 120.1143, plus a proton is **121.1215**, against a published 121.122. So the extracted ion was always the right one and only the name was wrong — and the `# 120` comment on that `DI` line in `Mods/tmt.txt` is where it came from, exactly as you guessed.

Worth stating plainly: the consolidation in this PR pointed `AnnotatePlexWindow` — which had the correct `121` — at the table that says `120`. Aligning them was right; I aligned them to the wrong one.

## The correspondence test — added, and I confirmed the TMT family

Your open question was whether the TMT tags pass, since their `DI` blocks are laid out differently in `tmt.txt`. **They do.** All eight tag types pass with the fix in place.

And it catches the bug it was written for — reverting the label produces:

```
iTRAQ8 label 120 at index 7 names channel 120, but the reporter ion at that index is 121.1215 (channel 121)
```

## The silent-success path — fixed, and the underlying cause too

Two changes rather than one, because the error alone would have left the real problem in place:

- A design whose rows match no file in the run is now an error naming how many rows were involved, instead of returning empty with no errors.
- A bare file name now resolves against **the design files own directory** before the working directory. Your point about `Path.GetFullPath` was the actual cause: a `TmtDesign.txt` written beside its raw files matched only when the process happened to be launched from that folder, so the same design worked or silently matched nothing depending on how MetaMorpheus was started.

Three tests: the no-match error, bare names resolving with the working directory deliberately set elsewhere, and rooted paths still taken as written.

## `Write` and `ToMzLibDesign` being test-only — your read is correct

The intent is what you guessed: land the format and validation now, wire it up separately. `ToMzLibDesign` is called by **#2755**, which is the PR that runs mzLibs `QuantificationEngine` — it was blocked on an mzLib release and is now unblocked, since 1.0.587 shipped this afternoon and master is on it.

So as merged, this reads and validates the design without it reaching quantification, and you are right that the release notes should say that rather than "TMT support". #2755 is what closes the gap.

## Coverage

Patch coverage was 85.85% — 17 missing and 15 partials in `TmtExperimentalDesign.cs`. Added tests for the validation paths that had none: a truncated row reported by line number, a non-numeric biological replicate, and one file carrying conflicting plex and fraction assignments.

76 TMT and isobaric tests green locally.

MetaMorpheus/GUI/Views/PlexModels.cs
nbollis, line 1: Nonblocking: Ideally, the TmtExperimentalDesign would serve as the model for the UI so that the UI share the same internal state as what search and quant do.

RayMSMS (dismissed):

MetaMorpheus/GUI/Views/PlexModels.cs
trishorts: Agreed, and it is more than cosmetic — it is the same defect this project keeps finding: two representations of one fact with nothing checking they agree. `PlexModels` and `TmtExperimentalDesign` are exactly that, and the split already has two live consequences on this branch:

- `AnnotatePlexWindow.SaveButton_Click` opens `StreamWriter(path, false)` — truncating — and writes only the selected plex, rather than going through `TmtExperimentalDesign.Write`. There is no path through the GUI to a working two-plex experiment.
- The same hand-rolled writer omits the Sample Type column, so `IsReferenceChannel` is always false for a GUI-authored design.

Both are what "the UI has its own copy of the design" costs, so your comment is the root and those two are symptoms.

I would rather not do it inside this PR, for reasons that are about landing it rather than about disagreeing:

1. This is 3,300 lines with two new WPF windows at 0/3. #2596 sat 77 days and #2602 sat 233, both smaller than this. Adding a model refactor makes that worse.
2. It is genuinely GUI work, and it deserves a reviewer looking at GUI code rather than finding it in the middle of a parser diff.

So I am splitting this PR three ways — design model + parser + tests (headless), CMD wiring, then GUI — and taking your comment as the defining piece of the GUI one: make `TmtExperimentalDesign` the model, which then fixes the truncating writer and the dropped Sample Type as a consequence rather than as two separate patches.

Flagging one thing I would rather you knew than discovered: that means the GUI PR is the one where the design actually becomes usable for multi-plex, so it should not be treated as the cosmetic tail of the three.

trishorts:
Superseded by a three-way split, as discussed in the `PlexModels.cs` thread:

| PR | scope | lines |
|---|---|---|
| **#2774** | design model + parser + tests, headless | ~2,170 |
| **#2775** | CMD wiring *(draft, stacked on #2774)* | ~190 + tests |
| **#2776** | GUI *(draft, stacked on #2774)* | ~1,000 |

#2774 is the one to review first; the other two show its commits until it merges, then rebase to their own files.

Two things did not survive the split, both deliberately:

- `ResolveExperimentalDesignTests` moved to #2775, because it exercises `CMD.Program` and would not compile without it. The compiler drew that boundary rather than me.
- A comment-only edit to `Multiplex_Labeling_TMT_iTRAQ.cs` was dropped — #2771 landed a fuller version of the same comment, so carrying it would have put pure churn in front of a reviewer.

Also resolved this branch's conflicts with master before splitting (`9a1102e5d`), so the partition was taken from a current tree rather than a stale one. Both conflicts were comment-only, from #2771 landing in `IsobaricMassTag.cs`.

Leaving this open until #2774 merges, so nothing is lost if the split needs revisiting — but it should not be reviewed in preference to the three.

acesnik (commented):
🤖 Now a draft, and worth writing down why so nobody has to reconstruct it.

**This is superseded, not abandoned.** Its 14 files are exactly #2774 plus #2775 plus #2776, minus one comment-only edit to `Multiplex_Labeling_TMT_iTRAQ.cs` that #2771 made redundant. I checked that as a set operation rather than by eye: nothing in the three is missing from here, and nothing here is missing from the three except that one file. So there is no work in this branch that the split lost, which is the thing a safety net is for.

**The four findings from my review of this PR were all addressed here**, and I verified them at this head rather than reading the reply. The iTRAQ 8-plex label is `121` in `IsobaricMassTag`, corrected at the table rather than by pointing `AnnotatePlexWindow` back at the wrong value. The correspondence between a label and the reporter ion at its index is now asserted rather than just the count, and the TMT family passes it — which was the open half of that note. And the silent-success path is fixed at the cause: `ResolveAgainstDesignFile` tries the design file's own directory before the process's working directory, so a `TmtDesign.txt` sitting beside its raw files no longer works or silently matches nothing depending on where MetaMorpheus was launched from.

**If the split ever does need revisiting, take #2774 and not this.** `TmtExperimentalDesign.cs` is 638 lines here against 715 there, because #2774 took three more commits after the partition — including `b1a269ef`, "Address review: two silent data losses, a rejected valid design". So this branch is behind the live one by a fix that is itself about silent data loss, and #2755, #2775 and #2776 all share this older copy for the same reason. That is an argument for rebasing them onto a merged #2774 rather than merging their own copies.

One loose thread: @Alexander-Sol is still the requested reviewer here, on a draft that should not be reviewed, while his approval on #2774 was dismissed by the push that fixed my last finding there. #2774 is at 2 of 3 and that request would do more good pointed at it.

Not verified by me, unchanged: `Test/Test.csproj` is `net10.0-windows`, so the 76-green figure is CI's rather than mine.

trishorts:
Closing this in favour of the three PRs it was split into on 2026-08-29.

- **#2774** — the model and its parser. **Merged 2026-09-02** as `d233c17b`.
- **#2775** — the CMD wiring. Open, updated onto the merged model, ready for review.
- **#2776** — the GUI. Open, updated onto the merged model, ready for review.

Together #2775 and #2776 reproduce this PR's file list exactly, minus the comment-only edit to `Multiplex_Labeling_TMT_iTRAQ.cs` that #2771 made redundant, so nothing here is lost by closing rather than merging.

Closing is also the safer end for it: this branch still carries a copy of `TmtExperimentalDesign.cs` that is 715 lines behind what merged, without the per-plex uniqueness key or the report when one channel is described twice with different samples. Merging it, or resolving a conflict toward it, would revert both — which is exactly what @acesnik flagged when he asked for this to be closed once #2774 landed.

@nbollis — your review comment about `TmtExperimentalDesign` being the model behind the UI has been **taken**, not just carried across: `AnnotatePlexWindow.SaveButton_Click` in #2776 now builds the whole design and writes it through `TmtExperimentalDesign.Write`. That fixed the truncating writer and the dropped `Sample Type` column together.

Issue #1213, opened 2018-06-08
consider default max peptide length. 


trishorts:
there are some whoppers in human canonical 

![image](https://user-images.githubusercontent.com/16841846/41182017-8e208b32-6b39-11e8-806d-2413e1756768.png)

zrolfs:
Default max of 50 (or 4500 Da) is pretty common in other search software.

rmillikin:
50 is fine with me. I tried to implement that a while ago and stefan vetoed it. I think these large-mass species are probably prone to deconvolution errors. though I suspect the scores would be quite poor. are you observing any high-scoring, very long peptides in your samples?

zrolfs:
There are a dozen or so high scoring 50-60 mer PSMs when digestion efficiency is low or a non-tryptic digest is used, but typically we don't observe things that long in bottom-up.

However...
There aren't any apparent benefits to using a max length. There's a tiny decrease in search time, but it's unlikely to improve the results, as I haven't observed a long, high scoring decoy.

trishorts:
I'm sheepish about  setting the max. I do see some longer identified peptides.

rmillikin:
So is the consensus to leave it as-is, with no upper limit to peptide length?

PR #2602, opened 2025-12-18
Calibration TMT

zhuoxinshi:


trishorts:
@zhuoxinshi closing this as stale — last updated 2026-01-06, no reviews, and GitHub can no longer compute whether it merges. Please reopen or re-push if it is still live; nothing here is a judgement on the work, and one part of it is worth keeping.

### What is being carried forward

The `CF` lines this PR adds to `Mods/tmt.txt` are the valuable part, and they fill a real gap. **Neither MetaMorpheus's `EngineLayer/Mods/tmt.txt` nor mzLib's `Omics/Resources/TMT.txt` currently contains a single `CF` line** — every TMT entry carries only `MM`, a bare monoisotopic mass. That is an odd asymmetry, because the reporter ions in the same entries are already formulas (`DI HCD:C8 N1 H15`) while the label itself is only a number, and a bare mass cannot give you an isotopic distribution.

These are being recorded on mzLib #1213, which covers moving the isobaric channel data into mzLib. The destination is mzLib's `Omics/Resources/TMT.txt` rather than MetaMorpheus's copy, because #2749 removes the MetaMorpheus copy in favour of the mzLib package.

### One correction, and please check me on it

The TMTpro formula is right; the TMT6/10/11 one appears to be one hydrogen heavy. Checked against the `MM` value already in the same entry:

```
C8 13C7 H25 N 15N2 O3  = 304.207145   vs MM 304.2071    diff +0.000045   ok
C8 13C4 H21 N 15N  O2  = 230.170757   vs MM 229.162932  diff +1.007825   <-- one H
C8 13C4 H20 N 15N  O2  = 229.162932   vs MM 229.162932  diff  0.000000
```

`H20` reproduces the recorded mass exactly. If you have a vendor source saying otherwise I would rather have that than my arithmetic — the entry disagreeing with itself is the part worth resolving either way.

Nothing checks a `CF` line against the `MM` line in the same entry today, which is why this would have gone unnoticed. A test asserting they agree across every entry is being added with the harvest.

### Two other notes for whenever this work resumes

- `IsobaricMassTag.GenerateIntensityFile` hardcodes `.Take(16)` on the label list. That silently truncates TMT18 and would break outright if TMTpro-16 lands (tracked in #1213). If the method comes back, the channel count should come from the tag.
- **TMT calibration itself is now unrepresented.** Closing this drops the capability, not just the branch. If low-resolution TMT calibration is still needed, it deserves an issue of its own so the requirement outlives the PR — happy to open one, or reopen this.

PR #2573, opened 2025-10-10
XicGrouping part2

zhuoxinshi:
This pull request introduces significant enhancements and new features to the DIA (Data Independent Acquisition) analysis engine, particularly focusing on improved precursor-fragment grouping algorithms and support for ISD (In-Source Decay) data. The main changes include the addition of a new `ISDEngine` class, integration of DIA-Umpire inspired grouping logic, expanded configuration options, and new correlation/overlap calculation methods. These updates provide more flexibility and accuracy in DIA data analysis workflows.

**New Engines and Grouping Algorithms:**

* Added a new `ISDEngine` class that inherits from `DIAEngine`, enabling ISD data analysis using a workflow similar to DIA, including ISD-specific scan grouping, relabeling, and pseudo-MS2 scan generation. (`MetaMorpheus/EngineLayer/DIA/ISDEngine.cs`)
* Introduced the `UmpirePfGroupingEngine` class, implementing precursor-fragment grouping algorithms based on DIA-Umpire (Nat Methods 2015) with custom overlap and correlation calculations. (`MetaMorpheus/EngineLayer/DIA/PrecursorFragmentGrouping/UmpirePfGroupingEngine.cs`)

**Precursor-Fragment Grouping Enhancements:**

* Refactored `XicGroupingEngine` to support configurable precursor and fragment rank thresholds, improved parallelization, and post-filtering of groups based on these ranks. (`MetaMorpheus/EngineLayer/DIA/PrecursorFragmentGrouping/XicGroupingEngine.cs`, `PfGroupingEngine.cs`) [[1]](diffhunk://#diff-d7940d5faa438193fe92572239f7c8297999ee1758855427d457f6e4550ed3dbR19-R47) [[2]](diffhunk://#diff-d7940d5faa438193fe92572239f7c8297999ee1758855427d457f6e4550ed3dbL43-R85) [[3]](diffhunk://#diff-d7940d5faa438193fe92572239f7c8297999ee1758855427d457f6e4550ed3dbL81-R109) [[4]](diffhunk://#diff-ca291e6c9c153bdceaf638bc5f78582657940dc48b3c6116f0514575a167244cR15-R39)
* Added static method `FilterPfPairsByRank` in `PfGroupingEngine` to filter precursor-fragment pairs within groups according to rank thresholds. (`MetaMorpheus/EngineLayer/DIA/PrecursorFragmentGrouping/PfGroupingEngine.cs`)

**Correlation and Overlap Calculation Methods:**

[...]

PR #2546, opened 2025-07-29
MetaMorpheusDIA

zhuoxinshi:
This is the first part of incorporating DIA analysis into MetaMorpheus. The DIA workflow being implemented here reconstructs DIA MS2 scans into pseudo Ms2WithSpecificMass that can be searched directly by our DDA search engines. This includes the following steps: 
1. Reading in raw scans and dividing MS2 scans with different isolation windows into different DIA window groups.
2. Constructing XICs for all peaks in the MS1 scans and MS2 scans in the same window.
3. Grouping the precursor XICs with fragment XICs in the same window into precursor-fragments groups based on their correlations.
4. Converting precursor-fragments groups into pseudo Ms2WithSpecificMass.

The DIAEngine class contains this complete workflow and outputs the final list of pseudo Ms2WithSpecificMass. Inside DIAEngine, the process of constructing XICs and grouping precursors and fragments are represented by abstract classes XicConstructor and PfGroupingEngine that can take in different strategies.

Issue #1027, opened 2018-03-10
Warning for attempting open-mass search using Classic Search mode


rmillikin:

rmillikin:
done in #1318

PR #2626, opened 2026-02-25
Dia

trishorts:


acesnik:
What's the story behind this one? DIA would be great to get in MM.

acesnik:
Comparing PTM localization performance in DIA with other tools would be interesting

acesnik (commented):
🤖 Reviewed the MetaMorpheus side of this while the architecture questions on mzLib #1027 are open, since most of what is here does not depend on how those land.

Same scale correction as over there, because it is putting people off: the header says +205,905 and **96% of that is three fixtures** — `decoy.msp` (75,065), `target.msp` (74,485) and an mzML (48,068). The reviewable content is about **2,900 lines of production code and 2,460 of tests**, which is an ordinary PR.

## One thing that needs fixing whatever the architecture decisions turn out to be

**This deletes `Test/DIATests/DIAEngineTest.cs` (411 lines) while keeping all 13 files of the `EngineLayer/DIA/` production code it tests.** That file covered `DIAEngine.ConstructMs2Groups`, both XIC constructors and their spline paths, plus the empty-scan `MetaMorpheusException` case. At this head:

```
ConstructMs2Groups          referenced in 0 test files
NeutralMassXicConstructor   referenced in 0 test files
MzPeakXicConstructor        referenced in 0 test files
```

The two surviving files in `Test/DIATests/` do not close it — `TestPfPairAndPfGroup` covers correlation, overlap ratio and ranking, `TestXicGrouping` covers PF and Umpire grouping. Neither touches scan-window construction or XIC construction. So merging as-is leaves three landed, in-use classes from #2546 and #2573 with no coverage at all, having had coverage before, in code this PR otherwise does not touch.

Given the branch is six months old and `dirty`, I read this as a merge resolution rather than a decision — which is the good news, since restoring the file is the whole fix. Worth checking the rest of the resolution while you are in there: `MetaMorpheusTask.cs` also comes out of the three-dot diff having lost a `using EngineLayer.SpectrumMatch;` that master carries, which has the same shape.

## Verified and good

`MyTask` gains `DiaSearch` properly, and the TOML side is genuinely done rather than assumed — `MetaMorpheusDiaSearchParameters` is primitives plus one enum, so it needs no converter in `tomlConfig`, and there are two round-trip tests (`MetaMorpheusDiaSearchParameters_TomlRoundTrip`, `DiaSearchTask_TomlRoundTrip`) that go through `MetaMorpheusTask.tomlConfig` rather than round-tripping in isolation. That is the failure mode I went looking for and it is covered.

The task structure mirrors the existing one — `DiaSearchTask` plus `PostDiaSearchAnalysisTask` and its parameters object, the same split as `SearchTask`/`PostSearchAnalysisTask` — which is what makes it feel like part of the codebase rather than beside it.

`DiaMetaDrawShim` earns its name and its remarks are the best thing in the PR: DIA has no single representative scan, evidence is accumulated across many MS2 scans as XICs, and MetaDraw's fragment panel wants a `List<MatchedFragmentIon>`, so one is synthesised from the library ion identities and the extracted intensities — with zero-signal fragments dropped so MetaDraw does not annotate ghost peaks. That last detail is the kind that only turns up when you look at the output.

`FdrAnalysisEngineDia` states why it is separate rather than leaving a reader to wonder: DIA results map one-to-one onto a library entry with a single continuous score, no notch ambiguity and no multi-protease grouping, so the DDA engine's machinery does not apply. Worth noting for whoever writes the release text that this path produces q-values but no PEP, since `PEPAnalysisEngine` is not in it — reasonable, and different from what a DDA user sees in the same columns.

## Deferred to #1027 rather than raised here

The two-subsystem question — `EngineLayer/DIA/` (landed, library-free) beside `EngineLayer/DiaSearch/` (this, library-based) — is a product decision, and I have put it on #1027 with the rest. The one thing I would say here is that those two directory names are a coin-flip apart, and whichever way coexistence goes, someone will open the wrong one. `DiaLibrarySearch` or similar would cost nothing now.

## Litter, all trivial

`MetaMorpheus/Program.cs` is a **0-byte file at the repo root**. `MetaMorpheusFileStructure.txt` is a 740-line directory dump. And `Development/Dia` adds ~2,090 lines plus its own `DiaDevBenchmarks.csproj` — same question as `Development/` on #1027, which is whether benchmarks belong in the shipped tree or on a branch.

## Not verified

`Test/Test.csproj` is `net*.0-windows`, so I ran nothing — the six new test files are CI's to confirm. The GUI half is unrunnable here by construction; `DiaSearchTaskWindow.xaml` would compile under `EnableWindowsTargeting` but I have not built it, and layout and enable/disable behaviour are unchecked either way. And I have not exercised the DIA path end to end, so nothing here is a claim about results.

Issue #1038, opened 2018-03-21
What is the difference between the .exe and .msi releases?


zrolfs:
.msi is a nice installer, .exe is giving me weird issues.

rmillikin:
Should be the same. I'll try out the .exe tonight on another machine

lonelu:
.exe seems like not working on certain computer.  Also, if .exe exist, people won't use .msi

rmillikin:
I've just nixed the .exe installer and recommended users use the .msi.

PR #2631, opened 2026-03-25
Circular Task and Engine and GUI

trishorts:
> **Project of origin:** `CircularPeptideSearch`


MetaMorpheus/EngineLayer/CircularSearch/CircularSearchEngine.cs
```diff
@@ -0,0 +1,294 @@
...
+    /// SCORING
+    /// -------
+    /// Scoring is identical to <see cref="ClassicSearch.ClassicSearchEngine"/>:
+    /// matched fragment ions are counted and a PSM score is computed.
+    ///
+    /// THREAD SAFETY
+    /// -------------
+    /// One lock object per scan index, matching the ClassicSearchEngine pattern.
+    /// </summary>
+    public class CircularSearchEngine : MetaMorpheusEngine
```
nbollis, line 46: If you make this inherit from classic search engine instead of metamorheus engine, you can reuse many of the functions that are rewritten here while retaining the separate circular search engine class.

MetaMorpheus/GUI/MainWindow.xaml
nbollis, line 1: We lost all of the comments :(

MetaMorpheus/GuiFunctions/GuiFunctions.csproj
```diff
@@ -16,7 +16,7 @@
   <ItemGroup>
     <PackageReference Include="itext7" Version="8.0.5" />
     <PackageReference Include="itext7.bouncy-castle-adapter" Version="8.0.5" />
-    <PackageReference Include="mzLib" Version="1.0.574" />
```
nbollis: Nuspec adjustment before final merge

MetaMorpheus/TaskLayer/CircularSearchTask/CircularSearchParameters.cs
```diff
@@ -0,0 +1,112 @@
...
+{
+    /// <summary>
+    /// Parameters governing the circular peptide search task.
+    ///
+    /// Kept intentionally minimal — the search honours the same
+    /// <see cref="CommonParameters"/> (protease, missed cleavages, tolerances,
+    /// dissociation type, fixed/variable mods) as the classic search.
+    /// Circular-specific knobs live here.
+    /// </summary>
+    public class CircularSearchParameters
```
nbollis, line 17: Does this class add any new functionality? Can you simply use SearchParameters in the circular search engine?

MetaMorpheus/TaskLayer/CircularSearchTask/CircularSearchTask.cs
```diff
@@ -0,0 +1,379 @@
...
+    ///    and populates a PSM array.
+    /// 4. Perform FDR calculation and write output files.
+    ///
+    /// OUTPUT FILES
+    /// ------------
+    /// • CircularPeptideSpectralMatches.psmtsv
+    /// • CircularPeptides.psmtsv  (unique peptide-level results)
+    /// • (Optional) .mzID
+    /// </summary>
+    public class CircularSearchTask : MetaMorpheusTask
```
nbollis, line 43: I think you can also make this inherit from search task to reuse a bunch of the methods.

MetaMorpheus/EngineLayer/CircularSearch/CircularSearchEngine.cs
```diff
@@ -0,0 +1,327 @@
...
+                    }
+
+                    for (int i = range.Item1; i < range.Item2; i++)
+                    {
+                        if (GlobalVariables.StopLoops) return;
+
+                        CircularProtein protein = CircularProteins[i];
+
+                        foreach (var bioPolymer in protein.Digest(
+                                     CommonParameters.DigestionParams,
```
trishorts, line 150: **Major** - Fragment cache skipped when dissociation type changes between scans

**Problem:**
The fragment generation logic uses `theorProducts.Count == 0` as a sentinel to decide whether fragments have already been generated for a given dissociation type. However, `theorProducts` is shared across all scans for a given `bioPolymer` via the `fragmentsPerDissociationType` dictionary, and lists are only cleared once per `bioPolymer` (line 139). If the first scan for a dissociation type produces zero fragments (e.g., `FragmentInternally` yields nothing for a short peptide), subsequent scans with the same dissociation type will re-invoke the fragmentation method on every iteration, because `Count` remains 0.

**Impact:**
Repeated fragmentation calls cause unnecessary CPU work proportional to the number of matching scans. While unlikely to produce incorrect results (re-fragmenting yields the same empty list), it is wasted computation that can degrade search performance on large datasets.

**Suggested fix:**
Use a `HashSet<DissociationType>` to track which types have been computed, rather than relying on `Count == 0` as a proxy for "not yet generated."

**Affected test scenarios:**
- CircularSearchEngine - short circular peptide with no internal fragments - should not re-fragment per scan
- CircularSearchEngine - multiple scans same dissociation type - fragmentation called once

---

MetaMorpheus/EngineLayer/CircularSearch/CircularSearchEngine.cs
```diff
@@ -0,0 +1,327 @@
...
+                    if (CommonParameters.DissociationType == DissociationType.Autodetect)
+                    {
+                        foreach (var kv in GlobalVariables.AllSupportedDissociationTypes
+                                     .Where(p => p.Value != DissociationType.Autodetect))
+                            fragmentsPerDissociationType[kv.Value] = new List<Product>();
+                    }
+                    else
+                    {
+                        fragmentsPerDissociationType[CommonParameters.DissociationType] =
+                            new List<Product>();
```
trishorts, line 140: **Major** - Fragment lists cleared per-peptide but not per-dissociation-type transition

**Problem:**
When `DissociationType` is `Autodetect`, the `fragmentsPerDissociationType` dictionary holds lists for multiple dissociation types. The lists are cleared at the start of each `bioPolymer` loop (line 139–140), which is correct. However, within a single `bioPolymer`, if scans arrive in mixed dissociation-type order (e.g., HCD scan, then ETD scan, then another HCD scan), the HCD fragment list will already be populated when the second HCD scan arrives — this is fine. But consider the case where a `bioPolymer` is circular: on first HCD encounter, internal fragments are generated. If a second dissociation type (ETD) is then encountered, ETD fragments are generated. Both lists coexist correctly. The subtle issue is that fragment generation for a given type is only triggered once (when `Count == 0`), but if `Autodetect` maps multiple `TheScan.DissociationType` values to the same key, fragments would only be generated for the first encountered type. This is actually handled correctly by keying on the actual dissociation type, but the coupling between "clear all" and "generate on first use" is fragile and undocumented.

**Impact:**
Currently functional but a maintenance hazard. Any future change to clearing logic or fragment caching could introduce silent scoring bugs that are hard to diagnose.

**Suggested fix:**
Add a comment documenting the invariant: lists are cleared once per peptide, and each dissociation type's list is populated lazily exactly once per peptide.

**Affected test scenarios:**
- CircularSearchEngine - Autodetect with mixed dissociation types - correct fragments per type

---

MetaMorpheus/EngineLayer/CircularSearch/CircularSearchEngine.cs
```diff
@@ -0,0 +1,327 @@
...
+                                        ? theScan.TheScan.DissociationType!.Value
+                                        : CommonParameters.DissociationType;
+
+                                if (!fragmentsPerDissociationType.TryGetValue(
+                                        dissociationType, out var theorProducts))
+                                    continue;
+
+                                // Generate fragments on first encounter for this type
+                                if (computedDissociationTypes.Add(dissociationType))
+                                {
```
trishorts, line 197: **Major** - Race condition on `proteinsSearched` counter in parallel loop

**Problem:**
`proteinsSearched` is a `double` variable captured by the parallel lambda and incremented with `proteinsSearched++` (line 196) without synchronization. The `++` operator on `double` is not atomic, and concurrent increments from multiple threads can lose updates. Additionally, `oldPercentProgress` (line 197) is also read and written from multiple threads without synchronization.

**Impact:**
Progress reporting may be inaccurate (skipped updates or duplicated reports). While this does not affect search correctness, in extreme cases the torn read/write on `double` could produce a corrupted value. The same pattern exists in ClassicSearchEngine, so this may be a known accepted trade-off, but it is still technically a data race.

**Suggested fix:**
Use `Interlocked.Increment` on an `int` counter instead of `++` on a captured `double`. For `oldPercentProgress`, either use `Interlocked.Exchange` or accept the benign race with a comment.

**Affected test scenarios:**
- CircularSearchEngine - large protein list with max parallelism - progress reported monotonically

---

MetaMorpheus/TaskLayer/CircularSearchTask/CircularSearchTask.cs
```diff
@@ -0,0 +1,412 @@
...
+                        databaseFilePath: target.DatabaseFilePath);
+
+                    circularProteins.Add(decoy);
+                }
+            }
+
+            // ── Mass-difference acceptor ──────────────────────────────────────
+            var massDiffAcceptor = GetMassDiffAcceptor(
+                CommonParameters.PrecursorMassTolerance,
+                CircularSearchParameters.MassDiffAcceptorType,
```
trishorts, line 171: **Major** - Integer division truncates progress percentage to zero

**Problem:**
The expression `completedFiles / currentRawFileList.Count` performs integer division because both operands are `int`. For any number of files greater than 1, the intermediate progress values will truncate to 0 (e.g., 1/3 = 0, 2/3 = 0). Only the final iteration yields 1. `ReportProgress` likely expects a percentage (0–100), but even if it expects a fraction (0.0–1.0), the int division still produces wrong values.

**Impact:**
The progress bar will remain stuck at 0% for the entire search and then jump to 100% on the last file. Users processing many files will see no meaningful progress feedback.

**Suggested fix:**
Cast one operand to `double` and multiply by 100 if the API expects a percentage: `(int)(completedFiles * 100.0 / currentRawFileList.Count)`. Verify what `ProgressEventArgs` expects (percentage vs fraction) and match accordingly.

**Affected test scenarios:**
- CircularSearchTask_MultipleFiles - 3 spectra files - progress reports should show incremental values (33, 66, 100)
- CircularSearchTask_SingleFile - 1 spectra file - progress should report 100

---

trishorts (commented):
Automated code review - 12 finding(s) at severity >= Major.

MetaMorpheus/TaskLayer/CircularSearchTask/CircularSearchTask.cs
```diff
@@ -0,0 +1,412 @@
...
+                taskId,
+                DecoyType.None,
+                CircularSearchParameters.SearchTarget,
+                localizeableModificationTypes,
+                TargetContaminantAmbiguity.RemoveContaminant);
+
+            var dbResults = (DatabaseLoadingEngineResults)dbLoader.Run();
+            var bioPolymers = dbResults.BioPolymers;
+
+            // Wrap every loaded protein as a CircularProtein.
```
trishorts, line 101: **Major** - Decoy shuffle mutates shared array, producing correlated decoys

**Problem:**
The `char[] arr` is initialized once from `target.BaseSequence` before the retry loop, but it is shuffled in-place on each attempt. If the first shuffle produces a canonical collision (rejected), the second attempt shuffles the already-shuffled array rather than starting from the original sequence. This is not truly a fresh shuffle of the target — it is a shuffle of the previous rejected shuffle. More critically, `arr` is **not reset between different targets**: after breaking out of the inner loop, the outer `foreach` moves to the next target and calls `target.BaseSequence.ToCharArray()` again, so the cross-target concern is fine. However, within the retry loop the bias remains.

**Impact:**
For short sequences where collisions are likelier, retries are not independent shuffles of the original sequence, slightly biasing the decoy distribution. The practical impact is small but violates the stated intent of "randomly shuffling the canonical ring sequence."

**Suggested fix:**
Re-initialize `arr` from `target.BaseSequence.ToCharArray()` at the top of each retry iteration to ensure each attempt is an independent shuffle of the original sequence.

**Affected test scenarios:**
- DecoyGeneration_ShortSequence - 3-4 residue protein - verify decoy distribution is uniform across retries
- DecoyGeneration_CollisionRetry - sequence where first shuffle collides - verify retry produces valid independent shuffle

---

MetaMorpheus/TaskLayer/CircularSearchTask/CircularSearchTask.cs
```diff
@@ -0,0 +1,412 @@
...
+                var rng = new Random(42); // fixed seed for reproducibility
+                foreach (var target in targetProteins)
+                {
+                    string decoySeq = null;
+
+                    for (int attempt = 0; attempt < 10; attempt++)
+                    {
+                        char[] arr = target.BaseSequence.ToCharArray();
+                        // Fisher-Yates shuffle
+                        for (int i = arr.Length - 1; i > 0; i--)
```
trishorts, line 131: **Major** - Pre-loading logic crashes on empty file list

**Problem:**
The pre-loading block at line 130 unconditionally accesses `currentRawFileList[0]` and `fileSettingsList[0]` before the loop. If `currentRawFileList` is empty, this throws an `ArgumentOutOfRangeException`. While the classic search tasks may guard against this upstream, this new task does not validate the input.

**Impact:**
An unhandled exception crashes the task if the user starts a circular search with no spectra files selected. The error message will be cryptic (index out of range) rather than a user-friendly warning.

**Suggested fix:**
Add an early return (with a warning) if `currentRawFileList` is null or empty, before the pre-load block.

**Affected test scenarios:**
- CircularSearchTask_NoSpectraFiles - empty file list - should return gracefully with warning, not throw

---

MetaMorpheus/EngineLayer/GlobalVariables.cs
```diff
@@ -53,7 +53,7 @@ public static class GlobalVariables
         public static GlobalSettings GlobalSettings { get; set; }
         public static IEnumerable<Modification> UnimodDeserialized { get; private set; }
         public static IEnumerable<Modification> UniprotDeseralized { get; private set; }
-        public static UsefulProteomicsDatabases.Generated.obo PsiModDeserialized { get; private set; }
+        public static Omics.Modifications.IO.obo PsiModDeserialized { get; private set; }
         public static IEnumerable<Modification> AllModsKnown { get { return _AllModsKnown.AsEnumerable(); } }
```
trishorts, line 61: **Major** - PsiModDeserialized type changed from fully-qualified to unqualified `obo`

**Problem:**
The property `PsiModDeserialized` was changed from `UsefulProteomicsDatabases.Generated.obo` to just `obo`. This relies on a `using` directive or a different `obo` type being in scope. The new file does not add a `using UsefulProteomicsDatabases.Generated;` directive. The type `obo` must resolve from somewhere — either the new `using Omics.Modifications.IO;` namespace happens to expose it, or the type was moved/re-exported. If the `obo` type resolved by the compiler is different from `UsefulProteomicsDatabases.Generated.obo`, this would be a breaking change in the public API surface of `GlobalVariables`.

**Impact:**
If `obo` does not resolve to the same type as `UsefulProteomicsDatabases.Generated.obo`, any code consuming `GlobalVariables.PsiModDeserialized` (e.g., `Loaders.GetFormalChargesDictionary`) will fail to compile or behave incorrectly at runtime. Even if it compiles, a type mismatch could cause subtle runtime errors.

**Suggested fix:**
Verify that the unqualified `obo` resolves to the exact same type as `UsefulProteomicsDatabases.Generated.obo`. If it does (e.g., because the type was moved to `Omics.Modifications.IO` or re-exported), the change is fine. If not, revert to the fully-qualified name or add the correct `using` directive. Add a comment explaining which namespace provides `obo` if it's non-obvious.

**Affected test scenarios:**
- Any test calling `GlobalVariables.SetUpGlobalVariables()` - should verify PsiModDeserialized is populated correctly
- Any test using `Loaders.GetFormalChargesDictionary(GlobalVariables.PsiModDeserialized)` - should verify type compatibility

MetaMorpheus/GUI/MainWindow.xaml.cs
```diff
@@ -915,12 +855,12 @@ private void RunAllTasks_Click(object sender, RoutedEventArgs e)
                 NotificationHandler(null, new StringEventArgs("You need to add at least one spectra file!", null));
                 return;
             }
-            if (!ProteinDatabases.Any() && PreRunTasks.Any(p => p.metaMorpheusTask.TaskType != MyTask.Average)) // will not throw this if averaging is the only task
+            if (!ProteinDatabases.Any() && PreRunTasks.Any(p => p.metaMorpheusTask.TaskType != MyTask.Average))
             {
```
trishorts, line 850: **Major** - NullReferenceException when defaultTomlName is null in OpenNewTaskWindow

**Problem:**
In the base code, `defaultTomlName` could be null for unrecognized task types, and `Path.Combine` was called unconditionally. The new code adds a `if (defaultTomlName != null)` guard around the file-existence check block, which is good. However, in the base code the same issue existed without the guard — the new code fixes it for the new task type but reveals that the original code had no guard either. Looking more carefully, the new code restructures the flow: it wraps the `File.Exists` check inside `if (defaultTomlName != null)`, but the base code called `Path.Combine(... , defaultTomlName)` unconditionally, which would throw `ArgumentNullException` if an unknown `MyTask` enum value were passed. The new code correctly guards this path for `CircularSearch` but the dialog creation block (`switch` for opening window) now has `dialog` potentially null if `taskType` doesn't match any case, and `dialog.Owner = this` is called unconditionally — though the new code adds `if (dialog == null) return;` which fixes a latent bug from the base.

**Impact:**
No runtime impact from the circular search addition itself. The null guard is actually an improvement over the base code.

**Suggested fix:**
No action needed — this is an improvement. Noting for completeness.

**Affected test scenarios:**
- OpenNewTaskWindow - unknown MyTask enum value - should not throw

MetaMorpheus/GUI/MainWindow.xaml.cs
```diff
@@ -529,20 +514,13 @@ private void AddSpectraFile_Click(object sender, RoutedEventArgs e)
             }
         }
 
-        /// <summary>
-        /// Event fires when a spectra file is selected.
-        /// </summary>
         private void AddSelectedSpectra(object sender, RoutedEventArgs e)
         {
             DataGridRow obj = (DataGridRow)sender;
-
```
trishorts, line 529: **Major** - Missing return after protein database validation allows run without databases

**Problem:**
In `RunAllTasks_Click`, when no protein databases are loaded and a non-Average task exists, the code calls `NotificationHandler` to display an error message but does NOT `return`. This means execution continues and the tasks will run without any protein databases, which will likely cause a crash or produce no results. This bug exists in both the base and new code, but the PR adds `CircularSearch` as a new task type that presumably also requires protein databases.

**Impact:**
A user could accidentally start a CircularSearch (or any non-Average task) without protein databases loaded. The run would proceed and either crash or silently produce empty/wrong results.

**Suggested fix:**
Add `return;` after the notification about needing a protein database, consistent with the other validation checks above it. Also consider whether `CircularSearch` should be excluded from this check like `Average` is.

**Affected test scenarios:**
- RunAllTasks_Click - no databases loaded with CircularSearch task - should block run
- RunAllTasks_Click - no databases loaded with Search task - should block run

MetaMorpheus/GUI/TaskWindows/CircularSearchTaskWindow.xaml.cs
```diff
@@ -0,0 +1,434 @@
...
+                else
+                {
+                    typeNode = new ModTypeForTreeViewModel(mod.Item1, true);
+                    FixedModTypeForTreeViewObservableCollection.Add(typeNode);
+                    typeNode.Children.Add(
+                        new ModForTreeViewModel("UNKNOWN MODIFICATION!", true,
+                            mod.Item2, true, typeNode));
+                }
+            }
+
```
trishorts, line 197: **Major** - Empty MissedCleavages text box causes use of uninitialized variable

**Problem:**
When `MissedCleavagesTextBox.Text` is empty (`""`), the `int.TryParse` call fails and returns `false`, but the condition `&& MissedCleavagesTextBox.Text != ""` makes the overall `if` false, so validation passes. Later at line 236, `missedCleavagesValue` is set to `int.MaxValue` for the empty case, which is correct. However, the variable `missedCleavages` was declared by the `out` parameter of the failed `TryParse` and defaults to `0`. If the text is non-empty but not a valid integer, the method correctly returns. But if the text is a valid negative integer (e.g. `"-1"`), `TryParse` succeeds, validation passes (no `< 0` check), and a negative missed cleavages value is used.

**Impact:**
A user entering a negative number for missed cleavages will not receive a validation error and will pass an invalid value to `DigestionParams`, potentially causing incorrect digestion behavior or downstream exceptions.

**Suggested fix:**
Add a `missedCleavages < 0` check to the validation block, similar to the `maxMods < 0` check on line 202.

**Affected test scenarios:**
- AddTaskButton_Click - MissedCleavages set to "-1" - should show validation error
- AddTaskButton_Click - MissedCleavages set to "0" - should be accepted as valid

---

MetaMorpheus/GUI/TaskWindows/CircularSearchTaskWindow.xaml.cs
```diff
@@ -0,0 +1,434 @@
...
+                    VariableModTypeForTreeViewObservableCollection.Add(typeNode);
+                    typeNode.Children.Add(
+                        new ModForTreeViewModel("UNKNOWN MODIFICATION!", true,
+                            mod.Item2, true, typeNode));
+                }
+            }
+
+            // Sync parent tri-state checkboxes
+            foreach (var node in FixedModTypeForTreeViewObservableCollection)
+                node.VerifyCheckState();
```
trishorts, line 226: **Major** - ProductMassTolerance and MinScore stored redundantly in CommonParameters and CircularSearchParameters

**Problem:**
`AddTaskButton_Click` sets `productMassTolerance` and `scoreCutoff` on `CommonParameters` (line 225-228) and then also sets them on `CircularSearchParameters` (lines 249-250). In `UpdateFieldsFromTask`, `ProductMassTolerance` is read from `CircularSearchParameters` (line 109) while `PrecursorMassTolerance` is read from `CommonParameters` (line 103). If any code path updates one copy but not the other, the values will diverge silently. The `UpdateFieldsFromTask` method reads `ProductMassTolerance` from `CircularSearchParameters` but the engine might use `CommonParameters.ProductMassTolerance`, or vice versa.

**Impact:**
If the two copies get out of sync (e.g., through deserialization or programmatic construction), the UI will display one value while the engine uses another, leading to incorrect search results that are difficult to diagnose.

**Suggested fix:**
Pick a single authoritative source for product tolerance and min score. Either always read/write from `CommonParameters` or always from `CircularSearchParameters`, and have the other derive from it. Document which is authoritative.

**Affected test scenarios:**
- RoundTrip - Create task, serialize, deserialize, verify ProductMassTolerance matches
- UpdateFieldsFromTask - CommonParameters.ProductMassTolerance differs from CircularSearchParameters.ProductMassTolerance - should use consistent source

---

MetaMorpheus/GUI/MainWindow.xaml
```diff
@@ -357,6 +365,7 @@
```
trishorts, line 342: **Major** - About tab lost all task-description bullet points and hyperlinks

**Problem:**
The base `AboutTab` contains detailed bullet-point descriptions of every task type (calibration, averaging, GPTMD, search, XL search, glycopeptide) with academic citation hyperlinks. The new file replaces this with only two short paragraphs, completely removing the `Grid.Row="2"` TextBlock that described XL search and glycopeptide tasks, as well as the inline bullet list in `Grid.Row="1"`. This appears to be an accidental regression from a full-file replacement rather than an intentional simplification.

**Impact:**
Users lose all in-app documentation about what each task type does and lose direct links to the relevant publications. This is a significant UX regression unrelated to the circular search feature.

**Suggested fix:**
Restore the original About tab content from the base branch and, optionally, append a new bullet point for the circular peptide search task with its citation link.

**Affected test scenarios:**
- AboutTab - verify all task descriptions and hyperlinks are present after PR merge
- AboutTab - RNA mode should still collapse XL/glyco descriptions

---

trishorts:
**Major (could not attach inline — targets removed lines)** — Databases tab lost RNA-aware detail text with `ProteinRnaTerminologyConverter`

**File:** `MetaMorpheus/GUI/MainWindow.xaml` (~line 268 in the reviewed snapshot)

**Problem:** The base Databases tab `TextBlock` uses a `Binding` with `ProteinRnaTerminologyConverter` to dynamically swap proteomics/RNA terminology (e.g., explaining PTM locations in .xml vs .fasta). The PR removes these converter bindings; in RNA mode the Databases tab will show protein-specific terminology, and users lose the .xml-vs-.fasta guidance.

**Suggested fix:** Restore the Databases tab `TextBlock` content with its `ProteinRnaTerminologyConverter` binding.

(Posted as a top-level comment because the change is a deletion and has no right-side diff line to anchor an inline comment.)

MetaMorpheus/GUI/MainWindow.xaml
trishorts: Restored — merging current `master` back into the branch brought all of these XAML comments back (69 now, vs 68 on `master`). Resolving.

MetaMorpheus/EngineLayer/CircularSearch/CircularSearchEngine.cs
trishorts: Good call — I dug into the overlap. The reusable part vs `ClassicSearchEngine` is the scan-matching infrastructure: `GetAcceptableScans`, the binary-search `GetFirstScanWithMassOverOrEqual`, the per-scan lock array, and the `AddOrReplace` guard — about 55 lines (~17% of the engine).

A few things make a literal `CircularSearchEngine : ClassicSearchEngine` awkward, though:
- Those helpers are all `private` in `ClassicSearchEngine`, so they aren't inheritable without opening them up.
- `ClassicSearchEngine`'s constructor takes SILAC labels, a `SpectralLibrary`, `writeSpectralLibrary`, etc. that circular search has no use for.
- `RunSpecific` has to be overridden completely anyway — circular does dual terminal **and** internal-ion scoring on the ring-opening linear products, which the code comments explicitly say must not be replicated in Classic.

So the options I see:
1. **Extract a small shared base** `SearchEngineBase : MetaMorpheusEngine` holding the scan-matching infra (the shared fields + `GetAcceptableScans` + `GetFirstScanWithMassOverOrEqual` + the locked `AddOrReplace` pattern as `protected`), and have both `ClassicSearchEngine` and `CircularSearchEngine` inherit it. Kills the duplication with no cross-coupling between the two engines.
2. **Make `ClassicSearchEngine`'s helpers `protected` and inherit directly** (`CircularSearchEngine : ClassicSearchEngine`). Works, but couples circular search to Classic's constructor/SILAC/spectral-library surface.
3. Leave as-is — it's ~55 lines and the two genuinely diverge in fragmentation.

I'm inclined toward (1) since it gives the reuse you're after without dragging the SILAC/spectral-library surface into circular search. Open to your take.

MetaMorpheus/TaskLayer/CircularSearchTask/CircularSearchTask.cs
trishorts: Most of what's duplicated here is actually already reused through the `protected static` helpers on `MetaMorpheusTask` — `GetMs2Scans`, `WritePsmsToTsv`, `LoadModifications`, and `SetAllFileSpecificCommonParams` are all shared, not re-implemented. The async file-load loop looks duplicated but is just the standard pattern around those helpers.

A literal `CircularSearchTask : SearchTask` is awkward because the two have different shapes: `SearchTask` runs multiple engines and delegates all output to `PostSearchAnalysisTask`, whereas circular search runs a single engine and writes its two `.psmtsv` files directly. Inheriting would force a full `RunSpecific` override, so there's little real reuse to gain.

The one piece of genuine duplication is `GetMassDiffAcceptor` + `ParseSearchMode` (which `SearchTask` also has).

Options:
1. **Extract `GetMassDiffAcceptor` + `ParseSearchMode` into a shared static factory** (e.g. `MassDiffAcceptorFactory`) used by both tasks — removes the real duplication without an inheritance relationship that doesn't quite fit.
2. Inherit from `SearchTask` and override `RunSpecific` — minimal practical gain given the architectural mismatch above.
3. Leave as-is.

Leaning toward (1). What do you think?

MetaMorpheus/GuiFunctions/GuiFunctions.csproj
trishorts: Noted. The csprojs currently point at a local mzLib build (`9.9.16`) for the CircularProtein APIs that come from mzLib #1038. Before final merge I'll repoint every `mzLib` PackageReference to the published version that lands when #1038 merges, so CI restores cleanly. Leaving this open as a pre-merge checklist item.

MetaMorpheus/TaskLayer/CircularSearchTask/CircularSearchParameters.cs
trishorts: I compared the two. `SearchParameters` is ~38 properties aimed at the search/quant pipeline (SILAC, label-free quant, spectral-library output, localization, histogram analysis, multiplex, `DecoyType`, etc.). `CircularSearchParameters` has 16, and only **3 are genuinely unique** to circular search:
- `MinScore` — no equivalent in `SearchParameters`.
- `GenerateDecoys` — a simple on/off, vs `SearchParameters.DecoyType` (the circular decoys are always random-shuffle, so the enum doesn't apply).
- `MinInternalFragmentLength` — enforced ≥2 in the setter, whereas `SearchParameters.MinAllowedInternalFragmentLength` defaults to 0 (= disabled).

Everything else already overlaps. So reusing `SearchParameters` would pull ~20 knobs that circular search doesn't use into the task's surface and serialization.

Options:
1. **Keep the dedicated minimal class** (the 3 unique fields justify it). One cleanup I can fold in: `ProductMassTolerance` here is just a serialization mirror — the authoritative copy lives in `CommonParameters` — so I can drop it from `CircularSearchParameters` to remove the redundancy.
2. Switch to `SearchParameters` and leave the ~20 search-only fields unused.

I'm leaning toward (1), keeping it minimal. Good with you?

trishorts:
### Verified end-to-end ✅

Ran the circular search through the **command line** on the bundled small yeast data (`Data/SmallCalibratible_Yeast.mzML` + `Data/SmallYeast.fasta`, 99 proteins):

```
CMD.exe -t CircularSearchTask.toml -s SmallCalibratible_Yeast.mzML -d SmallYeast.fasta -o <out>
```

Full pipeline ran to completion (exit 0, ~0.9 s):

| Stage | Result |
|---|---|
| DatabaseLoadingEngine | 99 proteins loaded |
| CircularSearchEngine | ran to completion (each protein wrapped circular + digested via doubled proxy; dual terminal+internal scoring on ring-opening products) |
| Decoys | shuffle-generated internally (`DecoyType.None` to the loader, by design) |
| FdrAnalysisEngine | **73 PSMs at q ≤ 0.01** |
| Output | `CircularPeptideSpectralMatches.psmtsv` (91 PSMs), `CircularPeptides.psmtsv` (63 unique peptides), `results.txt` — all populated |

Task summary: *"Circular search complete. 91 PSMs found (73 at q≤0.01), 63 unique peptides."*

This exercises the whole chain: CLI task-toml dispatch (the new `CircularSearch` case) → toml round-trip → `CircularProtein` wrapping → `CircularSearchEngine` → internal decoy generation → FDR → output writing.

Also on this branch:
- Full MetaMorpheus test suite green locally (**1525 passed, 1 skipped, 0 failed**).
- New tests added: CMD dispatch/round-trip (`TestCmd`).

Built against a local **mzLib 9.9.17** carrying the CircularProtein localized-PTM fixes (smith-chem-wisc/mzLib#1038). The csproj `mzLib` references will be repointed to the published mzLib version before final merge (so CI restore passes).

acesnik (commented):
🤖 Reviewed alongside mzLib #1038, which this consumes. The circular-search task itself is small and coherent once the fixtures are set aside — ~1,650 lines of production code across the engine, task, parameters and GUI window, with ~2,540 lines of tests over seven files. Two things I liked immediately: `MyTask` gains `CircularSearch` by **appending**, leaving every existing member's value alone, and the new menu item carries a `DataTrigger` on `IsRnaMode` that collapses it in RNA mode rather than offering a protein-only task to someone searching nucleic acids.

The TOML path is genuinely covered, which I nearly recorded as a gap because I looked in the wrong place. `CircularSearchTask_GeneratedTomlDispatchesAndRoundTrips` lives in `TestCmd.cs` rather than under `Test/CircularSearch/`, and it does the real thing: `GenerateDefaultTaskTomls` emits the template, reads it back through `MetaMorpheusTask.tomlConfig`, and asserts both `TaskType` and that the parameters survived. `CircularSearchTaskTests.cs:169` separately exercises `Tolerance` parsing from TOML text, which matters because `ProductMassTolerance` is one of the types that only round-trips because of a converter in that config.

## Two things I would fix before this can be reviewed properly

**Eight paper citations are deleted from `MainWindow.xaml`.** This is the one I would act on first, because it is user-visible and it is the lab's own published work. Measured against the merge base rather than inferred: `3f086fc6` (2026-05-26) had 14 `Hyperlink NavigateUri` entries, master still has 14, and this branch has 6. What goes missing is the task-description help panel and its links — calibration (`acs.jproteome.7b00873`), averaging (`pubmed/38487981`), GPTMD (`acs.jproteome.6b00034`), plus `7b00608`, `8b00141`, `9b00330`, the bioRxiv preprint and one `dx.doi.org` entry.

Worth being precise about what is *not* lost, because the raw diff reads worse than it is: buttons and the tree view are intact — 34 buttons and 3 `TreeView`s on both sides — and most of the −353 in that file is reindentation, including `<LineBreak />` pairs being collapsed onto single lines. Ignoring whitespace the file is +71/−281, so the citation panel is the substance of it.

**All five csprojs pin `mzLib 9.9.17`.** Master pins `1.0.587`, and `9.9.17` is a local build, so CI cannot restore this branch at all — which I suspect is most of why it has sat since June with one comment on it. That is not a criticism of the pin during development; it just means the PR is unreviewable in the ordinary sense until #1038's content ships as a real version and this points at it. If it is useful, the sequence is #1038 → mzLib release → repin here, and I would rather review the engine against a buildable branch than guess.

## Smaller

`EngineLayer.csproj` gains `InternalsVisibleTo("Test")`, which is a whole-assembly internals grant to the test project. Reasonable if the circular tests need to reach internals, but worth saying out loud since it widens what tests can bind to across all of `EngineLayer`, not just the new code — and once it exists, the next test that wants an internal will use it rather than asking whether the member should be public.

The fixtures are 98.6% of this PR: `Test/TestData/my.mgf` is **231,835 lines** and `d Tryp-SFYR MS2 CID35 plus2.mgf` is 84,611. Two notes on those. `my.mgf` is a placeholder name on a 231k-line file, which is the kind of thing that is impossible to reason about later. And the second file is committed **here and in mzLib #1038**, so the same 84,611-line fixture would land twice in two repos; whichever side genuinely needs it, one copy is enough.

One cross-reference rather than a finding: #2626 also edits `MyTask`, and it **reorders** the enum (`Average` and `GlycoSearch` swap) where this one appends. Nothing persists `MyTask` as an integer — I checked, there are no casts on master, and the TOML test compares the name — so it is not a correctness problem, but the two PRs will conflict there and whoever merges second inherits the reorder.

## Not verified

I ran nothing: `Test/Test.csproj` is `net*.0-windows`, and in any case the branch cannot restore against `mzLib 9.9.17`. So the seven test files are CI's to confirm once it builds, and I have not exercised the circular search end to end — nothing here is a claim about results. The GUI window would compile under `EnableWindowsTargeting` but I did not build it, and layout and enable/disable behaviour are unverifiable here either way.

PR #2635, opened 2026-03-31
Support Separate Fragment Mass Tolerance for Child Scans with Lower Resolution

RayMSMS:
**PR Summary**

In glycoproteomics workflows, the primary MS2 scan is often acquired on a high-resolution analyzer (e.g., Orbitrap), while ETD child scans are commonly acquired on a lower-resolution analyzer (e.g., ion trap). To better support this mixed-resolution acquisition, this PR introduces a dedicated low-resolution fragment tolerance parameter:

•	Added ProductMassTolerance_LowRes to CommonParameters
•	Applied low-res tolerance in fragment matching when scan analyzer is ion trap (isLowRes)
•	Kept parent/high-res matching behavior unchanged

Parent and Child

image

**Key Changes**

1. Engine

- Added CommonParameters.ProductMassTolerance_LowRes

- Updated fragment matching (MatchFragmentIons and MatchFragmentIonsOfAllCharges) to use:

  - ProductMassTolerance for high-res scans

  - ProductMassTolerance_LowRes for low-res scans (ion trap)

2. Defaults / compatibility

- If not set, ProductMassTolerance_LowRes defaults to 0.35 Da
- Existing behavior for high-res product tolerance remains unchanged

3. Task/TOML and file-specific flow

- Wired ProductMassTolerance_LowRes through task serialization/deserialization
- Added file-specific TOML support and propagation

4. GUI

- Added low-res tolerance input controls to GlycoSearch task window
- Added validation and save/load handling
- Updated tooltips and layout fixes (including XAML column alignment)

5. Validation

- Extended task settings validation to include low-res tolerance
- Added backward-compatible validator overload to avoid positional-argument regression in existing callers

6. Tests

- Added/updated tests for:
  - default ProductMassTolerance_LowRes behavior
  - Da/ppm low-res tolerance handling
  - TOML read/write and file-specific parsing behavior
  - engine matching behavior with low-res tolerance paths

**Observed Testing Outcome**
When comparing ProductMassTolerance settings (20 ppm vs 0.35 Da) for low-res child-scan matching, the wider low-res tolerance produced more matched c/z ions, including ions with larger mass deviations from theoretical values.
As a control, parent scan matching remained at 20 ppm, and b/y ion matching behavior stayed consistent across conditions.
Testing Result


nbollis (dismissed):

MetaMorpheus/EngineLayer/CommonParameters.cs
```diff
@@ -85,6 +86,7 @@ public CommonParameters(
             MaxThreadsToUsePerFile = maxThreadsToUsePerFile == -1 ? Environment.ProcessorCount > 1 ? Environment.ProcessorCount - 1 : 1 : maxThreadsToUsePerFile;
             ProductMassTolerance = productMassTolerance ?? new PpmTolerance(20);
             PrecursorMassTolerance = precursorMassTolerance ?? new PpmTolerance(5);
+            ProductMassTolerance_LowRes = productMassTolerance_LowRes ?? ProductMassTolerance;
```
trishorts: **Problem:**
In the constructor, `ProductMassTolerance_LowRes` defaults to `ProductMassTolerance` via `productMassTolerance_LowRes ?? ProductMassTolerance`. However, TOML deserialization calls the parameterless constructor first (which chains to the main constructor with all defaults), then sets properties via public setters. At parameterless-constructor time, `ProductMassTolerance` is the 20 ppm default, so `ProductMassTolerance_LowRes` is set to 20 ppm. If the user's TOML sets `ProductMassTolerance` to e.g. 10 ppm but omits `ProductMassTolerance_LowRes`, the low-res tolerance remains at 20 ppm rather than tracking the user's custom `ProductMassTolerance`.

**Impact:**
The PR states "if unset, `ProductMassTolerance_LowRes` defaults to `ProductMassTolerance`." This contract is violated for every TOML-loaded configuration that sets a non-default `ProductMassTolerance` without explicitly also setting `ProductMassTolerance_LowRes`. Users will silently get 20 ppm for low-res matching regardless of their configured product tolerance.

**Suggested fix:**
After TOML deserialization sets all properties, add a post-deserialization check: if `ProductMassTolerance_LowRes` was not explicitly provided in the TOML, reassign it to `ProductMassTolerance`. Alternatively, handle this in `SetAllFileSpecificCommonParams` or by using a sentinel/nullable pattern to detect "not set by user."

**Affected test scenarios:**
- LoadFromToml - ProductMassTolerance set to 10ppm, LowRes omitted - LowRes should equal 10ppm not 20ppm
- RoundTripToml - serialize then deserialize CommonParameters with custom ProductMassTolerance - LowRes should match

MetaMorpheus/EngineLayer/CommonParameters.cs
trishorts, line 146: **Problem:**
The source code comment at line 126 explicitly warns: "If you add a new property here, you must add it to `MetaMorpheusTask.cs/SetAllFileSpecificCommonParams`!! If you forget this, and the user uses file-specific parameters, your settings will be overwritten by default values!" The diff does not show changes to `SetAllFileSpecificCommonParams`, so it is unclear whether this was done. If omitted, `ProductMassTolerance_LowRes` will be reset to the default 20 ppm whenever file-specific parameters are applied.

**Impact:**
Any user employing file-specific parameters would silently lose their configured `ProductMassTolerance_LowRes`, reverting to the default. This would produce incorrect fragment matching tolerances for child scans in file-specific workflows.

**Suggested fix:**
Verify that `MetaMorpheusTask.cs/SetAllFileSpecificCommonParams` includes logic to propagate `ProductMassTolerance_LowRes` from file-specific settings, and add it if missing.

**Affected test scenarios:**
- FileSpecificParameters - set LowRes tolerance per file - should use file-specific value not default
- FileSpecificParameters - omit LowRes in file-specific - should fall back to main CommonParameters value

MetaMorpheus/EngineLayer/MetaMorpheusEngine.cs
```diff
@@ -128,11 +128,13 @@ private static double CalculateAllChargesPeptideScore(MsDataScan thisScan, List<
 
         }
 
-        public static List<MatchedFragmentIon> MatchFragmentIons(Ms2ScanWithSpecificMass scan, List<Product> theoreticalProducts, CommonParameters commonParameters, bool matchAllCharges = false)
+        public static List<MatchedFragmentIon> MatchFragmentIons(Ms2ScanWithSpecificMass scan, List<Product> theoreticalProducts, CommonParameters commonParameters, bool matchAllCharges = false, bool isLowRes = false)
```
trishorts: **Problem:**
The `MatchFragmentIons` method accepts `bool isLowRes = false` as an optional parameter defaulting to `false`. Every existing call site that does not pass this flag will silently use the standard tolerance even when processing ion-trap child scans. The method signature provides no compile-time enforcement that callers check the scan's mass analyzer. Since the parameter defaults to `false`, all pre-existing call sites are unchanged and will never use the low-res tolerance.

**Impact:**
If any call site that processes child scans was not updated to pass `isLowRes = true`, those scans will use the narrow Orbitrap tolerance for ion-trap data, defeating the entire purpose of this PR. The default-false pattern makes omission invisible.

**Suggested fix:**
Audit all call sites of `MatchFragmentIons` to confirm they pass the `isLowRes` flag when appropriate. Consider making the tolerance resolution happen inside `MatchFragmentIons` by inspecting the scan's `MassAnalyzer` property directly (e.g., `scan.TheScan.MassAnalyzer`), removing the need for callers to determine resolution externally.

**Affected test scenarios:**
- MatchFragmentIons - child scan from ion trap, isLowRes not passed - should still use wide tolerance
- MatchFragmentIons - parent scan from Orbitrap - should use narrow tolerance

MetaMorpheus/EngineLayer/MetaMorpheusEngine.cs
```diff
@@ -128,11 +128,13 @@ private static double CalculateAllChargesPeptideScore(MsDataScan thisScan, List<
...
         {
+            // if this is a child scan and it's an ion trap 2D scan, we want to use the wider tolerance for matching
+            var productMassTolerance = isLowRes? commonParameters.ProductMassTolerance_LowRes : commonParameters.ProductMassTolerance;
```
trishorts, line 134: **Problem:**
The XCorr scoring path (when `scan.TheScan.MassSpectrum.XcorrProcessed` is true) now also uses `productMassTolerance` (which may be the low-res tolerance). XCorr matching uses a fixed bin width of 1.0005079 Th, and tolerance checking in this path is a secondary filter. Applying a wide Da tolerance (e.g., 0.35 Da) in the XCorr bin-matching context could match bins that are incorrect for the XCorr algorithm's assumptions.

**Impact:**
If a child scan is both XCorr-processed and from a low-res analyzer, the wider tolerance could admit fragment matches from neighboring bins, inflating scores. This is likely a rare edge case but could produce false positive matches.

**Suggested fix:**
Determine whether XCorr-processed scans should always use the standard product tolerance regardless of mass analyzer, and if so, bypass the `isLowRes` selection for that code path.

**Affected test scenarios:**
- MatchFragmentIons - XCorr-processed ion-trap child scan - verify bin matching is correct with wide tolerance

MetaMorpheus/EngineLayer/CommonParameters.cs
```diff
@@ -160,6 +162,7 @@ public int DeconvolutionMaxAssumedChargeState
         [TomlIgnore] public Tolerance DeconvolutionMassTolerance { get; private set; }
         public int TotalPartitions { get; set; }
         public Tolerance ProductMassTolerance { get; set; } // public setter required for calibration task
+        public Tolerance ProductMassTolerance_LowRes { get; set; } // Wider mass tolerance for lower resolution analyzer (e.g. ion trap).
```
trishorts: **Problem:**
`ProductMassTolerance` has a public setter specifically for the calibration task, which recalculates and updates the tolerance after calibration. The new `ProductMassTolerance_LowRes` also has a public setter but is not coupled to calibration logic. If a user's low-res tolerance was initially set to default (tracking `ProductMassTolerance`), after calibration `ProductMassTolerance` changes but `ProductMassTolerance_LowRes` retains its pre-calibration value.

**Impact:**
Post-calibration searches would use a stale low-res tolerance that no longer reflects the calibrated instrument performance. This may produce slightly suboptimal matching for child scans in calibrated workflows.

**Suggested fix:**
In the calibration task, also update `ProductMassTolerance_LowRes` proportionally, or document that low-res tolerance is not subject to calibration adjustment.

**Affected test scenarios:**
- CalibrationTask - run calibration with default LowRes - LowRes should update or be documented as unchanged

MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
```diff
@@ -333,8 +332,8 @@ private GlycoSpectralMatch CreateGsm(Ms2ScanWithSpecificMass theScan, int scanIn
             foreach (var childScan in theScan.ChildScans)
             {
                 var childFragments = GlycoPeptides.OGlyGetTheoreticalFragments(CommonParameters.MS2ChildScanDissociationType, CommonParameters.CustomIons, peptide, peptideWithMod);
-
-                var matchedChildIons = MatchFragmentIons(childScan, childFragments, CommonParameters);
+                bool isIonTrapData = childScan.TheScan.MzAnalyzer == MZAnalyzerType.IonTrap2D;
```
trishorts: The expression `childScan.TheScan.MzAnalyzer == MZAnalyzerType.IonTrap2D` only identifies linear (2D) ion traps as low-resolution. Other ion-trap analyzer types such as `IonTrap3D` are also low-resolution instruments but will not trigger the wider `ProductMassTolerance_LowRes` tolerance. The same narrow check appears in `FindOGlycan` at the localization-scan tolerance selection.

**Impact:**
Any ETD child scan acquired on a 3D ion trap (or future low-res analyzer types) will silently use the high-resolution Orbitrap tolerance, producing fewer matched c/z ions and lower glycan localization scores — the exact problem this PR is meant to solve.

**Suggested fix:**
Extract a helper predicate (e.g., `IsLowResAnalyzer(MZAnalyzerType)`) that returns `true` for all ion-trap enum members. Use it in both `CreateGsm` and `FindOGlycan` to ensure consistent and complete coverage.

**Affected test scenarios:**
- CreateGsm - child scan with IonTrap3D analyzer - should apply low-res tolerance
- FindOGlycan - localizationScan from IonTrap3D child - should apply low-res tolerance

MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
```diff
@@ -439,12 +439,15 @@ private void FindOGlycan(Ms2ScanWithSpecificMass theScan, int scanIndex, int sco
...
             var localizationScan = theScan;
+            var toleranceForLocalizationScan = CommonParameters.ProductMassTolerance;
             List<Product> products = new List<Product>(); // product list for the theoretical fragment ions
 
             //For HCD-pd-ETD or CD-pd-EThcD type of data, we generate the different rpoducts.
             if (theScan.ChildScans.Count > 0 && GlycoPeptides.DissociationTypeContainETD(CommonParameters.MS2ChildScanDissociationType, CommonParameters.CustomIons))
             {
                 localizationScan = theScan.ChildScans.First();
+                // For the localization scan, if it is from ion trap, we will use a wider tolerance for the localization.
+                toleranceForLocalizationScan = localizationScan.TheScan.MzAnalyzer == MZAnalyzerType.IonTrap2D? CommonParameters.ProductMassTolerance_LowRes : CommonParameters.ProductMassTolerance;
```
trishorts: **Problem:**
In `FindOGlycan`, `toleranceForLocalizationScan` is computed and passed to `LocalizeOGlycan`, which is correct. However, the localization scan is always set to `theScan.ChildScans.First()` when child scans exist. If the first child scan happens to be an Orbitrap scan while a subsequent child scan is from an ion trap, the tolerance is derived from the first scan only. More importantly, there is no guarantee that the first child scan is the ETD scan; if the scan order varies, both the localization scan choice and the tolerance choice could be mismatched.

**Impact:**
For instrument methods that produce multiple child scans per parent (e.g., one CID child + one ETD child), the code may pick the wrong child scan for localization and apply the wrong tolerance, leading to degraded O-glycan localization accuracy.

**Suggested fix:**
Explicitly select the child scan whose dissociation type matches `MS2ChildScanDissociationType` (or contains ETD) rather than blindly taking `.First()`. Then derive `isLowRes` from that specific scan.

**Affected test scenarios:**
- FindOGlycan - parent with two child scans (CID first, ETD second) - should localize on ETD scan with correct tolerance
- FindOGlycan - parent with ETD child on Orbitrap - should use standard tolerance

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -541,13 +541,15 @@ public static CommonParameters SetAllFileSpecificCommonParams(CommonParameters c
             // set the rest of the file-specific parameters
             Tolerance precursorMassTolerance = fileSpecificParams.PrecursorMassTolerance ?? commonParams.PrecursorMassTolerance;
             Tolerance productMassTolerance = fileSpecificParams.ProductMassTolerance ?? commonParams.ProductMassTolerance;
+            Tolerance productMassTolerance_LowRes = fileSpecificParams.ProductMassTolerance_LowRes ?? commonParams.ProductMassTolerance;
```
trishorts: **Problem:**
The line `Tolerance productMassTolerance_LowRes = fileSpecificParams.ProductMassTolerance_LowRes ?? commonParams.ProductMassTolerance;` falls back to `commonParams.ProductMassTolerance` (the standard-resolution tolerance) when the file-specific parameter is not set. Every other parameter in this method falls back to its own task-level equivalent (e.g., `fileSpecificParams.ProductMassTolerance ?? commonParams.ProductMassTolerance`). The correct fallback should be `commonParams.ProductMassTolerance_LowRes`, which is the task-level low-resolution tolerance the user configured.

**Impact:**
When a user sets a task-level `ProductMassTolerance_LowRes` (e.g., 0.35 Da) but does not provide a file-specific override, the file-specific CommonParameters will silently receive the high-resolution `ProductMassTolerance` (e.g., 20 ppm) instead of the intended 0.35 Da. This defeats the entire purpose of the feature for any run that uses file-specific parameter resolution, which is the common path through `RunTask`.

**Suggested fix:**
Change the fallback from `commonParams.ProductMassTolerance` to `commonParams.ProductMassTolerance_LowRes`.

**Affected test scenarios:**
- SetAllFileSpecificCommonParams - fileSpecificParams.ProductMassTolerance_LowRes is null, task-level LowRes is 0.35 Da - expect file-specific LowRes to be 0.35 Da, not 20 ppm
- End-to-end search with task-level LowRes set and no file-specific TOML override - child scan matching should use task-level LowRes tolerance

MetaMorpheus/GUI/TaskWindows/GlycoSearchTaskWindow.xaml.cs
```diff
@@ -362,7 +366,31 @@ private void SaveButton_Click(object sender, RoutedEventArgs e)
                 ProductMassTolerance = new PpmTolerance(double.Parse(productMassToleranceTextBox.Text, CultureInfo.InvariantCulture));
             }
 
-            Tolerance PrecursorMassTolerance;
+            Tolerance ProductMassTolerance_lowRes;
```
trishorts, line 369: **Problem:**
In `SaveButton_Click`, the child scan mass tolerance parsing block uses `double.TryParse` and shows an error message saying "Please enter a positive number," but never actually validates that the parsed value `parsedChildTolerance` is greater than zero. A zero or negative tolerance value will be silently accepted and passed to `CommonParameters`, which would produce nonsensical fragment ion matching results.

**Impact:**
A user who accidentally enters `0` or `-1` in the child scan mass tolerance field will not receive any validation error. A zero tolerance means no ions would match; a negative tolerance is semantically invalid and could cause unexpected behavior in the matching engine.

**Suggested fix:**
After the `TryParse` succeeds, add a check: if `parsedChildTolerance <= 0`, show the existing error message and return. This mirrors the intent already expressed in the error string.

**Affected test scenarios:**
- SaveButton_Click - childScanMassToleranceTextBox contains "0" - should reject with error message
- SaveButton_Click - childScanMassToleranceTextBox contains "-5" - should reject with error message

MetaMorpheus/GUI/Util/TaskValidator.cs
```diff
@@ -16,6 +17,7 @@ public static class TaskValidator
         /// </summary>
         public static bool CheckTaskSettingsValidity(string precursorMassTolerance,
             string productMassTolerance,
```
trishorts, line 19: **Problem:**
The method `CheckTaskSettingsValidity` gains a new required parameter `string childScanMassTolerance` inserted at position 3 (between `productMassTolerance` and `maxMissedCleavages`). Every existing caller of this method must be updated to pass the new argument. Because the parameter is inserted in the middle (not appended at the end), any caller that uses positional arguments will silently shift all subsequent arguments by one position if it is not updated, leading to the wrong validation being applied to the wrong field.

**Impact:**
If any caller was missed during the update, the positional shift means `maxMissedCleavages` would be validated as `childScanMassTolerance`, `maxModificationIsoforms` as `maxMissedCleavages`, and so on — every subsequent field would be validated with the wrong check. This could allow invalid values through or reject valid ones. Since this is a GUI method with many string parameters of the same type, the compiler will not catch the mismatch.

**Suggested fix:**
Verify that every call site of `CheckTaskSettingsValidity` (search task window, calibration task window, GPTMD task window, XL search task window, etc.) has been updated to pass the new `childScanMassTolerance` argument in the correct position. Consider appending the new parameter at the end of the parameter list instead of inserting it in the middle to reduce the risk of positional errors, or consider using a parameter object / named arguments.

**Affected test scenarios:**
- SearchTaskWindow - launch with default settings - should call CheckTaskSettingsValidity with correct argument order
- GptmdTaskWindow - launch with default settings - validation should still apply to correct fields
- CalibrationTaskWindow - launch with default settings - no argument shift

MetaMorpheus/GUI/TaskWindows/GlycoSearchTaskWindow.xaml
```diff
@@ -238,23 +238,41 @@
...
+                                        <StackPanel Orientation="Horizontal" Margin="5">
+                                            <Label x:Name="lblChildMs2" Content="Product Mass Tolerance (Lower Res)" ToolTipService.ShowDuration="99999" ToolTipService.InitialShowDelay="500">
+                                                <Label.ToolTip>
+                                                    <TextBlock>
+                                                        Mass tolerance used for matching fragment ions in the MS2 child scan.
+                                                        Use this for lower-resolution analyzers (for example, ion trap).
+                                                    </TextBlock>
+                                                </Label.ToolTip>
+                                            </Label>
+                                            <local:DoubleTextBoxControl x:Name="childScanMassToleranceTextBox" HorizontalAlignment="Left"  TextWrapping="Wrap" Width="45" />
```
trishorts: **Problem:**
The XAML adds two new controls — `childScanMassToleranceTextBox` (DoubleTextBoxControl) and `childScanMassToleranceComboBox` (ComboBox) — but the diff for `GlycoSearchTaskWindow.xaml.cs` is empty. Without code-behind changes, these controls are never populated with values from `CommonParameters.ProductMassTolerance_LowRes` on load and never read back on save. The controls will always show their default/empty state and user input will be silently discarded.

**Impact:**
The GUI exposes the new "Product Mass Tolerance (Lower Res)" field but it is completely non-functional. Users will see the field, enter values, and believe they are configuring the low-res tolerance, but the search will always use the default fallback value. This is a silent data-loss bug for user configuration.

**Suggested fix:**
In `GlycoSearchTaskWindow.xaml.cs`, populate `childScanMassToleranceTextBox` and `childScanMassToleranceComboBox` from `CommonParameters.ProductMassTolerance_LowRes` in the constructor/load path, and read them back into the `CommonParameters` construction in `SaveButton_Click`. Follow the same pattern used for `productMassToleranceTextBox` / `productMassToleranceComboBox`.

**Affected test scenarios:**
- GlycoSearchTaskWindow_Load - existing task with LowRes tolerance set - control should display saved value
- GlycoSearchTaskWindow_Save - user enters 0.35 Da in low-res field - CommonParameters.ProductMassTolerance_LowRes should be AbsoluteTolerance(0.35)
- GlycoSearchTaskWindow_RoundTrip - save then reload - low-res tolerance value preserved

MetaMorpheus/Test/GlycoSearchEngineTest.cs
```diff
@@ -76,5 +69,103 @@ public void CreateGsm_WithWideProductTolerance_ScanInfo_p_IsCappedToOne()
...
+            // Default product tolerance is a PpmTolerance of 20 ppm; verify type and numeric width equivalence
+            Assert.That(commonParameters.ProductMassTolerance, Is.TypeOf<PpmTolerance>());
+            Assert.That(commonParameters.ProductMassTolerance.GetRange(1000).Width, Is.EqualTo(new PpmTolerance(20).GetRange(1000).Width));
+
+            // By default child tolerance should equal product tolerance
+            Assert.That(commonParameters.ProductMassTolerance_LowRes.GetRange(1000).Width, Is.EqualTo(commonParameters.ProductMassTolerance.GetRange(1000).Width));
+        }
+
+        [Test]
+        public static void TestChildScanTolerance_Da()
```
trishorts: **Problem:**
Both `TestChildScanTolerance_Da` and `TestChildScanTolerance_ppm` only test the `isLowRes: true` path. Neither test calls `MatchFragmentIons` with `isLowRes: false` on the same scan and then asserts that the standard `ProductMassTolerance` (20 ppm default) is used instead of the wider child-scan tolerance. Without this counter-assertion, the tests cannot distinguish between "the engine correctly selects the child tolerance" and "the engine always uses the child tolerance regardless of the flag."

**Impact:**
A regression where `MatchFragmentIons` ignores the `isLowRes` flag and unconditionally applies the wider tolerance would pass all existing tests, silently degrading specificity for high-resolution parent scans.

**Suggested fix:**
In each test, add a second call to `MatchFragmentIons` with `isLowRes: false`, then assert that matched ions fall within `ProductMassTolerance` and that at least some ions matched by the wide child tolerance are *not* matched by the narrow parent tolerance (or that the match count differs).

**Affected test scenarios:**
- TestChildScanTolerance_Da - isLowRes: false with 0.5 Da child tolerance - matched ions should satisfy default 20 ppm, not 0.5 Da
- TestChildScanTolerance_ppm - isLowRes: false with 200 ppm child tolerance - matched ions should satisfy default 20 ppm, not 200 ppm

MetaMorpheus/Test/GlycoSearchEngineTest.cs
```diff
@@ -76,5 +69,103 @@ public void CreateGsm_WithWideProductTolerance_ScanInfo_p_IsCappedToOne()
...
+                double matchedMass = ion.Mz.ToMass(ion.Charge);
+                double theoreticalMass = ion.NeutralTheoreticalProduct.NeutralMass;
+                // Verify matched ion is within child-scan tolerance
+                Assert.That(commonParameters.ProductMassTolerance_LowRes.Within(matchedMass, theoreticalMass));
+                Assert.That(childScanTolerance.Within(matchedMass, theoreticalMass));
+            }
+        }
+
+        [Test]
+        public static void TestChildScanTolerance_ppm()
```
trishorts: **Problem:**
`TestChildScanTolerance_Da` and `TestChildScanTolerance_ppm` assert `childMatchedIons.Count > 0` and that each matched ion is within the tolerance, but they never compare against the match count using the default (narrow) tolerance. A wider tolerance should match *more* fragment ions than a narrow one. Without this differential assertion, the test would pass even if the tolerance parameter were completely ignored and the engine always used the default 20 ppm.

**Impact:**
The core behavioral claim of the PR — that a wider child-scan tolerance yields more matched ions — is untested. A no-op implementation would pass.

**Suggested fix:**
Call `MatchFragmentIons` once with `isLowRes: false` (narrow tolerance) and once with `isLowRes: true` (wide tolerance), then assert `childMatchedIons.Count >= narrowMatchedIons.Count` and ideally `Is.GreaterThan` to prove the wider tolerance makes a measurable difference.

**Affected test scenarios:**
- TestChildScanTolerance_Da - compare match counts with isLowRes true vs false - wide tolerance should yield strictly more matches
- TestChildScanTolerance_ppm - compare match counts with isLowRes true vs false - wide tolerance should yield strictly more matches

trishorts (changes requested):

MetaMorpheus/EngineLayer/CommonParameters.cs
RayMSMS: Yes, the PR included the changes in line 544 and line 552 of the `MetaMorpheusTask.cs/SetAllFileSpecificCommonParams` to propagate the setting from different files.

MetaMorpheus/EngineLayer/MetaMorpheusEngine.cs
RayMSMS: Because the `ProductMassTolerance` is only used in a very specific case of the glycoSearch, it is only line 336 in `GlycoSearchEngine.cs/CreateGsm`.

MetaMorpheus/GUI/Util/TaskValidator.cs
```diff
@@ -12,12 +12,65 @@ namespace MetaMorpheusGUI
     /// </summary>
     public static class TaskValidator
     {
+        /// <summary>
+        /// Backward-compatible overload without low-res tolerance parameter.
+        /// </summary>
+        public static bool CheckTaskSettingsValidity(
```
nbollis: Why is there a new method? All it does is pass the properties to the old method. Please remove this and just update the calls

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -1577,5 +1577,29 @@ protected static void SanitizeBioPolymerDatabase<TBioPolymer>(List<TBioPolymer>
...
             }
         }
+
+        /// <summary>
+        /// Legacy TOML compatibility � when ProductMassTolerance_LowRes is omitted, the helper falls back to ProductMassTolerance to keep constant result.
+        /// </summary>
+        /// <typeparam name="TTask"></typeparam>
+        /// <param name="filePath"></param>
+        /// <returns></returns>
+        public static TTask ReadTaskTomlWithLowResFallback<TTask>(string filePath) where TTask : MetaMorpheusTask
```
nbollis, line 1587: What is the purpose of this function? Is this already taken care of by the common parameters constructor?

MetaMorpheus/EngineLayer/CommonParameters.cs
```diff
@@ -160,6 +162,7 @@ public int DeconvolutionMaxAssumedChargeState
         [TomlIgnore] public Tolerance DeconvolutionMassTolerance { get; private set; }
         public int TotalPartitions { get; set; }
         public Tolerance ProductMassTolerance { get; set; } // public setter required for calibration task
+        public Tolerance ProductMassTolerance_LowRes { get; set; }// Wider mass tolerance for lower resolution analyzer (e.g. ion trap). For now, this is a independent parameter, will not be modified by the calibration task.
```
nbollis: Update comment to say it is only used in glyco task

nbollis (commented):
This seems to add unnecessary complexity to toml reading. Instead of having one default when reading glyco tasks (set to product tolerance), and another when constructing common parameters (set to 0.35 Da). Could you simply use the common parameters constructor to set the default when null? That way you do not have to change anything about toml reading? 

Is there a need for two defaults?

Second, the task validator new method simply passes the parameters to the old method, instead we can simply call the old method.

trishorts (approved):

nbollis (approved):

Alexander-Sol (approved):

PR #2639, opened 2026-04-14
Use IRetentionTimePredictor in PEP Analysis Instead of Direct SSRCalc3 Calls

trishorts:
## Summary

`PepAnalysisEngine` previously instantiated `SSRCalc3` directly inside two methods to predict peptide hydrophobicity during PEP/FDR analysis. This change replaces those direct calls with the `IRetentionTimePredictor` interface introduced by Nic in mzLib (`RetentionTimePredictor.cs` / `IRetentionTimePredictor.cs`), making the retention time prediction step swappable without modifying this class.

## Changes

- Added `using Chromatography.RetentionTimePrediction` and `using Chromatography.RetentionTimePrediction.SSRCalc`
- `ComputeRetentionTimeEquivalentValues` (formerly `ComputeHydrophobicityValues`): removed local `SSRCalc3` instantiation; now accepts an `IRetentionTimePredictor` parameter and calls `predictor.PredictRetentionTime()` 
- `GetRetentionTimeEquivalentZscore` (formerly `GetSSRCalcHydrophobicityZScore`): same treatment — removed local `SSRCalc3` instantiation, now accepts and calls through `IRetentionTimePredictor`
- `BuildFileSpecificDictionaries`: updated both `ComputeRetentionTimeEquivalentValues` call sites to pass `new SSRCalc3RetentionTimePredictor()`
- `CreateOnePsmDataEntry`: updated `GetRetentionTimeEquivalentZscore` call site to pass `new SSRCalc3RetentionTimePredictor()`

## Behaviour

No change in behaviour. `SSRCalc3RetentionTimePredictor` wraps the same `SSRCalc3` algorithm that was previously called directly, so results are identical. The only difference is that the predictor is now passed as an `IRetentionTimePredictor`, making it straightforward to swap in a different predictor (e.g. Chronologer) in the future by changing the call sites in `BuildFileSpecificDictionaries` and `CreateOnePsmDataEntry`.

### DIFF
--- a/MetaMorpheus/EngineLayer/FdrAnalysis/PEPAnalysisEngine.cs
+++ b/MetaMorpheus/EngineLayer/FdrAnalysis/PEPAnalysisEngine.cs
@@ -1,4 +1,6 @@
 ﻿using Chemistry;
+using Chromatography.RetentionTimePrediction;
+using Chromatography.RetentionTimePrediction.SSRCalc;
 using EngineLayer.CrosslinkSearch;
 using EngineLayer.FdrAnalysis;
 using MathNet.Numerics.Statistics;
@@ -44,7 +46,7 @@ namespace EngineLayer
                 FeatureSelectionSeed = _randomSeed,
                 RandomStart = false
             };
-            
+
         private static readonly double AbsoluteProbabilityThatDistinguishesPeptides = 0.05;
 
         //These two dictionaries contain the average and standard deviations of hydrophobicitys measured in 1 minute increments accross each raw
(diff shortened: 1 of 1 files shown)


Alexander-Sol (commented):
Instead of instantiating a new predictor multiple times, can we create a property in the PEP Analysis Engine to hold the desired predictor?

trishorts:
i don't actually intend to use this interface. it is just a step in the path. my long term goal is to have a batch caller inside pep that only takes a list of full sequences and the name of a predictor and gets back retention times or hydrophobicities. Nic thought it would be good to move slowly. So step one instead of calling ssrcalc, we use his interface this gives us the same results. step 2, write the batch interface still with ssrcalc. verify again the results don't change. step 3 swap chronologer for ssrcalc. this is permanent upgrade. step 4 allow koina/prosit as an option.

nbollis (approved):

Alexander-Sol (approved):

pcruzparri (approved):

PR #2642, opened 2026-04-20
Pep model stability

nbollis:
Encountered a crash due to multiple threads reading model.zip at the same time. 

We can now avoid that by doing an in-memory copy instead of a from-disk copy. 

This pull request refactors the way prediction engines are managed for parallel processing in the `PEPAnalysisEngine`. The changes remove the need to save and reload the ML model for each thread, and instead use a thread-local prediction engine for thread safety and efficiency. This simplifies the code and improves performance, especially when an output folder is not specified.

**Parallel prediction engine management:**

* Replaced the previous approach of saving and loading the ML model per thread with the use of a `ThreadLocal<PredictionEngine>` to ensure each thread has its own thread-safe prediction engine, eliminating unnecessary file I/O and simplifying parallel execution logic.
* Added proper disposal of all thread-local prediction engines after parallel execution to prevent resource leaks.

**Code simplification and cleanup:**

* Removed the logic for saving the trained model to disk in `ComputePEPValuesForAllPSMs`, as it is no longer necessary for parallel thread safety.
* Removed the `_modelLock` object and associated locking logic, as model loading from disk per thread is no longer performed.
* Cleaned up comments to reflect the new approach and preserve clarity about single-threaded operation when no output folder is specified.

### DIFF
--- a/MetaMorpheus/EngineLayer/FdrAnalysis/PEPAnalysisEngine.cs
+++ b/MetaMorpheus/EngineLayer/FdrAnalysis/PEPAnalysisEngine.cs
@@ -163,13 +163,6 @@ namespace EngineLayer
                 var myPredictions = trainedModels[groupIndexNumber].Transform(mlContext.Data.LoadFromEnumerable(PSMDataGroups[groupIndexNumber]));
                 CalibratedBinaryClassificationMetrics metrics = mlContext.BinaryClassification.Evaluate(data: myPredictions, labelColumnName: "Label", scoreColumnName: "Score");
 
-                //Parallel operation of the following code requires the method to be stored and then read, once for each thread
-                //if not output directory is specified, the model cannot be stored, and we must force single-threaded operation
-                if (OutputFolder != null)
-                {
-                    mlContext.Model.Save(trainedModels[groupIndexNumber], dataView.Schema, Path.Combine(OutputFolder, "model.zip"));
-                }
-
                 //model is trained on peptides but here we can use that to compute PEP for all PSMs
                 int ambiguousPeptidesResolved = Compute_PSM_PEP(peptideGroups, peptideGroupIndices[groupIndexNumber], mlContext, trainedModels[groupIndexNumber], SearchType, OutputFolder);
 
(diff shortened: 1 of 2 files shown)


Alexander-Sol (dismissed):

MetaMorpheus/EngineLayer/FdrAnalysis/PEPAnalysisEngine.cs
```diff
@@ -389,82 +382,83 @@ public static string AggregateMetricsForOutput(List<CalibratedBinaryClassificati
...
             List<int> peptideGroupIndices,
             MLContext mLContext, TransformerChain<BinaryPredictionTransformer<Microsoft.ML.Calibrators.CalibratedModelParametersBase<Microsoft.ML.Trainers.FastTree.FastTreeBinaryModelParameters, Microsoft.ML.Calibrators.PlattCalibrator>>> trainedModel, string searchType, string outputFolder)
         {
             int maxThreads = FileSpecificParametersDictionary.Values.FirstOrDefault().MaxThreadsToUsePerFile;
             int ambiguousPeptidesResolved = 0;
 
-            //the trained model is not threadsafe. Therefore, to use the same model for each thread saved the model to disk. Then each thread reads its own copy of the model back from disk.
-            //If there is no output folder specified, then this can't happen. We set maxthreads eqaul to one and use the model that gets passed into the method.
+            // Preserve previous behavior for null/empty output folder by forcing single-threaded operation.
             if (String.IsNullOrEmpty(outputFolder))
```
trishorts: **Problem:**
The retained block
```
if (String.IsNullOrEmpty(outputFolder)) { maxThreads = 1; }
```
existed solely because, with the old design, threads loaded `model.zip` from disk and there was nowhere to write it when `outputFolder` was null. With the new `ThreadLocal<PredictionEngine>` approach the parallel path no longer touches disk at all, so this branch unnecessarily serializes the parallel loop in the common scenario of no output folder. The new comment ("Preserve previous behavior...") documents a constraint whose original justification has been removed by this same PR. Either the parallelism should now be enabled regardless of `outputFolder`, or the rationale for keeping the serialization should be documented (e.g., to avoid unbounded memory growth, etc.).

**Impact:**
Callers that invoke `ComputePEPValuesForAllPSMs` without an output folder (e.g., in-memory test runs, library consumers) lose the parallelism benefit that this PR is otherwise enabling. Future readers will be misled by a constraint whose reason has been removed but not explained. `outputFolder` becomes a latent perf flag with no documented semantics.

**Suggested fix:**
Either (a) drop the `IsNullOrEmpty(outputFolder) → maxThreads = 1` branch and let `Compute_PSM_PEP` always honor `MaxThreadsToUsePerFile`, since thread-safety is now provided in-memory; or (b) keep the branch but rewrite the comment to state the *current* reason for the restriction (not "preserve previous behavior"). At a minimum the parameter `outputFolder` should be documented or replaced with a boolean since its only remaining role is "is non-empty?".

trishorts (commented):
minor

trishorts (approved):

Alexander-Sol (approved):

pcruzparri (approved):

PR #2645, opened 2026-05-07
Metadraw Tweaks 

nbollis:
Did a few things. 

1. Added an option to truncate ion names (waterloss -> -H2O). 
2. Added RNA mods for custom coloring in MetaDraw
3. Fixed export of modified oligo spectral matches. 
4. Added 'Add All' and 'Remove All' to the spectral description options. 
5. Added additional spacing between ptm legend and spectral description. 
6. Surrounded our plotting with a try catch to avoid crashes on bad plot generation.


zhuoxinshi (dismissed):

MetaMorpheus/GUI/MetaDraw/MetaDrawSettingsWindow.xaml
```diff
@@ -204,19 +207,33 @@
...
+                                                              Content="{Binding DisplayName}">
+                                                        <CheckBox.ToolTip>
+                                                            <TextBlock Text="{Binding DisplayName, Converter={StaticResource SpectrumDescriptorToolTipConverter}}"/>
+                                                        </CheckBox.ToolTip>
+                                                    </CheckBox>
+                                                </DataTemplate>
+                                            </ItemsControl.ItemTemplate>
+                                        </ItemsControl>
+
+                                        <StackPanel Grid.Column="0" HorizontalAlignment="Right" VerticalAlignment="Top">
```
trishorts, line 225: **Problem:**
The new "Select All" / "Deselect All" buttons are placed in the same `Grid` cell (`Grid.Column="0"`) as the `ItemsControl` bound to `SpectrumDescriptors`. The parent `Grid` only declares row definitions and no column definitions, so it has a single implicit column; both children render in the same cell and the buttons sit on top of the checkbox list. The right-aligned `StackPanel` partially hides the checkbox content (especially long descriptor names that extend toward the right edge), since `ItemsControl` does not reserve space for the buttons.

**Impact:**
Long descriptor labels are visually clipped/overlapped by the action buttons; users may click a button thinking it is part of the checkbox row, and the descriptor list looks visually broken. Functionality of the buttons still works, but the layout is wrong.

**Suggested fix:**
Give the containing `Grid` two `ColumnDefinitions` (e.g., `*` and `Auto`) and place the `ItemsControl` in column 0 and the buttons `StackPanel` in column 1, or wrap them in a `DockPanel` where the buttons dock to the right before the list fills the remainder. Also remove the redundant `Grid.Column="0"` references that suggested intended columns that were never declared.

MetaMorpheus/Test/MetaDraw/MetaDrawSettingsAndViewsTest.cs
```diff
@@ -976,6 +978,50 @@ public static void TestSpectrumDescriptorViewModelSyncsWithStaticSettings()
                 Assert.That(descVm.IsSelected, Is.EqualTo(newValue));
             }
         }
+
+        [Test]
```
trishorts, line 982: **Problem:**
In `TestMetaDrawSettingsSnapshot`, the new property is exercised as:
```
snapshot.UseShortIonAnnotationsWhenPossible = true;
MetaDrawSettings.LoadSettings(snapshot, out bool flaggedError);
...
Assert.That(snapshot.UseShortIonAnnotationsWhenPossible);
```
The final assertion reads back the same field that was just set on the snapshot, so it cannot fail regardless of what `LoadSettings` does. Every other property in this block follows the pattern `Assert.That(snapshot.X.Equals(MetaDrawSettings.X))`; this one does not. The test does not actually verify the new setting is loaded into `MetaDrawSettings.UseShortIonAnnotationsWhenPossible`.

**Impact:**
A regression in `MetaDrawSettings.LoadSettings` (e.g., the new field is added to the snapshot but never copied through) would not be caught. This is the only assertion guarding the loader's handling of the new field, so the protection it appears to provide is illusory.

**Suggested fix:**
Compare against the static target: `Assert.That(MetaDrawSettings.UseShortIonAnnotationsWhenPossible, Is.EqualTo(snapshot.UseShortIonAnnotationsWhenPossible));`. Also assert the default-state baseline at the top of the test (before any mutations) for symmetry with the other fields.

MetaMorpheus/GUI/MetaDraw/MetaDrawSettingsWindow.xaml
nbollis: This is by design.

MetaMorpheus/Test/MetaDraw/MetaDrawSettingsAndViewsTest.cs
nbollis: fixed

zhuoxinshi (dismissed):

MetaMorpheus/Test/GuiTests/SpectralMatchComparerTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        var psmCopy = _psm.ReplaceFullSequence(_psm.FullSequence, _psm.BaseSeq);
+
+        Assert.That(SpectralMatchComparer.Instance.Equals(_psm, psmCopy), Is.True);
+        Assert.That(SpectralMatchComparer.Instance.GetHashCode(_psm), Is.EqualTo(SpectralMatchComparer.Instance.GetHashCode(psmCopy)));
+    }
+
+    [Test]
+    public void Equals_ReturnsFalse_WhenAnyComparedFieldDiffers()
+    {
+        var psmDifferent = _psm.ReplaceFullSequence(_psm.FullSequence + "_ALT", _psm.BaseSeq);
```
trishorts: **Major** - "WhenAnyComparedFieldDiffers" only exercises one of four compared fields

**Problem:**
`SpectralMatchComparer.Equals` compares four fields: `FullSequence`, `Ms2ScanNumber`, `FileNameWithoutExtension`, and `PrecursorScanNum`. The test `Equals_ReturnsFalse_WhenAnyComparedFieldDiffers` only mutates `FullSequence` (via `ReplaceFullSequence(_psm.FullSequence + "_ALT", ...)`). The three remaining fields are never independently varied, so a regression that, say, drops the `Ms2ScanNumber` clause from `Equals` would still make this test pass.

**Impact:**
The comparer's correctness contract is only verified for one quarter of the surface it claims to guard. Future refactoring of the equality logic can silently break OSM-export deduplication (the very fix this PR is shipping) without any red test.

**Suggested fix:**
Promote this to a `[TestCase]`-driven test (or four explicit tests) that vary each compared field one at a time and assert `Equals` returns false in every case. The OsmFromTsv ↔ PsmFromTsv "different types" case is already covered; the four-field matrix should be its own asymmetry.

**Affected test scenarios:**
- Equals_ReturnsFalse_WhenAnyComparedFieldDiffers - vary only FullSequence - expected false
- Equals_ReturnsFalse_WhenAnyComparedFieldDiffers - vary only Ms2ScanNumber - expected false
- Equals_ReturnsFalse_WhenAnyComparedFieldDiffers - vary only FileNameWithoutExtension - expected false
- Equals_ReturnsFalse_WhenAnyComparedFieldDiffers - vary only PrecursorScanNum - expected false

---

trishorts (commented):
Automated code review - 1 finding(s) at severity >= Major.

MetaMorpheus/Test/GuiTests/SpectralMatchComparerTests.cs
nbollis: Converted Equals_ReturnsFalse_WhenAnyComparedFieldDiffers to four [TestCase]-driven variants that independently vary each compared field (FullSequence, Ms2ScanNumber, FileNameWithoutExtension, PrecursorScanNum). Added a MutablePsmFromTsv helper that exposes the protected set properties on the mzLib base class for field-level isolation. All 9 tests pass.

Alexander-Sol (approved):

trishorts (approved):

zhuoxinshi (approved):

PR #2649, opened 2026-05-12
Parsimon fixes

nbollis:
Protein parsimony was originally designed to:
1. Take in All PSMs
2. Run Parsimony on the 1% set
3. Disambiguate All PSMs by removing non-parsimoneous proteins. 

In the past year, we have made edits that make it so only the 1% proteins are passed in so that the current flow is: 
1. Take in 1% PSMs
2. Run Parsimony on the 1% set
3. Disambiguate 1% PSMs by removing non-parsimoneous proteins. 

This PR revert that behavior and adds informative reporting to protein parsimony to state how many new connections it made and how many connections it removed (disambiguated)


MetaMorpheus/EngineLayer/ProteinParsimony/ProteinParsimonyEngine.cs
```diff
@@ -29,25 +30,29 @@ public class ProteinParsimonyEngine : MetaMorpheusEngine
         /// User-selectable option that treats differently-modified forms of a peptide as different peptides for the purposes of parsimony
         /// </summary>
         private readonly bool _treatModPeptidesAsDifferentPeptides;
+        public ProteinParsimonyEngine(FilteredPsms unFilteredPsmsForParsimony, bool modPeptidesAreDifferent, CommonParameters commonParameters, List<(string fileName, CommonParameters fileSpecificParameters)> fileSpecificParameters, List<string> nestedIds) : this(unFilteredPsmsForParsimony.FilteredPsmsList, modPeptidesAreDifferent, commonParameters, fileSpecificParameters, nestedIds) { }
```
trishorts, line 33: **Major** - `FilteredPsms` constructor overload silently breaks revert intent

**Problem:**

trishorts (commented):
Automated code review - 9 finding(s) at severity >= Major.

MetaMorpheus/EngineLayer/ProteinScoringAndFdr/ProteinScoringAndFdrEngine.cs
```diff
@@ -17,17 +16,30 @@ public class ProteinScoringAndFdrEngine : MetaMorpheusEngine
...
+        public ProteinScoringAndFdrEngine(List<ProteinGroup> proteinGroups, IList<SpectralMatch> filteredPsms, bool noOneHitWonders, bool treatModPeptidesAsDifferentPeptides, bool mergeIndistinguishableProteinGroups, 
             CommonParameters commonParameters, List<(string fileName, CommonParameters fileSpecificParameters)> fileSpecificParameters, List<string> nestedIds) : base(commonParameters, fileSpecificParameters, nestedIds)
         {
-            _FilteredPsms = filteredPsms.FilteredPsmsList;
+            _FilteredPsms = filteredPsms;
             ProteinGroups = proteinGroups;
             NoOneHitWonders = noOneHitWonders;
             TreatModPeptidesAsDifferentPeptides = treatModPeptidesAsDifferentPeptides;
             MergeIndistinguishableProteinGroups = mergeIndistinguishableProteinGroups;
             _decoyIdentifiers = proteinGroups.SelectMany(p => p.Proteins.Where(b => b.IsDecoy).Select(b => b.Accession.Split('_')[0])).ToHashSet();
```
trishorts, line 27: **Major** - Constructor ignores FilteredPsms.FilterType/FilterThreshold, re-derives from CommonParameters

**Problem:**

MetaMorpheus/EngineLayer/ProteinScoringAndFdr/ProteinScoringAndFdrEngine.cs
```diff
@@ -51,7 +63,7 @@ private void ScoreProteinGroups(List<ProteinGroup> proteinGroups, IEnumerable<Sp
         {
             // add each protein groups PSMs
             var peptideToPsmMatching = new Dictionary<IBioPolymerWithSetMods, HashSet<SpectralMatch>>();
-            foreach (var psm in psmList)
+            foreach (var psm in psmList.FilterByQValue(false, _filterThreshold, false, _filterType))
             {
```
trishorts, line 67: **Major** - FilterByQValue called with unnamed magic booleans hides intent and is easy to break

**Problem:**

MetaMorpheus/EngineLayer/ProteinScoringAndFdr/ProteinScoringAndFdrEngine.cs
```diff
@@ -17,17 +16,30 @@ public class ProteinScoringAndFdrEngine : MetaMorpheusEngine
         private readonly FilterType _filterType;
         private readonly double _filterThreshold;
 
-        public ProteinScoringAndFdrEngine(List<ProteinGroup> proteinGroups, FilteredPsms filteredPsms, bool noOneHitWonders, bool treatModPeptidesAsDifferentPeptides, bool mergeIndistinguishableProteinGroups, 
+        public ProteinScoringAndFdrEngine(List<ProteinGroup> proteinGroups, FilteredPsms filteredPsms, bool noOneHitWonders, bool treatModPeptidesAsDifferentPeptides, bool mergeIndistinguishableProteinGroups,
+            CommonParameters commonParameters, List<(string fileName, CommonParameters fileSpecificParameters)> fileSpecificParameters, List<string> nestedIds)
```
trishorts: **Major** - FilteredPsms overload re-filters an already-filtered list inside ScoreProteinGroups

**Problem:**

MetaMorpheus/EngineLayer/ProteinScoringAndFdr/ProteinScoringAndFdrEngine.cs
```diff
@@ -17,17 +16,30 @@ public class ProteinScoringAndFdrEngine : MetaMorpheusEngine
...
+            _FilteredPsms = filteredPsms;
             ProteinGroups = proteinGroups;
             NoOneHitWonders = noOneHitWonders;
             TreatModPeptidesAsDifferentPeptides = treatModPeptidesAsDifferentPeptides;
             MergeIndistinguishableProteinGroups = mergeIndistinguishableProteinGroups;
             _decoyIdentifiers = proteinGroups.SelectMany(p => p.Proteins.Where(b => b.IsDecoy).Select(b => b.Accession.Split('_')[0])).ToHashSet();
-            _filterType = filteredPsms.FilterType;
-            _filterThreshold = filteredPsms.FilterThreshold;
+
+            if (CommonParameters.QValueThreshold < CommonParameters.PepQValueThreshold)
```
trishorts: **Major** - Filter-type inequality picks PepQValue when thresholds tie, contradicting base FilteredPsms semantics

**Problem:**

MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -59,8 +59,10 @@ public MyTaskResults Run()
                 || Parameters.SearchParameters.MassDiffAcceptorType == MassDiffAcceptorType.Custom
                 )
             {
+                // RNA uses custom for adducts, we should only run mass diff localization if the user want to 
                 // This only makes sense if there is a mass difference that you want to localize. No use for exact and missed monoisotopic mass searches.
-                Parameters.SearchParameters.DoLocalizationAnalysis = true;
+                if (Parameters.SearchParameters.MassDiffAcceptorType == MassDiffAcceptorType.Custom && GlobalVariables.AnalyteType != AnalyteType.Oligo)
+                    Parameters.SearchParameters.DoLocalizationAnalysis = true;
```
trishorts: **Major** - ModOpen/Open searches no longer auto-enable DoLocalizationAnalysis

**Problem:**

MetaMorpheus/Test/StefanParsimonyTest.cs
```diff
@@ -205,6 +268,258 @@ public static void ParsimonyWeirdCatch()
...
+        public static void TestProteinScoringAndFdrEngine_ProteinRescuePath()
+        {
+            // Test the protein rescue path in DoProteinFdr (line 210)
+            // rescuedProteins.Where(x => !x.IsDecoy) are added back to sortedProteinGroups
+
+            // Create proteins with same sequence (target and decoy)
+            Protein targetProtein = new Protein("MATSIK", "TARGET_protein");
+            Protein decoyProtein = new Protein("MATSIK", "DECOY_protein", isDecoy: true);
+
+            DigestionParams digestionParams = new DigestionParams(minPeptideLength: 5);
```
trishorts, line 419: **Major** - Vacuous Count >= 0 assertion makes filter-type-selection test prove nothing

**Problem:**

MetaMorpheus/Test/StefanParsimonyTest.cs
```diff
@@ -205,6 +268,258 @@ public static void ParsimonyWeirdCatch()
...
+                new List<string>());
+
+            var results = (ProteinScoringAndFdrResults)proteinScoringEngine.Run();
+
+            // Verify that the target protein was rescued and is in the output
+            var targetGroups = results.SortedAndScoredProteinGroups.Where(p => !p.IsDecoy).ToList();
+            Assert.That(targetGroups.Count >= 1, "Target protein should be rescued and included in output");
+        }
+
+        [Test]
```
trishorts, line 470: **Major** - PepQValue filter-type-selection test has identical vacuous assertion

**Problem:**

MetaMorpheus/Test/StefanParsimonyTest.cs
```diff
@@ -153,6 +153,69 @@ public static void ParsimonyDontTreatModifiedFormsAsUnique()
...
+                0, 1, true, Polarity.Positive, double.NaN, null, null, MZAnalyzerType.Orbitrap, double.NaN, null,
+                null, "scan=1", double.NaN, null, null, double.NaN, null, DissociationType.AnyActivationType, 0, null);
+
+            Ms2ScanWithSpecificMass scan = new Ms2ScanWithSpecificMass(fakeScan, 2, 0, "File", new CommonParameters());
+
+            SpectralMatch psm1 = new PeptideSpectralMatch(modifiedPeptide, 0, 10, 1, scan, new CommonParameters(), new List<MatchedFragmentIon>());
+            psm1.SetFdrValues(0, 0, 0, 0, 0, 0, 0, 0);
+            psm1.ResolveAllAmbiguities();
+
+            SpectralMatch psm2 = new PeptideSpectralMatch(unmodifiedPeptide, 0, 10, 2, scan, new CommonParameters(), new List<MatchedFragmentIon>());
```
trishorts, line 197: **Major** - Parsimony reporting test only asserts counts are non-negative

**Problem:**

trishorts (commented):
Automated code review - 9 finding(s) at severity >= Major.

MetaMorpheus/EngineLayer/ProteinParsimony/ProteinParsimonyEngine.cs
trishorts, line 33: **Major** - `FilteredPsms` constructor overload silently breaks revert intent

**Problem:**
The new overload `ProteinParsimonyEngine(FilteredPsms unFilteredPsmsForParsimony, ...)` delegates to the list-based constructor via `: this(unFilteredPsmsForParsimony.FilteredPsmsList, ...)`. The parameter is named `unFilteredPsmsForParsimony` but its type is `FilteredPsms` (semantically a filtered wrapper). Any existing caller that still constructs a `FilteredPsms` containing only the 1%-FDR set will pass that pre-filtered list as `_allPsms`. Stage 5's disambiguation loop (`foreach (SpectralMatch psm in _allPsms)`) will then only see the 1% PSMs — exactly the bug the PR claims to revert — but the code will compile and run silently.

**Impact:**
Future callers (or any caller that wasn't updated in this PR) will get the broken behavior the PR is trying to fix, with no error and no test failure unless the test explicitly checks low-confidence PSMs were disambiguated. The reported `HypothesesRemoved` count will also under-count.

**Suggested fix:**
Either drop the `FilteredPsms` overload entirely (force callers to pass an explicit `List<SpectralMatch>` of all PSMs), or rename the parameter and add an XML doc making the "must be the unfiltered set" precondition explicit. Consider an assertion/guard that throws if the inner list looks already-filtered.

**Affected test scenarios:**
- ProteinParsimonyEngine_FilteredPsmsCtor — passes wrapper containing only 1% PSMs — expect either intentional behavior documented, or rejection
- ProteinParsimonyEngine_ListCtor_AllPsms — verify _allPsms holds unfiltered list and stage 5 trims sub-1% PSMs

---

MetaMorpheus/EngineLayer/ProteinScoringAndFdr/ProteinScoringAndFdrEngine.cs
trishorts, line 27: **Major** - Constructor ignores FilteredPsms.FilterType/FilterThreshold, re-derives from CommonParameters

**Problem:**
In the base code, the `FilteredPsms`-taking constructor assigned `_filterType = filteredPsms.FilterType;` and `_filterThreshold = filteredPsms.FilterThreshold;` directly from the `FilteredPsms` object that the caller built. In the new code, the `FilteredPsms` overload now chains into the `IList<SpectralMatch>` overload (line 19), which silently re-derives `_filterType` and `_filterThreshold` from `CommonParameters.QValueThreshold`/`CommonParameters.PepQValueThreshold` using its own inequality heuristic. The authoritative source of truth (the `FilteredPsms` instance that was actually used to produce the filtered list) is dropped on the floor.

**Impact:**
If `FilteredPsms` was constructed against different effective thresholds or with a different filter-type decision than the inequality `QValueThreshold < PepQValueThreshold` produces (e.g. when both thresholds are equal, when one was overridden, or when `FilteredPsms` applied its own precedence rules), the protein scoring/FDR step will now use thresholds that do not match the input PSM set. This silently changes protein-level scoring and q-value assignment and is a high-blast-radius regression for the very flow this PR claims to revert.

**Suggested fix:**
Have the `FilteredPsms` overload set `_filterType`/`_filterThreshold` directly from `filteredPsms.FilterType`/`filteredPsms.FilterThreshold` (as the base did), rather than chaining into the `IList` overload's CommonParameters-driven derivation. Only the new "all PSMs" path through the `IList<SpectralMatch>` constructor should be using the parameter-derived fallback.

**Affected test scenarios:**
- ProteinScoringAndFdrEngine_ctor - FilteredPsms with FilterType=PepQValue and QValueThreshold<PepQValueThreshold in CommonParameters - engine uses PepQValue (matches FilteredPsms), not QValue
- ProteinScoringAndFdrEngine_ctor - FilteredPsms with FilterThreshold differing from CommonParameters.QValueThreshold - engine uses FilteredPsms.FilterThreshold
- ScoreProteinGroups - mismatched threshold path - identical PSM scoring to base behavior

MetaMorpheus/EngineLayer/ProteinScoringAndFdr/ProteinScoringAndFdrEngine.cs
trishorts, line 67: **Major** - FilterByQValue called with unnamed magic booleans hides intent and is easy to break

**Problem:**
`psmList.FilterByQValue(false, _filterThreshold, false, _filterType)` passes two positional `false` literals at a step that determines exactly which PSMs feed protein scoring. The reader cannot tell from this call site what those flags control (decoys-included? ambiguous-included? something else?), and an inadvertent flip of a single boolean would silently change which PSMs participate in protein scoring/FDR. Given that this PR is specifically about changing the input contract (all PSMs in, filter internally), this is the most safety-critical line in the diff and it has the weakest readability/correctness signal.

**Impact:**
A wrong-value or wrong-position boolean here would silently include decoys or ambiguous PSMs (or exclude things that should be included) in the dictionary built for protein scoring, which then propagates into `AllPsmsBelowOnePercentFDR`, `BestPeptideScore`, `BestPeptideQValue`, `BestPeptidePEP`, and ultimately protein-level q-values. There is no compile-time or runtime check that catches it.

**Suggested fix:**
Use named arguments at this call (e.g. `psmList.FilterByQValue(includeDecoys: false, filterThreshold: _filterThreshold, includeAmbiguousPsms: false, filterType: _filterType)`, matching the actual `FilterByQValue` signature). This makes the intent explicit and lets future readers verify the call against the contract `FilteredPsms` used historically.

**Affected test scenarios:**
- ScoreProteinGroups - input contains decoy PSMs at q<threshold - decoys excluded from peptideToPsmMatching
- ScoreProteinGroups - input contains ambiguous PSMs at q<threshold - ambiguous PSMs excluded from peptideToPsmMatching
- ScoreProteinGroups - input mixes filter types - only PSMs passing `_filterType`/`_filterThreshold` reach scoring

MetaMorpheus/EngineLayer/ProteinScoringAndFdr/ProteinScoringAndFdrEngine.cs
trishorts: **Major** - FilteredPsms overload re-filters an already-filtered list inside ScoreProteinGroups

**Problem:**
The `FilteredPsms`-taking constructor now forwards `filteredPsms.FilteredPsmsList` to the `IList<SpectralMatch>` constructor, which stores it as `_FilteredPsms`. `RunSpecific` then calls `ScoreProteinGroups(ProteinGroups, _FilteredPsms)`, which iterates `_FilteredPsms.FilterByQValue(false, _filterThreshold, false, _filterType)`. Since `FilteredPsmsList` was already filtered by `FilteredPsms`, this is redundant filtering work on every call through the legacy code path. Worse, combined with FINDING-1, the second filter uses possibly-different thresholds from the first.

**Impact:**
On large datasets this is wasted CPU per PSM (re-evaluating q-value/PEP filtering for every PSM in `FilteredPsmsList`), and combined with FINDING-1 it can also reject PSMs that `FilteredPsms` had already accepted. For purely-redundant cases it is a silent O(N) tax with no behavior change; for divergent-threshold cases it is a correctness bug.

**Suggested fix:**
Either skip the `FilterByQValue` call entirely when the constructor input is known to be already filtered (e.g. take a "needs filtering" flag, or have the `FilteredPsms` overload set a different field/path), or unify on a single contract where the engine always receives all PSMs and always filters internally and the `FilteredPsms` overload becomes redundant.

**Affected test scenarios:**
- ProteinScoringAndFdrEngine - FilteredPsms overload - no PSMs lost vs. base behavior on identical input
- ProteinScoringAndFdrEngine - IList overload with all PSMs - filtering produces same set as FilteredPsms would have

MetaMorpheus/EngineLayer/ProteinScoringAndFdr/ProteinScoringAndFdrEngine.cs
trishorts: **Major** - Filter-type inequality picks PepQValue when thresholds tie, contradicting base FilteredPsms semantics

**Problem:**
The new derivation uses `if (CommonParameters.QValueThreshold < CommonParameters.PepQValueThreshold)` to choose `FilterType.QValue`, falling through to `FilterType.PepQValue` on equality. In the base code this decision lived inside `FilteredPsms` and was authoritative. The new heuristic is duplicating that decision but with an arbitrary tie-break (PepQValue wins ties). There is no comment or test explaining why the tie should favor PepQValue, and no guarantee `FilteredPsms` resolves ties the same way.

**Impact:**
For any run where users set `QValueThreshold == PepQValueThreshold` (a perfectly reasonable configuration — e.g. both 0.01), the engine will pick PEP-based filtering even if `FilteredPsms` would have picked q-value-based filtering, silently changing both which PSMs feed scoring and how protein groups are sorted in `SortProteinGroupsByFilterType`.

**Suggested fix:**
Either (a) remove the duplication entirely by passing the chosen `FilterType`/`FilterThreshold` in from the caller (which already constructs `FilteredPsms` with a definitive choice), or (b) make the tie-break explicit and documented (`<=` vs `<`) and add a test pinning the equality case so a future change cannot silently flip it.

**Affected test scenarios:**
- ProteinScoringAndFdrEngine_ctor - QValueThreshold == PepQValueThreshold - engine matches FilteredPsms FilterType resolution
- ProteinScoringAndFdrEngine_ctor - QValueThreshold > PepQValueThreshold - selects PepQValue with PepQValueThreshold
- ProteinScoringAndFdrEngine_ctor - QValueThreshold < PepQValueThreshold - selects QValue with QValueThreshold

MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
trishorts: **Major** - ModOpen/Open searches no longer auto-enable DoLocalizationAnalysis

**Problem:**
The `Run()` method changed the localization-enable logic. The base code unconditionally set `Parameters.SearchParameters.DoLocalizationAnalysis = true` whenever the mass diff acceptor was `ModOpen`, `Open`, or `Custom`. The new code only sets it when the acceptor is `Custom` AND the analyte type is not `Oligo`. The outer `if` still gates on all three acceptor types but no longer assigns anything for `ModOpen` or `Open` - those branches now leave `DoLocalizationAnalysis` at whatever value the search-task config provided. The accompanying comment ("RNA uses custom for adducts, we should only run mass diff localization if the user want to") only justifies the Custom+Oligo carve-out, not removing the auto-enable for ModOpen/Open.

**Impact:**
ModOpen and Open searches will silently stop running mass-difference localization unless the user explicitly opted in elsewhere, regressing a previously implicit behavior. This is unrelated to the parsimony revert claimed by the PR and is a likely unintended side effect of the conditional rewrite.

**Suggested fix:**
Preserve the unconditional assignment for `ModOpen` and `Open`, and only make the Custom branch conditional on `AnalyteType != Oligo`. Restructure as separate cases (e.g., switch or chained ifs) so each acceptor type explicitly sets the flag the way it was set before.

**Affected test scenarios:**
- PostSearchAnalysisTask.Run - MassDiffAcceptorType.Open - DoLocalizationAnalysis becomes true
- PostSearchAnalysisTask.Run - MassDiffAcceptorType.ModOpen - DoLocalizationAnalysis becomes true
- PostSearchAnalysisTask.Run - MassDiffAcceptorType.Custom + AnalyteType.Oligo - DoLocalizationAnalysis stays at user-supplied value
- PostSearchAnalysisTask.Run - MassDiffAcceptorType.Custom + AnalyteType.Proteoform - DoLocalizationAnalysis becomes true

MetaMorpheus/Test/StefanParsimonyTest.cs
trishorts, line 419: **Major** - Vacuous Count >= 0 assertion makes filter-type-selection test prove nothing

**Problem:**
`TestProteinScoringAndFdrEngine_FilterTypeSelection_QValue` is named for and commented as testing the QValue branch of filter type selection inside `ProteinScoringAndFdrEngine`. The only assertion at the end is `Assert.That(results.SortedAndScoredProteinGroups.Count >= 0);`. `Count` is a non-negative int, so the assertion is tautological and the test cannot fail regardless of what the engine does (or doesn't do) with its filter-type selection. The internal `_filterType` field referenced in the comment is never inspected.

**Impact:**
This test gives the appearance of coverage for one of the two new filter-type branches but exercises no observable behavior. A future regression that breaks QValue selection in the constructor would not be caught.

**Suggested fix:**
Either expose the chosen filter type (e.g., a property) and assert it, or assert a downstream output that differs between QValue and PepQValue selection (e.g., a protein group's QValue when only one branch would compute it correctly). At minimum, assert a specific group count derived from the test inputs rather than `>= 0`.

**Affected test scenarios:**
- TestProteinScoringAndFdrEngine_FilterTypeSelection_QValue - QValueThreshold < PepQValueThreshold - constructor selects FilterType.QValue and produces expected protein-group output
- TestProteinScoringAndFdrEngine_FilterTypeSelection_QValue - regression that swaps branches - test fails

MetaMorpheus/Test/StefanParsimonyTest.cs
trishorts, line 470: **Major** - PepQValue filter-type-selection test has identical vacuous assertion

**Problem:**
`TestProteinScoringAndFdrEngine_FilterTypeSelection_PepQValue` is the symmetric twin of FINDING-1: it claims to verify the PepQValue branch but ends with `Assert.That(results.SortedAndScoredProteinGroups.Count >= 0);`, which is always true. The test cannot distinguish FilterType.PepQValue from FilterType.QValue (or from no filter type being applied at all).

**Impact:**
Combined with FINDING-1, both new branches of the filter-type selection logic the PR adds in `ProteinScoringAndFdrEngine` are effectively untested even though dedicated tests exist for them.

**Suggested fix:**
Same as FINDING-1: assert against an observable that actually differs by branch (e.g., expose the selected filter type, or use inputs whose PEP_QValue vs. QValue would change which PSMs survive and assert on the surviving set).

**Affected test scenarios:**
- TestProteinScoringAndFdrEngine_FilterTypeSelection_PepQValue - PepQValueThreshold < QValueThreshold - constructor selects FilterType.PepQValue and PEP-based filtering is applied to PSMs
- TestProteinScoringAndFdrEngine_FilterTypeSelection_PepQValue - regression that always picks QValue - test fails

MetaMorpheus/Test/StefanParsimonyTest.cs
trishorts, line 197: **Major** - Parsimony reporting test only asserts counts are non-negative

**Problem:**
`TestParsimonyResultsTrackingPropertiesAndToString` is the only new test exercising the PR's core feature: the `HypothesesAdded` / `HypothesesRemoved` counters on `ProteinParsimonyResults`. Its assertions are `Assert.That(result.HypothesesAdded >= 0)` and `Assert.That(result.HypothesesRemoved >= 0)`. As `int` counters those are tautological. The only other assertions check that the `ToString()` output contains two literal substrings; the actual numbers reported are never inspected.

**Impact:**
The headline reporting feature claimed by this PR ("how many new connections it made and how many connections it removed") has no behavioral test. A regression that always reports 0/0, swaps the two counters, or double-counts would not be caught.

**Suggested fix:**
Construct an input where the expected added/removed counts are known (e.g., a shared peptide deliberately set up so parsimony adds N hypotheses and removes M), and assert `HypothesesAdded == N` and `HypothesesRemoved == M`. Also assert that `ToString()` includes those numeric values, not just the labels.

**Affected test scenarios:**
- TestParsimonyResultsTrackingPropertiesAndToString - input designed for additions only - HypothesesAdded equals expected non-zero N, HypothesesRemoved equals 0
- TestParsimonyResultsTrackingPropertiesAndToString - input designed for removals only - HypothesesRemoved equals expected non-zero M
- TestParsimonyResultsTrackingPropertiesAndToString - ToString output - includes the numeric counts (not just labels)

MetaMorpheus/EngineLayer/ProteinParsimony/ProteinParsimonyEngine.cs
nbollis: Fixed. Removed the FilteredPsms constructor overload entirely ? callers now pass explicit PSM lists so filtered input cannot silently defeat the revert intent.

MetaMorpheus/EngineLayer/ProteinScoringAndFdr/ProteinScoringAndFdrEngine.cs
nbollis: Fixed. Removed the FilteredPsms constructor overload ? all callers now pass explicit filtered lists, so the engine cannot silently diverge from the actual FilteredPsms semantics.

MetaMorpheus/EngineLayer/ProteinScoringAndFdr/ProteinScoringAndFdrEngine.cs
nbollis: Fixed. Converted the FilterByQValue call to use named arguments (includeHighQValuePsms, qValueThreshold, filterAtPeptideLevel, filterType) so the intent is explicit.

MetaMorpheus/EngineLayer/ProteinScoringAndFdr/ProteinScoringAndFdrEngine.cs
nbollis: Fixed. Production scoring paths now pass unfiltered PSM lists directly (Parameters.AllSpectralMatches for main path, psmsForThisFile for per-file scoring, raw psms for glyco). FilteredPsms is only used for export/reporting, so there is no double-filtering in the scoring path.

MetaMorpheus/EngineLayer/ProteinScoringAndFdr/ProteinScoringAndFdrEngine.cs
nbollis: Fixed. Aligned tie behavior with FilteredPsms: PepQValue is now selected only when PepQValueThreshold < QValueThreshold. Ties stay on QValue, matching the base FilteredPsms logic.

MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
nbollis: Fixed. Restored automatic DoLocalizationAnalysis = true for ModOpen and Open, preserving the Custom && analyte != Oligo behavior.

MetaMorpheus/Test/StefanParsimonyTest.cs
nbollis: Fixed. Replaced the vacuous Count >= 0 assertion with branch-proving checks: the QValue test now uses a PSM that passes QValue but fails PepQValue filtering, and asserts the protein group survives under QValue selection.

MetaMorpheus/Test/StefanParsimonyTest.cs
nbollis: Fixed. Same fixture as the QValue test, but with PepQValueThreshold < QValueThreshold. Asserts the protein group is empty because the PSM fails PepQValue filtering.

MetaMorpheus/Test/StefanParsimonyTest.cs
nbollis: Fixed. Now asserts exact HypothesesAdded=2 and HypothesesRemoved=0, verifies PSM association counts, and checks ToString() renders the actual values rather than just label presence.

MetaMorpheus/EngineLayer/ProteinParsimony/ProteinParsimonyEngine.cs
nbollis: Fixed. Removed the FilteredPsms constructor overload entirely ? callers now pass explicit PSM lists so filtered input cannot silently defeat the revert intent.

MetaMorpheus/EngineLayer/ProteinScoringAndFdr/ProteinScoringAndFdrEngine.cs
nbollis: Fixed. Removed the FilteredPsms constructor overload ? all callers now pass explicit filtered lists, so the engine cannot silently diverge from the actual FilteredPsms semantics.

MetaMorpheus/EngineLayer/ProteinScoringAndFdr/ProteinScoringAndFdrEngine.cs
nbollis: Fixed. Converted the FilterByQValue call to use named arguments (includeHighQValuePsms, qValueThreshold, filterAtPeptideLevel, filterType) so the intent is explicit.

MetaMorpheus/EngineLayer/ProteinScoringAndFdr/ProteinScoringAndFdrEngine.cs
nbollis: Fixed. Production scoring paths now pass unfiltered PSM lists directly (Parameters.AllSpectralMatches for main path, psmsForThisFile for per-file scoring, raw psms for glyco). FilteredPsms is only used for export/reporting, so there is no double-filtering in the scoring path.

MetaMorpheus/Test/StefanParsimonyTest.cs
nbollis: Fixed. Same fixture as the QValue test, but with PepQValueThreshold < QValueThreshold. Asserts the protein group is empty because the PSM fails PepQValue filtering.

MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
nbollis: Fixed. Restored automatic DoLocalizationAnalysis = true for ModOpen and Open, preserving the Custom && analyte != Oligo behavior.

MetaMorpheus/Test/StefanParsimonyTest.cs
nbollis: Fixed. Replaced the vacuous Count >= 0 assertion with branch-proving checks: the QValue test now uses a PSM that passes QValue but fails PepQValue filtering, and asserts the protein group survives under QValue selection.

MetaMorpheus/EngineLayer/ProteinScoringAndFdr/ProteinScoringAndFdrEngine.cs
nbollis: Fixed. Aligned tie behavior with FilteredPsms: PepQValue is now selected only when PepQValueThreshold < QValueThreshold. Ties stay on QValue, matching the base FilteredPsms logic.

MetaMorpheus/Test/StefanParsimonyTest.cs
nbollis: Fixed. Now asserts exact HypothesesAdded=2 and HypothesesRemoved=0, verifies PSM association counts, and checks ToString() renders the actual values rather than just label presence.

trishorts (approved):

Alexander-Sol (approved):

pcruzparri (approved):

PR #2650, opened 2026-05-13
Wire external MS1 feature files into MetaMorpheus precursor decon

trishorts:
> **Project of origin:** `consensus-deconvolution-paper`

## Summary

This PR adds the option to feed MetaMorpheus's precursor-deconvolution pipeline with envelopes derived from an externally-supplied whole-file MS1 deconvolution result (`<basename>_ms1.feature`, e.g. TopFD / FlashDeconv output) instead of — or **in addition to** — on-the-fly classic deconvolution.

It piggybacks on [mzLib PR 1064 (`FromFile deconvolution algorithm`)](https://github.com/smith-chem-wisc/mzLib/pull/1064), which exposes the FromFile envelope-join as a `DeconvolutionParameters` polymorphism slot. MetaMorpheus needs **no new MS1-reader code** — the feature file is loaded by mzLib, joined to MS2 scans by mzLib, and surfaced as `IsotopicEnvelope` objects exactly as classic deconvolution does today. The MM-side surface area is:

- One nullable string on `FileSpecificParameters` (`Ms1FeatureFilePath`).
- One nullable `DeconvolutionParameters` slot on `CommonParameters` (`AdditionalPrecursorDeconvolutionParameters`).
- A second per-MS2 envelope foreach in `_GetMs2Scans` that runs when the slot is non-null.
- Auto-discovery of `<basename>_ms1.feature` next to the raw file.

End-to-end benchmark on the supplied yeast top-down file (Cal → GPTMD → Search): combining the FromFile and classic sources yields **+44% PSMs and +10% protein groups at 1% FDR** over the today-baseline classic-only pipeline.

## Design (the additive-source pattern)

Today, `_GetMs2Scans` builds the per-scan `precursorSet` HashSet from up to two sources:

1. Classic on-the-fly deconvolution (`PrecursorDeconvolutionParameters`), gated on `DoPrecursorDeconvolution`.
2. The scan-header precursor info (`SelectedIonChargeStateGuess` / `SelectedIonMonoisotopicGuessMz`), gated on `UseProvidedPrecursorInfo`.

Both write into the same `HashSet`, so cross-source duplicates dedupe naturally.

This PR adds a **third additive source**, gated independently on `AdditionalPrecursorDeconvolutionParameters != null`:

3. External MS1 features loaded from a `_ms1.feature` file via mzLib's `FromFileDeconvolutionParameters`.

Because the new block reuses the same `precursorSet`, the user gets a clean toggle:

| `DoPrecursorDeconvolution` | `Ms1FeatureFilePath` set / discovered | Effective precursor source |
|---|---|---|
| `true`  | no  | Classic only (today's behavior, unchanged) |
| `true`  | yes | Classic **and** FromFile, dedup'd |
| `false` | yes | FromFile only |
| `false` | no  | Header info only (or no precursors at all) |

### Why the `MsDataScan` overload of `GetIsolatedMassesAndCharges`

mzLib PR 1064 contains one breaking change relevant here: the `MzSpectrum` overload of `GetIsolatedMassesAndCharges` now **throws** when given FromFile params, because FromFile needs an RT anchor that only an `MsDataScan` carries. The existing classic-decon call at `MetaMorpheusTask.cs:252` (passing `precursorSpectrum.MassSpectrum`) is **unchanged** — it never sees FromFile params. The new additive call uses the `MsDataScan` overload (passing `precursorSpectrum` directly), which is the only path that works with FromFile params.

### Where the resolution lives

`SetAllFileSpecificCommonParams` owns the FromFile resolution. `RunTask` only does path discovery: it auto-discovers `<basename>_ms1.feature` next to the raw file and stamps the path onto `fileSettingsList[i].Ms1FeatureFilePath` (creating a minimal `FileSpecificParameters` if one didn't already exist). Every subclass already calls `SetAllFileSpecificCommonParams` to derive its per-file `CommonParameters`, so the FromFile slot gets populated uniformly across all task types without touching any subclass call site.

### Charge bounds

`FromFileDeconvolutionParameters(filePath, minCharge, maxCharge)` inherits min/max charge from the existing primary decon params on the same `CommonParameters`. The two sources thus agree on which charges to consider — no risk of FromFile emitting envelopes outside the classic-decon search range.

## Coverage across task types

All five `MetaMorpheusTask` subclasses pick up the FromFile path with **zero subclass-side changes** because of where the wiring sits:

| Task subclass | Calls `SetAllFileSpecificCommonParams`? | Routes through `_GetMs2Scans`? | Result |
|---|---|---|---|
| `SearchTask` | yes (4 call sites) | yes (`GetMs2Scans` at SearchTask.cs:255) | covered |
| `CalibrationTask` | yes (CalibrationTask.cs:98) | yes (`GetMs2Scans` at CalibrationTask.cs:201) | covered |
| `GPTMDTask` | yes (3 call sites) | yes (`GetMs2Scans` at GPTMDTask.cs:133) | covered |
| `GlycoSearchTask` | yes (2 call sites) | yes (`GetMs2Scans` at GlycoSearchTask.cs:109) | covered |
| `XLSearchTask` | yes (2 call sites) | yes (`GetMs2ScansWrapByScanNum` at XLSearchTask.cs:106) | covered |

`grep -n PrecursorDeconvolutionParameters\\|Deconvoluter\\|Deconvolute\\|GetIsolatedMassesAndCharges` returns **no matches** outside `_GetMs2Scans` and `MsDataScan.GetIsolatedMassesAndCharges` in the production task layer — no orphan decon paths to wire separately.

### FromFile path persists through Calibration's per-file toml

Calibration emits a `<basename>-calib.toml` next to the calibrated mzML. That writer already serialises every non-`[TomlIgnore]` `FileSpecificParameters` field, so `Ms1FeatureFilePath = "..."` is **automatically persisted forward**. In a chained `Calibrate → GPTMD → Search` run, GPTMD and Search find the calibrated mzML in `Task1CalibrationTask/`, read its companion `-calib.toml`, and inherit the FromFile path even though the calibrated mzML lives in a different folder than the original feature file. No extra user action required.

## Files touched

| File | Change |
|---|---|
| `EngineLayer/CommonParameters.cs` | New `additionalPrecursorDeconParams` ctor arg + `[TomlIgnore] public DeconvolutionParameters AdditionalPrecursorDeconvolutionParameters { get; set; }`. Threaded through `CloneWithNewTerminus`. |
| `TaskLayer/FileSpecificParameters.cs` | New `Ms1FeatureFilePath` toml-readable property. |
| `TaskLayer/MetaMorpheusTask.cs` | `SetAllFileSpecificCommonParams` constructs `Readers.FromFileDeconvolutionParameters` when the file-specific path is set. `RunTask` auto-discovers the adjacent `_ms1.feature`. `_GetMs2Scans` calls `GetIsolatedMassesAndCharges` a second time per MS2 scan via the `MsDataScan` overload when `AdditionalPrecursorDeconvolutionParameters` is non-null, merging envelopes into the same `precursorSet`. New private helper `TryFindAdjacentMs1FeatureFile`. |

Total: **3 files, ~113 insertions / ~10 deletions** across two commits on `ExternalMs1Features`.

## Test methodology

Empirical verification at two granularities, both using the same yeast top-down test file. Data files: `05-26-17_B7A_yeast_td_fract7_rep1.mzML` (165 MB, 2509 MS2 scans, top-down LCMS yeast), its co-produced `05-26-17_B7A_yeast_td_fract7_rep1_ms1.feature` (TopFD output, 4866 features), and the reviewed yeast UniProt XML.

| Mode | `DoPrecursorDeconvolution` | Feature file present? | Description |
|---|---|---|---|
| **A** | `false` | yes | FromFile only |
| **B** | `true`  | no  | Classic only (today's baseline) |
| **C** | `true`  | yes | Both, dedup'd |

The "feature file present?" distinction was enforced by hardlinking the mzML into two sibling directories (`with_feature/`, `no_feature/`) so the auto-discovery saw the feature file only where intended. Common search parameters across all runs: top-down protease, `DeconvolutionMaxAssumedChargeState = 60`, 20 ppm precursor & product mass tolerance, three-MM mass-diff acceptor.

## Results, level 1: Search-only

Single-task Search against the original (uncalibrated) mzML. Shows the raw effect of switching precursor sources, before any feedback from calibration.

```
                              A (FromFile only)   B (Classic only)   C (Both)
─────────────────────────────────────────────────────────────────────────────
Elapsed                       9.5 s               14.7 s             15.5 s
MS2 scans                     2,509               2,509              2,509
Precursors built              17,824              62,145             75,967
PSMs (q ≤ 0.01)               457                 580                777
Proteoforms (q ≤ 0.01)        69                  0                  0
Protein groups (q ≤ 0.01)     44                  67                 71
```

- **Precursors:** C produced 75,967 envelopes — close to A (17,824) + B (62,145) = 79,969 minus the ~4,000 dedup'd duplicates (5%). **The two sources are mostly orthogonal** on this file.
- **PSMs at 1% FDR:** C > B > A (777 > 580 > 457). Adding FromFile to Classic gained **+197 PSMs / +34%**.
- **Protein groups at 1% FDR:** 71 (C) vs 67 (B) — +4 from the additive source.
- **Timing:** FromFile alone is fastest by ~5s (skips classic decon entirely). Adding FromFile to Classic costs ~0.8s — the binary-search-on-sorted-features hot path that PR 1064 documents is doing what it claims.

## Results, level 2: Full pipeline (Cal → GPTMD → Search)

End-to-end chain — calibration first, then GPTMD on the calibrated mzML, then a final Search on the calibrated mzML + GPTMD-augmented database. Each pipeline run uses three tomls (`top_down_calibration*.toml top_down_gptmd*.toml top_down_*.toml`) in a single CMD invocation. Same A/B/C mode definitions as above.

```
                                        A (FromFile only)   B (Classic only)   C (Both)
─────────────────────────────────────────────────────────────────────────────────────────
Calibration converged precursor tol.    ±3.1 PPM            ±2.4 PPM           ±2.5 PPM
Calibration converged product tol.      ±17.1 PPM           ±17.0 PPM          ±17.7 PPM
GPTMD modifications added               540                 532                946
Final Search PSMs (q ≤ 0.01)            629                 785                1,131
Final Search proteoforms (q ≤ 0.01)     90                  155                158
Final Search protein groups (q ≤ 0.01)  57                  73                 80
Final Search precursors built           17,825              62,147             76,180
Final Search MS2 scans                  2,509               2,509              2,509
```

### Interpretation

- **C wins the whole pipeline.** Final-Search PSMs come in at C (1,131) > B (785) > A (629). C delivers **+346 PSMs / +44%** over the today-baseline B pipeline, **+7 protein groups**, and **+3 proteoforms**, end-to-end.
- **Calibration converges looser with FromFile-only.** Pipeline A converged to `±3.1 PPM` precursor tolerance — looser than B's `±2.4` or C's `±2.5` — because A's precursor pool (17,825 envelopes) gives the calibration engine fewer high-confidence PSMs to anchor mass-accuracy corrections on. This is the cost of running FromFile alone on this dataset; the additive C mode gives the calibration engine **both** sources and lands tight tolerances close to the B baseline.
- **GPTMD modifications scale with combined coverage.** A and B end up with nearly identical mod counts (540 vs 532) despite very different precursor pools — interesting orthogonality. C produces **946 mods**, roughly A + B with dedup. The newer modifications discovered by FromFile-only (e.g. Phosphorylation on S) and by Classic-only (e.g. additional Acetylation on K) **complement** rather than duplicate each other on this file.
- **The Search-only proteoform anomaly disappears under the full pipeline.** Search-only mode B and C both reported **0** proteoforms at 1% FDR (Search-only A had 69, but in isolation). After the calibration step tightens mass tolerances and GPTMD enriches the database, all three pipelines report meaningful proteoform counts: A=90, B=155, C=158. The earlier "anomaly" was the search engine choking on the original ±20 PPM tolerance combined with multi-charge-ambiguous classic-decon envelopes — calibrating to ±2.5 PPM resolves it.

### Where the FromFile signal shows up at each stage

- **Calibration:** with the feature file present, the calibration engine sees more (and different) precursor candidates → different convergence point. The calibrated `-calib.toml` includes `Ms1FeatureFilePath = "..."`, which is the mechanism that carries the path forward into stages 2 and 3.
- **GPTMD:** picks up the `-calib.toml`, sees the path, runs its candidate-PSM search with the additive precursor source, finds more candidate modifications → augmented database has +414 mods (C) vs Classic-only (B).
- **Search:** picks up the same `-calib.toml`, runs against the GPTMD-augmented database with the additive source → +346 PSMs over Classic-only on this dataset.

The FromFile path's effect compounds through the chain: more precursors at Cal → tighter mass calibration anchors + larger search pool → more candidate mods discovered at GPTMD → richer search database + more precursors at Search → more PSMs and protein groups.

## Reproduction

1. Place an `mzML` (or `raw`) next to a `<basename>_ms1.feature` file (TopFD- or FlashDeconv-formatted).
2. Either leave them adjacent (auto-discovery picks the feature file up) or supply `Ms1FeatureFilePath = "..."` in the file-specific `<basename>.toml`.
3. Run any task — Search, Calibration, GPTMD, GlycoSearch, or XLSearch — with `DoPrecursorDeconvolution = true` (additive mode) or `false` (FromFile-only). The Search-only single-task case and the chained Cal → GPTMD → Search multi-task case are both supported transparently.

## Known prerequisite — needs an upstream mzLib companion PR

mzLib PR 1064 ships a `Ms1Feature` reader that recognises only the older `Sample_ID / ID / Time_begin / Minimum_charge_state` column schema. Newer TopFD versions write the same `_ms1.feature` extension but use a different header set (`File_name / Fraction_ID / Feature_ID / Min_time / Min_charge` plus `Envelope_num / EC_score`); format detection picks the right reader by extension, but CsvHelper then throws on the missing columns.

A companion mzLib PR adds `[Name(...)]` aliases on the existing `Ms1Feature` record fields and marks fields absent from the newer schema `[Optional]`. The change is **~15 lines on the production class + a 4-row fixture and four test additions / extensions**. Local mzLib commits:

- `ce24771c` Accept newer-TopFD column names in `Ms1Feature`
- `3aa2fe04` Add test coverage for newer-TopFD `_ms1.feature` schema

That mzLib PR should land first; once mzLib master ships with those commits and MetaMorpheus is bumped to the resulting mzLib version, this MM PR builds and runs cleanly on a fresh clone.

## Test plan

- [x] Build clean (0 errors, 53 warnings — all pre-existing nullable-ref / obsolete-API noise) against mzLib master + the `Ms1Feature` schema patch.
- [x] **Search-only Mode A** (FromFile only) — 457 PSMs at 1% FDR, 9.5 s.
- [x] **Search-only Mode B** (Classic only, baseline) — 580 PSMs at 1% FDR, 14.7 s.
- [x] **Search-only Mode C** (Both, dedup'd) — 777 PSMs at 1% FDR, 15.5 s; +34% over baseline; HashSet dedup'd ~5% of envelopes across the two sources.
- [x] **Calibration A/B isolated test** — with the feature file present, calibration converges to `±2.5 PPM / ±17.7 PPM` (vs `±2.4 / ±17.0` without); the resulting `-calib.toml` includes the auto-discovered `Ms1FeatureFilePath`.
- [x] **GPTMD A/B isolated test** — with the feature file present, GPTMD produces **+274 modifications / +31%** vs without.
- [x] **Full pipeline Mode A** (FromFile only, Cal → GPTMD → Search) — 629 PSMs / 90 proteoforms / 57 protein groups at 1% FDR; calibration converges to `±3.1 PPM` precursor.
- [x] **Full pipeline Mode B** (Classic only, baseline) — 785 / 155 / 73; calibration converges to `±2.4 PPM` precursor.
- [x] **Full pipeline Mode C** (Both) — 1,131 PSMs / 158 proteoforms / 80 protein groups at 1% FDR; **+44% PSMs / +10% protein groups** over the Classic-only baseline; calibration converges to `±2.5 PPM` precursor.
- [x] **`Ms1FeatureFilePath` persists through Calibration's per-file toml** — verified in pipelines A and C; pipeline B's calibrated toml correctly omits the field.
- [x] **All 5 task subclasses covered via the `SetAllFileSpecificCommonParams` + `_GetMs2Scans` seams** — verified empirically for Search, Calibration, and GPTMD; covered structurally for GlycoSearch and XLSearch (which use the same call paths). No subclass-side wiring changes required.
- [ ] Companion mzLib PR (`Accept newer-TopFD column names in Ms1Feature` + test coverage) landed upstream and mzLib version bump applied to MetaMorpheus.
- [ ] GUI surface to set `Ms1FeatureFilePath` per-file (auto-discovery covers the common case; explicit toml field works today).


acesnik (dismissed):
🤖 This is the one I have read today that I would merge on its merits, and I want to be specific about why rather than just approving, because several of the decisions in it are the ones that go wrong elsewhere.

Letting an external whole-file MS1 deconvolution feed the precursor set is a real capability — TopFD and FlashDeconv results exist and are often better than what we derive per-scan — and doing it *additively* rather than as a replacement is the right shape, since it does not force a user to trust the external tool exclusively.

**The PEP consequence is identified and handled, in the PR itself.** The comment at the scoring site says it plainly: `PrecursorDeconvolutionScore` is a real PEP feature, so a FromFile envelope that fed the default 0 would be systematically penalised against a classic-decon envelope, and `GetOrComputeGenericScore` puts both on a method-agnostic scale. That is exactly the trap I flagged on #2640 a few minutes ago — a value quietly reaching `PEPAnalysisEngine` and moving q-values — and here it was found and closed before anyone reviewed it. Worth saying out loud.

The same care shows up in the `fractionalIntensity` handling. Classic-decon envelopes compute it as a fraction of scan intensity; a whole-file feature has no comparable denominator, and rather than mixing scales the code reports 1.0 with the reasoning written down. That is the kind of thing that would otherwise surface months later as an unexplained bias.

Things I checked that are right, since each is a place this could have gone wrong:

`AdditionalPrecursorDeconvolutionParameters` is `[TomlIgnore]` with a stated reason — resolved per-file at run time, not persisted — which is the correct answer to "does this need a `tomlConfig` converter" rather than adding one unnecessarily. The `OxyriboAveragine` assignment is inside `if (digestionParams is RnaDigestionParams)`, matching what the existing two decon sources already do, so the new third source does not silently use the protein averagine for RNA. And the "merged into the same precursor set and thus dedup'd" claim holds: `PrecursorSet` is tolerance-aware, buckets by rounded m/z, and consolidates lazily via `Sanitize()` on enumeration, so near-duplicate envelopes from two sources collapse rather than double-counting.

The failure handling is also the right way round. A configured `Ms1FeatureFilePath` that does not exist warns. A corrupt feature file is caught per-file so one bad input degrades to skip-and-continue instead of aborting the task, with the comment noting it matches the resilience contract master had before the refactor. Compare that with an unguarded external call in the same file and the difference is stark.

## Two small things

**I would make the auto-discovery announcement a `Warn` rather than a `Log`.** Your own comment is the argument: an externally-supplied feature file "can swing PSM counts substantially", and a user who left a stale `_ms1.feature` next to their raw deserves to see it. A `Log` line satisfies that for someone who reads `results.txt`; `Warn` is what reaches a GUI user who does not. Given that the trigger is the mere presence of a file, and the effect is on reported identifications, the louder channel seems right — and it is the house channel for exactly this "you may not have meant this" case.

**Related, for the description rather than the code:** auto-discovery is opt-out. Someone upgrading, with a `_ms1.feature` sitting beside their raw from an earlier TopFD run, gets different results and only a log line saying so. That is a defensible default and I am not asking you to change it, but it is a behaviour change worth stating where release notes get written from.

Lower still: the per-file `catch` is broad enough to swallow a programming error alongside the IO and parse failures it is aimed at. Justified by the contract it is restoring, so I would leave it, but narrowing to the expected types would make a genuine bug in that path visible rather than silently skipping a file.

## Approving

CI is green across the board and it is `dirty`, so this needs a rebase before it goes anywhere regardless; the approval is 1 of 3. Nothing above is a reason to hold it.

Not verified: `Test/Test.csproj` is `net*.0-windows`, so `ExternalMs1FeatureTests` and the `ParameterTest` additions are CI's rather than mine — and CI is green on them. I have not run a search with an external feature file, so I am making no claim about whether the additional precursors improve identifications, only about how they are scored, deduped and reported.

PR #2653, opened 2026-05-16
Multiple Decon UI

nbollis:
Added Multiple Decon to the UI. This allows a user to run multiple decon's on the same spectrum (e.g. IsoDec and Classic)

### DIFF
--- a/MetaMorpheus/GUI/Views/Deconvolution/MultipleDeconParamControl.xaml
+++ b/MetaMorpheus/GUI/Views/Deconvolution/MultipleDeconParamControl.xaml
@@ -60,7 +60,6 @@
                     </Grid.ColumnDefinitions>
                     <Grid.RowDefinitions>
                         <RowDefinition Height="Auto"/>
-                        <RowDefinition Height="Auto"/>
                     </Grid.RowDefinitions>
 
                     <Label Content="Max Assumed Charge State:" Grid.Column="0" Grid.Row="0" />
@@ -84,11 +83,6 @@
                             </TextBlock>
                         </TextBox.ToolTip>
                     </local:IntegerTexBoxControl>
-
-                    <Label Content="Polarity:" Grid.Column="0" Grid.Row="1" />
-                    <ComboBox Grid.Column="1" Grid.Row="1" Width="100" Height="25"
-                              ItemsSource="{Binding Polarities}"
-                              SelectedItem="{Binding SharedPolarity}" />
                 </Grid>
             </GroupBox>
 
--- a/MetaMorpheus/GUI/Views/Deconvolution/MultipleDeconParamControl.xaml
+++ b/MetaMorpheus/GUI/Views/Deconvolution/MultipleDeconParamControl.xaml
@@ -8,6 +8,45 @@
              d:DesignHeight="600" d:DesignWidth="800">
     <UserControl.Resources>
         <local:DeconvolutionTypeToControlConverter x:Key="DeconTypeToControlConverter"/>
+
+        <Style x:Key="CircleRemoveButtonStyle" TargetType="Button">
+            <Setter Property="Width" Value="18"/>
+            <Setter Property="Height" Value="18"/>
+            <Setter Property="Content" Value="X"/>
+            <Setter Property="Foreground" Value="White"/>
+            <Setter Property="FontSize" Value="11"/>
+            <Setter Property="FontWeight" Value="Bold"/>
+            <Setter Property="BorderThickness" Value="0"/>
+            <Setter Property="Template">
+                <Setter.Value>
+                    <ControlTemplate TargetType="Button">
+                        <Grid>
+                            <Ellipse Fill="Red" Width="18" Height="18"/>
+                            <ContentPresenter HorizontalAlignment="Center" VerticalAlignment="Center"
+                                              Content="{TemplateBinding Content}"
+                                              TextElement.Foreground="White"
+                                              TextElement.FontSize="11"
+                                              TextElement.FontWeight="Bold"/>
+                        </Grid>
+                    </ControlTemplate>
+                </Setter.Value>
+            </Setter>
+        </Style>
+
+        <Style x:Key="SubParamGroupBoxStyle" TargetType="GroupBox">
+            <Setter Property="Margin" Value="0 0 0 10"/>
+            <Setter Property="HeaderTemplate">
+                <Setter.Value>
+                    <DataTemplate>
+                        <StackPanel Orientation="Horizontal" VerticalAlignment="Center">
+                            <Button Style="{StaticResource CircleRemoveButtonStyle}"
+                                    Click="RemoveSubType_Click" Tag="{Binding}" />
+                            <TextBlock Text="{Binding}" VerticalAlignment="Center" FontWeight="Bold" Margin="5 0 0 0" />
+                        </StackPanel>
+                    </DataTemplate>
+                </Setter.Value>
+            </Setter>
+        </Style>
     </UserControl.Resources>
     <ScrollViewer VerticalScrollBarVisibility="Auto">
         <StackPanel Margin="10">
@@ -64,17 +103,8 @@
             <ItemsControl ItemsSource="{Binding SubParameters}">
                 <ItemsControl.ItemTemplate>
                     <DataTemplate>
-                        <GroupBox Margin="0 0 0 10" Header="{Binding}">
-                            <Grid>
-                                <Grid.RowDefinitions>
-                                    <RowDefinition Height="Auto"/>
-                                    <RowDefinition Height="*"/>
-                                </Grid.RowDefinitions>
-                                <Button Grid.Row="0" Content="X" Width="22" Height="22"
-                                        HorizontalAlignment="Right" Margin="0 0 5 5"
-                                        Click="RemoveSubType_Click" Tag="{Binding}" />
-                                <ContentControl Grid.Row="1" Content="{Binding Converter={StaticResource DeconTypeToControlConverter}}" />
-                            </Grid>
+                        <GroupBox Header="{Binding}" Style="{StaticResource SubParamGroupBoxStyle}">
+                            <ContentControl Content="{Binding Converter={StaticResource DeconTypeToControlConverter}}" />
                         </GroupBox>
                     </DataTemplate>
                 </ItemsControl.ItemTemplate>
--- a/MetaMorpheus/GUI/Views/Deconvolution/MultipleDeconParamControl.xaml
+++ b/MetaMorpheus/GUI/Views/Deconvolution/MultipleDeconParamControl.xaml
@@ -1,0 +1,84 @@
+﻿<UserControl x:Class="MetaMorpheusGUI.MultipleDeconParamControl"
+             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
+             xmlns:d="http://schemas.microsoft.com/expression/blend/2008" 
+             xmlns:local="clr-namespace:MetaMorpheusGUI"
+             mc:Ignorable="d" 
+             d:DesignHeight="600" d:DesignWidth="800">
+    <UserControl.Resources>
+        <local:DeconvolutionTypeToControlConverter x:Key="DeconTypeToControlConverter"/>
+    </UserControl.Resources>
+    <ScrollViewer VerticalScrollBarVisibility="Auto">
+        <StackPanel Margin="10">
+            <GroupBox Header="Shared Parameters" Margin="0 0 0 10">
+                <Grid>
+                    <Grid.ColumnDefinitions>
+                        <ColumnDefinition Width="2*"/>
+                        <ColumnDefinition Width="*"/>
+                        <ColumnDefinition Width="2*"/>
+                        <ColumnDefinition Width="*"/>
+                    </Grid.ColumnDefinitions>
+                    <Grid.RowDefinitions>
+                        <RowDefinition Height="Auto"/>
+                        <RowDefinition Height="Auto"/>
+                    </Grid.RowDefinitions>
+
+                    <Label Content="Max Assumed Charge State:" Grid.Column="0" Grid.Row="0" />
+                    <local:IntegerTexBoxControl Grid.Column="1" Grid.Row="0" Width="45"
+                                                Text="{Binding SharedMaxAssumedChargeState}"
+                                                HorizontalContentAlignment="Center" VerticalContentAlignment="Center" AllowNegative="True">
+                        <TextBox.ToolTip>
+                            <TextBlock>
+                                Shared maximum charge state applied to all sub-deconvolution types.
+                            </TextBlock>
+                        </TextBox.ToolTip>
+                    </local:IntegerTexBoxControl>
+
+                    <Label Content="Min Assumed Charge State:" Grid.Column="2" Grid.Row="0" />
+                    <local:IntegerTexBoxControl Grid.Column="3" Grid.Row="0" Width="45"
+                                                Text="{Binding SharedMinAssumedChargeState}"
+                                                HorizontalContentAlignment="Center" VerticalContentAlignment="Center" AllowNegative="True">
+                        <TextBox.ToolTip>
+                            <TextBlock>
+                                Shared minimum charge state applied to all sub-deconvolution types.
+                            </TextBlock>
+                        </TextBox.ToolTip>
+                    </local:IntegerTexBoxControl>
+
+                    <Label Content="Polarity:" Grid.Column="0" Grid.Row="1" />
+                    <ComboBox Grid.Column="1" Grid.Row="1" Width="100" Height="25"
+                              ItemsSource="{Binding Polarities}"
+                              SelectedItem="{Binding SharedPolarity}" />
+                </Grid>
+            </GroupBox>
+
+            <StackPanel Orientation="Horizontal" HorizontalAlignment="Left" Margin="0 0 0 10">
+                <ComboBox ItemsSource="{Binding AvailableSubTypes}"
+                          SelectedItem="{Binding SelectedAddType}"
+                          Width="150" Height="25" />
+                <Button Content="Add Deconvolution Type" Height="25" Margin="10 0 0 0"
+                        Click="AddSubType_Click" />
+            </StackPanel>
+
+            <ItemsControl ItemsSource="{Binding SubParameters}">
+                <ItemsControl.ItemTemplate>
+                    <DataTemplate>
+                        <GroupBox Margin="0 0 0 10" Header="{Binding}">
+                            <Grid>
+                                <Grid.RowDefinitions>
+                                    <RowDefinition Height="Auto"/>
+                                    <RowDefinition Height="*"/>
+                                </Grid.RowDefinitions>
+                                <Button Grid.Row="0" Content="X" Width="22" Height="22"
+                                        HorizontalAlignment="Right" Margin="0 0 5 5"
+                                        Click="RemoveSubType_Click" Tag="{Binding}" />
+                                <ContentControl Grid.Row="1" Content="{Binding Converter={StaticResource DeconTypeToControlConverter}}" />
+                            </Grid>
+                        </GroupBox>
+                    </DataTemplate>
+                </ItemsControl.ItemTemplate>
+            </ItemsControl>
+        </StackPanel>
+    </ScrollViewer>
+</UserControl>
--- a/MetaMorpheus/GUI/Views/Deconvolution/MultipleDeconParamControl.xaml
+++ b/MetaMorpheus/GUI/Views/Deconvolution/MultipleDeconParamControl.xaml
@@ -1,4 +1,4 @@
-﻿<UserControl x:Class="MetaMorpheusGUI.MultipleDeconParamControl"
+<UserControl x:Class="MetaMorpheusGUI.MultipleDeconParamControl"
              xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
              xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
@@ -65,7 +65,7 @@
                     <Label Content="Max Assumed Charge State:" Grid.Column="0" Grid.Row="0" />
                     <local:IntegerTexBoxControl Grid.Column="1" Grid.Row="0" Width="45"
                                                 Text="{Binding SharedMaxAssumedChargeState}"
-                                                HorizontalContentAlignment="Center" VerticalContentAlignment="Center" AllowNegative="True">
+                                                HorizontalContentAlignment="Center" VerticalContentAlignment="Center" AllowNegative="False">
                         <TextBox.ToolTip>
                             <TextBlock>
                                 Shared maximum charge state applied to all sub-deconvolution types.
@@ -76,7 +76,7 @@
                     <Label Content="Min Assumed Charge State:" Grid.Column="2" Grid.Row="0" />
                     <local:IntegerTexBoxControl Grid.Column="3" Grid.Row="0" Width="45"
                                                 Text="{Binding SharedMinAssumedChargeState}"
-                                                HorizontalContentAlignment="Center" VerticalContentAlignment="Center" AllowNegative="True">
+                                                HorizontalContentAlignment="Center" VerticalContentAlignment="Center" AllowNegative="False">
                         <TextBox.ToolTip>
                             <TextBlock>
                                 Shared minimum charge state applied to all sub-deconvolution types.
--- a/MetaMorpheus/GUI/Views/Deconvolution/MultipleDeconParamControl.xaml
+++ b/MetaMorpheus/GUI/Views/Deconvolution/MultipleDeconParamControl.xaml
@@ -65,7 +65,7 @@
                     <Label Content="Max Assumed Charge State:" Grid.Column="0" Grid.Row="0" />
                     <local:IntegerTexBoxControl Grid.Column="1" Grid.Row="0" Width="45"
                                                 Text="{Binding SharedMaxAssumedChargeState}"
-                                                HorizontalContentAlignment="Center" VerticalContentAlignment="Center" AllowNegative="False">
+                                                HorizontalContentAlignment="Center" VerticalContentAlignment="Center" AllowNegative="True">
                         <TextBox.ToolTip>
                             <TextBlock>
                                 Shared maximum charge state applied to all sub-deconvolution types.
@@ -76,7 +76,7 @@
                     <Label Content="Min Assumed Charge State:" Grid.Column="2" Grid.Row="0" />
                     <local:IntegerTexBoxControl Grid.Column="3" Grid.Row="0" Width="45"
                                                 Text="{Binding SharedMinAssumedChargeState}"
-                                                HorizontalContentAlignment="Center" VerticalContentAlignment="Center" AllowNegative="False">
+                                                HorizontalContentAlignment="Center" VerticalContentAlignment="Center" AllowNegative="True">
                         <TextBox.ToolTip>
                             <TextBlock>
                                 Shared minimum charge state applied to all sub-deconvolution types.
--- a/MetaMorpheus/GUI/Views/Deconvolution/MultipleDeconParamControl.xaml.cs
+++ b/MetaMorpheus/GUI/Views/Deconvolution/MultipleDeconParamControl.xaml.cs
@@ -1,0 +1,30 @@
+using System.Windows;
+using System.Windows.Controls;
+using GuiFunctions;
+
+namespace MetaMorpheusGUI;
+
+public partial class MultipleDeconParamControl : UserControl
+{
+    public MultipleDeconParamControl()
+    {
+        InitializeComponent();
+    }
+
+    private void RemoveSubType_Click(object sender, RoutedEventArgs e)
+    {
+        if (sender is Button button && button.Tag is DeconParamsViewModel vm
+            && DataContext is MultipleDeconParamsViewModel hostVm)
+        {
+            hostVm.RemoveSubType(vm);
+        }
+    }
+
+    private void AddSubType_Click(object sender, RoutedEventArgs e)
+    {
+        if (DataContext is MultipleDeconParamsViewModel hostVm)
+        {
+            hostVm.AddSubType(hostVm.SelectedAddType);
+        }
+    }
+}
--- a/MetaMorpheus/GuiFunctions/MzLibExtensions.cs
+++ b/MetaMorpheus/GuiFunctions/MzLibExtensions.cs
@@ -67,7 +67,7 @@ namespace GuiFunctions
                 // Default to only classic
                 case DeconvolutionType.Multiple:
                     var inner = DeconvolutionType.ClassicDeconvolution.GetDefaultDeconParams(analyteType, isPrecursor);
-                    return new MultipleDeconParameters([inner], inner.MinAssumedChargeState, inner.MaxAssumedChargeState, inner.Polarity, inner.AverageResidueModel); e.ClassicDeconvolution.GetDefaultDeconParams(analyteType, isPrecursor)], 0, 0);
+                    return new MultipleDeconParameters([inner], inner.MinAssumedChargeState, inner.MaxAssumedChargeState, inner.Polarity, inner.AverageResidueModel, inner.ExpectedIsotopeSpacing);
 
                 case DeconvolutionType.ClassicDeconvolution:
 
--- a/MetaMorpheus/GuiFunctions/MzLibExtensions.cs
+++ b/MetaMorpheus/GuiFunctions/MzLibExtensions.cs
@@ -1,4 +1,5 @@
-﻿using EngineLayer;
+﻿using Chemistry;
+using EngineLayer;
 using MassSpectrometry;
 using MzLibUtil;
 using Omics;
@@ -9,9 +10,10 @@ using Readers;
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using Chemistry;
+using System.Runtime.InteropServices.WindowsRuntime;
 using Transcriptomics;
 using Transcriptomics.Digestion;
+using UsefulProteomicsDatabases;
 
 namespace GuiFunctions
 {
@@ -20,6 +22,12 @@ namespace GuiFunctions
     /// </summary>
     public static class MzLibExtensions
     {
+
+        public static DeconParamsViewModel GetDefaultViewModel(this DeconvolutionType deconvolutionType, AnalyteType? analyteType = null, bool isPrecursor = true)
+        {
+            var defaultParams = deconvolutionType.GetDefaultDeconParams(analyteType, isPrecursor);
+            return defaultParams.ToViewModel();
+        }
 
         /// <summary>
         /// Converts the given <see cref="DeconvolutionParameters"/> to a <see cref="DeconParamsViewModel"/>.
@@ -43,6 +51,75 @@ namespace GuiFunctions
             {
                 throw new NotImplementedException();
             }
+        }
+
+        public static DeconvolutionParameters GetDefaultDeconParams(this DeconvolutionType type, AnalyteType? analyteType = null, bool isPrecursor = true)
+        {
+            analyteType ??= GlobalVariables.AnalyteType;
+
+            switch (type)
+            {
+                case DeconvolutionType.ExampleNewDeconvolutionTemplate:
+                    return null;
+
+                case DeconvolutionType.ClassicDeconvolution:
+
+                    // Precursor
+                    if (isPrecursor)
+                    {
+                        return analyteType switch
+                        {
+                            AnalyteType.Peptide => new ClassicDeconvolutionParameters(1, 12, 4, 3),
+                            AnalyteType.Proteoform => new ClassicDeconvolutionParameters(1, 60, 4, 3),
+                            AnalyteType.Oligo => new ClassicDeconvolutionParameters(-20, -1, 4, 3, Polarity.Negative,
+                                new OxyriboAveragine()),
+                            _ => throw new ArgumentOutOfRangeException()
+                        };
+                    }
+                    else
+                    {
+                        return analyteType switch
+                        {
+                            AnalyteType.Peptide => new ClassicDeconvolutionParameters(1, 10, 4, 3),
+                            AnalyteType.Proteoform => new ClassicDeconvolutionParameters(1, 10, 4, 3),
+                            AnalyteType.Oligo => new ClassicDeconvolutionParameters(-10, -1, 4, 3, Polarity.Negative,
+                                new OxyriboAveragine()),
+                            _ => throw new ArgumentOutOfRangeException()
+                        };
+                    }
+
+                case DeconvolutionType.IsoDecDeconvolution:
+
+                    // Precursor
+                    if (isPrecursor)
+                    {
+                        return analyteType switch
+                        {
+                            AnalyteType.Peptide => new IsoDecDeconvolutionParameters() { MaxAssumedChargeState = 12 },
+                            AnalyteType.Proteoform => new IsoDecDeconvolutionParameters()
+                                { MaxAssumedChargeState = 60 },
+                            AnalyteType.Oligo => new IsoDecDeconvolutionParameters(Polarity.Negative)
+                                { MaxAssumedChargeState = -20, MinAssumedChargeState = -1 },
+                            _ => throw new ArgumentOutOfRangeException()
+                        };
+                    }
+                    else
+                    {
+                        return analyteType switch
+                        {
+                            AnalyteType.Peptide => new IsoDecDeconvolutionParameters(reportMultipleMonoisos: false)
+                                { MaxAssumedChargeState = 10 },
+                            AnalyteType.Proteoform => new IsoDecDeconvolutionParameters(reportMultipleMonoisos: false)
+                                { MaxAssumedChargeState = 10 },
+                            AnalyteType.Oligo => new IsoDecDeconvolutionParameters(Polarity.Negative,
+                                    reportMultipleMonoisos: false)
+                                { MaxAssumedChargeState = -10, MinAssumedChargeState = -1 },
+                            _ => throw new ArgumentOutOfRangeException()
+                        };
+                    }
+            }
+
+            return null;
         }
 
         public static bool IsCrossLinkedPeptide(this SpectrumMatchFromTsv sm)
--- a/MetaMorpheus/GuiFunctions/MzLibExtensions.cs
+++ b/MetaMorpheus/GuiFunctions/MzLibExtensions.cs
@@ -40,17 +40,18 @@ namespace GuiFunctions
         public static DeconParamsViewModel ToViewModel(this DeconvolutionParameters parameters)
         {
             if (parameters is ClassicDeconvolutionParameters classicParams)
-            {
                 return new ClassicDeconParamsViewModel(classicParams);
-            }
-            else if (parameters is IsoDecDeconvolutionParameters isoParams)
-            {
+            
+            if (parameters is IsoDecDeconvolutionParameters isoParams)
                 return new IsoDecDeconParamsViewModel(isoParams);
-            }
-            else
-            {
-                throw new NotImplementedException();
-            }
+            
+            if (parameters is MultipleDeconParameters multipleParams)
+                return new MultipleDeconParamsViewModel(multipleParams);
+
+            //if (parameters is FromFileDeconvolutionParameters file)
+            //    return new FromFileDeconvolutionParametersViewModel(file);
+
+            throw new NotImplementedException();
         }
 
         public static DeconvolutionParameters GetDefaultDeconParams(this DeconvolutionType type, AnalyteType? analyteType = null, bool isPrecursor = true)
@@ -59,8 +60,14 @@ namespace GuiFunctions
 
             switch (type)
             {
+                case DeconvolutionType.FromFile:
                 case DeconvolutionType.ExampleNewDeconvolutionTemplate:
                     return null;
+
+                // Default to only classic
+                case DeconvolutionType.Multiple:
+                    var inner = DeconvolutionType.ClassicDeconvolution.GetDefaultDeconParams(analyteType, isPrecursor);
+                    return new MultipleDeconParameters([inner], inner.MinAssumedChargeState, inner.MaxAssumedChargeState, inner.Polarity, inner.AverageResidueModel); e.ClassicDeconvolution.GetDefaultDeconParams(analyteType, isPrecursor)], 0, 0);
 
                 case DeconvolutionType.ClassicDeconvolution:
 
--- a/MetaMorpheus/GuiFunctions/MzLibExtensions.cs
+++ b/MetaMorpheus/GuiFunctions/MzLibExtensions.cs
@@ -1,4 +1,4 @@
-﻿using Chemistry;
+using Chemistry;
 using EngineLayer;
 using MassSpectrometry;
 using MzLibUtil;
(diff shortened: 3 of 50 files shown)


MetaMorpheus/GUI/Views/Deconvolution/MultipleDeconParamControl.xaml.cs
```diff
@@ -0,0 +1,30 @@
...
+
+    private void RemoveSubType_Click(object sender, RoutedEventArgs e)
+    {
+        if (sender is Button button && button.Tag is DeconParamsViewModel vm
+            && DataContext is MultipleDeconParamsViewModel hostVm)
+        {
+            hostVm.RemoveSubType(vm);
+        }
+    }
+
```
trishorts, line 22: **Critical** - AddSubType_Click always adds precursor defaults, never product defaults

**Problem:**
`AddSubType_Click` invokes `hostVm.AddSubType(hostVm.SelectedAddType)`, which falls through to the optional parameter `bool isPrecursor = true` on `MultipleDeconParamsViewModel.AddSubType`. The control is reused for both the precursor and the product `MultipleDeconParamsViewModel` (the same `MultipleDeconParamControl` appears on both sides via `DeconvolutionTypeToControlConverter`). When the user clicks "Add" on the product side, the new sub-VM is still constructed from `GetDefaultDeconParams(... isPrecursor: true)`, so it inherits precursor-tuned charge ranges and isotope settings rather than product ones.

**Impact:**
Any sub-type the user adds to a product-side Multiple Decon configuration silently uses precursor defaults (e.g., MaxAssumedChargeState of 12/60 instead of 10/10 for peptide/proteoform). Wrong charge ranges propagate into the search task and skew product deconvolution results - exactly the kind of high-risk wrong-result behavior this PR is supposed to wire up correctly.

**Suggested fix:**
The view model must know whether it is the precursor or product host, and `AddSubType` must default to that. Pass an `isPrecursor` flag (or `bool forPrecursor`) into the `MultipleDeconParamsViewModel` constructor from `DeconHostViewModel` when it is created for the product list, store it on the VM, and use that stored flag inside `AddSubType` instead of the hard-coded `true` default. Alternatively, infer it from the existing sub-parameters (they all share the same role), but the explicit flag is safer.

**Affected test scenarios:**
- `MultipleDeconParamsViewModel_AddSubType_OnProductHost_UsesProductDefaults` - construct VM in product mode, AddSubType(ClassicDeconvolution), assert MaxAssumedChargeState matches product defaults (10 for peptide), not precursor defaults (12).
- `MultipleDeconParamControl_AddSubType_OnProductDataContext_AddsProductDefaultedSubVm` - host the control with a product-mode VM, simulate the Add click, assert resulting sub-VM has product charge defaults.
- `MultipleDeconParamsViewModel_AddSubType_OnPrecursorHost_UsesPrecursorDefaults` - regression-cover the existing path.

---

MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/MultipleDeconParamsViewModel.cs
```diff
@@ -0,0 +1,117 @@
...
+            MinAssumedChargeState = value;
+            foreach (var sub in _subParameters)
+                sub.MinAssumedChargeState = value;
+        }
+    }
+
+    public int SharedMaxAssumedChargeState
+    {
+        get => MaxAssumedChargeState;
+        set
```
trishorts, line 77: **Major** - Polarity shadowed with `new` breaks shared-polarity propagation via base reference

**Problem:**
`public new Polarity Polarity` shadows the base `DeconParamsViewModel.Polarity` rather than overriding it. The shared-polarity feature this PR introduces only fires when the property is accessed through the most-derived static type. Any code that touches polarity through a `DeconParamsViewModel` reference - for example a parent host iterating `PrecursorDeconvolutionParametersList` and setting `Polarity` polymorphically, or a WPF binding declared against the base type via `{Binding Polarity}` on an item template that resolves through the base - calls the base setter, which updates `_parameters.Polarity` but never copies the new polarity into `_subParameters`.

**Impact:**
Polarity is silently de-synchronized between the Multiple Decon container and its children depending on which code path mutated it. The user can see a positive polarity in the parent yet still have negative-polarity sub-parameters used during deconvolution, producing wrong charge assignments. This is a per-method-pair, non-obvious failure mode that QA will struggle to reproduce.

**Suggested fix:**
Make the base `DeconParamsViewModel.Polarity` setter `virtual` (if it is not already) and `override` it here, so polarity propagation runs regardless of the static type used at the call site. Remove the `new` keyword.

**Affected test scenarios:**
- `MultipleDeconParamsViewModel_SetPolarityViaBaseReference_PropagatesToSubParameters` - cast instance to `DeconParamsViewModel`, set Polarity, assert each sub-VM's Polarity changed.
- `MultipleDeconParamsViewModel_SetPolarityViaDerivedReference_PropagatesToSubParameters` - retain existing coverage.
- `DeconHostViewModel_SharedPolarityAcrossList_AppliesToMultipleDeconChildren` - verify host-level polarity changes reach Multiple Decon children.

---

MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/MultipleDeconParamsViewModel.cs
```diff
@@ -0,0 +1,117 @@
...
+        subVm.MaxAssumedChargeState = _parameters.MaxAssumedChargeState;
+        _subParameters.Add(subVm);
+        RebuildParameters();
+    }
+
+    public void RemoveSubType(DeconParamsViewModel vm)
+    {
+        if (_subParameters.Count <= 1)
+            return;
+
```
trishorts, line 130: **Major** - RebuildParameters reads from stale _parameters fields after sub-mutations

**Problem:**
`RebuildParameters` constructs a new `MultipleDeconParameters` from `_parameters.MinAssumedChargeState`, `_parameters.MaxAssumedChargeState`, `_parameters.Polarity`, `_parameters.AverageResidueModel`, and `_parameters.ExpectedIsotopeSpacing`. Sub-VMs, however, are appended/removed via `AddSubType`/`RemoveSubType` *after* the user may have mutated sub-VM charges or polarity directly (through the per-sub-control bindings), and those mutations live on the sub-VM's own parameters, not on `_parameters`. After Add/Remove, the new `MultipleDeconParameters` is rebuilt with the parent's values, overwriting nothing per se, but the rebuilt object's `Parameters` collection comes from `s.Parameters` references - which means if a sub's polarity was changed but the parent's was not, the rebuilt container's polarity will diverge from at least one sub's polarity, breaking the invariant the "shared polarity" commit is supposed to enforce.

**Impact:**
Multiple Decon state can become internally inconsistent across an Add/Remove. The pipeline downstream of `Parameters` (which is what the task uses) sees a polarity/charge tuple on the container that does not match what the children carry, producing wrong-result deconvolution that is hard to reproduce.

**Suggested fix:**
Either (a) reset all sub-VMs to the parent's shared values inside `RebuildParameters` (mirroring what `AddSubType` already does for new subs), or (b) drop the per-sub-VM mutability for the shared fields entirely - bind sub-controls' charge/polarity to the parent's `Shared*` properties so they can never diverge. Option (b) is the more defensible design given the PR's intent.

**Affected test scenarios:**
- `MultipleDeconParamsViewModel_RebuildParameters_AfterSubMutation_SubsRemainConsistent` - mutate a sub-VM's polarity directly, AddSubType, assert all subs share the parent polarity.
- `MultipleDeconParamsViewModel_ParametersGetter_ReflectsCurrentSharedState_AfterEdits` - assert `((MultipleDeconParameters)Parameters).Polarity` matches every child's polarity after a sequence of edits.

---

MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/DeconHostViewModel.cs
```diff
@@ -41,78 +41,27 @@ public DeconHostViewModel(DeconvolutionParameters? initialPrecursorParameters =
...
+            {
+                if (initialPrecursorParameters != null && initialPrecursorParameters.DeconvolutionType == deconType)
+                    PrecursorDeconvolutionParametersList.Add(initialPrecursorParameters.ToViewModel());
+                else
+                    PrecursorDeconvolutionParametersList.Add(
+                        deconType.GetDefaultViewModel(GlobalVariables.AnalyteType, true));
+
+                if (initialProductParameters != null && initialProductParameters.DeconvolutionType == deconType)
+                    ProductDeconvolutionParametersList.Add(initialProductParameters.ToViewModel());
+                else
```
trishorts: **Major** - Catch of NotImplementedException silently swallows real bugs

**Problem:**
The new loop wraps the per-type defaulting in `try { ... } catch (NotImplementedException e) { Console.WriteLine(...); }`. The XML comment claims this is intended to skip types without defaults, but `NotImplementedException` is also the standard exception thrown by stub code and partially-implemented branches. Any actual bug in `GetDefaultViewModel`/`GetDefaultDeconParams` for a real type (Classic, IsoDec, Multiple) will be silently swallowed - the type will disappear from the dropdown and the user will see "the option I expected is missing" rather than a stacktrace. `Console.WriteLine` in a WPF app goes nowhere visible.

**Impact:**
Regressions in defaulting code for in-scope types fail silently. The first symptom is a missing combo box entry that users may interpret as a feature, not a crash, delaying detection of broken defaults indefinitely.

**Suggested fix:**
Replace `NotImplementedException` with a dedicated sentinel exception thrown deliberately by extension methods that lack a default for a type, e.g., `NoDefaultDeconParamsException`. Keep the catch narrow to that sentinel. Log via the project's existing GUI/diagnostic logger rather than `Console.WriteLine`, and discard the unused `e` variable or include it in the log.

**Affected test scenarios:**
- `DeconHostViewModel_DeconTypeWithoutDefaults_SkippedFromList` - keep happy-path skip behavior under the new sentinel exception.
- `DeconHostViewModel_DeconTypeWithBuggyDefaults_DoesNotSwallowGenericFailure` - injecting a NotImplemented-throwing default should now fail loudly, not be skipped.

---

MetaMorpheus/GuiFunctions/MzLibExtensions.cs
```diff
@@ -21,6 +23,12 @@ namespace GuiFunctions
     public static class MzLibExtensions
     {
 
+        public static DeconParamsViewModel GetDefaultViewModel(this DeconvolutionType deconvolutionType, AnalyteType? analyteType = null, bool isPrecursor = true)
```
trishorts, line 26: **Major** - `GetDefaultViewModel` throws `NotImplementedException` for valid enum inputs

**Problem:**
`GetDefaultViewModel` unconditionally calls `defaultParams.ToViewModel()` on the result of `GetDefaultDeconParams`. For `DeconvolutionType.FromFile` and `DeconvolutionType.ExampleNewDeconvolutionTemplate` the latter intentionally returns `null` (line 60-62). Since extension-method dispatch is static, the call proceeds and inside `ToViewModel` none of the `is`-pattern checks match `null`, so it falls through to `throw new NotImplementedException();`. The same happens if `GetDefaultDeconParams` returns the unreachable trailing `return null;` for a future enum value. The result is a confusing `NotImplementedException` thrown from `ToViewModel` rather than a coherent failure or documented null result.

**Impact:**
Any caller (including the new `DeconHostViewModel` wiring) that hands `GetDefaultViewModel` a `FromFile` or `ExampleNewDeconvolutionTemplate` selection will crash with `NotImplementedException` whose message gives no hint that the deconvolution type itself is the problem. If the new "Multiple Decon" UI is wired so the user can pick any `DeconvolutionType` from a dropdown, this is reachable from the GUI.

**Suggested fix:**
Either (a) short-circuit in `GetDefaultViewModel` when `GetDefaultDeconParams` returns `null` (return `null` or throw a typed exception that names the unsupported `DeconvolutionType`), or (b) make `GetDefaultDeconParams` throw a typed exception for the unsupported cases so the failure mode is explicit and discoverable.

**Affected test scenarios:**
- `GetDefaultViewModel_FromFile_ReturnsNullOrTypedException` - FromFile - does not throw `NotImplementedException`
- `GetDefaultViewModel_ExampleNewDeconvolutionTemplate_ReturnsNullOrTypedException` - ExampleNewDeconvolutionTemplate - does not throw `NotImplementedException`
- `GetDefaultViewModel_UnknownEnumValue_ThrowsTypedException` - undefined enum value - throws a typed, named exception

---

MetaMorpheus/GUI/Views/Deconvolution/MultipleDeconParamControl.xaml
```diff
@@ -0,0 +1,108 @@
...
+                        <ColumnDefinition Width="*"/>
+                        <ColumnDefinition Width="2*"/>
+                        <ColumnDefinition Width="*"/>
+                    </Grid.ColumnDefinitions>
+                    <Grid.RowDefinitions>
+                        <RowDefinition Height="Auto"/>
+                    </Grid.RowDefinitions>
+
+                    <Label Content="Max Assumed Charge State:" Grid.Column="0" Grid.Row="0" />
+                    <local:IntegerTexBoxControl Grid.Column="1" Grid.Row="0" Width="45"
```
trishorts, line 66: **Major** - AllowNegative="True" on shared charge-state inputs contradicts shared-polarity refactor

**Problem:**
Both `SharedMaxAssumedChargeState` and `SharedMinAssumedChargeState` `IntegerTexBoxControl`s are declared with `AllowNegative="True"`. Per the PR's `shared polarity` commit, polarity is now hoisted into a separate piece of shared state across the decon param view models. The numeric charge magnitude entered here should therefore always be a positive integer; the sign is supposed to come from the polarity selector. Allowing the user to type `-3` for "Max Assumed Charge State" lets two independent inputs disagree about polarity, and downstream code that interprets the value as a signed integer will silently double-apply the sign in negative mode (and invert it in positive mode).

**Impact:**
A user who fills in `-10` as max charge in negative-mode runs will get charge ranges that are either inverted, empty, or pass through to mzLib with a sign mismatch, depending on how each consumer combines the magnitude with polarity. Min > Max validation also becomes meaningless once signs can flip.

**Suggested fix:**
Remove `AllowNegative="True"` from both `IntegerTexBoxControl`s on this control (and assert positivity in the setter on the VM). Let polarity be the sole source of sign.

**Affected test scenarios:**
- MultipleDeconParamsViewModelTest - setting SharedMaxAssumedChargeState to a negative value - rejected or clamped to positive
- MultipleDeconParamsViewModelTest - polarity Negative + SharedMaxAssumedChargeState 10 - emitted decon params carry charge magnitude 10 with negative polarity (not -10)
- MultipleDeconParamsViewModelTest - SharedMinAssumedChargeState > SharedMaxAssumedChargeState - validation fires regardless of polarity

---

MetaMorpheus/Test/GuiTests/Deconvolution/MultipleDeconParamsViewModelTest.cs
```diff
@@ -0,0 +1,235 @@
...
+        _viewModel.AddSubType(DeconvolutionType.IsoDecDeconvolution);
+        var originalParameters = _viewModel.Parameters;
+        _viewModel.RemoveSubType(_viewModel.SubParameters.Last());
+
+        Assert.That(_viewModel.Parameters, Is.Not.SameAs(originalParameters));
+    }
+
+    [Test]
+    public void SharedMinAssumedChargeState_PropagatesToSubParameters()
+    {
```
trishorts, line 124: **Major** - `RemoveSubType_BlockWhenOnlyOne` does not assert that removal is blocked

**Problem:**
The test asserts only `_viewModel.SubParameters.Count, Is.EqualTo(1)` after calling `RemoveSubType` on the lone sub-parameter. That passes whether removal was blocked or whether removal succeeded followed by a re-add of a default sub-parameter. The test name implies the *block* behavior, but the assertions don't distinguish the two cases.

**Impact:**
A regression that changes the policy from "block when one" to "remove then auto-add a fresh one" would still pass this test. The shared polarity/charge values, the underlying `Parameters` reference, and the identity of the surviving sub-parameter are all untested.

**Suggested fix:**
Capture the sub-parameter reference before `RemoveSubType`, then assert `Is.SameAs` afterward to verify it was not actually removed. Alternatively assert that `Parameters` is unchanged (same reference) before and after the call.

**Affected test scenarios:**
- RemoveSubType_BlockWhenOnlyOne - remove the only sub-parameter - the same sub-parameter instance remains and Parameters reference is unchanged

trishorts (commented):
Automated code review - 7 finding(s) at severity >= Major.

trishorts:
**[Major / Bug]** `SetAllPrecursorMaxChargeState` / `SetAllProductMaxChargeState` skip MultipleDecon sub-parameters

`MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/DeconHostViewModel.cs:109`

> _Posting as a top-level comment because the affected line isn't in this PR's diff._

**Problem:**
`SetAllPrecursorMaxChargeState` and `SetAllProductMaxChargeState` iterate `PrecursorDeconvolutionParametersList`/`ProductDeconvolutionParametersList` and assign `precursorParams.MaxAssumedChargeState = newMaxCharge`. When the element is a `MultipleDeconParamsViewModel`, this goes through the base property (the new VM does not override `MaxAssumedChargeState`), updating only the container's `_parameters.MaxAssumedChargeState` and leaving every sub-VM's max charge unchanged. The "shared" wrappers (`SharedMaxAssumedChargeState`) are only invoked through their own bindings, not through this method.

**Impact:**
External code paths (e.g., the loader/UI flow that calls `SetAllPrecursorMaxChargeState` when the user picks a higher max charge globally) leave Multiple Decon children stale. The container reports the new value, but actual deconvolution uses the children's old (lower) value, producing wrong charge ranges. This contradicts the PR's "shared polarity / shared charges" intent.

**Suggested fix:**
Either route the loop through the shared wrappers (cast to `MultipleDeconParamsViewModel` and call `SharedMaxAssumedChargeState = newMaxCharge` when applicable), or override `MaxAssumedChargeState` in `MultipleDeconParamsViewModel` so the base setter also propagates to `_subParameters`. The override is the cleaner fix because it also fixes any other call site that touches `MaxAssumedChargeState` on the base property.

trishorts:
**[Major / TestCoverage]** Duplicate test methods `TestDeconHostViewModel_GlobalVariables_Proteoform`

`MetaMorpheus/Test/GuiTests/Deconvolution/DeconHostViewModelTest.cs:162`

> _Posting as a top-level comment because the affected line isn't in this PR's diff._

**Problem:**
Two test methods exist with identical names and identical bodies: `TestDeconHostViewModel_GlobalVariables_Proteoform_Classic` (line ~136) and `TestDeconHostViewModel_GlobalVariables_Proteoform` (line ~162). They both set `GlobalVariables.AnalyteType = AnalyteType.Proteoform` and assert the same things. NUnit will silently treat these as two distinct tests, but the second is pure duplication and adds no coverage. The naming suggests one of them was intended to cover a different scenario (perhaps IsoDec proteoform) that was never written.

**Impact:**
Wasted test execution time, misleading test count, and an apparent IsoDec proteoform path that is actually never exercised. Future readers will be confused about which test is authoritative.

**Suggested fix:**
Delete one of the duplicate methods, or rewrite `TestDeconHostViewModel_GlobalVariables_Proteoform` to cover the IsoDec proteoform default behavior (the case that the `_Classic` suffix implies is missing).

trishorts:
**[Major / Correctness]** `TestEquals_SameParameters` asserts inequality despite identical parameter values

`MetaMorpheus/Test/GuiTests/Deconvolution/DeconParamsViewModelTest.cs:218`

> _Posting as a top-level comment because the affected line isn't in this PR's diff._

**Problem:**
The test name and arrangement clearly express the intent "two view models built from value-equal parameters should be equal," yet the assertions are `Is.False`. Either the assertion is wrong (the test should pass equality), or the test name is wrong. Given that `TestGetHashCode_UniqueIdentifier` and `TestEquals_DifferentUniqueIdentifier` document that equality is identity-based, the test is asserting the right thing but is named in a way that contradicts its assertions.

**Impact:**
A reader reasonably expects this test to verify value equality; it instead verifies identity inequality. Confusing test name silently encodes the opposite behavior, which masks real bugs if the equality contract is later changed.

**Suggested fix:**
Rename the test to something like `TestEquals_SameParameterValues_DifferentInstances_NotEqual`, or document with a comment that equality is identity-based and value-equal instances must compare unequal.

trishorts:
**[Major / TestCoverage]** `TestEquals_DifferentParameters` test cases collapse to the same comparison

`MetaMorpheus/Test/GuiTests/Deconvolution/DeconParamsViewModelTest.cs:174`

> _Posting as a top-level comment because the affected line isn't in this PR's diff._

**Problem:**
`TestEquals_DifferentParameters` parameterizes `viewModel1`'s charge/polarity but always compares against a fixed `parameters2 = (2, 15, Polarity.Negative)`. Because equality is identity-based (per `TestGetHashCode_UniqueIdentifier`), every case is guaranteed to return false regardless of the actual `minCharge`/`maxCharge`/`polarity` arguments — even the `(2, 15, Polarity.Negative)` case where the parameters are value-equal. The `TestName` labels ("MinChargeDiffers_NotEqual", etc.) suggest the test was intended to verify value-based equality semantics but actually proves nothing about field comparisons.

**Impact:**
The test names mislead readers into believing field-level equality checks exist. The same applies to `TestEquals_DifferentParameters_obj`. If the equality contract changes from identity to value-based, these tests will not catch regressions in field comparisons.

**Suggested fix:**
Either delete the parameterized `TestCase` attributes (since identity equality makes them redundant) and keep one example, or rewrite the test to genuinely vary `viewModel2`'s underlying parameters and assert value-based equality.

MetaMorpheus/GUI/Views/Deconvolution/MultipleDeconParamControl.xaml.cs
nbollis: Fixed. MultipleDeconParamsViewModel now stores an _isPrecursor flag passed from DeconHostViewModel at construction. AddSubType uses this flag instead of the hardcoded true default, so product-side additions inherit product charge defaults. Added tests covering both precursor and product modes.

MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/MultipleDeconParamsViewModel.cs
nbollis: Fixed. Made DeconParamsViewModel.Polarity setter virtual and changed MultipleDeconParamsViewModel.Polarity from new to override. Base-reference writes now propagate to all sub-VMs regardless of static type. Added tests for base-reference propagation.

MetaMorpheus/GuiFunctions/MzLibExtensions.cs
nbollis: Fixed. GetDefaultViewModel now returns null when GetDefaultDeconParams returns null, avoiding the fallthrough to NotImplementedException. The host constructor handles FromFile initial parameters as null and skips ExampleNewDeconvolutionTemplate explicitly.

MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/MultipleDeconParamsViewModel.cs
nbollis: Fixed after design reconsideration. Shared polarity and charge values are a user-controlled 'set all at once' convenience (via the Shared* wrappers and overridden Polarity), not a continuous invariant. RebuildParameters no longer overwrites per-sub customizations—it only reconstructs MultipleDeconParameters from current sub-VM states. Removed the test that asserted the old reset-behavior.

MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/DeconHostViewModel.cs
nbollis: Fixed. Removed the broad catch (NotImplementedException) from DeconHostViewModel. GetDefaultViewModel now returns null for unsupported types (FromFile, ExampleNewDeconvolutionTemplate), and the host loop explicitly skips those types without a catch block.

MetaMorpheus/Test/GuiTests/Deconvolution/MultipleDeconParamsViewModelTest.cs
nbollis: Fixed. RemoveSubType_BlockWhenOnlyOne now asserts Is.SameAs on the remaining sub-parameter reference, verifying the block is genuine and not a remove-then-re-add.

MetaMorpheus/GUI/Views/Deconvolution/MultipleDeconParamControl.xaml
nbollis: We initially applied the suggested fix (AllowNegative='False'), but reverted it after testing with oligo mode. Oligo defaults use negative charge states, so the shared charge controls must allow negative input to support all analyte types. Polarity remains managed via the dedicated polarity setter; the shared charge textboxes need the flexibility for negative-mode defaults.

nbollis:
Fixed. SetAllPrecursorMaxChargeState and SetAllProductMaxChargeState now route through SharedMaxAssumedChargeState when the element is a MultipleDeconParamsViewModel, so sub-VMs synchronize correctly.

nbollis:
Fixed. Removed the duplicate TestDeconHostViewModel_GlobalVariables_Proteoform test. The _Classic variant remains as the authoritative proteoform test.

nbollis:
Fixed. Renamed to TestEquals_SameParameters_DifferentInstancesNotEqual to clarify the test verifies reference inequality for value-equal instances.

nbollis:
Fixed. Collapsed TestEquals_DifferentParameters and TestEquals_DifferentParameters_obj into a single TestEquals_DifferentInstancesAreNotEqual test, since identity equality makes the parameterized cases redundant.

MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/DeconHostViewModel.cs
```diff
@@ -38,94 +38,41 @@ public DeconHostViewModel(DeconvolutionParameters? initialPrecursorParameters =
...
-                        };
-                        ProductDeconvolutionParametersList.Add(toAdd.ToViewModel());
-                    }
-
-                    break;
-            }
+            if (deconType == DeconvolutionType.ExampleNewDeconvolutionTemplate || deconType == DeconvolutionType.FromFile)
+                continue;
+
+            if (initialPrecursorParameters != null && !precursorIsFromFile && initialPrecursorParameters.DeconvolutionType == deconType)
```
trishorts, line 51: **Major** - NotImplementedException no longer caught; constructor will throw

**Problem:**
The base file wrapped the per-type `ToViewModel`/`GetDefaultViewModel` calls in a `try/catch (NotImplementedException)` block so a decon type without a default would simply be skipped. The new constructor removes the `try/catch` entirely and only filters out `ExampleNewDeconvolutionTemplate` and `FromFile`. The PR scope explicitly lists "Fix: `NotImplementedException` catch handling" but the resulting code has no catch at all — any new `DeconvolutionType` added to mzLib in future without a registered default will now propagate `NotImplementedException` out of `DeconHostViewModel`'s constructor and crash the task window on open.

**Impact:**
The class becomes brittle to additions in the underlying `DeconvolutionType` enum. Where the old behavior was "log and skip", the new behavior is "crash the GUI on open". This is a regression in robustness for a change the PR description claims as a fix.

**Suggested fix:**
Either restore a `try/catch (NotImplementedException)` around the per-type adds (with a Console/log message and `continue`), or extend the skip list to enumerate the exact types this PR intends to support. If the intent of the "fix" was to remove the catch because it masked real bugs, document that intent and ensure every non-skipped `DeconvolutionType` actually has a default registered for both precursor and product use.

**Affected test scenarios:**
- DeconHostViewModelTest - constructor when a non-skipped DeconvolutionType lacks a default - should not throw
- DeconHostViewModelTest - constructor with a new fake DeconvolutionType without defaults - should skip and continue, not propagate

---

MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/DeconHostViewModel.cs
```diff
@@ -38,94 +38,41 @@ public DeconHostViewModel(DeconvolutionParameters? initialPrecursorParameters =
         PrecursorDeconvolutionParametersList = new ObservableCollection<DeconParamsViewModel>();
         ProductDeconvolutionParametersList = new ObservableCollection<DeconParamsViewModel>();
 
+        // FromFile parameters are not supported in the GUI; treat them as null for list construction
+        bool precursorIsFromFile = initialPrecursorParameters?.DeconvolutionType == DeconvolutionType.FromFile;
```
trishorts, line 42: **Major** - Initial `MultipleDeconParameters` is downgraded to a Classic VM on load

**Problem:**
When `initialPrecursorParameters` is a `MultipleDeconParameters` instance, the loop enters the branch `initialPrecursorParameters.DeconvolutionType == deconType`, then calls `initialPrecursorParameters.ToViewModel(true)`. However, the enum loop iterates only `ClassicDeconvolution` and `IsoDecDeconvolution`-style entries (skipping `ExampleNewDeconvolutionTemplate` and `FromFile`). If `MultipleDecon` is not a `DeconvolutionType` enum value, the multi-decon parameters will never match any iteration, and the constructor falls through to selecting the Classic default. If `MultipleDecon` *is* an enum value, then no `MultipleDeconParamsViewModel` will be created in the list for users who did not start with multi-decon, so they cannot pick "Multiple" from the combo box — yet the scope says this PR is wiring multi-decon into the GUI.

**Impact:**
Either (a) saved-task settings that round-trip a `MultipleDeconParameters` won't restore the multi-decon view on load and silently revert to Classic, or (b) the multi-decon option won't be selectable from the combo box at all because it's never added to `PrecursorDeconvolutionParametersList`/`ProductDeconvolutionParametersList`. Either way the GUI feature this PR introduces is broken for the canonical use case.

**Suggested fix:**
Explicitly add a `MultipleDeconParamsViewModel` to both lists during construction (independent of the enum loop, since multi-decon is a composite of others), and select it when `initialPrecursorParameters`/`initialProductParameters` is a `MultipleDeconParameters`. Verify by walking the saved/loaded path end-to-end.

**Affected test scenarios:**
- DeconHostViewModelTest - constructor with MultipleDeconParameters as initialPrecursor - selected precursor VM should be MultipleDeconParamsViewModel
- DeconHostViewModelTest - constructor with null initial params - MultipleDeconParamsViewModel should appear in the parameters list so user can pick it
- DeconHostViewModelTest - round-trip MultipleDeconParameters through save/load - same parameters retrieved

---

MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/MultipleDeconParamsViewModel.cs
```diff
@@ -0,0 +1,129 @@
...
+        RebuildParameters();
+    }
+
+    private void RebuildParameters()
+    {
+        _parameters = new MultipleDeconParameters(
+            _subParameters.Select(s => s.Parameters),
+            _parameters.MinAssumedChargeState,
+            _parameters.MaxAssumedChargeState,
+            _parameters.Polarity,
```
trishorts, line 141: **Major** - `RebuildParameters` snapshots stale shared fields, losing user edits to Min/Max/Polarity

**Problem:**
`RebuildParameters` constructs a new `MultipleDeconParameters` using `_parameters.MinAssumedChargeState`, `_parameters.MaxAssumedChargeState`, `_parameters.Polarity`, etc. — i.e., the values from the *old* underlying `_parameters` reference. But the inherited `MinAssumedChargeState`/`MaxAssumedChargeState`/`Polarity` setters write directly into `Parameters` (the same reference), so before `RebuildParameters` runs they may already reflect the user's edits, *or* they may have been bypassed if `SharedMinAssumedChargeState` propagated to subs and somewhere else updated `_parameters` via reassignment. After `AddSubType` calls `RebuildParameters`, `_parameters` is replaced with a new object, but `base.Parameters` (via the abstract getter) still returns `_parameters` — that's fine — yet the property-change notification fires `nameof(Parameters)`, not `nameof(MinAssumedChargeState)`/`nameof(MaxAssumedChargeState)`/`nameof(Polarity)`, so UI bindings on those base properties may not re-read the now-different backing object. Combined with the new `PropertyChanged += …` re-fire wiring in the constructor (which fires `Shared*` when base changes), there is no symmetric fire of base property notifications when `Parameters` is swapped wholesale by `RebuildParameters`.

**Impact:**
After Add/Remove of a sub-type, the bound Min/Max/Polarity controls on the shared header may stop reflecting the underlying object, causing the displayed numbers to desync from what the next pipeline run will actually use. The PR scope lists "RebuildParameters no longer overwrites per-sub customizations" and "keeps shared-field consistency" as fixes — this path still has a stale-snapshot smell that warrants explicit verification.

**Suggested fix:**
After replacing `_parameters` in `RebuildParameters`, raise property-changed for `MinAssumedChargeState`, `MaxAssumedChargeState`, `Polarity`, `SharedMinAssumedChargeState`, `SharedMaxAssumedChargeState` so bindings refresh. Alternatively, mutate the existing `_parameters.Parameters` collection in place instead of constructing a new `MultipleDeconParameters`, which avoids the snapshot/notification problem entirely.

**Affected test scenarios:**
- MultipleDeconParamsViewModelTest - AddSubType after editing SharedMaxAssumedChargeState - rebuilt parameters reflect edited value, UI gets notification
- MultipleDeconParamsViewModelTest - RemoveSubType after editing Polarity - rebuilt parameters reflect edited polarity
- MultipleDeconParamsViewModelTest - PropertyChanged events fire for MinAssumedChargeState/MaxAssumedChargeState/Polarity after AddSubType

---

MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/MultipleDeconParamsViewModel.cs
```diff
@@ -0,0 +1,129 @@
...
+        get => base.Polarity;
+        set
+        {
+            base.Polarity = value;
+            foreach (var sub in _subParameters)
+                sub.Polarity = value;
+        }
+    }
+
+    public void AddSubType(DeconvolutionType type)
```
trishorts, line 115: **Major** - `SharedMaxAssumedChargeState` setter order can be rejected when shrinking

**Problem:**
The setter assigns `MaxAssumedChargeState = value` first (base setter), then iterates subs setting their `MaxAssumedChargeState`. The base `MaxAssumedChargeState` setter's switch rejects positive values that are `< MinAssumedChargeState`. If the user is *decreasing* the shared max from, say, 50 to 5 while a sub-VM's `MinAssumedChargeState` is currently 10 (because a prior `SharedMinAssumedChargeState` propagation only updated some subs), the per-sub assignment `sub.MaxAssumedChargeState = value` is silently rejected by that sub's validator, and the sub's max remains 50. There is no symmetric "set the min first, then max (or vice versa) depending on direction" handling. `SharedMinAssumedChargeState` has the same problem in the opposite direction.

**Impact:**
Shared charge updates can silently fail to propagate to one or more subs, producing inconsistency between the displayed shared value and the actual per-sub deconvolution parameters that will run. The user sees a single shared field but the underlying objects diverge.

**Suggested fix:**
In the shared setters, when shrinking the range, write the constraining side last (e.g., when lowering `Max`, lower `Min` first if needed; when raising `Min`, raise `Max` first if needed). Or, more simply, push directly into each sub's underlying `Parameters.Max/MinAssumedChargeState` to bypass the per-sub validator since the shared wrapper has already validated. Add a test that drops the shared max well below several subs' current min and asserts every sub ends up consistent.

**Affected test scenarios:**
- MultipleDeconParamsViewModelTest - SharedMaxAssumedChargeState lowered below current sub MinAssumedChargeState - all sub max values updated consistently
- MultipleDeconParamsViewModelTest - SharedMinAssumedChargeState raised above current sub MaxAssumedChargeState - all sub min/max updated consistently

---

MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/MultipleDeconParamsViewModel.cs
```diff
@@ -0,0 +1,129 @@
...
+                sub.MinAssumedChargeState = value;
+        }
+    }
+
+    public int SharedMaxAssumedChargeState
+    {
+        get => MaxAssumedChargeState;
+        set
+        {
+            MaxAssumedChargeState = value;
```
trishorts: **Major** - Constructor's `PropertyChanged += …` handler creates a self-recursive loop

**Problem:**
The constructor subscribes to its own `PropertyChanged` event. When `MinAssumedChargeState` is set, `OnPropertyChanged(nameof(MinAssumedChargeState))` fires, the handler runs and calls `OnPropertyChanged(nameof(SharedMinAssumedChargeState))`. So far benign. But the `SharedMinAssumedChargeState` setter itself calls `MinAssumedChargeState = value` (which fires `MinAssumedChargeState`, which fires `SharedMinAssumedChargeState` again via the handler) — no infinite loop *yet* because the handler doesn't re-set, only notifies. However: each `OnPropertyChanged` from `SharedMin/MaxAssumedChargeState` itself triggers the same handler again, and the handler will notify *yet again* for the *same* event. The handler is unconditional re-fire, with no guard against re-entry, and is wired permanently for the lifetime of the VM. This is fragile and confusing — a future maintainer adding any cross-property notification will create a real infinite loop.

**Impact:**
Each base `MinAssumedChargeState` change now causes UI re-binding work N+1 times (once for the base property, once for the shared property — and any future cross-property notification compounds this). More importantly the construct invites a real infinite-recursion bug in the next edit. The commit message "Shared charge state wrappers raise property-change notifications on direct base property changes" describes a legitimate need but the implementation conflates "notify" with "re-fire arbitrary chains" via the event itself.

**Suggested fix:**
Override the protected `OnPropertyChanged` (or the base property setters via `new`) and, when the property is `MinAssumedChargeState`/`MaxAssumedChargeState`, raise the corresponding `Shared*` notification inline — without subscribing to your own event. This avoids the event-loop hazard and makes the relationship explicit at the call site.

**Affected test scenarios:**
- MultipleDeconParamsViewModelTest - setting MinAssumedChargeState fires PropertyChanged for both MinAssumedChargeState and SharedMinAssumedChargeState exactly once each
- MultipleDeconParamsViewModelTest - rapid alternating sets do not produce exponential notification counts

---

MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/MultipleDeconParamsViewModel.cs
```diff
@@ -0,0 +1,129 @@
...
+    public void AddSubType(DeconvolutionType type)
+    {
+        var defaultParams = type.GetDefaultDeconParams(GlobalVariables.AnalyteType, _isPrecursor);
+        var subVm = defaultParams.ToViewModel(_isPrecursor);
+        subVm.Polarity = _parameters.Polarity;
+        subVm.MinAssumedChargeState = _parameters.MinAssumedChargeState;
+        subVm.MaxAssumedChargeState = _parameters.MaxAssumedChargeState;
+        _subParameters.Add(subVm);
+        RebuildParameters();
+    }
```
trishorts, line 124: **Major** - Sub-parameters created via `parameters.Parameters.Select(p => p.ToViewModel())` lose the `isPrecursor` flag

**Problem:**
In the constructor, sub-VMs are constructed via `parameters.Parameters.Select(p => p.ToViewModel())` — no argument is passed. Compare with the new `AddSubType`, which uses `defaultParams.ToViewModel(_isPrecursor)`. If `ToViewModel`'s parameterless overload defaults to `isPrecursor: true`, then any `MultipleDeconParamsViewModel` constructed with `isPrecursor: false` (i.e., for product/MS2 deconvolution) will end up with sub-VMs that were configured as precursor VMs. This mismatch can affect things like the per-VM defaults for max charge, polarity handling, and which UI controls show. The PR explicitly added the `isPrecursor` parameter to this VM yet only uses it in `AddSubType`, not when wrapping the initial sub-parameters.

**Impact:**
A multi-decon configuration loaded for product deconvolution will display/edit its initial sub-VMs as if they were precursors, while newly added sub-VMs are correctly tagged as products. Behavior diverges depending on whether a sub came from the initial load or was added later.

**Suggested fix:**
Pass `_isPrecursor` to `ToViewModel` for the initial sub-VMs in the constructor as well, matching what `AddSubType` does. Add a regression test that constructs a `MultipleDeconParamsViewModel(parameters, isPrecursor: false)` and verifies the wrapped sub-VMs reflect the product-side defaults.

**Affected test scenarios:**
- MultipleDeconParamsViewModelTest - constructed with isPrecursor=false - sub-VMs use product-side defaults
- MultipleDeconParamsViewModelTest - sub-VM created in ctor vs sub-VM added later - both behave consistently with respect to isPrecursor

---

MetaMorpheus/Test/GuiTests/Deconvolution/MultipleDeconParamsViewModelTest.cs
```diff
@@ -0,0 +1,306 @@
...
+        _viewModel.SharedMaxAssumedChargeState = -5;
+
+        _viewModel.RemoveSubType(_viewModel.SubParameters.Last());
+
+        Assert.That(_viewModel.MinAssumedChargeState, Is.EqualTo(-10));
+        Assert.That(_viewModel.MaxAssumedChargeState, Is.EqualTo(-5));
+        Assert.That(_viewModel.Polarity, Is.EqualTo(Polarity.Negative));
+        Assert.That(_viewModel.SubParameters.Count, Is.EqualTo(1));
+        var sub = _viewModel.SubParameters.Single();
+        Assert.That(sub.MinAssumedChargeState, Is.EqualTo(-10));
```
trishorts, line 233: **Major** - AddSubType precursor/product default tests don't actually verify defaults

**Problem:**
`AddSubType_OnPrecursorHost_UsesPrecursorDefaults` and `AddSubType_OnProductHost_UsesProductDefaults` are intended to verify that the production-side fix "AddSubType on MultipleDeconParamsViewModel sets correct product defaults" (commit ab1072e8) actually applies the right defaults for the host (precursor vs product). However, both tests construct the view model where the shared `MaxAssumedChargeState` already equals the asserted "default" (12 for the precursor `_viewModel` whose `_innerParams` is `(1, 12, 4, 3)`; 10 for `productVm` whose `productInner` is `(1, 10, 4, 3)`). Because `AddSubType_InheritsCurrentSharedValues` proves the implementation simply inherits the current shared values, both tests pass identically whether `AddSubType` truly applies the host-specific default or just inherits the shared value. They cannot detect a regression where `AddSubType` ignores the `isPrecursor` flag.

**Impact:**
The PR's claimed fix for product defaults in `AddSubType` is unverified. A regression that always uses precursor defaults (or always inherits) would not be caught by these tests, defeating the purpose of adding them.

**Suggested fix:**
Make the shared/inner values deliberately different from the host's defaults before constructing the view model (e.g., construct a precursor `MultipleDeconParamsViewModel` whose inner params have `Max=7`, then call `AddSubType` and assert the added sub has `Max=12` — the precursor default — rather than 7). The product variant should do the symmetric construction so that inheriting vs. defaulting produce distinct values.

**Affected test scenarios:**
- AddSubType - inner params differ from host defaults on precursor host - added sub gets precursor default (e.g. 12), not the inner/shared value
- AddSubType - inner params differ from host defaults on product host - added sub gets product default (e.g. 10), not the inner/shared value

trishorts (commented):
Automated code review - 7 finding(s) at severity >= Major.

trishorts:
**Major** — Global `AnalyteType` mutated on dialog open and never reset on cancel/non-top-down

*(Posted as a top-level comment because the affected line is not part of this PR's diff. Same pattern was introduced in three task windows by this PR; the third one — `CalibrateTaskWindow.xaml.cs:65` — was caught by the engine but couldn't be anchored inline.)*

**File:** `MetaMorpheus/GUI/TaskWindows/CalibrateTaskWindow.xaml.cs:65` (also present in `GptmdTaskWindow.UpdateFieldsFromTask` ~line 86 and `SearchTaskWindow.UpdateFieldsFromTask` ~line 320)

**Problem:**
`UpdateFieldsFromTask` now writes `GlobalVariables.AnalyteType = AnalyteType.Proteoform` whenever the loaded task uses the "top-down" protease. This runs during window construction — before the user has confirmed anything — and there is no corresponding `else` branch that resets the global back when the task is bottom-up. As a result: (a) opening a top-down task and then clicking Cancel leaves the global in `Proteoform`, and (b) subsequently opening a bottom-up task does not restore it.

**Impact:**
`GlobalVariables.AnalyteType` is a process-wide singleton. Once it flips to `Proteoform` from merely *viewing* a top-down task, every downstream consumer (other windows, result formatting, default settings) sees a stale value that does not match the task the user is actually editing or running. Users who experiment with a default top-down task and then switch to a peptide workflow without restarting the GUI will get behaviour appropriate to proteoform mode.

**Suggested fix:**
Either move the assignment into `SaveButton_Click` (so it only fires when the user actually commits the task) and add an explicit `else` that sets the analyte back to `Peptide`, or perform the assignment in a single helper that takes both branches and is invoked from save, not from dialog initialisation. Apply the same correction in all three task windows.

MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/DeconHostViewModel.cs
nbollis: DeconvolutionType.Multiple is a real enum member in the current mzLib build. The enum loop in DeconHostViewModel iterates over it and both the initial-params match and the GetDefaultViewModel fallback handle it correctly. No downgrade occurs.

MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/MultipleDeconParamsViewModel.cs
nbollis: The PropertyChanged handler only calls OnPropertyChanged (never sets properties), so there's no infinite loop at runtime. It does cause double notification per change. Overriding OnPropertyChanged instead would be cleaner; we can refactor in a follow-up.

MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/DeconHostViewModel.cs
nbollis: Fixed. Added a null guard in the constructor loop — future DeconvolutionType values without defaults are skipped rather than creating a null entry that would crash First().

MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/MultipleDeconParamsViewModel.cs
nbollis: Fixed. Added OnPropertyChanged for SharedMinAssumedChargeState and SharedMaxAssumedChargeState after RebuildParameters so the UI re-reads shared values.

MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/MultipleDeconParamsViewModel.cs
nbollis: Fixed. Added a TrySetAllSubParameters helper that snapshots sub values, attempts to push the shared value to all subs, and rolls back if any sub rejects (via base class validators). The shared property only updates when all subs accept.

MetaMorpheus/GuiFunctions/ViewModels/Deconvolution/MultipleDeconParamsViewModel.cs
nbollis: Fixed. Constructor now passes _isPrecursor to ToViewModel() so initial sub-parameters use the correct precursor or product defaults.

MetaMorpheus/Test/GuiTests/Deconvolution/MultipleDeconParamsViewModelTest.cs
nbollis: Fixed. Refactored both tests to use VMs with shared values deliberately different from expected defaults (Max=5 vs defaults 12/10). Renamed to InheritsSharedValues to match actual AddSubType behavior. Tests now genuinely verify the isPrecursor flag works.

nbollis:
Fixed. Replaced hardcoded Proteoform assignment with task.CommonParameters.DetermineAnalyteType() in SearchTaskWindow, GPTMDTaskWindow, and CalibrateTaskWindow. This correctly sets the analyte type for all task variants (Peptide, Proteoform, Oligo) and no longer mutates global state on dialog open.

Alexander-Sol (dismissed):

trishorts (dismissed):

RayMSMS (approved):

trishorts (approved):

Alexander-Sol (approved):

PR #2654, opened 2026-05-16
Deconvolution Parameters in FileSpecificParameters

nbollis:
added decon precursor and product to file specific parameters. 

This will come in handy later when we add in from file deconvolution.

### DIFF
--- a/MetaMorpheus/GUI/Views/Deconvolution/HostDeconParamControl.xaml
+++ b/MetaMorpheus/GUI/Views/Deconvolution/HostDeconParamControl.xaml
@@ -100,6 +100,7 @@
                 <!-- Precursor Decon Type ComboBox -->
                 <ComboBox Grid.Column="2" x:Name="PrecursorDeconvolutionTypeComboBox" Width="100" Height="30"
                           VerticalContentAlignment="Center" HorizontalContentAlignment="Center"
+                          IsEnabled="{Binding DoPrecursorDeconvolution}"
                           ItemsSource="{Binding PrecursorDeconvolutionParametersList}"
                           SelectedItem="{Binding PrecursorDeconvolutionParameters}"
                           Visibility="{Binding DisplayDeconSelectionComboBox, Converter={StaticResource BooleanToVisibilityConverter}}"
@@ -115,11 +116,13 @@
             <!-- Specific Precursor Parameters -->
             <ContentControl Grid.Row="1" Margin="5" x:Name="PrecursorSpecificParams" Content="{
                         Binding PrecursorDeconvolutionParameters,
-                        Converter={StaticResource DeconTypeToControlConverter }}" />
+                        Converter={StaticResource DeconTypeToControlConverter }}"
+                            IsEnabled="{Binding DoPrecursorDeconvolution}" />
 
             <!-- Specific Product Parameters -->
             <Expander Grid.Row="2" Header="MS2 Deconvolution" 
-                      Visibility="{Binding ShowMs2Section, RelativeSource={RelativeSource AncestorType=UserControl}, Converter={StaticResource BooleanToVisibilityConverter}}">
+                      Visibility="{Binding ShowMs2Section, RelativeSource={RelativeSource AncestorType=UserControl}, Converter={StaticResource BooleanToVisibilityConverter}}" 
+                      IsEnabled="{Binding DoPrecursorDeconvolution}">
                 <Border BorderBrush="{StaticResource BorderColor}" BorderThickness="1" CornerRadius="10">
                     <StackPanel>
                         <Grid HorizontalAlignment="Right">
--- a/MetaMorpheus/GUI/Views/Deconvolution/HostDeconParamControl.xaml
+++ b/MetaMorpheus/GUI/Views/Deconvolution/HostDeconParamControl.xaml
@@ -121,8 +121,7 @@
 
             <!-- Specific Product Parameters -->
             <Expander Grid.Row="2" Header="MS2 Deconvolution" 
-                      Visibility="{Binding ShowMs2Section, RelativeSource={RelativeSource AncestorType=UserControl}, Converter={StaticResource BooleanToVisibilityConverter}}" 
-                      IsEnabled="{Binding DoPrecursorDeconvolution}">
+                      Visibility="{Binding ShowMs2Section, RelativeSource={RelativeSource AncestorType=UserControl}, Converter={StaticResource BooleanToVisibilityConverter}}">
                 <Border BorderBrush="{StaticResource BorderColor}" BorderThickness="1" CornerRadius="10">
                     <StackPanel>
                         <Grid HorizontalAlignment="Right">
--- a/MetaMorpheus/TaskLayer/FileSpecificParameters.cs
+++ b/MetaMorpheus/TaskLayer/FileSpecificParameters.cs
@@ -70,6 +70,10 @@ namespace TaskLayer
                     case nameof(CustomIons):
                         CustomIons = keyValuePair.Value.Get<List<ProductType>>(); break;
 
+                    case nameof(PrecursorDeconvolutionParameters):
+                        PrecursorDeconvolutionParameters = keyValuePair.Value.Get<DeconvolutionParameters>(); break;
+                    case nameof(ProductDeconvolutionParameters):
+                        ProductDeconvolutionParameters = keyValuePair.Value.Get<DeconvolutionParameters>(); break;
 
                     default:
                         throw new MetaMorpheusException("Unrecognized parameter \"" + keyValuePair.Key + "\" in file-specific parameters toml");
@@ -94,6 +98,9 @@ namespace TaskLayer
         public string SeparationType { get; set; }
 
         public List<ProductType> CustomIons { get; set; }
+
+        public DeconvolutionParameters PrecursorDeconvolutionParameters { get; set; }
+        public DeconvolutionParameters ProductDeconvolutionParameters { get; set; }
 
         public FileSpecificParameters Clone()
         {
--- a/MetaMorpheus/TaskLayer/FileSpecificParameters.cs
+++ b/MetaMorpheus/TaskLayer/FileSpecificParameters.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using EngineLayer;
 using MzLibUtil;
 using Nett;
@@ -104,7 +104,22 @@ namespace TaskLayer
 
         public FileSpecificParameters Clone()
         {
-            return (FileSpecificParameters)this.MemberwiseClone();
+            return new FileSpecificParameters
+            {
+                PrecursorMassTolerance = this.PrecursorMassTolerance,
+                ProductMassTolerance = this.ProductMassTolerance,
+                ProductMassTolerance_LowRes = this.ProductMassTolerance_LowRes,
+                DigestionAgent = this.DigestionAgent,
+                MinPeptideLength = this.MinPeptideLength,
+                MaxPeptideLength = this.MaxPeptideLength,
+                MaxMissedCleavages = this.MaxMissedCleavages,
+                MaxModsForPeptide = this.MaxModsForPeptide,
+                DissociationType = this.DissociationType,
+                SeparationType = this.SeparationType,
+                CustomIons = this.CustomIons != null ? new List<ProductType>(this.CustomIons) : null,
+                PrecursorDeconvolutionParameters = this.PrecursorDeconvolutionParameters,
+                ProductDeconvolutionParameters = this.ProductDeconvolutionParameters
+            };
         }
     }
 }
--- a/MetaMorpheus/TaskLayer/FileSpecificParameters.cs
+++ b/MetaMorpheus/TaskLayer/FileSpecificParameters.cs
@@ -14,6 +14,9 @@ namespace TaskLayer
     {
         public FileSpecificParameters(TomlTable tomlTable)
         {
+            if (tomlTable == null)
+                throw new MetaMorpheusException("File-specific parameters TOML table cannot be null");
+
             foreach (var keyValuePair in tomlTable)
             {
                 switch (keyValuePair.Key)
--- a/MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
+++ b/MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
@@ -544,6 +544,8 @@ namespace TaskLayer
             Tolerance productMassTolerance_LowRes = fileSpecificParams.ProductMassTolerance_LowRes ?? commonParams.ProductMassTolerance_LowRes;
             DissociationType dissociationType = fileSpecificParams.DissociationType ?? commonParams.DissociationType;
             string separationType = fileSpecificParams.SeparationType ?? commonParams.SeparationType;
+            DeconvolutionParameters precursorDeconParams = fileSpecificParams.PrecursorDeconvolutionParameters ?? commonParams.PrecursorDeconvolutionParameters;
+            DeconvolutionParameters productDeconParams = fileSpecificParams.ProductDeconvolutionParameters ?? commonParams.ProductDeconvolutionParameters;
 
             CommonParameters returnParams = new CommonParameters(
                 dissociationType: dissociationType,
@@ -581,8 +583,8 @@ namespace TaskLayer
                 maxHeterozygousVariants: commonParams.MaxHeterozygousVariants,
                 minVariantDepth: commonParams.MinVariantDepth,
                 addTruncations: commonParams.AddTruncations,
-                precursorDeconParams: commonParams.PrecursorDeconvolutionParameters,
-                productDeconParams: commonParams.ProductDeconvolutionParameters,
+                precursorDeconParams: precursorDeconParams,
+                productDeconParams: productDeconParams,
                 useMostAbundantPrecursorIntensity: commonParams.UseMostAbundantPrecursorIntensity,
                 fragmentationParams: commonParams.FragmentationParameters);
 
@@ -1579,7 +1581,7 @@ namespace TaskLayer
         }
 
         /// <summary>
-        /// Legacy TOML compatibility — when ProductMassTolerance_LowRes is omitted, the helper falls back to ProductMassTolerance to keep constant result.
+        /// Legacy TOML compatibility ï¿½ when ProductMassTolerance_LowRes is omitted, the helper falls back to ProductMassTolerance to keep constant result.
         /// </summary>
         /// <typeparam name="TTask"></typeparam>
         /// <param name="filePath"></param>
--- a/MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
+++ b/MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
@@ -1585,7 +1585,7 @@ namespace TaskLayer
         }
 
         /// <summary>
-        /// Legacy TOML compatibility � when ProductMassTolerance_LowRes is omitted, the helper falls back to ProductMassTolerance to keep constant result.
+        /// Legacy TOML compatibility when ProductMassTolerance_LowRes is omitted, the helper falls back to ProductMassTolerance to keep constant result.
         /// </summary>
         /// <typeparam name="TTask"></typeparam>
         /// <param name="filePath"></param>
--- a/MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
+++ b/MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
@@ -544,12 +544,13 @@ namespace TaskLayer
             Tolerance productMassTolerance_LowRes = fileSpecificParams.ProductMassTolerance_LowRes ?? commonParams.ProductMassTolerance_LowRes;
             DissociationType dissociationType = fileSpecificParams.DissociationType ?? commonParams.DissociationType;
             string separationType = fileSpecificParams.SeparationType ?? commonParams.SeparationType;
+
             DeconvolutionParameters precursorDeconParams = fileSpecificParams.PrecursorDeconvolutionParameters ?? commonParams.PrecursorDeconvolutionParameters;
             DeconvolutionParameters productDeconParams = fileSpecificParams.ProductDeconvolutionParameters ?? commonParams.ProductDeconvolutionParameters;
 
-            // DoPrecursorDeconvolution flows from CommonParameters (not FSP) — deconv params from FSP are stored
-            // independently and will be used when DoPrecursorDeconvolution is true.
-
+            // DoPrecursorDeconvolution and DoProductDeconvolution flow from CommonParameters only;
+            // file-specific PrecursorDeconvolutionParameters / ProductDeconvolutionParameters are stored
+            // independently and take effect when the corresponding Do* flag is true.
             CommonParameters returnParams = new CommonParameters(
                 dissociationType: dissociationType,
                 precursorMassTolerance: precursorMassTolerance,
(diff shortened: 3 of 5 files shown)


MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -1579,7 +1581,7 @@ protected static void SanitizeBioPolymerDatabase<TBioPolymer>(List<TBioPolymer>
```
trishorts, line 1587: **Minor** - File-specific decon params can desync from scalar decon settings passed alongside

**Problem:**
The new code lets `fileSpecificParams.PrecursorDeconvolutionParameters` / `ProductDeconvolutionParameters` override the common decon params, but the constructor call still passes `deconvolutionMaxAssumedChargeState: commonParams.DeconvolutionMaxAssumedChargeState`, `deconvolutionIntensityRatio: commonParams.DeconvolutionIntensityRatio`, and `deconvolutionMassTolerance: commonParams.DeconvolutionMassTolerance` from the common params unchanged. Previously these three scalar settings and the decon params object always originated from the same `CommonParameters` instance, so they were guaranteed to agree. Now a user can ship a file-specific TOML whose `PrecursorDeconvolutionParameters.MaxAssumedChargeState` (or intensity ratio / mass tolerance) differs from the per-file `DeconvolutionMaxAssumedChargeState` settings being passed in, which leaves the returned `CommonParameters` internally inconsistent depending on which field downstream code reads.

**Impact:**
If downstream code (or the `CommonParameters` constructor itself) ever re-derives or recomputes anything from the scalar fields rather than from the decon params object — for example a guard, a log message, or a fallback construction — the per-file deconvolution will silently use a different charge range / intensity ratio / mass tolerance than the user configured in the file-specific TOML, producing wrong precursor envelopes without any warning.

**Suggested fix:**
When `fileSpecificParams.PrecursorDeconvolutionParameters` is supplied, pull the corresponding scalar fields (`MaxAssumedChargeState`, intensity ratio, deconvolution mass tolerance) from that object rather than from `commonParams`, so the returned `CommonParameters` is internally consistent. Same treatment for the product variant.

**Affected test scenarios:**
- SetAllFileSpecificCommonParams - file-specific TOML supplies `PrecursorDeconvolutionParameters` with `MaxAssumedChargeState` different from common - returned `CommonParameters.DeconvolutionMaxAssumedChargeState` should match the file-specific value
- SetAllFileSpecificCommonParams - file-specific TOML supplies `ProductDeconvolutionParameters` with a different intensity ratio - returned `CommonParameters.DeconvolutionIntensityRatio` should match the file-specific value

---

trishorts, line 1587: **Minor** - File-specific decon override is independent of `DoPrecursorDeconvolution`

**Problem:**
`SetAllFileSpecificCommonParams` now applies `fileSpecificParams.PrecursorDeconvolutionParameters` even when the file-specific TOML disables precursor deconvolution (the constructor still receives `doPrecursorDeconvolution: commonParams.DoPrecursorDeconvolution`, not anything from `fileSpecificParams`). Conversely, the file cannot toggle `DoPrecursorDeconvolution` on/off per file while it can replace the params used when deconvolution does run. The two switches were previously coupled through the single common source; the new override widens the file-specific surface without exposing the matching boolean.

**Impact:**
A user who sets file-specific `PrecursorDeconvolutionParameters` will reasonably expect that the file is "the source of truth" for precursor deconvolution behavior on that file, but the on/off boolean is still inherited from common. Either the override is silently inert (boolean off in common) or runs with the file-specific params despite the user expecting the common gate to apply — both are surprising.

**Suggested fix:**
Decide which behavior is intended: (a) also expose `DoPrecursorDeconvolution` (and the symmetric product setting) as file-specific so the override is coherent, or (b) ignore file-specific decon params when the corresponding `Do*Deconvolution` flag in `commonParams` is false and document that. At minimum, add a comment in `FileSpecificParameters` and a test that pins down the chosen semantics.

**Affected test scenarios:**
- SetAllFileSpecificCommonParams - common has `DoPrecursorDeconvolution = false`, file-specific supplies `PrecursorDeconvolutionParameters` - assert the documented behavior (either decon stays off or the per-file params drive a now-on decon)
- SetAllFileSpecificCommonParams - file-specific supplies only `ProductDeconvolutionParameters` - verify precursor decon is unchanged

---

MetaMorpheus/TaskLayer/FileSpecificParameters.cs
```diff
@@ -95,6 +99,9 @@ public FileSpecificParameters()
 
         public List<ProductType> CustomIons { get; set; }
```
trishorts, line 103: **Minor** - Clone() shallow-copies new DeconvolutionParameters reference, aliasing mutable state

**Problem:**
`Clone()` uses `MemberwiseClone()`, which performs a shallow copy. Previously, `FileSpecificParameters` held only value-typed nullables, `DigestionAgent` (effectively immutable per-instance), tolerance value objects, and `CustomIons` (already a shared reference). With this PR, two new reference-type properties `PrecursorDeconvolutionParameters` and `ProductDeconvolutionParameters` are added; after `Clone()`, both the original and the clone point at the same `DeconvolutionParameters` instance. If any consumer mutates the cloned params (e.g., adjusting `AverageResidueModel`, `Polarity`, `PhaseRes` on `IsoDecDeconvolutionParameters`), the change leaks back to the original — and to every other clone made before the mutation.

**Impact:**
Future code that calls `Clone()` to derive a per-file or per-task variant of deconv params and then tweaks one field on the clone will silently corrupt the source `FileSpecificParameters`. This is the kind of aliasing bug the existing `CustomIons` pre-existing behavior already invites; doubling down on it with two more mutable graph nodes increases the surface area, especially given the PR description that more "from file deconvolution" work is coming on top of this foundation.

**Suggested fix:**
Either deep-clone the two new deconv properties inside `Clone()` (e.g., delegate to a `DeconvolutionParameters.Clone()` if available), or document in an inline comment that `PrecursorDeconvolutionParameters`/`ProductDeconvolutionParameters` are intentionally shared and must be treated as immutable after assignment.

**Affected test scenarios:**
- TestFileSpecificParametersClone - mutate clone.PrecursorDeconvolutionParameters property - original PrecursorDeconvolutionParameters unchanged
- TestFileSpecificParametersClone - mutate clone.ProductDeconvolutionParameters property - original ProductDeconvolutionParameters unchanged

---

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -1579,7 +1581,7 @@ protected static void SanitizeBioPolymerDatabase<TBioPolymer>(List<TBioPolymer>
         }
 
         /// <summary>
-        /// Legacy TOML compatibility � when ProductMassTolerance_LowRes is omitted, the helper falls back to ProductMassTolerance to keep constant result.
+        /// Legacy TOML compatibility � when ProductMassTolerance_LowRes is omitted, the helper falls back to ProductMassTolerance to keep constant result.
         /// </summary>
```
trishorts, line 1598: **Minor** - Revert cosmetic-only change (CommentReword)

**Problem:**
A single XML-doc comment line on the legacy-TOML-compatibility helper was rewritten to identical-looking text (only an invisible character/encoding difference between the `-` and `+` lines).

**Impact:**
The documented sentence reads the same and the code it describes is unchanged, so compiled IL, runtime behaviour, and the public API surface are all identical. The author explicitly does not want stylistic churn mixed with substantive changes - it makes the PR harder to review.

**Suggested fix:**
Revert these lines to match the base branch. No functional impact - this is purely undoing noise.

**Revert patch:**
```diff
diff --git a/MetaMorpheus/TaskLayer/MetaMorpheusTask.cs b/MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
--- a/MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
+++ b/MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
@@ -1581,7 +1581,7 @@ namespace TaskLayer
         }
 
         /// <summary>
-        /// Legacy TOML compatibility � when ProductMassTolerance_LowRes is omitted, the helper falls back to ProductMassTolerance to keep constant result.
+        /// Legacy TOML compatibility � when ProductMassTolerance_LowRes is omitted, the helper falls back to ProductMassTolerance to keep constant result.
         /// </summary>
         /// <typeparam name="TTask"></typeparam>
         /// <param name="filePath"></param>
```

**Affected test scenarios:**
- _No tests needed - revert restores behaviour already covered by existing tests._

trishorts (commented):
Automated code review - 4 finding(s) at severity >= Informational.

trishorts:
**[Major / TestCoverage]** Clone test for new deconv properties is vacuous — both sides are always null

`MetaMorpheus/Test/ParameterTest.cs:38-39`

> _Posting as a top-level comment because the affected line isn't in this PR's diff._

**Problem:**
In `TestFileSpecificParametersClone`, the new assertions `Assert.That(fsp.PrecursorDeconvolutionParameters, Is.EqualTo(fspClone.PrecursorDeconvolutionParameters))` and the matching `ProductDeconvolutionParameters` assertion are added, but none of the six `TestCase` TOML fixtures (`testFileSpecfic_Protease.toml`, `testFileSpecfic_DigestionAgent*.toml`, `testFileSpecfic_RNA_*.toml`) include `PrecursorDeconvolutionParameters` or `ProductDeconvolutionParameters` keys. Both `fsp.PrecursorDeconvolutionParameters` and `fspClone.PrecursorDeconvolutionParameters` are therefore always `null`, so the assertion reduces to `Assert.That(null, Is.EqualTo(null))` and does not actually validate cloning of the new properties.

**Impact:**
The new code paths in `Clone()` for the deconvolution properties are not exercised. A regression that drops or mishandles either property during cloning would not be caught by this test, defeating the purpose of adding the assertions.

**Suggested fix:**
Either populate the test fixtures with deconvolution sections, or construct a `FileSpecificParameters` instance in-test that explicitly sets `PrecursorDeconvolutionParameters` and `ProductDeconvolutionParameters` to non-null `ClassicDeconvolutionParameters`/`IsoDecDeconvolutionParameters` values and clone that, then assert equality (and ideally non-null) on both properties.

trishorts:
**[Major / TestCoverage]** No positive test that file-specific deconv params override CommonParameters

`MetaMorpheus/Test/ParameterTest.cs:165-166, 282-283`

> _Posting as a top-level comment because the affected line isn't in this PR's diff._

**Problem:**
In both `TestFileSpecifcParameterOverwrite` and `TestFileSpecifcParameterOverwrite_WithRnaDigestionParams`, the only new assertions for the deconv properties (`updatedParameters.PrecursorDeconvolutionParameters`, `updatedParameters.ProductDeconvolutionParameters`) verify the "no-op" case: they assert that when the `FileSpecificParameters` does NOT set the deconv params, `updatedParameters` equals `notDefaultParameters`. The `basicFileSpecificParameters` instance also never sets `PrecursorDeconvolutionParameters` or `ProductDeconvolutionParameters`, so both code paths only test "fsp is null → don't overwrite". There is no in-process assertion that setting `FileSpecificParameters.PrecursorDeconvolutionParameters` to a non-null value actually causes `SetAllFileSpecificCommonParams` to replace `CommonParameters.PrecursorDeconvolutionParameters`.

**Impact:**
The core purpose of this PR — wiring file-specific deconv params through to `CommonParameters` — is not directly verified in unit form. If `MetaMorpheusTask.SetAllFileSpecificCommonParams` silently drops the deconv params (or never reads them), the unit tests pass; only the heavier `TestFileSpecificDeconvolutionParams` integration test would catch it, and only via indirect output diffing.

**Suggested fix:**
Add a third FSP construction in each test that sets `PrecursorDeconvolutionParameters` and `ProductDeconvolutionParameters` to instances distinct from `notDefaultParameters` (e.g., `IsoDec` vs `Classic`), call `SetAllFileSpecificCommonParams`, and assert that `updatedParameters.PrecursorDeconvolutionParameters` and `ProductDeconvolutionParameters` equal the file-specific values, not the not-default ones.

trishorts:
**[Minor / TestCoverage]** `TestFileSpecificDeconvolutionParams` can trivially pass if PSM lists are empty

`MetaMorpheus/Test/TestToml.cs:471-475, 504-508`

> _Posting as a top-level comment because the affected line isn't in this PR's diff._

**Problem:**
`TestFileSpecificDeconvolutionParams` validates the new file-specific deconvolution wiring by running two `EverythingRunnerEngine` configurations (one with the deconv params set directly on the task, one with them injected via file-specific TOML overriding the opposite type) and asserting `resultsClassicT.SequenceEqual(resultsClassicD)` and `resultsIsoDecT.SequenceEqual(resultsIsoDecD)` on the resulting `AllPSMs.psmtsv` files. If either deconvolution path returns zero PSMs for `PrunedDbSpectra.mzml` (or both fail and produce only an identical header row), `SequenceEqual` returns `true` and the test passes without proving anything about the override.

**Impact:**
A regression that causes either deconv path to silently produce empty results — or worse, causes the file-specific TOML to be ignored such that both runs use the same in-task params and yield identical (but wrong) output — could go undetected. The test is the only positive end-to-end verification of the new feature, so its assertion strength matters more than usual.

**Suggested fix:**
Before the `SequenceEqual` checks, assert that both result line arrays have length greater than the header row count (e.g., `> 1`), and ideally assert that `resultsClassicD` and `resultsIsoDecD` differ from each other so the "TOML override actually changed deconv behavior" claim is meaningful.

trishorts:
**[Informational / TestCoverage]** Cross-type decon override scenarios should be pinned down by a test

`MetaMorpheus/Test/ParameterTest.cs` (test additions not visible in diff)

> _Posting as a top-level comment because the affected line isn't in this PR's diff._

**Problem:**
The PR's whole point is enabling per-file decon params, and the TOML converter already supports switching between `ClassicDeconvolutionParameters` and `IsoDecDeconvolutionParameters` based on the `DeconvolutionType` key. Without a test that explicitly mixes types — e.g. common uses `IsoDecDeconvolution` and the file-specific TOML uses `ClassicDeconvolution` (or vice versa) — the round-trip and the "type swap" path are unverified, even though they are the most interesting new capability this PR enables.

**Impact:**
A user could write a file-specific TOML that intentionally swaps deconvolution algorithm for a single file and silently get the wrong algorithm if the converter or `SetAllFileSpecificCommonParams` fails to honor the type swap; that would not be detected by the suite as it stands.

**Suggested fix:**
Add a TestToml/ParameterTest case that (a) loads a file-specific TOML whose `PrecursorDeconvolutionParameters.DeconvolutionType` differs from the task's common decon type, (b) runs `SetAllFileSpecificCommonParams`, and (c) asserts the returned `CommonParameters.PrecursorDeconvolutionParameters` is of the file-specific concrete type. Add the symmetric test for `ProductDeconvolutionParameters`.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
nbollis: This is intended and in alignment with the current operations. If the users sets their normal decon params, and chooses not to run deconvolution, the parameters are not used. 

That said, I added some checks that disable the setting of decon params if the user does not select DoPrecursorDeconvolution

nbollis: This is intended and in alignment with the current operations. If the users sets their normal decon params, and chooses not to run deconvolution, the parameters are not used. 

That said, I added some checks that disable the setting of decon params if the user does not select DoPrecursorDeconvolution

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
nbollis: fixed

MetaMorpheus/Test/ParameterTest.cs
```diff
@@ -44,9 +44,40 @@ public static void TestFileSpecificParametersClone(string filePath)
...
+                ProductDeconvolutionParameters = new IsoDecDeconvolutionParameters(),
+                CustomIons = new List<ProductType> { ProductType.b, ProductType.y }
+            };
+
+            FileSpecificParameters fspClone = fsp.Clone();
+            Assert.That(fspClone, Is.Not.Null);
+
+            Assert.That(fspClone.PrecursorDeconvolutionParameters, Is.Not.Null);
+            Assert.That(fspClone.PrecursorDeconvolutionParameters, Is.EqualTo(fsp.PrecursorDeconvolutionParameters));
+            Assert.That(fspClone.PrecursorDeconvolutionParameters.MaxAssumedChargeState,
```
trishorts, line 67: **Major** - Clone "deep copy" test never verifies deep-copy semantics for deconv params

**Problem:**
`TestFileSpecificParametersClone_DeepCopiesDeconvolutionParams` claims to verify that `Clone()` deep-copies the deconvolution parameter objects, but every assertion against `PrecursorDeconvolutionParameters` and `ProductDeconvolutionParameters` is reference-equality between the same object - the current `Clone()` does `PrecursorDeconvolutionParameters = this.PrecursorDeconvolutionParameters`, so `fsp.PrecursorDeconvolutionParameters` and `fspClone.PrecursorDeconvolutionParameters` are literally the same reference. `Is.EqualTo(...)` and `MaxAssumedChargeState`-equality therefore pass trivially regardless of clone semantics. Only the `CustomIons` list mutation at the end actually proves independence, and that proves nothing about deconv params.

**Impact:**
The PR explicitly lists "Clone performs a deep copy of deconv params" as something the new test verifies. As written, the test would still pass if `Clone()` had `PrecursorDeconvolutionParameters = null` swapped in (no - that one would fail Is.Not.Null), but it would happily pass if a future change broke deep copy by sharing references. The named contract is not actually under test.

**Suggested fix:**
After cloning, mutate a writable property on `fspClone.PrecursorDeconvolutionParameters` (or `ProductDeconvolutionParameters`) and assert the corresponding property on `fsp.PrecursorDeconvolutionParameters` is unchanged - mirroring exactly what the existing `CustomIons.Add(...)` check does for the list. Add `Is.Not.SameAs` assertions for both deconv objects.

**Affected test scenarios:**
- TestFileSpecificParametersClone_DeepCopiesDeconvolutionParams - mutate clone's PrecursorDeconvolutionParameters - original unaffected
- TestFileSpecificParametersClone_DeepCopiesDeconvolutionParams - clone's deconv references are not the same instance as the original's

---

MetaMorpheus/TaskLayer/FileSpecificParameters.cs
```diff
@@ -95,9 +102,27 @@ public FileSpecificParameters()
...
```
trishorts, line 103: **Major** - Clone() shallow-copies deconv params despite scope stating deep copy

**Problem:**
The PR scope states: "Replace `MemberwiseClone` with explicit property copy in `FileSpecificParameters.Clone()` to ensure deep copy of deconvolution parameter objects." The new explicit-copy implementation does deep-copy `CustomIons` via `new List<ProductType>(this.CustomIons)`, but `PrecursorDeconvolutionParameters` and `ProductDeconvolutionParameters` are still assigned by reference (`= this.PrecursorDeconvolutionParameters`). Functionally this is identical to the old `MemberwiseClone` for those two fields - they're shared between original and clone.

**Impact:**
Anyone relying on the documented contract ("deep copy of deconvolution parameter objects") to safely mutate a cloned file-specific configuration without bleeding back into the original will observe shared state. If downstream from-file deconvolution code ever mutates the deconv-params object (e.g., assigning an `AverageResidueModel` per file), every other file sharing the same clone source will see the mutation.

**Suggested fix:**
Invoke a `Clone()`/`Copy()` on the deconv parameter objects when populating the clone (if `DeconvolutionParameters` exposes one in mzLib), or null-guard and clone explicitly per concrete type. If the deconv-params classes are intentionally immutable and reference-sharing is acceptable, update the PR description/scope to drop the "deep copy of deconvolution parameter objects" claim and tighten Finding-1's test accordingly.

**Affected test scenarios:**
- TestFileSpecificParametersClone_DeepCopiesDeconvolutionParams - mutate clone's deconv params - original unaffected (currently would fail under a strengthened test)

---

trishorts (commented):
Automated code review - 2 finding(s) at severity >= Major.

MetaMorpheus/TaskLayer/FileSpecificParameters.cs
nbollis: The current approach is good enough because we don't actually need a deep copy, a shallow is good enough. A proper deep copy will be enabled by https://github.com/smith-chem-wisc/mzLib/pull/1070 and a subsequent mzlib release. 

Renamed the testing method for clarity

MetaMorpheus/Test/ParameterTest.cs
nbollis: The current approach is good enough because we don't actually need a deep copy, a shallow is good enough. A proper deep copy will be enabled by https://github.com/smith-chem-wisc/mzLib/pull/1070 and a subsequent mzlib release. 

Renamed the testing method for clarity

Alexander-Sol (dismissed):

MetaMorpheus/GUI/Views/Deconvolution/HostDeconParamControl.xaml
```diff
@@ -115,11 +116,13 @@
...
-                        Converter={StaticResource DeconTypeToControlConverter }}" />
+                        Converter={StaticResource DeconTypeToControlConverter }}"
+                            IsEnabled="{Binding DoPrecursorDeconvolution}" />
 
             <!-- Specific Product Parameters -->
             <Expander Grid.Row="2" Header="MS2 Deconvolution" 
-                      Visibility="{Binding ShowMs2Section, RelativeSource={RelativeSource AncestorType=UserControl}, Converter={StaticResource BooleanToVisibilityConverter}}">
+                      Visibility="{Binding ShowMs2Section, RelativeSource={RelativeSource AncestorType=UserControl}, Converter={StaticResource BooleanToVisibilityConverter}}" 
+                      IsEnabled="{Binding DoPrecursorDeconvolution}">
                 <Border BorderBrush="{StaticResource BorderColor}" BorderThickness="1" CornerRadius="10">
```
trishorts, line 125: **Major** - MS2 (product) deconvolution section is gated on the precursor decon flag

**Problem:**
The "Specific Product Parameters" `Expander` (Header="MS2 Deconvolution") gained `IsEnabled="{Binding DoPrecursorDeconvolution}"`. This binds the entire MS2/product deconvolution UI — the product polarity combo (`ProductPolarityComboBox`), the product decon-type combo (`ProductDeconvolutionTypeComboBox`), and the `ProductSpecificParams` content — to the *precursor* deconvolution flag. Product (fragment) deconvolution is conceptually independent of whether precursor deconvolution is performed, and the available bindings expose a separate `ProductDeconvolutionParameters` / `ProductDeconvolutionParametersList` with no `DoPrecursorDeconvolution` relationship. This looks like the precursor `IsEnabled` binding was copy-pasted onto the product `Expander`.

**Impact:**
A user who disables "Deconvolute Precursors" can no longer view, expand, or edit any MS2/product deconvolution parameters, even though those parameters are independent and may still be needed. This also directly conflicts with the PR's stated invariant that the precursor/product `DeconvolutionParameters` are independent of `DoPrecursorDeconvolution`. Because `IsEnabled=false` on an `Expander` also blocks the expand/collapse toggle, the product section becomes completely inaccessible.

**Suggested fix:**
Bind the product `Expander`'s `IsEnabled` to a product-specific flag (or leave the product section always enabled, matching the prior behavior), rather than to `DoPrecursorDeconvolution`. The precursor flag should only gate precursor controls.

**Affected test scenarios:**
- DeconHostModel - DoPrecursorDeconvolution set false - product deconvolution parameters remain editable/accessible
- DeconHostModel - DoPrecursorDeconvolution toggled - precursor controls enable/disable but product (MS2) controls are unaffected

trishorts (commented):
Code review - 1 finding(s) posted.

MetaMorpheus/GUI/Views/Deconvolution/HostDeconParamControl.xaml
nbollis: fixed by removing the enabled binding in the ms2 decon params

trishorts (approved):

Alexander-Sol (approved):

zhuoxinshi (approved):

PR #2661, opened 2026-05-19
Add PrecursorDeconvolutionScore feature to PEP (bottom-up first)

trishorts:
# Add PrecursorDeconvolutionScore feature to PEP (bottom-up first)

## Summary

Wires mzLib's method-agnostic envelope-quality score (mzLib [#1054](https://github.com/smith-chem-wisc/mzLib/pull/1054), shipped in mzLib 1.0.579 as `IsotopicEnvelope.GenericScore`) into MetaMorpheus's PEP analysis pipeline as a new `PsmData.PrecursorDeconvolutionScore` feature. The score is computed once per precursor envelope using mzLib's cached `envelope.GetOrComputeGenericScore(deconvolutionParameters)` extension method and flows through the existing precursor-scalar plumbing (the same path that already carries `PrecursorScanIntensity`, `PrecursorScanEnvelopePeakCount`, and `PrecursorFractionalIntensity`) into the ML.NET training feature vector.

**Scope.** This PR adds the feature to the **bottom-up search type only**. Once it's approved and lands, the identical wiring will be extended to top-down, crosslink, and RNA in a follow-up using this PR as the template. The bottom-up-only restriction keeps the diff small, the test footprint focused, and the baseline shift readable.

## Measured impact (bottom-up benchmark)

`TaGe_SA_A549_3_snip` ×2 files, target counts at pep-q ≤ 0.01:

| Metric | Master | This PR | Δ |
|---|---:|---:|---:|
| All target PSMs | 382 | **384** | +2 |
| All target peptides | 153 | 153 | 0 |
| All target protein groups | 145 | **144** | −1 |
| Per-file PSMs | 190 | **191** | +1 |
| Per-file peptides | 153 | 153 | 0 |
| Per-file protein groups | 145 | **144** | −1 |

A small, mixed effect on this near-saturation dataset. The new feature reshuffles which PSMs land above the 1% pep-q threshold; per-file PSM count nudges up while the aggregate protein-group count drops by one (a few PSMs flip into a group already covered, no new group). The signal will probably matter more on larger or noisier datasets — same pattern observed when wiring Chronologer through the existing z-score path on this same benchmark.

### Data path (top → bottom)

| File | Change |
|---|---|
| `EngineLayer/Util/Precursor.cs` | New `double DeconvolutionScore { get; init; }` field on the record. |
| `TaskLayer/MetaMorpheusTask.cs` (line 274 deconvolution loop) | For each yielded `IsotopicEnvelope`, call `envelope.GetOrComputeGenericScore(commonParameters.PrecursorDeconvolutionParameters)` and stash on the `Precursor` via object initializer. mzLib's `GetOrComputeGenericScore` is idempotent and caches on the envelope, so re-asking the same envelope is free. |
| `TaskLayer/MetaMorpheusTask.cs` (line 331 PSM construction) | Pass `precursor.DeconvolutionScore` to the new `Ms2ScanWithSpecificMass` constructor arg. |
| `EngineLayer/Ms2ScanWithSpecificMass.cs` | New constructor parameter `double precursorDeconvolutionScore = 0` (source-compatible — existing callers unaffected); new `PrecursorDeconvolutionScore` property. |
| `EngineLayer/SpectralMatch.cs` | New `PrecursorScanDeconvolutionScore` property, populated from `scan.PrecursorDeconvolutionScore` in the constructor — mirrors the existing `PrecursorScanIntensity` / `PrecursorScanEnvelopePeakCount` / `PrecursorFractionalIntensity` pattern at lines 30–36. |
| `EngineLayer/FdrAnalysis/PsmData.cs` | New `[LoadColumn(29)] public float PrecursorDeconvolutionScore { get; set; }`. Added to `trainingInfos["standard"]` only. `assumedAttributeDirection["PrecursorDeconvolutionScore"] = 1` — higher generic score = better Averagine match = more likely target. |
| `EngineLayer/FdrAnalysis/PEPAnalysisEngine.cs::CreateOnePsmDataEntry` | One line in the `PsmData` initializer: `PrecursorDeconvolutionScore = (float)psm.PrecursorScanDeconvolutionScore`. |

### Sentinel-value handling

When the precursor was constructed from scan-header data alone (the `UseProvidedPrecursorInfo` path at `MetaMorpheusTask.cs:297–304`, used by some `.msalign` inputs), no envelope is deconvoluted, so there is nothing to score. The PSM's `PrecursorDeconvolutionScore` falls through as `0`.

This conflates "no envelope was computed" with "envelope was computed and is maximally bad" (both score 0). Per scope discussion, an accepted limitation for now; the model learns to treat `0` as a generic "no information" value either way. A future improvement could use `-1` to distinguish the two cases.

### Search types

`PrecursorDeconvolutionScore` is added to `trainingInfos["standard"]` only. The other entries in `PsmData.trainingInfos` — `"top-down"`, `"crosslink"`, `"RNA"` — are untouched. Top-down was the original use case the mzLib weights were calibrated against (IsoDec on yeast), so it's the highest-priority follow-up.

## Tests

**Full `Test.csproj` suite: 1,446 passed, 0 failed, 1 skipped, 7m 27s.**

### Updated baselines

- `FdrTest.TestPsmData` — `trainingInfo["standard"]` array gains `"PrecursorDeconvolutionScore"`; `positiveAttributes` gains the same entry; the test's `PsmData` initializer gains `PrecursorDeconvolutionScore = 28`; `standardToString` gains a trailing `\t28`. (Test still passes against the actual `PsmData.ToString("standard")` output.)
- `PostSearchAnalysisTaskTests.AllResultsAndResultsTxtContainsCorrectValues_PepQValue_BottomUp` — count assertions updated to match the new model outputs (see "Measured impact" table). No structural changes to the test, just the expected numeric values.

### Unchanged tests

- `PostSearchAnalysisTaskTests.AllResultsAndResultsTxtContainsCorrectValues_QValue_BottomUp` — passes without modification. Target-decoy q-value counts (vs PEP q-value counts) are not affected by this feature on this dataset.
- All other test classes pass without modification.

## Compatibility / rollout

- **Existing `Ms2ScanWithSpecificMass` call sites are source-compatible.** The new constructor parameter `double precursorDeconvolutionScore = 0` has a default value, so test mocks and external code that construct an `Ms2ScanWithSpecificMass` without supplying it continue to compile and get the sentinel `0`.
- **No `PsmData` schema break for downstream consumers.** The new `[LoadColumn(29)]` is appended; existing column indices 0–28 are unchanged. Pre-PR-trained ML.NET models will need re-training (expected for any feature addition that's actually wired into training).
- **mzLib bump.** `1.0.578 → 1.0.579` is already in master via [#2652](https://github.com/smith-chem-wisc/MetaMorpheus/pull/2652) — no further bump required from this PR.

## Plan after merge

Once this lands and is approved as the bottom-up model, the same wiring will be extended to the three remaining search types in `PsmData.trainingInfos` — adding `"PrecursorDeconvolutionScore"` to each, re-baselining whichever tests assert per-search-type counts (top-down has its own `AllResultsAndResultsTxt…` integration in `PostSearchAnalysisTaskTests`, and crosslink/RNA have their own count-asserting tests too). The data-path plumbing (`Precursor` → `Ms2ScanWithSpecificMass` → `SpectralMatch` → `PsmData`) is search-type-agnostic and won't need to change; the only follow-up edits will be the `PsmData.trainingInfos` entries and test baselines.

## Companion reading

- mzLib PR #1054 — [Generic Deconvolution Scoring for IsotopicEnvelope](https://github.com/smith-chem-wisc/mzLib/pull/1054) — the source of the score and the `GetOrComputeGenericScore` API used here. The mzLib PR's "Three access patterns / One-line ergonomic" section is precisely the pattern adopted in `MetaMorpheusTask.GetMs2Scans` here.

### DIFF
--- a/MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
+++ b/MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
@@ -1588,7 +1588,7 @@ namespace TaskLayer
         }
 
         /// <summary>
-        /// Legacy TOML compatibility � when ProductMassTolerance_LowRes is omitted, the helper falls back to ProductMassTolerance to keep constant result.
+        /// Legacy TOML compatibility when ProductMassTolerance_LowRes is omitted, the helper falls back to ProductMassTolerance to keep constant result.
         /// </summary>
         /// <typeparam name="TTask"></typeparam>
         /// <param name="filePath"></param>
--- a/MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
+++ b/MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
@@ -275,8 +275,8 @@ namespace TaskLayer
                                     precursorSpectrum.MassSpectrum, commonParameters.PrecursorDeconvolutionParameters))
                                 {
                                     double? intensity = null;
-                                    if (commonParameters.UseMostAbundantPrecursorIntensity) 
-                                        intensity = envelope.Peaks.Max(p => p.intensity); 
+                                    if (commonParameters.UseMostAbundantPrecursorIntensity)
+                                        intensity = envelope.Peaks.Max(p => p.intensity);
 
                                     var fractionalIntensity = envelope.TotalIntensity /
                                           precursorSpectrum.MassSpectrum.YArray
@@ -286,7 +286,15 @@ namespace TaskLayer
                                               precursorSpectrum.MassSpectrum.GetClosestPeakIndex(ms2scan.IsolationRange.Maximum)
                                           ].Sum();
 
-                                    precursorSet.Add(new(envelope, intensity, fractionalIntensity));
+                                    // Method-agnostic envelope-quality score from mzLib (idempotent: caches on the
+                                    // envelope, so re-asking the same envelope is cheap).
+                                    double genericScore = envelope.GetOrComputeGenericScore(
+                                        commonParameters.PrecursorDeconvolutionParameters);
+
+                                    precursorSet.Add(new Precursor(envelope, intensity, fractionalIntensity)
+                                    {
+                                        DeconvolutionScore = genericScore
+                                    });
                                 }
                             }
                         }
@@ -330,7 +338,8 @@ namespace TaskLayer
                             // assign precursor for this MS2 scan
                             var scan = new Ms2ScanWithSpecificMass(ms2scan, precursor.MonoisotopicPeakMz,
                                 precursor.Charge, fullFilePath, commonParameters, neutralExperimentalFragments,
-                                precursor.Intensity, precursor.EnvelopePeakCount, precursor.FractionalIntensity);
+                                precursor.Intensity, precursor.EnvelopePeakCount, precursor.FractionalIntensity,
+                                precursor.DeconvolutionScore);
 
                             // assign precursors for MS2 child scans
                             if (ms2ChildScans != null)
@@ -1579,7 +1588,7 @@ namespace TaskLayer
         }
 
         /// <summary>
-        /// Legacy TOML compatibility — when ProductMassTolerance_LowRes is omitted, the helper falls back to ProductMassTolerance to keep constant result.
+        /// Legacy TOML compatibility ï¿½ when ProductMassTolerance_LowRes is omitted, the helper falls back to ProductMassTolerance to keep constant result.
         /// </summary>
         /// <typeparam name="TTask"></typeparam>
         /// <param name="filePath"></param>
--- a/MetaMorpheus/EngineLayer/FdrAnalysis/PEPAnalysisEngine.cs
+++ b/MetaMorpheus/EngineLayer/FdrAnalysis/PEPAnalysisEngine.cs
@@ -640,6 +640,7 @@ namespace EngineLayer
                 MostAbundantPrecursorPeakIntensity = mostAbundantPrecursorPeakIntensity,
                 PrecursorFractionalIntensity = fractionalIntensity,
                 InternalIonCount = internalMatchingFragmentCount,
+                PrecursorDeconvolutionScore = (float)psm.PrecursorScanDeconvolutionScore,
             };
 
             return psm.PsmData_forPEPandPercolator;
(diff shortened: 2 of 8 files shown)


MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -286,7 +286,15 @@ public static List<Ms2ScanWithSpecificMass>[] _GetMs2Scans(MsDataFile myMSDataFi
                                               precursorSpectrum.MassSpectrum.GetClosestPeakIndex(ms2scan.IsolationRange.Maximum)
                                           ].Sum();
 
-                                    precursorSet.Add(new(envelope, intensity, fractionalIntensity));
+                                    // Method-agnostic envelope-quality score from mzLib (idempotent: caches on the
+                                    // envelope, so re-asking the same envelope is cheap).
+                                    double genericScore = envelope.GetOrComputeGenericScore(
```
nbollis, line 291: Should we pass in the spectrum as well to take advantage of the spectrum aware scoring you did?

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -1579,7 +1588,7 @@ protected static void SanitizeBioPolymerDatabase<TBioPolymer>(List<TBioPolymer>
         }
 
         /// <summary>
-        /// Legacy TOML compatibility � when ProductMassTolerance_LowRes is omitted, the helper falls back to ProductMassTolerance to keep constant result.
+        /// Legacy TOML compatibility � when ProductMassTolerance_LowRes is omitted, the helper falls back to ProductMassTolerance to keep constant result.
```
nbollis: Revert comment change. Both of our AI's keep making this weird comment changed where they add an unknown character.

nbollis (dismissed):

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
trishorts: i have a separate pr for that

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
trishorts: reverted

nbollis (approved):

Alexander-Sol (approved):

pcruzparri (approved):

PR #2663, opened 2026-05-24
feat: write a ProForma 2.0 column in search output (+ MetaDraw)

trishorts:
## Overview
This PR surfaces **ProForma 2.0** in MetaMorpheus search output. `PsmTsvWriter` now writes a `ProForma` column immediately after `Full Sequence` for every PSM/proteoform — bottom-up and top-down alike — by calling the new mzLib `IBioPolymerWithSetMods.ToProFormaString()`, which resolves each modification to its Unimod/PSI-MOD/RESID accession (or name) and emits a standards-compliant proteoform string. **MetaDraw** displays the column in its PSM grid. Because the value flows through the existing `DataDictionary`, it appears in both the header (`GetTabSeparatedHeader`) and every row with no change to column-ordering logic; crosslink and glyco searches keep their own headers and are untouched. The feature was verified end-to-end on a real top-down human (Jurkat) search, where carbamidomethyl renders as `C[UNIMOD:4]`. This PR requires the companion mzLib ProForma changes.

## What's included
- `EngineLayer/SpectrumMatch/PsmTsvWriter.cs` — populate `s[SpectrumMatchFromTsvHeader.ProForma]` via `Resolve(pepsWithMods.Select(b => b.ToProFormaString()))` (ambiguity handled like Full Sequence).
- `GUI/MetaDraw/MetaDraw.xaml` — `DataGridTextColumn` bound to `ProForma` next to Full Sequence.
- `Test/PsmTsvWriterTests.cs` — `ProFormaColumn_IsWrittenForPeptide` (header contains the column; row emits `PEM[Oxidation]TIDEK`).

## Dependency
**Requires mzLib with the ProForma reader/writer** (companion PR: smith-chem-wisc/mzLib `feature/proforma-reader-writer`). The local-feed PackageReference version bumps used during development are **not** committed; this branch should reference the published mzLib version once the mzLib PR merges. **CI will be red until then** — this is expected.

## Notes for reviewers
- Existing header tests use name-based `IndexOf`, so the added column does not break them.
- **Verified live:** CMD top-down classic search (Jurkat `fract1` + human reviewed PTM DB) produced `AllProteoforms.psmtsv`/`AllPSMs.psmtsv` with a populated ProForma column.

### DIFF
--- a/MetaMorpheus/EngineLayer/SpectrumMatch/PsmTsvWriter.cs
+++ b/MetaMorpheus/EngineLayer/SpectrumMatch/PsmTsvWriter.cs
@@ -11,6 +11,7 @@ using Omics.Modifications;
 using Proteomics;
 using Proteomics.ProteolyticDigestion;
 using Readers;
+using Readers.ProForma;
 
 namespace EngineLayer
 {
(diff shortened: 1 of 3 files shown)


trishorts:
Depends on smith-chem-wisc/mzLib#1072 — CI will be red until that merges and a published mzLib version is referenced.

trishorts:
Closing in favor of #2675, which is the same feature on a cleaner base.

The MetaDraw column from this branch has been ported over (`63232178` on `proformaIntegration`), so nothing here is lost. #2675 is also now rebased onto current master.

The dependency note in the description above is stale: mzLib #1072 merged 2026-07-15, and `ToProFormaString()` plus `SpectrumMatchFromTsv.ProForma` ship in mzLib 1.0.583 — the version MetaMorpheus master already pins. No version bump is needed and CI should be green.

One deliberate scope difference: #2675 gates the column on `AnalyteType.Proteoform`, so it is written for top-down results only. Keeping the initial surface narrow; extending it to other search types is straightforward if anyone wants it.

PR #2664, opened 2026-05-25
Custom oxonium ions via the monosaccharide diagnostic-ion column

trishorts:
﻿# Custom oxonium ions via the monosaccharide diagnostic-ion column

## Summary

Lets users turn the diagnostic ions they attach to a custom monosaccharide into a **strict diagnostic filter** during glyco search. The data source is the existing `DiagnosticIonMasses` column (column 4) of `Glycan_Mods/MonosaccharidesCustom.tsv` added in #2660 â€” there is no new file and no new registry. When the glyco **Oxonium Ion Filter** is enabled, each such ion must co-occur with its linked monosaccharide: an ion observed without its monosaccharide (or a monosaccharide present without its ion) rejects the candidate.

Default behavior is unchanged. The shipped TSV registers no custom monosaccharides, so the new code path is gated off and the combined oxonium-ion array is the built-in array itself.

## How to add a custom oxonium ion

1. Open `Glycan_Mods/MonosaccharidesCustom.tsv` in the MetaMorpheus data directory.
2. Add (or uncomment) a monosaccharide row and list one or more diagnostic ion m/z values in column 4 (`DiagnosticIonMasses`), comma-separated.
3. Restart MetaMorpheus and run a glyco search with **Oxonium Ion Filter** enabled.

## TSV format (column 4)

| Column | Name | Notes |
|---|---|---|
| 1 | Monosaccharide name | built-in or custom |
| 2 | Single-char code | unique ASCII letter |
| 3 | Monoisotopic mass (Da) | residue mass |
| 4 | **DiagnosticIonMasses** | comma-separated observed m/z; these are the custom oxonium ions |
| 5 | Description | free text |

## Worked example

```
Name	SingleCharCode	MonoisotopicMass	DiagnosticIonMasses	Description
HexA	U	176.03209	175.02482,157.01425	Hexuronic acid (GlcA, GalA)
```

With the oxonium filter on, a candidate glycopeptide is rejected if 175.02482 or 157.01425 is observed but the candidate has no HexA, or if it has HexA but neither ion is observed.

## What the strict filter does

For each registered custom diagnostic ion, with the oxonium filter enabled:

- ion observed **and** linked monosaccharide present â†’ keep
- ion observed **but** monosaccharide absent â†’ reject
- monosaccharide present **but** ion absent â†’ reject
- ion absent **and** monosaccharide absent â†’ keep

[...]

PR #2660, opened 2026-05-19
Support user-defined custom monosaccharides via TSV

trishorts:
# Custom monosaccharides via TSV

## Summary

Adds a new shared file `Glycan_Mods/MonosaccharidesCustom.tsv` that lets users register monosaccharides MetaMorpheus does not ship with (e.g. HexA, Kdo, generic pentoses, heavy-labeled variants). Once registered, the new monosaccharide is usable in **both** composition format (`HexNAc(2)Hex(5)HexA(1)`) and structure format (`(N(H)(U))`) of glycan database files, contributes correct mass to total glycan mass and to `Composition` strings, and emits user-supplied diagnostic ions when present in a glycan.

The shipped TSV contains only a documented header banner and commented-out worked examples, so **no new monosaccharides are registered by default** and existing glyco search behavior is byte-identical to master.

This PR was split off from #2657 (which bundled custom monosaccharides together with custom glycan databases) so each feature can be reviewed and shipped independently. The glycan-database portion of that work stays in #2657.

---

### Quick steps

1. Open `Glycan_Mods\MonosaccharidesCustom.tsv` inside your MetaMorpheus install folder.
2. Below the `CUSTOM DEFINITIONS` banner near the bottom, add one tab-separated row per monosaccharide.
3. Save and **restart MetaMorpheus**.
4. Use the new name in composition-format glycan files (e.g. `HexNAc(2)Hex(5)HexA(1)`) or the new single-char code in structure-format files (e.g. `(N(H)(U))`).

### TSV file format

Tab-separated, one entry per data line. The optional column-header row beginning with `Name<TAB>…` is recognized and skipped. Comment (`#`) lines and blank lines are ignored.

| Column | Required? | Meaning |
|---|---|---|
| `Name` | yes | Unique. Must not collide with a built-in (`Hex`, `HexNAc`, `NeuAc`, `NeuGc`, `Fuc`, `dHex`, `Phospho`, `Sulfo`, `Na`, `Ac`, `Xylose`, `Kdn`) or with another entry. |
| `SingleCharCode` | yes | Exactly one ASCII letter (A–Z or a–z). Unique. Must not collide with a built-in code (`H`, `N`, `A`, `G`, `F`, `P`, `S`, `Y`, `C`, `X`, `K`). |
| `MonoisotopicMass` | yes | Decimal Daltons. The mass of the monosaccharide **residue** (the form integrated into the glycan chain — after water loss). |
[...]

PR #2665, opened 2026-05-26
Custom oxonium ions via the monosaccharide diagnostic-ion column

trishorts:
> **Project of origin:** `Oxonium`

# Custom oxonium ions via the monosaccharide diagnostic-ion column

## Summary

Lets users turn the diagnostic ions they attach to a custom monosaccharide into a **strict diagnostic filter** during glyco search. The data source is the existing `DiagnosticIonMasses` column (column 4) of `Glycan_Mods/MonosaccharidesCustom.tsv` added in #2660 — there is no new file and no new registry. When the glyco **Oxonium Ion Filter** is enabled, each such ion must co-occur with its linked monosaccharide: an ion observed without its monosaccharide (or a monosaccharide present without its ion) rejects the candidate.

Default behavior is unchanged. The shipped TSV registers no custom monosaccharides, so the new code path is gated off and the combined oxonium-ion array is the built-in array itself.

> Builds on #2660 (now merged). Supersedes #2664, which was closed pending the #2660 merge; this PR is the same work rebased cleanly onto `master`.

## How to add a custom oxonium ion

Either route works; both write the same row to the same file.

- **GUI** — Settings tab → **Create new monosaccharide**, filling in the **Diagnostic Ions (m/z)** box. Registers immediately, no restart.
- **File** — add (or uncomment) a monosaccharide row in `MonosaccharidesCustom.tsv` at the top level of the MetaMorpheus data directory (Settings tab → **Open mods/data folder**), listing one or more diagnostic ion m/z values in column 4 (`DiagnosticIonMasses`), comma-separated. Restart to pick it up.

Then run a glyco search with **Oxonium Ion Filter** enabled.

## TSV format (column 4)

| Column | Name | Notes |
|---|---|---|
| 1 | Monosaccharide name | built-in or custom |
| 2 | Single-char code | unique ASCII letter |
| 3 | Monoisotopic mass (Da) | residue mass |
| 4 | **DiagnosticIonMasses** | comma-separated observed m/z; these are the custom oxonium ions |
| 5 | Description | free text |

## Worked example

```
Name	SingleCharCode	MonoisotopicMass	DiagnosticIonMasses	Description
HexA	U	176.03209	175.02482,157.01425	Hexuronic acid (GlcA, GalA)
```

With the oxonium filter on, a candidate glycopeptide is rejected if 175.02482 or 157.01425 is observed but the candidate has no HexA, or if it has HexA but neither ion is observed.

## What the strict filter does

For each registered custom diagnostic ion, with the oxonium filter enabled:

- ion observed **and** linked monosaccharide present → keep
- ion observed **but** monosaccharide absent → reject
- monosaccharide present **but** ion absent → reject
- ion absent **and** monosaccharide absent → keep

"Observed" is the same relative-intensity test the built-in rules use: a peak matched within the product mass tolerance whose intensity clears 2% of the HexNAc 138.055 ion (`OxoniumRelativeIntensityThreshold`, now shared by both paths). A spectrum with no 138.055 at all has no denominator and so no evidence either way — the strict rule stands down for that spectrum and rejects nothing, rather than reading the ion as absent, which would reject every candidate carrying the monosaccharide. A custom ion may not duplicate a built-in oxonium ion, or one already claimed by another custom monosaccharide; either is rejected at load time. The filter is gated by the existing Oxonium Ion Filter parameter; with it off, column-4 ions keep their prior scoring-only role.

## What changed under the hood

- `GlycoSearch/OxoniumIonReservedIndices.cs` (new): named constants for the six built-in oxonium-array indices the engine had hardcoded (4, 5, 9, 10, 12, 14).
- `GlycoSearch/GlycoSearchEngine.cs`: the hardcoded indices now use those constants (pure rename).
- `GlycoSearch/Glycan.cs`: `HasCustomOxoniumIons`, `CustomOxoniumIons` ((m/z, Kind-index) pairs), and `AllOxoniumIonsIncludingCustoms` (built-ins + appended customs; returns the built-in array unchanged when there are no customs).
- `GlycoSearch/GlycoPeptides.cs`: `ScanOxoniumIonFilter` probes the combined ion list; `DiagonsticFilter` applies the strict rule after the unchanged built-in rules, gated on `HasCustomOxoniumIons`; helpers `CheckOxoniumPresence` and `ApplyStrictMonoFilter`.
- `Glycan_Mods/MonosaccharidesCustom.tsv`: banner documents the column-4 ions' new strict-filter role, the intensity threshold, the duplicate-ion rule, and the stand-down case.
- `docs/Custom-Monosaccharides-and-Diagnostic-Ions.md` (new): end-user guide — both the GUI dialog and the TSV, the current file location, the dual role of column-4 ions, a worked example, loader errors, and an FAQ.

## Tests

1. Ordering-freeze guard for the six reserved indices.
2. Accessor behavior: empty default (combined array is the built-in array reference); registration appears in `CustomOxoniumIons` and the combined array.
3. Helper units: `ApplyStrictMonoFilter` (four cases), `CheckOxoniumPresence`.
4. Strict filter matrix: the four observed×present combinations, a two-custom single-mismatch reject, and a built-in-reject test proving the legacy path is intact with no customs.
5. Defensive guards: undersized intensity/Kind inputs throw rather than silently rejecting.
   5b. Absent 138.055 reference: strict rule stands down in both directions, and the guards still fire.
6. Round-trip: `LoadCustomMonosaccharides` column-4 ions appear in `CustomOxoniumIons`.
7. Performance bounds (`[Explicit]`, `Category=Performance`): negligible per-call overhead at zero customs; under 10 µs/call at 50 custom ions.

## Compatibility / rollout

- No new code files outside GlycoSearch, no schema change, no GUI change (the **Create new monosaccharide** dialog already carries a Diagnostic Ions field; this PR only gives those ions their filtering role and documents it). Built-in `AllOxoniumIons` ordering is unchanged and guarded by a test.
- With the shipped (no-customs) TSV, search output is byte-identical to the base branch.
- Builds on #2660 (custom monosaccharides). A GUI surface for managing these ions can follow in a later PR.


RayMSMS (dismissed):

MetaMorpheus/EngineLayer/GlycoSearch/Glycan.cs
```diff
@@ -328,6 +328,61 @@ public static void ResetCustomMonosaccharides()
...
+        /// pairs in a deterministic order (by Kind index, then declaration order). Empty unless customs
+        /// were registered. Consumed by the strict custom-oxonium branch of GlycoPeptides.DiagonsticFilter
+        /// and used to size the oxonium intensity list. No hydrogen offset is applied: values are observed
+        /// m/z, matching the convention of AllOxoniumIons and the MonosaccharidesCustom.tsv documentation.
+        /// </summary>
+        public static IReadOnlyList<(int MzScaled, int KindIndex)> CustomOxoniumIons
+        {
+            get
+            {
+                var list = new List<(int MzScaled, int KindIndex)>();
```
acesnik: This getter allocates two lists and runs a sort every time it's read, and `DiagonsticFilter` reads it once per candidate inside the `while (iDLow < GlycanBoxes.Count() && ...)` loop at `GlycoSearchEngine.cs:479` — the innermost loop of glyco search. So for anyone who actually registers a custom monosaccharide, every candidate pays a `List<int>` allocation, a `Sort()`, a `List<(int,int)>` allocation and the dictionary walk. Your own perf test puts the ceiling at 10 µs/call at 50 ions, which over millions of candidate evaluations is real time, and all of it is avoidable.

The contents only change in `RegisterCustomMonosaccharide` and `ResetCustomMonosaccharides`, so this wants to be a cached field rebuilt in those two places and simply returned here. `AllOxoniumIonsIncludingCustoms` below has the same shape — it allocates a fresh combined array per call, once per scan via `ScanOxoniumIonFilter` — and the same fix applies. Neither affects the no-customs path, which is presumably why the benchmarks look fine; the cost lands only on the users this feature is for.

MetaMorpheus/EngineLayer/GlycoSearch/GlycoPeptides.cs
```diff
@@ -522,13 +523,56 @@ public static bool DiagonsticFilter(double[] oxoniumIonsintensities, GlycanBox g
...
+        internal static bool CheckOxoniumPresence(double[] oxoniumIonsintensities, int index)
+        {
+            return oxoniumIonsintensities[index] > 0;
+        }
+
+        /// <summary>
+        /// Strict custom-oxonium rule: accept only when the ion's observed state matches the candidate's
+        /// possession of the linked monosaccharide. A mismatch in either direction rejects.
+        /// </summary>
+        internal static bool ApplyStrictMonoFilter(bool hasSignal, bool hasMono)
```
acesnik, line 591: This is the part I'd most like your and @RayMSMS's read on, because it's a science question rather than a code one.

`hasSignal == hasMono` treats bare presence as truth in both directions, with no intensity threshold — `CheckOxoniumPresence` is `> 0`, and `ScanOxoniumIonFilter` writes a non-zero intensity whenever *any* envelope matches within the product tolerance. So a single noise peak at a custom ion's m/z makes `hasSignal` true, and every candidate lacking that monosaccharide is then rejected for the whole scan. In the other direction, a genuine glycopeptide whose diagnostic ion just didn't fly at the collision energy used gets rejected too. Oxonium yields are notoriously energy- and structure-dependent, which the TSV banner you added says as much about.

What makes me raise it is that the built-in rules immediately above deliberately don't work this way — they use ratios against a 0.02 threshold rather than bare presence, presumably for exactly this reason. The new path is the odd one out.

A concrete footgun: nothing stops a custom ion duplicating a built-in m/z. Your own test constant `Ion657 = 65723544` *is* `AllOxoniumIons[15]`. If a user lists something ubiquitous like 204.087 as their custom sugar's diagnostic ion, `hasSignal` is true on essentially every glycopeptide spectrum and the filter rejects every candidate that lacks that sugar — a near-total wipeout of results, silently. Worth either rejecting duplicate-of-built-in ions at load time, or applying a relative-intensity threshold rather than `> 0`, or both.

MetaMorpheus/EngineLayer/GlycoSearch/GlycoPeptides.cs
```diff
@@ -522,13 +523,56 @@ public static bool DiagonsticFilter(double[] oxoniumIonsintensities, GlycanBox g
...
+            // candidate -- or its monosaccharide present while the ion is absent -- rejects the spectrum.
+            // Gated on HasCustomOxoniumIons so the default (no customs) path is unchanged.
+            if (Glycan.HasCustomOxoniumIons)
+            {
+                var customOxoniumIons = Glycan.CustomOxoniumIons;
+                int builtInCount = Glycan.AllOxoniumIons.Length;
+                for (int j = 0; j < customOxoniumIons.Count; j++)
+                {
+                    int idx = builtInCount + j;
+                    bool hasSignal = idx < oxoniumIonsintensities.Length
```
acesnik: These two guards fail in the dangerous direction. An index past the end of `oxoniumIonsintensities` — or past the end of `glycanBox.Kind` on the next line — is silently read as "absent", and under strict semantics "absent" is a *rejection* whenever the other side is present.

So if `oxoniumIonsintensities` was ever built without the custom slots (an array from a caller that ran before registration, a cached array, a future call site), every candidate carrying the custom monosaccharide is rejected and the search quietly returns nothing for them. Note that `DiagonsticFilter_UndersizedInputs_GuardsTreatAsAbsentAndDoNotThrow` passes only because it undersizes *both* arrays, so `hasMono` is false too and the mismatch cancels out — it doesn't cover the asymmetric case, which is the one that bites.

A length mismatch here is a programming error, not a data condition. I'd rather it throw, or skip the custom branch entirely, than silently reject.

MetaMorpheus/Test/CustomOxoniumFilterTests.cs
```diff
@@ -0,0 +1,270 @@
...
+
+namespace Test
+{
+    /// <summary>
+    /// Strict custom-oxonium filter (option B): the diagnostic ions a user attaches to a custom
+    /// monosaccharide (MonosaccharidesCustom.tsv column 4) feed GlycoPeptides.DiagonsticFilter with
+    /// strict bidirectional semantics. All tests that register custom monosaccharides restore global
+    /// state in a finally block via Glycan.ResetCustomMonosaccharides().
+    /// </summary>
+    [TestFixture]
```
acesnik, line 16: Two notes on global state, same family as the ones on #2700.

This fixture mutates process-wide `Glycan` statics in most of its tests and isn't `[NonParallelizable]`. NUnit defaults to serial and I found no assembly-level `[Parallelizable]` or `.runsettings` in this repo, so it isn't flaky today — but it matches the convention the markers in `PsmTsvWriterTests.cs` exist for.

More substantive: the `finally` blocks call `ResetCustomMonosaccharides()`, which resets to *empty* rather than restoring whatever startup loaded. `GlobalVariables.LoadGlycans()` calls `LoadCustomMonosaccharides` on the shipped TSV, which registers nothing today, so empty happens to equal the startup state and this is fine. The moment a real row is added to the shipped file, these tests will wipe it for every test that runs after them. A save-and-restore, or a note pinning that assumption, would keep it honest.

MetaMorpheus/Test/CustomOxoniumFilterTests.cs
```diff
@@ -0,0 +1,270 @@
...
+    public class CustomOxoniumFilterTests
+    {
+        private const int Ion512 = 51219700; // 512.197 m/z * 1e5
+        private const int Ion657 = 65723544; // 657.23544 m/z * 1e5
+
+        // Build a GlycanBox with a chosen Kind[] without touching the glycan database. DiagonsticFilter
+        // only reads glycanBox.Kind, so an uninitialized instance with the private setter invoked is enough.
+        private static GlycanBox BoxWithKind(byte[] kind)
+        {
+            var box = (GlycanBox)RuntimeHelpers.GetUninitializedObject(typeof(GlycanBox));
```
acesnik: `GetUninitializedObject` plus a reflected private setter is doing a lot of work here, and it breaks at run time rather than compile time if `Kind` is ever renamed or turned into a field. The `Assert.That(setter, Is.Not.Null, ...)` above softens that into a clear message, which I appreciate.

Not asking you to change it — building a real `GlycanBox` needs the glycan database and that's a worse trade for a unit test. But the comment says "`DiagonsticFilter` only reads `glycanBox.Kind`", and that's an invariant nothing enforces; the day the filter reads a second member, these tests fail on an uninitialized object in a thoroughly confusing way. Worth an `internal` test-only constructor at some point.

acesnik (commented):
:robot: I like the shape of this a lot. Reusing the column-4 ions that #2660 already parses instead of inventing a second registry is the right instinct, gating the whole branch on `HasCustomOxoniumIons` keeps the default path genuinely untouched, and naming the six magic indices is a real improvement on its own — I checked all six against `AllOxoniumIons` and 4/5/9/10/12/14 do map to 138.055, 144.066, 204.087, 274.093, 292.103 and 366.140, so the rename is faithful and the ordering-freeze test is a good guard to have added.

Holding off on approving for two things rather than requesting changes outright, since one of them is a question for you and Ray rather than a defect:

1. `Glycan.CustomOxoniumIons` rebuilds and sorts a list on every access, and `DiagonsticFilter` calls it once per glycan-box candidate in the innermost loop of glyco search. Cheap to fix by caching at registration time.
2. The strict rule uses bare presence with no intensity threshold, which sits oddly beside the ratio-plus-threshold convention of the built-in rules right above it, and has a sharp failure mode I've described inline.

The rest are small. Details on each below.

MetaMorpheus/EngineLayer/GlycoSearch/Glycan.cs
trishorts: Fixed in 8506693. Both are now built once in `RebuildCustomOxoniumCaches`, called only from `RegisterCustomMonosaccharide` and `ResetCustomMonosaccharides` — the only two places the set can change — and the properties just return the cached field. The no-customs path is unchanged in kind as well as in cost: `_allOxoniumIonsIncludingCustoms` stays null and the property hands back the built-in `AllOxoniumIons` reference itself.

`CustomOxoniumIons_AreCachedAndRebuiltOnlyOnRegistrationChange` pins it with `Is.SameAs` on repeated reads, asserts a new registration does invalidate both, and asserts reset restores the built-in reference.

MetaMorpheus/EngineLayer/GlycoSearch/GlycoPeptides.cs
trishorts: You're right on both counts, and I've taken both fixes rather than picking one. 8506693.

**Threshold.** The new path is no longer the odd one out. `CheckOxoniumPresence` now takes the 138.055 HexNAc intensity as a reference and requires the custom ion to clear 2% of it — the same test, and now literally the same constant, as the built-in NeuAc and HexHexNAc rules. I pulled the magic `0.02` out into `OxoniumRelativeIntensityThreshold` and pointed the built-in comparisons at it too, so there is one definition of "observed" in the file instead of two.

One deliberate divergence: when the 138.055 reference is itself zero, the built-in comparisons see `x/0` = +Infinity and call the ion observed. For the strict rule I return "not observed" instead. Under strict semantics "observed" is a rejection lever, and a spectrum with no HexNAc signal at all should not be able to pull it. Called out in the XML doc so it doesn't read as an oversight.

`DiagonsticFilter_NoiseLevelCustomIon_IsNotObservedAndDoesNotReject` covers exactly your first failure mode — a peak at 0.5% of the reference, candidate lacking the sugar, now kept.

**Duplicate ions.** Rejected at load time, as you suggested. `RegisterCustomMonosaccharide` throws if a diagnostic ion falls within 0.01 Da of any built-in oxonium ion, or of one already claimed by another custom monosaccharide (two sugars sharing an ion is unsatisfiable under the strict rule — no candidate carrying only one of them could pass). `GlycanDatabase.LoadCustomMonosaccharides` already wraps `ArgumentException`, so the user gets a `MetaMorpheusException` naming the file and line; there's a test asserting that.

The 0.01 Da window rather than exact equality is because `204.087` and the built-in `204.08720` are the same ion typed two ways, and exact matching would wave the first one through. 0.01 Da is wider than any realistic product tolerance at these masses (20 ppm at m/z 204 is 0.004 Da) and far narrower than the spacing between built-in ions.

Good catch on `Ion657` — it was `AllOxoniumIons[15]`, so the test constant would have tripped the new check. Replaced with 733.25, which is genuinely free. The perf fixture's 50 synthetic ions were spaced 0.00001 Da apart and are now 0.02 Da apart for the same reason.

**Still your and @RayMSMS's call:** whether 2%-of-138.055 is the right bar for a *custom* sugar's ion, or whether these want their own (looser or per-ion) threshold. Sharing the built-in constant is the conservative choice — it makes the new rule no more trigger-happy than the rules that ship today — but it is a choice, not a measurement, and I have no data on custom-sugar oxonium yields to defend a different number with. Happy to make it a parameter if you'd rather it be tunable.

The TSV banner and `docs/Custom-Monosaccharides-and-Diagnostic-Ions.md` both document the threshold and the duplicate rule now, including the full list of built-in ions a user must avoid.

MetaMorpheus/EngineLayer/GlycoSearch/GlycoPeptides.cs
trishorts: Agreed, and your read of why the test passed is exactly right. Fixed in 8506693 — both now throw.

The array length is checked once before the loop rather than per-iteration, and the `Kind[]` bound per custom ion; both messages say what was expected, what was found, and which producer is supposed to have sized it. Silently rejecting is the one outcome that can't happen now.

Split the test in two so the asymmetric case is actually covered:

- `DiagonsticFilter_UndersizedIntensityArray_Throws` — array short, monosaccharide **present**. This is the case the old test couldn't see.
- `DiagonsticFilter_UndersizedKindArray_Throws` — ion observed, `Kind[]` one slot short of the custom index. (Sized to `KindCapacity - 1` rather than 2, so the built-in rules that read `Kind[0..2]` run normally and only the custom branch sees the mismatch — with a length-2 array the legacy code throws `IndexOutOfRange` first, which is pre-existing behavior and not what this test is about.)

MetaMorpheus/Test/CustomOxoniumFilterTests.cs
trishorts: Both taken. 8506693.

`[NonParallelizable]` on the fixture, and on `CustomOxoniumPerformanceTests` too — it mutates the same statics and had the same gap.

On the restore: you're right that "empty" only happens to equal startup today. The `finally` blocks now call `RestoreStartupMonosaccharides()`, which resets and then re-loads the shipped `MonosaccharidesCustom.tsv` from `GlobalVariables.DataDir` — the same file `LoadGlycans()` reads at startup. So it restores whatever startup actually loaded rather than pinning an assumption about what that is, and adding a real row to the shipped file won't quietly break every test that runs after these. The `ResetCustomMonosaccharides()` calls at the *top* of each test stay, since those are establishing the known-empty precondition the test needs.

MetaMorpheus/Test/CustomOxoniumFilterTests.cs
trishorts: Did it now rather than "at some point" — it was three lines. 8506693.

`GlycanBox` has an `internal GlycanBox(byte[] kind)` that chains to `base(new int[0])` and sets `Kind`; `Test` already has `InternalsVisibleTo`. `BoxWithKind` in both fixtures is now just `new GlycanBox(kind)`, and `System.Reflection` / `System.Runtime.CompilerServices` are gone from both files.

That removes the invariant you flagged along with the reflection: it's a real, fully-constructed object, so the day `DiagonsticFilter` reads a second member these tests fail on the assertion rather than on an uninitialized field — and a rename of `Kind` is a compile error again.

trishorts:
Thanks @acesnik — all five addressed in 8506693, replies inline on each thread. Summary of what changed:

**1. Hot-path allocation.** `CustomOxoniumIons` and `AllOxoniumIonsIncludingCustoms` are now built once in `RebuildCustomOxoniumCaches`, called only from `RegisterCustomMonosaccharide` and `ResetCustomMonosaccharides`. No per-candidate list, sort, or array. The no-customs path still returns the built-in `AllOxoniumIons` reference itself.

**2. Presence semantics (the science one).** Took both of your suggestions rather than choosing between them:

- `CheckOxoniumPresence` is now relative-intensity, not bare presence: the custom ion must clear 2% of the HexNAc 138.055 ion. That is the same rule *and now literally the same constant* as the built-in NeuAc/HexHexNAc comparisons — I pulled `0.02` out into `OxoniumRelativeIntensityThreshold` and pointed the built-ins at it too, so there is one definition of "observed" in the file. The new path is no longer the odd one out.
- Duplicate ions are rejected at load time. A diagnostic ion within 0.01 Da of any built-in oxonium ion, or of one already claimed by another custom monosaccharide, throws — surfaced to the user as a `MetaMorpheusException` naming the file and line. 0.01 Da rather than exact equality because `204.087` and `204.08720` are the same ion typed two ways.

There is one open question I'd still like your and @RayMSMS's read on, left in the thread: whether 2%-of-138.055 is the right bar for a *custom* sugar's ion, or whether these want their own (or a tunable) threshold. Sharing the built-in constant is the conservative default, but it's a choice rather than a measurement.

**3. Guards failing in the dangerous direction.** Both now throw instead of silently reading "absent" (= reject). The array-length check runs once before the loop, the `Kind[]` bound per ion. The old test passed only because it undersized both sides; it is split into two tests that each cover the asymmetric case.

**4. Test global state.** Both fixtures are `[NonParallelizable]`, and the `finally` blocks restore the *startup* registration state (reset + re-load the shipped TSV) rather than resetting to empty, so a real row added to the shipped file won't wipe itself for every test that follows.

**5. Reflection in the test helper.** Replaced with an `internal GlycanBox(byte[] kind)` test-only constructor. Both fixtures drop `System.Reflection` and `System.Runtime.CompilerServices`; a rename of `Kind` is a compile error again.

Also caught by your review: the test constant `Ion657` *was* `AllOxoniumIons[15]` and would have tripped the new duplicate check, and the perf fixture's 50 synthetic ions were 0.00001 Da apart. Both respaced.

Docs updated to match — the TSV banner and `docs/Custom-Monosaccharides-and-Diagnostic-Ions.md` now state the intensity threshold and list every built-in oxonium ion a custom ion must avoid.

Local: 20/20 in the custom-oxonium fixtures, 105/105 across the glyco suite, and both `[Explicit]` perf tests still pass.

trishorts:
CI is green on 8506693 — all 11 checks pass, including `codecov/patch` and `codecov/project`.

I've resolved four of the five threads, each of which is fully addressed and has a reply describing what changed:

- hot-path allocation in `CustomOxoniumIons` / `AllOxoniumIonsIncludingCustoms`
- guards failing toward silent rejection
- fixture global state (`[NonParallelizable]` + restore-to-startup)
- `GetUninitializedObject` + reflected setter

**Left open on purpose:** the presence-semantics thread on `GlycoPeptides.cs`. The two concrete defects in it are fixed — relative-intensity threshold instead of bare presence, and load-time rejection of duplicate ions — but the question you raised for me and @RayMSMS is still genuinely open: is 2%-of-138.055 the right bar for a *custom* sugar's diagnostic ion, or do these want their own, looser, or tunable threshold? Reusing the built-in constant is the conservative default rather than a measured answer, and I'd rather leave the thread open until one of you weighs in than close it over my own choice.

@RayMSMS — you approved before this round; the presence rule has changed since, so worth a re-look at that one thread.

MetaMorpheus/EngineLayer/GlycoSearch/GlycoPeptides.cs
```diff
@@ -481,13 +482,98 @@ public static bool DiagonsticFilter(double[] oxoniumIonsintensities, GlycanBox g
...
+        /// peak at the ion's m/z reject every candidate lacking the linked monosaccharide, since
+        /// ScanOxoniumIonFilter records an intensity for any envelope inside the product tolerance.
+        /// A missing 138.055 reference is treated as "not observed" rather than dividing by zero. This
+        /// is a deliberate, and the only, divergence from the built-in comparisons, which would see
+        /// +Infinity there and call the ion observed: under strict semantics "observed" is a rejection
+        /// lever, and a spectrum with no HexNAc signal at all should not be able to pull it.
+        /// </summary>
+        internal static bool CheckOxoniumPresence(double[] oxoniumIonsintensities, int index, double hexNAcReferenceIntensity)
+        {
+            if (hexNAcReferenceIntensity <= 0)
```
acesnik, line 580: 🤖 *(Edited: my original version of this note attributed a phrase to the doc comment that is not in it. The comment does not claim equivalence with the built-in rules — it says the opposite, and the correction sharpens the point rather than withdrawing it.)*

The doc comment above already anticipates the direction I would have worried about first, and states the right principle: with a zero reference the built-in comparison would see `+Infinity`, call the ion observed, and reject every candidate *lacking* the monosaccharide — so "a spectrum with no HexNAc signal at all should not be able to pull" the rejection lever. That is correct, and it is the reason `return false` is here rather than a division.

**But `return false` does not achieve it.** Under `hasSignal == hasMono` at `:530`, false is not neutral either: it is a mismatch for any candidate that *has* the monosaccharide. So a spectrum with no 138.055 still pulls the lever — just against the opposite population. Before, candidates without the custom monosaccharide were rejected; now candidates with it are. On any such spectrum, every candidate carrying a custom monosaccharide is refused.

Reachable rather than theoretical: `ScanOxoniumIonFilter` fills each ion slot independently with no 138 precondition, and `DiagonsticFilter` is gated only on `OxoniumIonFilter` at `GlycoSearchEngine.cs:479`. A low-energy HCD spectrum with strong 204.087 and absent 138.055 is ordinary — 138 is a secondary fragment of 204 and is favoured at higher energy.

The form that delivers what the comment describes is to skip the custom branch entirely when the reference is `<= 0`, rather than calling the ion unobserved: with no reference there is no evidence either way, and the strict rule has nothing to say. That is neutral in the sense the comment intends; `false` only looks neutral because the built-in rules use it that way, and under `hasSignal == hasMono` it is not.

`CheckOxoniumPresence_UsesRelativeIntensityNotBarePresence` pins the unit behaviour but not this consequence — a case with reference 0 and the monosaccharide present would be the one to add.

acesnik (commented):
🤖 Thanks — that's a thorough round, and I checked it against the code rather than taking the replies on trust. All four mechanical points hold up: the caches are rebuilt from exactly the two mutation sites and the no-customs path hands back the built-in array reference itself; the guards throw with the producer named, and the split test does now cover the asymmetric case; the reflection is gone in favour of a real `internal GlycanBox(byte[])`; and both fixtures are `[NonParallelizable]` with the shipped TSV reloaded in the `finally`.

Two of the duplicate-ion claims I could check against the data, and both hold: the smallest gap between any two built-in oxonium ions is 0.984 Da, which is 98x the 0.01 Da collision window, so the window cannot swallow a genuinely distinct ion — and 733.25 really is free.

I also ran the three fixtures: 20 of 20 pass, on macOS arm64 through a throwaway net10.0 harness, including every test the replies cite.

One thing the threshold change brought with it, inline. It is the same failure mode as the guards you just fixed, arriving from the other side, so I would rather raise it now than after it merges.

On the open question — I have no data on custom-sugar oxonium yields either, and sharing the built-in constant is the right default in the absence of any. It makes the new rule no more trigger-happy than what ships today, which is the property that matters. I would leave it there rather than add a knob nobody can set well. If Ray has numbers, that changes.

docs/Custom-Monosaccharides-and-Diagnostic-Ions.md
```diff
@@ -0,0 +1,263 @@
...
+- [End-to-end worked example](#end-to-end-worked-example)
+- [Validation & error messages](#validation--error-messages)
+- [Tips & caveats](#tips--caveats)
+- [FAQ](#faq)
+
+---
+
+## Quick start
+
+1. Open `Glycan_Mods/MonosaccharidesCustom.tsv` in the MetaMorpheus data directory.
```
RayMSMS: Maybe we can add the "GUI" part to this Markdown file.

docs/Custom-Monosaccharides-and-Diagnostic-Ions.md
```diff
@@ -0,0 +1,263 @@
...
+```
+Name	SingleCharCode	MonoisotopicMass	DiagnosticIonMasses	Description
+HexA	U	176.03209	175.02482,157.01425	Hexuronic acid (GlcA, GalA)
+```
+
+---
+
+## Where the file lives
+
+`MonosaccharidesCustom.tsv` sits in the **`Glycan_Mods`** folder of the MetaMorpheus data directory, alongside the `OGlycan` and `NGlycan` database folders.
```
RayMSMS: This line can be revised because the recent PR already moved the TSV to datadir.

MetaMorpheus/EngineLayer/Glycan_Mods/MonosaccharidesCustom.tsv
```diff
@@ -44,6 +44,37 @@
...
+#
+#                          An ion may NOT duplicate a built-in oxonium ion, nor
+#                          one already claimed by another custom monosaccharide
+#                          in this file. Anything within 0.01 Da of one of those
+#                          is an error and the file will fail to load, naming
+#                          the offending line. The built-in oxonium ions are:
+#                              109.02895  115.03951  126.05550  127.03952
+#                              138.05550  144.06607  163.06064  168.06607
+#                              186.07663  204.08720  274.09268  290.08759
+#                              292.10324  308.09816  366.14002  657.23544
```
RayMSMS, line 75: Thanks for this change!

RayMSMS (dismissed):

MetaMorpheus/EngineLayer/GlycoSearch/GlycoPeptides.cs
trishorts: You're right, and the correction is the important half: `return false` was never neutral under `hasSignal == hasMono`, it just moved the rejection from one population to the other. Fixed in b5989b3.

`DiagonsticFilter` now skips the custom branch outright when the 138.055 reference is absent, rather than asking `CheckOxoniumPresence` for a verdict it has no basis to give:

```csharp
bool hexNAcReferenceObserved = HexNAc_diagnostic > 0;
...
    if (!hexNAcReferenceObserved)
    {
        continue;
    }
```

No denominator, no evidence, so the strict rule says nothing. The bounds guards still run on those spectra — the array-length check before the loop and the `Kind[]` check ahead of the `continue` — since a mis-sized array is a programming error whatever the spectrum looks like, and I didn't want the neutral path to be a hole in them.

`CheckOxoniumPresence` keeps returning false for a zero reference — as a unit that answer is correct, "the ratio does not clear the threshold" — but its doc comment no longer claims that this achieves neutrality. It now says the opposite: neither verdict is neutral under the strict rule, so a caller applying that rule must not use the result when the reference is absent, and points at the skip in `DiagonsticFilter`. The rationale lives at the call site that depends on it.

Three tests, and the first is the case you named:

- `DiagonsticFilter_NoHexNAcReference_MonoPresent_StrictRuleSkippedAndCandidateKept` — reference 0, candidate carries the sugar. Rejected before this change, kept now.
- `DiagonsticFilter_NoHexNAcReference_IonPeakPresentMonoAbsent_StrictRuleSkippedAndCandidateKept` — the same spectrum from the other side: a peak at the custom ion's m/z with nothing to size it against must not reject the other population either.
- `DiagonsticFilter_NoHexNAcReference_UndersizedIntensityArray_StillThrows` — pins that standing down doesn't disarm the guard.

23/23 in the two custom-oxonium fixtures, 141/141 across the glyco suite, Release build clean.

On the threshold question you closed out above: agreed, and I've left it at the shared built-in constant with no knob.

docs/Custom-Monosaccharides-and-Diagnostic-Ions.md
trishorts: Added in b5989b3 — you're right that a guide describing only the TSV was missing the shorter route in for most people.

New section, **Two ways in: the GUI dialog or the file**, sitting between Quick start and Where the file lives. It documents Settings tab → **Create new monosaccharide**, with a table mapping every field in the dialog to its TSV column:

| Dialog field | TSV column | Required |
|---|---|:---:|
| Name | 1 · `Name` | ✅ |
| Single-Char Code | 2 · `SingleCharCode` | ✅ |
| Chemical Formula | — (converted to a mass on save; wins over the mass box) | ⬜ |
| Monoisotopic Mass | 3 · `MonoisotopicMass` | ✅ unless a formula was given |
| **Diagnostic Ions (m/z)** | 4 · `DiagnosticIonMasses` | ⬜ |
| Description | 5 · `Description` | ⬜ |

Diagnostic Ions (m/z) is called out as the field this whole guide is about, and links to Part 3 for the two roles those ions play.

Also threaded the GUI through the rest of the page rather than bolting the section on:

- **Quick start** now opens with the dialog and offers the file as the alternative.
- The **worked example** gives both routes for step 1.
- The restart rule is split, because it differs: hand-editing the TSV needs a restart, the dialog registers the sugar for the running session as well as writing it to the file. Fixed in the Tips list and the FAQ too, where it was stated unconditionally.
- A note on the "usable this session but could not be saved" warning `PersistCustomMonosaccharide` returns, since that is the one dialog outcome that looks like success and isn't.
- **Open mods/data folder**, on the same Settings tab, is named as the way to reach the file — which ties into your other comment.

I left the GUI out of Part 3 itself; the scoring-vs-strict-filter distinction is the same however the ion got registered.

docs/Custom-Monosaccharides-and-Diagnostic-Ions.md
trishorts: Caught something worse than a stale line — thanks. Fixed in b5989b3.

The section now says the file sits at the **top level of the data directory**, next to `proteases_custom.tsv` and `rnase_custom.tsv`, which is what `GlobalVariables.CustomMonosaccharidePath` resolves to. The path table is rewritten:

| Install type | Typical path |
|---|---|
| Windows GUI install | `%LOCALAPPDATA%\MetaMorpheus\MonosaccharidesCustom.tsv` |
| Built from source / portable | `<build output folder>\MonosaccharidesCustom.tsv` |
| Custom data directory | `<your data dir>\MonosaccharidesCustom.tsv` |

I kept one sentence about `Glycan_Mods` rather than deleting the history, because anyone reading this guide today may have a file sitting there: it notes the move, why (so install/repair/upgrade can't overwrite it), and that `EnsureCustomMonosaccharideFileExists` copies a legacy file over once on first launch and leaves the original alone. Also noted that a missing file gets the documented template written out on startup, and pointed at **Open mods/data folder** on the Settings tab as the quick way there. The Quick start and worked example no longer say `Glycan_Mods/` either.

The same staleness had gotten into the code: both custom-oxonium test fixtures built the shipped-TSV path by hand as `Path.Combine(DataDir, "Glycan_Mods", "MonosaccharidesCustom.tsv")`, so after the move `RestoreStartupMonosaccharides` hit a `File.Exists` that is now always false and silently restored nothing — the exact "empty only happens to equal startup" failure the restore-to-startup change was added to prevent, arriving by a different door. Both now ask `GlobalVariables.CustomMonosaccharidePath`.

MetaMorpheus/EngineLayer/Glycan_Mods/MonosaccharidesCustom.tsv
trishorts: Thanks Ray — glad the banner reads right. One addition since you looked: the column-4 note now also states that a custom ion must clear 2% of the HexNAc 138.055 reference to count as observed, and that the strict rule stands down entirely on a spectrum with no 138.055 at all.

MetaMorpheus/EngineLayer/GlycoSearch/GlycoPeptides.cs
trishorts: Closing this one out. Both concrete defects were fixed in 8506693 — relative-intensity threshold instead of bare presence, and load-time rejection of duplicate ions — and the open science question has both answers now: @acesnik above ("sharing the built-in constant is the right default in the absence of any [data]... I would leave it there rather than add a knob nobody can set well") and @RayMSMS's approval. So the threshold stays as the shared `OxoniumRelativeIntensityThreshold`, with no per-ion or tunable knob. If measurements on custom-sugar oxonium yields ever turn up, that is the moment to revisit it, not before.

One consequence of the threshold change was caught later and is fixed in b5989b3 — see the thread at `GlycoPeptides.cs:561` for the spectrum-with-no-138.055 case.

trishorts:
Round three. Every open thread is addressed and resolved; the branch is even with `master` with no conflicts.

**1. The absent-138.055 case** (@acesnik, `GlycoPeptides.cs:561`) — you were right that `return false` isn't neutral under `hasSignal == hasMono`; it just moved the rejection onto the candidates that *carry* the custom monosaccharide. `DiagonsticFilter` now skips the custom branch outright when the reference is absent — no denominator, no evidence, nothing for the strict rule to say — instead of asking `CheckOxoniumPresence` for a verdict it has no basis to give. The bounds guards still fire on those spectra, so standing down isn't a hole in them. `CheckOxoniumPresence` keeps returning false as a unit, but its doc comment no longer claims that achieves neutrality; it says the opposite and points at the skip.

**2. The threshold question** (@acesnik, `GlycoPeptides.cs:572`) — closed on your and @RayMSMS's read: the shared `OxoniumRelativeIntensityThreshold`, no knob. Revisit if measurements on custom-sugar oxonium yields ever appear.

**3. GUI documentation** (@RayMSMS) — new **Two ways in: the GUI dialog or the file** section, mapping every field of **Create new monosaccharide** to its TSV column, with **Diagnostic Ions (m/z)** called out as the one this guide is about. Quick start and the worked example now lead with the dialog, and the restart rule is split, because it differs: hand-edits need a restart, the dialog doesn't.

**4. The moved TSV** (@RayMSMS) — the guide described the old `Glycan_Mods` layout throughout; rewritten to the data-directory root, with the one-time legacy migration noted rather than the history deleted. The same staleness had reached the code: both custom-oxonium fixtures rebuilt the shipped-TSV path by hand under `Glycan_Mods`, so after the move `RestoreStartupMonosaccharides` hit an always-false `File.Exists` and silently restored nothing — the exact failure the restore-to-startup change was added to prevent, arriving by a different door. Both now ask `GlobalVariables.CustomMonosaccharidePath`.

**Tests.** Three added, the first being the case @acesnik named:

- `DiagonsticFilter_NoHexNAcReference_MonoPresent_StrictRuleSkippedAndCandidateKept` — rejected before this change, kept now
- `DiagonsticFilter_NoHexNAcReference_IonPeakPresentMonoAbsent_StrictRuleSkippedAndCandidateKept`
- `DiagonsticFilter_NoHexNAcReference_UndersizedIntensityArray_StillThrows`

Local: Release build of the full solution clean (0 errors), 23/23 in the two custom-oxonium fixtures, 49/49 across the monosaccharide tests, 141/141 across the glyco suite.

**Currency and coverage.** Merged to `7c9f3e7` this morning, 0 commits behind `master`, `MERGEABLE` with no conflicts. Patch coverage was 98.82% before this round with codecov/patch and codecov/project both green; the single partial is one branch of `if (oxoniumIonIntensities[R144] <= 1e-8)` in `GlycoSearchEngine`, which this PR only *renamed* — the literal `5` became `OxoniumIonReservedIndices.R144` — and which was equally untested on `master`. It sits inside a private method reachable only through a full glyco search, so covering it means building an integration fixture for a branch whose behaviour this PR does not touch. Happy to add one if you'd rather not carry it, but I'd rather not pad the patch to chase the number. Every line this PR actually adds is covered.

The PR description is updated to match: both routes in, the corrected file location, the intensity threshold, the duplicate rule and the stand-down case.

https://claude.ai/code/session_01Cw7mJPfKbte5nuyDqVUDyH

trishorts:
Note on the merge state, since it will look like a step backwards: `master` has `dismiss_stale_reviews` on, so pushing b5989b3 dismissed @RayMSMS's approval from this morning. Not a change of mind — just the branch rule. Re-requested from both of you.

The bar is `required_approving_review_count: 3`, so this needs a third reviewer beyond @acesnik and @RayMSMS whenever the two of you are happy with the round.

RayMSMS (approved):

trishorts:
@acesnik @nbollis — could you take a look when you get a chance? All nine review threads are resolved, CI is 13/13 green, and @RayMSMS approved `ad66fca`. `master` needs three approvals, so this is waiting on two more. Thanks!

acesnik (commented):
🤖 Round three did what I asked and did it better than the way I suggested it: skipping the custom branch outright rather than asking `CheckOxoniumPresence` for a verdict it has no basis to give. The skip condition at `GlycoPeptides.cs:528` is the exact complement of the `<= 0` guard at `:580`, so the two can't drift apart, and putting the `continue` at `:541` *below* the `Kind[]` check at `:533` means standing down doesn't quietly disarm the guards — which was the previous round's failure mode arriving from the other side. Since `ScanOxoniumIonFilter` writes a literal `0.0` for an unmatched ion (`:32`), `> 0` is the right test here rather than the `1e-8` epsilon the engine uses for R144.

I rebuilt `ad66fca` on macOS arm64 (.NET 10, `UbuntuMac`, 0 errors, no new warnings) and ran the three fixtures through a throwaway harness: 23 pass, plus both `[Explicit]` perf tests. Then I mutated the fix away — deleted `:528` and `:541-544` — and got 1 failure of 23. Only `DiagonsticFilter_NoHexNAcReference_MonoPresent_StrictRuleSkippedAndCandidateKept` has power over this change. `..._IonPeakPresentMonoAbsent_...` passes either way, because with the reference at zero `CheckOxoniumPresence`'s own guard already returns false and `false == false` accepts; it duplicates the direct assertion at `CustomOxoniumFilterTests.cs:120`. `..._UndersizedIntensityArray_StillThrows` also passes either way, but it isn't idle — I hoisted the stand-down above the length check as a second mutation and it's the one test that catches that, so it's an ordering guard worth keeping under a name that says so. One fidelity note on the test that does bite: `intensities[OxoniumIndex_R138] = 0` at `:325` is a no-op write into a fresh array, so index 9 is zero too, and the engine's own gate at `GlycoSearchEngine.cs:667` would have dropped that scan before `DiagonsticFilter` ever saw it. Setting `HexNAc204` non-zero would make the fixture match the scenario its comment describes.

The rest is the new documentation, and the first item under it is a real bug. `docs/…:188` says column-4 ions are "matched against the spectrum, and folded into the diagnostic-ion score", and `:294` says to enter observed m/z with "No hydrogen offset is applied". Those can't both hold. Every built-in in `GlycanDiagnosticIons` is emitted as `13805550 - hydrogenAtomMonoisotopicMass` (`Glycan.cs:159-186`) — m/z converted to neutral mass — while the custom loop at `:188-199` adds the user's number raw. That set becomes `Modification.DiagnosticIons` (`:72-74`), and mzLib puts the value straight into `Product.NeutralMass`, with the comment "the diagnostic ion is assumed to be annotated in the mod info as the *neutral mass*" (`PeptideWithSetModifications.cs:547-556`). So `175.02482` gets searched as a neutral mass, at m/z 176.0327 — roughly 287× a 20 ppm window there, so it never matches and scores nothing. This PR's filter path reads the same number as m/z and converts it correctly (`GlycoPeptides.cs:28`). The filter role is right; the scoring role is a proton high. It came in with #2660 and is in 1.1.8 through 1.1.10, so the fix probably wants its own PR — but the docs here assert the broken half works, and the FAQ repeats it at `:288`.

The worked example also picks the one search type the filter never runs in. `DiagonsticFilter` has a single production call site, `GlycoSearchEngine.cs:479`, inside `FindOGlycan` (`:435`), reached only through `MatchGlycopeptide` (`:733`), which `:206-208` calls for `OGlycanSearch` and `N_O_GlycanSearch` only; `NGlycanSearch` goes to `FindNGlycan`, whose sole oxonium gate is the 204 check. But `docs/…:219` sets the goal as N-glycans, `:237` says "set the search type to N-glycan", and `:239` promises the rejections. Follow it exactly and nothing is filtered and nothing says why. Either the example should use N+O, or the scope limit needs stating — the TSV banner doesn't mention it either.

Then the FAQ at `:290`: "Can custom ions tie to a built-in monosaccharide? Yes — put a built-in name (e.g. `HexNAc`) in column 1". `Glycan.cs:308-309` throws when `NameCharDic` already has the name, `GlycanDatabase.cs:139-143` wraps it, and `GlobalVariables.LoadGlycans` (`:497`) runs from `SetUpGlobalVariables` at `MainWindow.xaml.cs:45`, before `InitializeComponent()`, with no try/catch and no `DispatcherUnhandledException` handler in `App.xaml.cs`. The window never opens, so "Open mods/data folder" is unreachable. The same page says "do not redefine" built-ins at `:253`.

That mechanism has a second trigger that isn't hypothetical, and it's the one I'd most want settled before merge. `DiagnosticIonMasses` shipped in 1.1.8, 1.1.9 and 1.1.10, so users can already have column 4 populated, and `EnsureCustomMonosaccharideFileExists` only writes the template when the file is *missing* — an upgrade preserves theirs. If any ion they listed is within 0.01 Da of a built-in, a file that loaded fine yesterday now takes the GUI down at startup. If it isn't, those ions silently become hard reject gates, since `OxoniumIonFilt` defaults true (`GlycoSearchParameters.cs:14`) — and the new banner explaining that never reaches them, precisely because their file already exists. Warn-and-skip at load instead of a throw would cover both directions, and this probably deserves a line in the release notes.

Smaller, but it undercuts a comment: the justification for the 0.01 Da collision window (`Glycan.cs:334-339`, "Wider than any realistic product tolerance at these masses (20 ppm at m/z 204 is 0.004 Da)") fails at the top of the array it checks against. `AllOxoniumIons` ends with 657.23544 and 673.23035, where 20 ppm is 0.0131 and 0.0135 Da — so an ion ~0.0096 Da from either passes the duplicate check and still matches the same peak as the built-in, which is the failure the check exists to prevent. Wider again at low-res product tolerance. My earlier check was the other direction (the 0.984 Da minimum gap *between* built-ins), so it didn't cover this.

The length here is the documentation surface, not doubt about the direction — the engine change itself is three lines and I think they're the right three. What I did not do: run a search on real spectra. Everything above about results is read off the source, including "N-glycan output is unaffected", which I'd want confirmed on real input before either of us states it; the glyco suite won't run on macOS (backslash path literals in the fixtures), so Windows CI stays the only real signal. The mutation counts are measured; the proton and collision-window items are code reading and arithmetic.

Holding the checkmark for Role 1 and the upgrade path rather than requesting changes, since the first is arguably #2660's problem and the second is a rollout judgement call. Happy to be argued out of either.

nbollis (dismissed):

trishorts:
Round four, in `2a5616e2f`. Every item addressed; two of them changed my mind about decisions made in earlier rounds, so I've said why below rather than just claiming them fixed.

**1. Role 1 is a proton high.** You're right, and I fixed it here rather than deferring it. `GlycanDiagnosticIons` now emits custom ions as `ion - hydrogenAtomMonoisotopicMass`, the same conversion the built-in literals have always carried, because that set becomes `Modification.DiagnosticIons` and mzLib assigns it straight to `Product.NeutralMass`. It's arguably #2660's bug, but this PR is the one that writes down what Role 1 does, and documenting a scoring path that scores nothing seemed worse than a one-line fix in a file this PR already touches. Nobody can have tuned around an ion that never matched.

The reason it survived review twice: `TestNGlyco.LoadCustomMonosaccharides_DiagnosticIonsEmittedWhenPresent` asserted the raw m/z came back out, so the bug had a test holding it in place. That test now asserts the neutral-mass form and anchors it against a built-in ion emitted from the same glycan, so it fails if either convention drifts. Users who added ions under 1.1.8–1.1.10 will see diagnostic-ion scores rise for glycans carrying their sugar; that's noted in the docs.

**2. The upgrade path.** Taken, and I think you're right that this was the one worth settling before merge. A collision found *in the file* no longer throws: the offending ion is skipped, the monosaccharide loads without it, and a warning naming file, line and ion goes to `GlobalVariables.ErrorsReadingMods` — which `LoadModifications` initialises one line before `LoadGlycans`, and which the GUI drains in `PrintErrorsReadingMods` once the window is up. The dialog path still throws, since there the user is typing the ion and can correct it; the two now share one collision rule via `Glycan.DescribeDiagnosticIonCollision` and differ only in what they do with the answer.

This reverses the round-two decision that duplicates are rejected at load time. That decision was right about a *new* file and wrong about an *existing* one, which is the case you found: refusing to start is not corrective when starting is how you reach the file.

I also took your second direction, which warn-and-skip doesn't cover on its own. Any custom monosaccharide with column-4 ions now produces one notification saying those ions act as strict gates under a filter that defaults on. That's the only way the information reaches a user whose file predates the banner, since their file is never rewritten. It stays quiet for an empty column 4, including the shipped template — there's a test pinning that, because a warning nobody needs every startup would get ignored along with the one they do.

**3. The collision window.** Also right, and my 0.984 Da check didn't cover it either. The window is now the wider of a 0.01 Da floor and 20 ppm of the heavier ion. The floor still governs below ~500 m/z, where it's already several times the tolerance; 20 ppm takes over at 657.235 and 673.230, where it's 0.0131 and 0.0135 Da. Three tests: one ion 0.012 Da from 657.235 now rejected, the same 0.012 Da offset from 204.087 still accepted so the widening isn't mistaken for widening everywhere, and a sweep asserting the window is never narrower than 20 ppm at any built-in.

**4. Test fidelity at `:325`.** Correct — a no-op write, and with 204 at zero the engine's own gate would have dropped that scan before `DiagonsticFilter` saw it. `HexNAc204` is now set non-zero, and the absence of 138.055 is asserted rather than written, so the fixture matches the low-energy-HCD scenario its comment describes. I made the same change to the `..._IonPeakPresentMonoAbsent_...` sibling, which had the same shape.

Thanks also for the mutation testing — the note that `..._UndersizedIntensityArray_StillThrows` is the only test with power over the stand-down's *ordering* is why it kept its name rather than getting folded into the others.

**5. Search-type scope.** The worked example did pick the one search type the filter never runs in. It's now N+O-glycan, with a callout saying why, and Role 2's heading and body state the scope: the gate is reached only through `MatchGlycopeptide`, so `OGlycanSearch` and `N_O_GlycanSearch` only, and a pure N-glycan search keeps Role 1 and filters nothing. The TSV banner says it too.

**6. The FAQ on reusing a built-in name.** Deleted and replaced with the opposite, which is both the truth and a hazard: it's rejected, and rejected before the window exists. It now points at the "do not redefine the built-ins" rule three sections up that it was contradicting.

**What I have and haven't verified.** Release build clean; 43 tests across the custom-oxonium and monosaccharide fixtures and 118 across the glyco suite, all green locally, CI running now. Still no search on real spectra — so "N-glycan output is unaffected" remains read off the source for both of us, and the Role 1 change is the one I'd most want a real-data look at before we both state what it does to scores, since it moves ions that previously matched nothing into the matcher.

One judgement left for you: the Role 1 fix changes results for anyone who set column 4 under 1.1.8–1.1.10. I've documented it, but if you'd rather it land as its own PR against #2660 so it gets its own release note and its own bisect point, say so and I'll split it back out.

Issue #21, opened 2017-01-07
Validate calibrated mzML files give better results in other software


stefanks:

stefanks:
Depends on https://github.com/smith-chem-wisc/mzLib/issues/33

trishorts:
probably need to write indexed .mzml

stefanks:
Let's try to validate by MsFragger

PR #2666, opened 2026-05-26
Truncation search: follow-on task for N-/C-terminal proteoform truncations (core)

trishorts:

# Truncation search: a follow-on task for proteoform N-/C-terminal truncations

## Summary

Adds a new **`TruncationSearchTask`** that runs after a normal top-down search and discovers **N- and C-terminally truncated proteoforms** of the parents identified in that search. It is a standard MetaMorpheus task (run-list `[Search, Truncation]`, dispatchable from CMD) with its own TOML schema and per-class FDR, and it writes results using the existing PSM/proteoform `.psmtsv` columns.

The task is **self-contained and depends on no unreleased code** — it builds against the current released mzLib and uses only standard MetaMorpheus/mzLib APIs (digestion, fragmentation, `MatchFragmentIons`/`CalculatePeptideScore`, `FdrAnalysisEngine`, the standard PSM writer). It works on either classic deconvolution or an external precursor source; nothing in the truncation code requires a particular precursor pipeline.

> **Scope reduced.** Per @acesnik's review this PR was three separable features. It is now **the core terminal-truncation task only**. The other two are split into follow-on PRs, stacked on this one:
> - sequence-tag candidate filtering — `truncation-search-tags`
> - direct internal-fragment search — `truncation-search-internal`
>
> Nothing was dropped: the top of the stack is byte-identical to the tree that was under review before the split.

## How it works (three passes)

1. **Parents from the upstream search.** The Search task deposits its resolved, FDR'd PSMs into a small `TaskChainContext` (in-memory hand-off; a disk `AllProteoforms.psmtsv` fallback also exists). The truncation task dedups those to proteoform level, filters to confident parents, expands pipe-ambiguous identifications into individual parents, and adds reverse decoys for any parent lacking one.
2. **Pass 2 — candidate location.** A fragment-bin index (N- and C-series of every parent) plus a truncation-aware mass acceptor locate candidate parents per scan; each candidate is scored twice with the standard scorer restricted to the N-series and to the C-series, and the highest (parent, series) wins. Scans already explained intact are skipped.
3. **Pass 3 — chopping + scoring.** The terminus opposite the winning series is chopped one-or-more residues until the truncated mass matches at an allowed notch; surviving mods are re-indexed onto the shorter form (with terminal-mod retention), both series are scored with the standard scorer, and duplicate truncations collapse into one ambiguous PSM. The truncation type (`N-/C-terminal truncation(start-end)`) rides the existing **Description** column, mirroring the existing `chain(2-121)` descriptors.

Intact PSMs are inherited as `full-length`, then everything is pooled into one **per-class FDR/PEP** pass and written to `AllTruncatedPSMs.psmtsv` / `AllTruncatedProteoforms.psmtsv`.

`SeedParentsFromDatabase` (opt-in, off by default) seeds parents by digesting the database instead, removing the requirement that a truncation's parent be observed intact.

## Design doc

`docs/Truncation-Search.md` ships with the code, following the pattern #2665 set. It carries the numbered decisions the doc-comments cite, so a `(decision #14)` in the source resolves to something a reader can actually open.

## Tests

64 NUnit tests covering the acceptor, the Pass 2 engine (dual-series scoring, parent filtering, oversized-parent handling, pipe-split, q-filter), the chopper (C/N chop, mod retention, collapse, notch, early-exit bound), reverse-decoy generation, pooled FDR + output columns, TOML round-trip / CMD dispatch, and an end-to-end `[Search, Truncation]` run that asserts well-formed output. Full MetaMorpheus suite green locally.

## Notes

- New `MyTask.Truncation` enum member; `CMD/Program.cs` dispatches `Toml.ReadFile<TruncationSearchTask>`, and `-g` writes a default `TruncationSearchTask.toml`.
- `TaskChainContext` is a generic deposit/retrieve slot. `EverythingRunnerEngine` creates it **only when the run list contains a task that reads it**, and clears it as soon as that task has run, so a search does not pin its PSM graph through run lists that will never consume it.
- Benchmarking instrumentation is deliberately not on the TOML surface: the perf log is driven by the `MetaMorpheusTruncationPerfLog` environment variable.
- Also removes a stray committed VS hot-reload temp file (`GUI_h05aw1sq_wpftmp.csproj`).


trishorts:
### Manual unit-test review

Every unit test this PR adds was reviewed individually by the author, rather than relied on solely as automated/AI-generated checks. Of the 65 tests added:

- 55 were approved as written;
- 8 received explanatory doc-comments (the Met-excision / internal-fragment cases in `TruncationChopperTests`);
- 1 was strengthened (`Acceptor_ProseString_DescribesRule` now asserts the full acceptance rule, not just that the prose mentions "observed");
- 1 was removed as a tautology (`Scaffolding_TestProjectWired`, a leftover `Assert.True` placeholder that tested nothing) — leaving 64 retained.

This satisfies the author's standing AI-use policy that AI-assisted tests receive human review before being relied upon. Recorded 2026-06-22.

MetaMorpheus/EngineLayer/Truncation/InternalTruncationSearch.cs
```diff
@@ -0,0 +1,163 @@
...
+                            int notch = chopAcceptor.Accepts(precursorMass, form.MonoisotopicMass);
+                            if (notch < 0) continue;
+
+                            form.Fragment(dissociationType, FragmentationTerminus.Both, products, commonParameters.FragmentationParameters);
+                            List<MatchedFragmentIon> matched = MetaMorpheusEngine.MatchFragmentIons(scan, products, commonParameters);
+
+                            int nIons = 0, cIons = 0;
+                            foreach (MatchedFragmentIon ion in matched)
+                            {
+                                if (ion.NeutralTheoreticalProduct.Terminus == FragmentationTerminus.N) nIons++;
```
pcruzparri, line 123: **Critical** — `products` list accumulates across parent/span candidates, inflating match counts and scores

_Note: this is in the internal-truncation search path, gated behind `SearchInternalTruncations` (default false); it does not affect the default terminal-truncation pipeline._

**Problem:**
`var products = new List<Product>()` is allocated once per scan (line 96) and reused for every parent, start position, and end position in the triple-nested loop. Each `form.Fragment(..., products, ...)` call (this line) **appends** theoretical fragments to the same list without clearing it. `MatchFragmentIons` (line 124) then scores the scan against the accumulated superset of fragments from all prior candidates, not just the current internal span.

**Impact:**
Later candidates inherit matching fragments from unrelated parents/spans, letting them clear the bilateral ion gate (`minIonsPerTerminus`) on false matches and producing inflated propensity-weighted scores. The per-scan best-candidate selection (`if (best == null || score > best.Score)`) is therefore unreliable — an internal span can win because fragments from a *previous* parent happened to match the scan.

**Suggested fix:**
Add `products.Clear();` immediately before this `Fragment` call, inside the innermost candidate loop. Alternatively, move `var products = new List<Product>();` inside the `for (int pi = 0; ...)` loop so each parent starts fresh.

**Affected test scenarios:**
- Internal search — two parents sharing a fragment mass — later candidate's matched-ion count includes earlier parent's fragments
- Internal search — bilateral gate false-positive — a candidate with few own matches clears `minIonsPerTerminus` via fragments inherited from a prior parent
- Internal search — single parent, multiple spans — score of span N includes fragments from span N-1 of the same parent

MetaMorpheus/EngineLayer/Truncation/TruncationSearchEngine.cs
```diff
@@ -0,0 +1,507 @@
...
+                {
+                    TruncationParent parent = _parents[candidateIds[rank]];
+
+                    parent.Proteoform.Fragment(dissociationType, FragmentationTerminus.N, nTermProducts, _commonParameters.FragmentationParameters);
+                    double nScore = MetaMorpheusEngine.CalculatePeptideScore(scan.TheScan,
+                        MetaMorpheusEngine.MatchFragmentIons(scan, nTermProducts, _commonParameters));
+
+                    parent.Proteoform.Fragment(dissociationType, FragmentationTerminus.C, cTermProducts, _commonParameters.FragmentationParameters);
+                    double cScore = MetaMorpheusEngine.CalculatePeptideScore(scan.TheScan,
+                        MetaMorpheusEngine.MatchFragmentIons(scan, cTermProducts, _commonParameters));
```
pcruzparri, line 242: **Critical** — `nTermProducts`/`cTermProducts` accumulate across Pass-2 rescoring candidates, biasing the winning (parent, series) pair

**Problem:**
`nTermProducts` and `cTermProducts` are created once per scan (lines 223–224) before the candidate rescoring loop (`for (int rank = 0; ...)`). This `Fragment` call and its C-series counterpart (line 246) **append** to the same two lists every iteration without clearing. When candidate N is scored, `MatchFragmentIons` sees candidates 0..N combined, so candidate N gets credit for fragments from earlier-ranked parents that happen to coincide with scan peaks.

**Impact:**
This is the authoritative final score that determines the Pass-2 winner, not the approximate index tally used for pruning. Later-ranked candidates are systematically inflated, so the wrong (parent, series) pair can win and feed incorrect input into Pass 3 chopping. The bias grows with the shortlist size (`MaxCandidatesToScore = 50`).

**Suggested fix:**
Clear both lists at the top of each candidate iteration: add `nTermProducts.Clear();` before this line and `cTermProducts.Clear();` before line 246. Or declare fresh lists inside the `for` loop body.

**Affected test scenarios:**
- Pass 2 — two candidate parents sharing a b-ion mass — rank-1 candidate's N-series score includes rank-0's b-ions
- Pass 2 — shortlist of 50 — rank-49 parent scored against union of all 50 parents' fragments
- Pass 2 — winner selection — a lower-ranked parent wins only because inherited fragments push its score above the higher-ranked parent's true score

MetaMorpheus/EngineLayer/Truncation/TruncationSearchEngine.cs
```diff
@@ -0,0 +1,507 @@
...
+                    binsC.Clear();
+                    foreach (DissociationType dissociationType in _dissociationTypes)
+                    {
+                        _parents[id].Proteoform.Fragment(dissociationType, FragmentationTerminus.Both, products, _commonParameters.FragmentationParameters);
+                        foreach (Product product in products)
+                        {
+                            if (double.IsNaN(product.NeutralMass))
+                            {
+                                continue;
+                            }
```
pcruzparri, line 442: **High** — `products` list accumulates across parents during fragment-index build, contaminating the index

**Problem:**
In `BuildFragmentIndex`, `var products = new List<Product>()` (line 432) is allocated once per parallel chunk. This `Fragment` call appends each parent's fragments to the same list without clearing it. Although `binsN`/`binsC` are cleared per parent (lines 428–429), they are populated by `foreach (Product product in products)` (line 443), which iterates the accumulated products from *all* preceding parents in the chunk — not just the current parent.

**Impact:**
Parent *id* is indexed under the union of its own fragment bins and every earlier parent's fragment bins. A scan peak matching parent 0's fragment mass also tallies as a hit for parents 1, 2, ... in the same chunk, inflating their match counts in `RankCandidateParents`. This corrupts the candidate shortlist that feeds the authoritative rescoring (see the Critical finding above), letting wrong parents through and pushing the true best parent deeper in the ranking.

**Suggested fix:**
Add `products.Clear();` at the start of each parent iteration, right after `binsN.Clear(); binsC.Clear();` (line 428), so each parent's bins come only from its own fragments.

**Affected test scenarios:**
- Index build — two parents in the same chunk sharing a fragment mass — parent 1's bin set includes parent 0's fragment bin
- Index build — chunk of 1000 parents — parent 999's index entry references bins from all 999 prior parents' fragments
- Candidate ranking — scan with a peak matching only parent 0's fragment — all later parents in parent 0's chunk get an extra tally they shouldn't

pcruzparri (commented):
K-LLM ensemble code review — 3 finding(s) at severity >= High, posted as inline comments.

All three share the same root cause: a single `List<Product>` is created outside a candidate loop and passed to `PeptideWithSetModifications.Fragment(...)`, which **appends** without clearing, so `MatchFragmentIons` scores each candidate against an ever-growing union of fragments from all prior candidates. This inflates matched-ion counts and biases winner selection.

Method: 9 independent open-source LLM reviewers (Big Pickle, DeepSeek V4 Flash/Pro, MiMo-V2.5, Nemotron 3 Ultra, North Mini Code, Qwen 3.5/3.6 Plus, Kimi K2.6) each analyzed a differently-shuffled copy of this PR's diff. Findings were clustered by (file region, root cause), majority-voted, then validated against the PR branch source. 8/9 reviewers returned results; 3 findings survived validation.

All three are in code added by this PR. One reviewer (nemotron-3-ultra-free) returned empty output on two attempts; it is excluded from the vote denominator.

MetaMorpheus/EngineLayer/Truncation/InternalTruncationSearch.cs
trishorts: Fixed in 3aafb51b. Added `products.Clear();` immediately before the `form.Fragment(...)` call inside the innermost span loop, so each internal span is scored only against its own b/y ladder.

MetaMorpheus/EngineLayer/Truncation/TruncationSearchEngine.cs
trishorts: Fixed in 3aafb51b. Both `nTermProducts` and `cTermProducts` are now cleared at the top of each candidate iteration (before the N- and C-series `Fragment(...)` calls), so each rescored candidate's score reflects only its own fragments.

MetaMorpheus/EngineLayer/Truncation/TruncationSearchEngine.cs
trishorts: Fixed in 3aafb51b. `products.Clear();` is now called at the top of the per-dissociation-type loop in `BuildFragmentIndex`, so each parent's index bins are populated only from its own fragments (this also removes the redundant re-processing of earlier dissociation types within the same parent).

trishorts:
Updated the branch and addressed the review:

- Merged current `smith-chem-wisc/master` into the branch (now current with master).
- Fixed all three K-LLM ensemble findings (the shared `List<Product>` accumulation root cause) in `3aafb51b`: cleared the reused fragment list before each `Fragment(...)` call at the three sites — internal-span loop, Pass-2 N/C rescoring, and the fragment-index build. Per-comment replies inline.
- Truncation test suite green locally (76/76, Release/x64).

MetaMorpheus/EngineLayer/Truncation/InternalTruncationSearch.cs
trishorts: On closer inspection this is a false positive, and I've reverted the earlier change (commit `a9b35ac3`). `PeptideWithSetModifications.Fragment(...)` calls `products.Clear();` as its first statement (mzLib `PeptideWithSetModifications.cs:240`), so the list is reset on every call — reusing one `List<Product>` across candidates is the intended, memory-efficient MetaMorpheus idiom and does not accumulate. Verified at runtime against mzLib 1.0.579: fragmenting one parent into a list yields 79 products; fragmenting a second parent into the **same** list yields 29 (its own count), not 108. `MatchFragmentIons` is always called immediately after `Fragment(...)`, so it only ever sees the current candidate's fragments. No score inflation occurs.

MetaMorpheus/EngineLayer/Truncation/TruncationSearchEngine.cs
trishorts: On closer inspection this is a false positive, and I've reverted the earlier change (commit `a9b35ac3`). `PeptideWithSetModifications.Fragment(...)` calls `products.Clear();` as its first statement (mzLib `PeptideWithSetModifications.cs:240`), so the list is reset on every call — reusing one `List<Product>` across candidates is the intended, memory-efficient MetaMorpheus idiom and does not accumulate. Verified at runtime against mzLib 1.0.579: fragmenting one parent into a list yields 79 products; fragmenting a second parent into the **same** list yields 29 (its own count), not 108. `MatchFragmentIons` is always called immediately after `Fragment(...)`, so it only ever sees the current candidate's fragments. No score inflation occurs.

MetaMorpheus/EngineLayer/Truncation/TruncationSearchEngine.cs
trishorts: On closer inspection this is a false positive, and I've reverted the earlier change (commit `a9b35ac3`). `PeptideWithSetModifications.Fragment(...)` calls `products.Clear();` as its first statement (mzLib `PeptideWithSetModifications.cs:240`), so the list is reset on every call — reusing one `List<Product>` across candidates is the intended, memory-efficient MetaMorpheus idiom and does not accumulate. Verified at runtime against mzLib 1.0.579: fragmenting one parent into a list yields 79 products; fragmenting a second parent into the **same** list yields 29 (its own count), not 108. `MatchFragmentIons` is always called immediately after `Fragment(...)`, so it only ever sees the current candidate's fragments. No score inflation occurs.

trishorts:
Correction / follow-up on the K-LLM review: after verifying against the source and at runtime, all three findings are **false positives**, and I've reverted my earlier change (`a9b35ac3`).

The findings assumed `PeptideWithSetModifications.Fragment(...)` appends to the passed list. It does not — it calls `products.Clear();` as its first line (mzLib `PeptideWithSetModifications.cs:240`), in the perf-critical path that is deliberately called once per candidate. Reusing a single `List<Product>` across candidates is therefore the intended, memory-efficient idiom; there is no cross-candidate accumulation and no score inflation. `MatchFragmentIons` is always invoked immediately after `Fragment(...)`, so it only sees the current candidate's products.

Runtime check (mzLib 1.0.579): fragment parent A into a list → 79 products; fragment parent B into the **same** list → 29 (B's own count), not 108.

Net state of the branch: synced to current `smith-chem-wisc/master`; the three engine files are unchanged from their reviewed form; truncation test suite green locally (76/76, Release/x64). Marking the three threads resolved.

pcruzparri (dismissed):

MetaMorpheus/TaskLayer/TruncationSearchTask/TruncationSearchParameters.cs
```diff
@@ -0,0 +1,106 @@
...
+        public string Pass1ProteoformsFilePath { get; set; } = null;
+
+        /// <summary>
+        /// Parent inclusion q-value threshold (decision #3). Deliberately permissive (10% FDR) to cast
+        /// a wide net of parents; the truncation search's own FDR analysis polices the final output.
+        /// </summary>
+        public double ParentQValueThreshold { get; set; } = 0.10;
+
+        /// <summary>
+        /// Notch acceptor used when chopping a parent down to the precursor mass in Pass 3 (decision #80).
```
acesnik: `decision #80` looks like a typo — every other reference in this PR is in the #1–#20 range.

Which points at something bigger. I count over twenty comments across this PR citing `01_Architecture.md` (decisions #1 through #20) and `03_Benchmarks.md`, and neither document is in the repository — this PR adds no docs at all. So the stated rationale for most of the non-obvious design choices here is unavailable to anyone reading the code, now or in two years. This one is the sharp edge of it: I can't tell whether `#80` is a typo for `#8` or `#10` or something else, *because* the document isn't there to check.

#2665 shipped `docs/Custom-Monosaccharides-and-Diagnostic-Ions.md` alongside the feature, which is exactly the right pattern. Either commit the architecture doc with this PR, or inline the handful of decisions the comments actually lean on. A dangling `(decision #14)` is worse than no citation, because it reads like the reasoning is written down somewhere.

MetaMorpheus/TaskLayer/TruncationSearchTask/TruncationSearchParameters.cs
```diff
@@ -0,0 +1,106 @@
...
+        public MassDiffAcceptorType MassDiffAcceptorType { get; set; } = MassDiffAcceptorType.ThreeMM;
+
+        /// <summary>Custom mass-diff-acceptor string, used only when <see cref="MassDiffAcceptorType"/> is Custom.</summary>
+        public string CustomMdac { get; set; } = null;
+
+        /// <summary>Custom output-folder name used for benchmark runs (decision #20). Null = default naming.</summary>
+        public string CustomOutputFolderName { get; set; } = null;
+
+        /// <summary>
+        /// Optional path to the rolling, append-only <c>perf_log.tsv</c> (03_Benchmarks.md). When set, the
```
acesnik: This is benchmarking scaffolding shipping in the product. `PerfLogPath` and `CustomOutputFolderName` ("used for benchmark runs") are user-facing TOML parameters that exist for your benchmarking rather than for anyone running a search, and they're backed by `PerfLogger.cs` — 162 lines in `TaskLayer` — plus `TruncationTimings.cs`.

It's all opt-in and the null path is a genuine no-op, so nothing misbehaves. But these land in the generated TOML that every user sees and has to ignore, and they're the kind of thing that outlives the benchmarking campaign that motivated it. I'd rather see this as a separate PR, or kept out of the serialized parameter surface — an environment variable or a test-only hook would carry the same benchmarking capability without becoming part of the task's public schema.

MetaMorpheus/EngineLayer/Truncation/ProteoformChopper.cs
```diff
@@ -0,0 +1,202 @@
...
+            int length = parent.BaseSequence.Length;
+
+            for (int chopped = 1; chopped <= length - 1; chopped++)
+            {
+                int chopFromN = terminusToChop == FragmentationTerminus.N ? chopped : 0;
+                int chopFromC = terminusToChop == FragmentationTerminus.C ? chopped : 0;
+
+                PeptideWithSetModifications truncated = BuildTruncated(parent, chopFromN, chopFromC);
+
+                int notch = massDiffAcceptor.Accepts(targetMass, truncated.MonoisotopicMass);
```
acesnik: This loop has no early exit, and its sibling `ChopInternalCandidates` below does.

Chopping is monotonic — every residue removed lowers the mass, and dropped terminal mods lower it further — so once `truncated.MonoisotopicMass` falls below the lowest mass the acceptor could accept, no further chop can ever match. As written this runs all the way to `length - 1` on every non-matching scan, and each iteration builds a whole `PeptideWithSetModifications`, which is not cheap: the constructor runs `DetermineFullSequence()` and `UpdateCleavageSpecificity()`, and then `MonoisotopicMass` is computed on top. On a 500-residue proteoform that's ~499 full object constructions to conclude nothing matched, once per scan. `ChopInternalCandidates` already does the right thing with its `break`.

Related, in that sibling: `InternalChopMassMargin = 50.0` is a magic number with no derivation in the comment. The bound you actually want is the acceptor's minimum acceptable mass, which it can tell you; 50 Da is presumably a safe over-estimate for the notch span, but ThreeMM notches span ~3 Da, so it's unclear what the other 47 are insuring against.

acesnik (commented):
:robot: I read the whole history here first — pcruzparri's K-LLM ensemble review, your two follow-ups, and the six resolved threads — so this doesn't retread settled ground.

**On the three reverted findings: your analysis is right, and I confirmed it independently.** `PeptideWithSetModifications.Fragment(...)` does call `products.Clear()` as its first statement (mzLib `PeptideWithSetModifications.cs:240`, inside the "be very careful about allocating" block), so reusing one `List<Product>` across candidates is the intended idiom and nothing accumulates. I also checked the inverse failure the clear makes possible — that a second `Fragment(...)` call could wipe an earlier one's output — and it doesn't bite here: in `BuildFragmentIndex` the `foreach (Product product in products)` sits *inside* the `foreach (DissociationType ...)` loop, so each dissociation type's products are consumed before the next call clears them, and `binsN`/`binsC` correctly accumulate across types while being cleared per parent. Pass-2 rescoring uses a single resolved dissociation type, so it's unaffected either way. Reverting `a9b35ac3` was the right call on all three.

The engineering here is careful and it shows — the acceptor's asymmetric interval logic is right, the mod re-indexing arithmetic in `BuildTruncated` checks out key-by-key against the one-is-N-terminus convention, and reusing the existing `Description` column for the truncation descriptor rather than inventing a new one is exactly the kind of restraint that keeps output readable. Removing `GUI_h05aw1sq_wpftmp.csproj` is also correct; that's a Visual Studio hot-reload temp file, not a project.

I'm commenting rather than approving, for one reason that isn't about any line of code: **at 4,414 lines across 29 files this contains three separable features** — the core terminal-truncation task, the sequence-tag filter (`SequenceTagExtractor` + `ProteinTagIndex` + two parameters), and the internal-truncation search (`InternalTruncationSearch` + `FragmentationPropensity` + its own FDR pass and output file). Your own summary files the latter two under "Additional pieces". Each is independently reviewable and independently testable, and split up they'd get real scrutiny instead of a skim. I'd gently offer the K-LLM pass as evidence rather than a criticism of it: three high-severity findings that all turned out to be the same false positive, costing you two round-trips to refute, is what reviewing a diff this size looks like.

Nothing below is a correctness bug. Five notes inline.

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
```diff
@@ -516,7 +516,17 @@ protected override MyTaskResults RunSpecific(string OutputFolder, List<DbForTask
...
-            return postProcessing.Run();
+            MyTaskResults postSearchResults = postProcessing.Run();
+
+            // Hand the resolved, FDR'd PSM set to a downstream TruncationSearchTask via the shared
+            // in-memory task-chain context (decision #1). The consumer dedups to proteoform level and
+            // applies its own permissive parent filter; no-op when no context/consumer is present.
+            // Enforce the no-nulls half of the contract at the deposit site rather than relying on a side
+            // effect inside PostSearchAnalysisTask: AllSpectralMatches is assembled from fixed-length per-file
+            // arrays that hold null slots for unmatched scans. (The consumer also null-filters defensively.)
+            TaskChainContext?.Deposit(taskId, parameters.AllSpectralMatches?.Where(p => p != null).ToList());
```
acesnik: This deposits unconditionally, whether or not anything downstream will ever read it — and it allocates a full filtered copy of the PSM list to do so.

`EverythingRunnerEngine` holds the one `TaskChainContext` for the entire run list and never clears it, so from here to the end of the run every `SpectralMatch` from this search stays reachable, along with the object graph each one pins. For a top-down search feeding a `[Calibrate, GPTMD, Search, ...]` chain that's a large allocation kept alive well past the point the GC would otherwise have taken it, for the benefit of a consumer that usually isn't in the run list at all.

Cheap improvements, in rough order of preference: have the runner clear the context entry once a task has consumed it; or only deposit when a downstream consumer is present in the run list; or at minimum deposit the list you already have rather than a filtered copy and let the consumer null-filter, since the comment notes it does that anyway.

MetaMorpheus/TaskLayer/TaskChainContext.cs
```diff
@@ -0,0 +1,65 @@
+using System.Collections.Generic;
+
+namespace TaskLayer
+{
+    /// <summary>
+    /// In-memory hand-off between tasks in a run list (01_Architecture.md decision #1). A finishing
+    /// task deposits a typed result keyed by its task id; a later task retrieves it by the upstream
+    /// task id it was configured with. Used by <see cref="SearchTask"/>'s post-search analysis to pass
+    /// the deduped proteoform-level SpectralMatch list to <see cref="TruncationSearchTask"/> without a
```
acesnik: Doc and code disagree here. This says the context carries "the deduped proteoform-level SpectralMatch list", but the actual deposit in `SearchTask.cs` is `parameters.AllSpectralMatches` null-filtered — the full PSM-level set, not deduped and not proteoform-level. The dedup to proteoform level happens on the consumer side, which the `SearchTask` comment correctly describes.

Worth fixing the wording, because this is the summary a future reader of `TaskChainContext` will trust when deciding what they can assume about a retrieved payload — and "already deduped" is precisely the wrong thing to assume.

MetaMorpheus/TaskLayer/TruncationSearchTask/TruncationSearchParameters.cs
trishorts: Both fixed.

`#80` was a typo for the chop-notch decisions — it now reads `(decisions #9, #15)`.

On the bigger point, you were right and I took the #2665 pattern: **`docs/Truncation-Search.md` now ships with the code**. It carries the numbered decisions the comments cite (#1–#21, gaps and all), plus the optional features and the benchmarking hook, and all 28 citations are repointed at it by path and section. `01_Architecture.md`, `03_Benchmarks.md`, `02_Implementation.md` and the dangling `doc §11.2.4` no longer appear anywhere in the source. Nothing dangles now.

MetaMorpheus/TaskLayer/TruncationSearchTask/TruncationSearchParameters.cs
trishorts: Off the serialized surface.

- `CustomOutputFolderName` is deleted outright. Nothing ever read it — the declaration was the only occurrence in the repo, so it was pure schema.
- `PerfLogPath` keeps the capability but is now `[TomlIgnore]` and defaults from the `MetaMorpheusTruncationPerfLog` environment variable. Benchmark runs set one variable; tests set the property directly. It no longer appears in the TOML anyone sees.

`PerfLogger.cs` and `TruncationTimings.cs` stay, since they are what the hook drives, but nothing about them is user-facing now.

Two tests hold the line: `PerfLogPath_IsNotPartOfTheTomlSurface` asserts the key is absent from a written TOML and that a stale key in an existing benchmarking TOML still loads (ignored, not fatal), and `PerfLogPath_DefaultsFromEnvironmentVariable` covers the new hook.

MetaMorpheus/EngineLayer/Truncation/ProteoformChopper.cs
trishorts: Both fixed, and the magic number is gone rather than justified.

`ChopUntilMassMatches` now asks the acceptor for its own lower bound once — `GetAllowedPrecursorMassIntervalsFromObservedMass(targetMass)`, minimum across notches — and breaks as soon as the truncated mass falls below it. The monotonicity holds from the first step onwards regardless of mods: the terminal mod is gone after step 1, so every later step only removes a residue. An acceptor with no bounded interval (open search) yields `-inf`, which simply disables the early exit.

`ChopInternalCandidates` now uses that same bound, and `InternalChopMassMargin = 50.0` is deleted. You were right that the other 47 Da were insuring against nothing I could name.

Two tests: `ChopUntilMassMatches_StopsOnceBelowLowestAcceptableMass` counts `Accepts` calls through a wrapping acceptor and asserts **1** where the old loop made 49 on a 50-residue parent; `ChopUntilMassMatches_EarlyExitDoesNotSkipAReachableMatch` proves a 40-residue chop still resolves.

(That method now lives in #2738 along with the rest of the internal search, but the shared bound helper and the terminal early exit are here.)

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
trishorts: Took all three, since they compose.

- `MetaMorpheusTask` gained `ConsumesTaskChainContext` — false by default, true on `TruncationSearchTask`. `EverythingRunnerEngine` now constructs the context **only when the run list contains a consumer**, so for `[Calibrate, GPTMD, Search, ...]` the context is null and this deposit is the existing `?.` no-op.
- The runner calls `TaskChainContext.Clear()` as soon as a consuming task returns, so the payload is not held for the rest of the run list.
- The deposit is `parameters.AllSpectralMatches` as-is. The filtered copy was redundant: both consumer sites null-filter as they go.

MetaMorpheus/TaskLayer/TaskChainContext.cs
trishorts: Fixed. The summary now says the payload is the search's **PSM-level** `SpectralMatch` list as searched — neither deduped nor reduced to proteoform level — and that the consumer does the dedup itself. You are right that "already deduped" is precisely the wrong thing for a future reader to assume.

I also documented the new lifecycle there, since it is the other thing a reader of this class would want to know: the context is handed out only when the run list contains a consumer, and cleared once that consumer has run.

trishorts:
Thanks @acesnik — all five inline notes are addressed, and I took the split.

**The split.** This PR is now the **core terminal-truncation task only**, 25 files. The other two features are follow-on PRs stacked on it:

- #2737 — sequence-tag candidate filtering (9 files, ~585 lines)
- #2738 — direct internal-fragment search (10 files, ~437 lines)

Nothing was dropped or quietly rewritten in the process: the top of the stack is **byte-identical** to the tree you reviewed (`git diff` against the pre-split tree is empty). Each branch builds and its tests pass on its own.

One wrinkle worth flagging: GitHub only opens cross-fork PRs against `master`, so #2737 and #2738 currently show this PR's commits in their diffs. They are drafts, and each reduces to its own feature once the one below it merges. Review order is #2666 → #2737 → #2738.

**The five notes:**

| | |
|---|---|
| `decision #80` + citations to docs that aren't in the repo | `docs/Truncation-Search.md` ships with the code (the #2665 pattern); all 28 citations repointed at it; `#80` → `#9, #15` |
| benchmarking scaffolding on the TOML surface | `CustomOutputFolderName` deleted (nothing read it); `PerfLogPath` is `[TomlIgnore]` + the `MetaMorpheusTruncationPerfLog` env var |
| chopper has no early exit; `InternalChopMassMargin = 50.0` | loop breaks at the acceptor's own lower bound; the magic number is deleted and the internal loop uses the same bound |
| unconditional deposit + filtered copy | `ConsumesTaskChainContext`; the runner builds the context only when a consumer is present and clears it once that consumer has run; deposit is the list as-is |
| `TaskChainContext` doc contradicts the code | summary corrected to PSM-level, not deduped, not proteoform-level |

**One thing that wasn't in the review.** Merging master turned `TestGenerateDefaultTaskTomlsCoversEveryTaskType` (from #2708) red: adding `MyTask.Truncation` without a matching `-g` default toml is exactly the gap that test exists to catch. Fixed here, and `TruncationSearchTask.toml` added to its expected list.

Full MetaMorpheus suite green locally on the merged tree (1839 passed), and on each branch of the stack.

acesnik (dismissed):
🤖 Approving — all five addressed and CI is green.

acesnik (dismissed):

acesnik (approved):

Alexander-Sol:
## Agent review — Approve with nits (please address #1 before merge or document it)

**Summary.** A large, well-documented, self-contained addition: a new `TruncationSearchTask` (three passes — indexed dual single-series parent selection, terminus-directed chopping, pooled FDR) wired into the CMD/run-list layer via `MyTask.Truncation` and an in-memory `TaskChainContext` hand-off. The core algorithms (mod re-indexing on chop, acceptor bounds, dual-series parent selection, decoy generation) are correct and match the mzLib APIs they call. **No confirmed correctness bug in the truncation enumeration/mass logic.** Test coverage is unusually strong for a PR this size.

### 🟠 Medium (plausible, methodological) — pooling inherited full-length targets without their decoys biases truncation FDR optimistic
`TruncationSearchTask.cs:1817` + `TruncationFdr.cs` + `TruncationPass3.InheritAsFullLength`. Confident Pass-1 intact winners are re-injected into the pooled FDR as full-length **targets** (`InheritAsFullLength` runs only on Pass-1 winners — overwhelmingly targets); their decoy counterparts are not re-injected. The truncation subset is target/decoy-balanced, but the full-length block is target-dominated and sits atop the score ranking. Since q = cumulativeDecoy/cumulativeTarget over the pooled ranked list, the inflated target count depresses q for lower-scoring truncation PSMs.

**Scenario:** 500 confident intact IDs (~0 decoys) + 100 truncation targets + 100 truncation decoys → a truncation at the score where 20 truncation decoys have accumulated shows q ≈ 20/(500+X) instead of ≈ 20/X. Decision #18 documents the pooling intent but not this asymmetry — worth validating against a native combined search or justifying explicitly, since it affects the statistical validity of the headline output.

### 🟠 Medium (confirmed) — fragment index sized to the heaviest parent, allocated unconditionally per file
`TruncationSearchEngine.BuildFragmentIndex:1076`: `indexLength = ceil(maxParentMass * 1000) + 2`, and it allocates **two** `List<int>[]` of that length. At default `MaxParentMass = 30000` that's ~480 MB of mostly-null arrays per file, even for a handful of proteoforms; the docs advise *raising* `MaxParentMass` for database-seeded runs (100 kDa → ~1.6 GB). ModernSearchEngine uses this pattern but as an opt-in indexed mode; here it's unconditional. The build already uses thread-local `Dictionary<int,List<int>>` partials — a sparse bin map for lookup would avoid the dense allocation.

### Low
- **Perf:** `ProteoformChopper.ChopUntilMassMatches:85` builds a fresh PWSM (mod re-index + mass compute) for every chop depth from 1 — a 30 kDa parent winning a 2 kDa scan builds ~250 PWSMs before early-exit. Bounded, winners-only, but compounds #2 on raised-mass runs.
- **Plausible:** the early-exit (`cs:99-105`) assumes chopping is monotonically mass-decreasing; a large negative-mass mod on an interior residue could break monotonicity and skip a valid deeper chop. Realistic top-down mods never trigger it.
- **Confirmed:** candidate-prune threshold `floor(ScoreCutoff/2)+1` (`:846`) is over-aggressive only for `ScoreCutoff ≤ 2` (the justifying comment misreads `CalculatePeptideScore` — the fractional term is global, not per-ion). Safe minimum is `max(1, ScoreCutoff − 1)`. Default cutoff 5 is fine.
- **UX/scope (confirmed):** loading a `TruncationSearchTask.toml` in the GUI silently no-ops — `MainWindow.xaml.cs:1895` has no `"Truncation"` case and no default (also `:2046/:2079/:2189`). Consistent with the stated CMD-only scope; no crash and enum/TOML back-compat is preserved.
- **Nit:** duplicate intact inheritance across precursor-expanded scans (`:1817`) — harmless, collapsed by the `(FullFilePath, ScanNumber, mass)` dedup.

### Verified NOT bugs
Mod re-indexing on chop (N/residue/C keys 1 / r+1 / len+2, tested), `GetReverseDecoyFromTarget(new int[len])` usage, per-proteoform decoy generation with existing-decoy seeding, unconditional peptide/proteoform-level q-values, enum/TOML back-compat, dual-series winner→opposite-terminus direction, acceptor bound `0 < obs ≤ theo + tol`.

trishorts:
Thanks @Alexander-Sol — I checked all six against the code and against the two canonical runs the manuscript reports. **#1 does not reproduce**; #2, #5 and #6 are real, and I've said below exactly what I'd change for each.

### 🟠 #1 — pooled FDR: the inherited full-length block *does* carry its decoys

The premise that `InheritAsFullLength` sees "overwhelmingly targets" doesn't hold. The per-scan Pass-1 lookup is built at `TruncationSearchTask.cs:88-90` as

```csharp
pass1ByFile?[rawFilePath].GroupBy(p => p.ScanNumber)
    .ToDictionary(g => g.Key, g => g.OrderByDescending(p => p).First());
```

— the best Pass-1 PSM for that scan, **target or decoy**, with no target/decoy and no q-value filter on this path. (`ParentQValueThreshold` filters *parent seeding* in `BuildParentsFromPsms`, not inheritance.) `InheritAsFullLength` rebuilds from `source.Protein`, so a decoy that wins its scan in Pass 1 is re-injected as a decoy full-length PSM.

Measured in the two canonical runs (`AllTruncatedPSMs.psmtsv`, split on the `full-length` Description prefix):

| run | full-length T | full-length D | %D | truncation T | truncation D |
|---|---|---|---|---|---|
| E01 (SPE) | 5563 | 652 | 10.5% | 1033 | 151 |
| E02 (1DLC) | 4348 | 602 | 12.2% | 1326 | 96 |

And the bias runs the *opposite* way from the scenario. Recomputing q over the truncation subset alone (same scores, same competition, monotonised) versus the q the pooled run actually assigns:

| run | truncation targets at q<0.01, pooled | truncation-only |
|---|---|---|
| E01 | 630 | 680 |
| E02 | 881 | 930 |

Pooling is **conservative** here — it costs ~7% of truncation IDs, because the inherited decoys rank ahead of most truncation PSMs rather than being absent. Worth documenting either way: I'll extend decision #18 in `docs/Truncation-Search.md` to state what is inherited, that decoys come with it, and this pooled-vs-subset delta.

### 🟠 #2 — dense fragment index: confirmed, with one correction to the sizing

`indexLength = ceil(maxParentMass * 1000) + 2` where `maxParentMass` is `_parents[^1].MonoisotopicMass` (`TruncationSearchEngine.cs:377`) — the heaviest parent **actually included**, not the `MaxParentMass` parameter; parents above the parameter are excluded before the build. So a handful of small proteoforms allocates a small index, not 480 MB.

Your headline number stands as soon as one real ~30 kDa proteoform is in the parent list, which is the ordinary top-down case: two `List<int>[]` × 30M references ≈ 480 MB, ~1.6 GB for a 100 kDa database-seeded run. For scale, peak RSS in the E07 benchmark runs is 10.4 GB (SPE) / 9.5 GB (1DLC), so this is a real but not dominant slice.

The build already produces sparse `Dictionary<int,List<int>>` partials, so the fix is to keep the merge sparse and turn `Tally` into a dictionary probe. My hesitation is the hot loop: `Tally` runs once per tolerance bin per experimental fragment per scan, twice (N and C), so array index → dictionary probe is paid millions of times per file. I'd want an E07 before/after on wall time rather than land it blind.

### #5 — prune threshold: the comment misreads the score, but the threshold is *under*-aggressive, not over

Agreed the justification is wrong. `CalculatePeptideScore` is `Σ(1 + intensity/TotalIonCurrent)` = N + f with **f < 1 globally**, not "<2 per matched ion". With the correct reading, the smallest ion count that can reach cutoff C is essentially `ceil(C)`, so the current `floor(C/2)+1` = 3 at the default cutoff 5 *keeps* parents that can never clear 5 — it under-prunes rather than dropping real hits. Your `ScoreCutoff ≤ 2` edge is the one place it bites, and only for a degenerate spectrum (a 1-ion match at C=2 needs f=1, i.e. one peak carrying the entire TIC).

Fix I'd make: `Math.Max(1, (int)Math.Ceiling(ScoreCutoff) - 1)` plus a rewritten comment. Safe at every cutoff, and strictly more pruning than today at the default.

### #3 / #4 — chopper: accepted as documentation

The early exit does assume chopping is monotonically mass-decreasing. That breaks only if an interior residue carries a mod more negative than the residue's own mass (≥ 57.02 Da), which no realistic top-down modification is. I'd rather state the assumption at the `break` than change behaviour. Same reasoning for the depth-1 rebuild: it's winners-only and bounded, and a binary search over chop depth would rest on exactly the same monotonicity assumption, so it trades a bounded cost for a sharper edge case.

### #6 — GUI no-op: confirmed and intentional

`MainWindow.xaml.cs:1895` has cases for Search/Calibrate/Gptmd/XLSearch/GlycoSearch/Average and no `default:`, so a `TruncationSearchTask.toml` dropped on the GUI silently does nothing. CMD-only is the stated scope of this PR. The cheapest honest fix is a `default:` that raises the existing "Cannot read task toml" style notification, which also covers any future task type — happy to add it here if you'd rather not leave the silent path.

### #7 — nit

Agreed; the `(FullFilePath, ScanNumber, mass)` dedup in `TruncationFdr.RunPooledFdr` collapses the duplicates.

---

One process note: `dismiss_stale_reviews` is on for this repo, so any commit I push here drops the approvals already on the head (this branch has lost two that way). Unless you want #2/#5/#6 in this PR, I'd prefer to land them as a follow-up on the merged core, and keep the change here to the decision-#18 documentation. Tell me which you'd rather have.

Also, if the nits are fine as follow-ups: master requires three approving reviews, and this one is currently a comment rather than a submitted approval.

Issue #1097, opened 2018-04-13
[PeptideTermMod:Lysine not cleaved at N-term]


zrolfs:
This seems like a trypsin-specific modification. Should we uncheck it when searching other enzymes? It's popping up in nonspecific searches...

rmillikin:
Maybe we should have protease-specific mods?

zrolfs:
Why is this even a thing? Does anybody know?

trishorts:
GG is  a protease specific mod that occurs upon cleaveage of ubiquitin side chains. There are others, and these tags change with other proteases.

zrolfs:
So lysine not being cleaved is indicative of ubiquitin? I am confused. Isn't this just a missed cleavage that happens to be on the N-terminus?

trishorts:
No lysine still gets cleaved but there is an addition two amino acid tag hanging off the side

trishorts:
![image](https://user-images.githubusercontent.com/16841846/41316067-eb03324e-6e56-11e8-89c4-22c9b5382cc6.png)

zrolfs:
I'm getting it for nonspecific searches where 
"[PeptideTermMod:Lysine not cleaved at N-term]LWGKQKNK"
has the same mass as
"KLWGKQKNK"
so there doesn't appear to be a GG?

trishorts:
I don't know what that is. I was talking about protease specific mods. I don't think there should be any for non-specific. 
The notation you write looks like a uniprot mod. check uniprot ptmlist

rmillikin:
I removed this mod in #1512 along with a few other ones.

These seem weird to treat as mods for the above reasons (ambiguity, handling it in 2 places [mods and missed cleavages]) but also because we didn't even check to see if the previous amino acid is a K, to my knowledge...

We could write code for interpreting mass-differences as truncations, unsearched missed cleavages, etc. in an open-mass search. not sure how we should approach it in a notch search, if at all

rmillikin:
we have protease-specific mods now ("AspN Digested", "Trypsin Digested"). "Trypsin Digested" includes GG on K. I think this issue has already been resolved but maybe I'm misinterpreting...

Issue #1086, opened 2018-04-09
Modifications not making it into ProteinGroups


zrolfs:
The below peptide was not included:
TSAK[UniProt:N6-acetyllysine]EEDAFHFVSYVPVNGR|TSAK[Mod:Acetyl on K]EEDAFHFVSYVPVNGR

Unsure if ambiguity issue or other. Unmodified version made it in.

zrolfs:
Is there another issue open for Uniprot Acetyl vs Mod acetyl? We could tag it as that.

rmillikin:
#819

PR #2671, opened 2026-06-16
Select proteoform candidates by the most abundant isotopic peak

trishorts:
## Summary
Adds an opt-in `PrecursorMassMatchMode` (default `Monoisotopic`) that, when set to `MostAbundant`, selects top-down candidates by matching the observed **most abundant** isotopic peak against each candidate's theoretical apex (exact monoisotopic + averagine offset), instead of the inferred monoisotopic mass. This directly removes the monoisotopic off-by-N errors that cap high-mass proteoform identification.

## Why a notch set
A single tight-ppm point at the averagine-predicted apex under-identifies, because the predicted apex misses the experimental apex by ±1–2 neutrons (~67 ppm/neutron at 15 kDa). `MostAbundantMassDiffAcceptor` therefore emits a small notch set at `apex + k·(13C−12C)`, k ∈ {−2..+2}, each at tight ppm, keeping FDR controlled per notch.

## Changes
- `PrecursorMassMatchMode` enum; `CommonParameters.PrecursorMassMatchMode` (+ toml/clone/file-specific plumbing). Default `Monoisotopic`, so baseline behaviour is untouched.
- **Scans and PSMs store observations only.** `Ms2ScanWithSpecificMass.PrecursorMostAbundantMass` and `SpectralMatch.ScanPrecursorMostAbundantMass` hold the observed apex (0 when no envelope was deconvoluted); `PrecursorMass` always means the monoisotopic mass. Neither property changes meaning with the search being run, so reading one can never silently return something else. `GetMs2Scans` records the apex in **every** search, because it is an observation rather than a search decision.
- **`PrecursorMassExtensions`** is the single place the search mode is interpreted, and it answers two different questions — using the wrong one is a bug:
  - `GetPrecursorMassForSearch(scan|psm, commonParameters)` — which observed mass candidates are **selected** with. Falls back to the monoisotopic mass when no apex was observed.
  - `GetObservedMonoisotopicMass(psm, theoreticalMono, commonParameters)` — how to read a mass **difference** without mistaking an isotope-assignment offset for chemistry.
  - `MatchesCandidateMass(psm, candidateMass, tolerance, commonParameters)` — per-candidate apex test, for searches carrying discrete mass shifts.
- **`MostAbundantMassDiffAcceptor`** (notch identifies `k`) and **`MostAbundantDotMassDiffAcceptor`** (notch identifies the mass *shift*, preserving G-PTM-D's per-shift FDR grouping). `SearchTask.GetMassDiffAcceptor` returns the former when the mode is set; `GPTMDTask` selects the latter.
- **Bug fixed (pre-existing in the first revision of this PR):** a most-abundant match may sit on a nonzero apex notch, so `(ScanPrecursorMass − theoretical)` can carry a whole-isotopologue offset. Read raw it becomes a phantom ~1–2 Da modification. `LocalizationEngine` was localizing it, `GptmdEngine` was assigning a mod from it, and calibration was reading it as m/z drift. All now route through `GetObservedMonoisotopicMass`, which is the identity in monoisotopic mode. A follow-up audit fixed three more sites: `NonSpecificEnzymeSearchEngine` (two selection sites where the mode was a silent no-op), `BinTreeStructure`, and the PSM-ranking tie-breakers. G-PTM-D is the documented exception — it cannot use the rounding correction, because an unknown modification mass shares the residual with `k`; it uses `MatchesCandidateMass` instead.
- **psmtsv columns** `Precursor Most Abundant Mass` and `Most Abundant Mass Diff (ppm)`, both under the `includeMostAbundantColumn` gate, with headers owned by mzLib (#1086, #1097) so they round-trip through mzLib's reader.
- **Deliberately left monoisotopic** (disclosed rather than silently changed): fragment-bin and complementary-ion math, all reporting/writing, and XL/glyco search — where most-abundant mode is currently a silent no-op.

## Validation (yeast top-down, 1% FDR, pooled across all 24 fractions)
MostAbundant **326 proteoforms / 10,850 PSMs** vs monoisotopic ±3-notch baseline **281 / 9,314** — **+16.0% / +16.5%**, using fewer notches (5 vs 7). Consistent at 8-fraction (+15.1%) and single-fraction scale.

## Notes
- **mzLib dependency is released, not a local build.** smith-chem-wisc/mzLib#1078 is merged (merge commit `4d815866`), and the psmtsv column headers landed in #1086 (`9467ef5b`) and #1097 (`fc2889cc`). This PR references **released mzLib 1.0.583**, which carries all three. (An earlier revision of this description said "draft until #1078 merges and a release is available" — that is now satisfied.)
- `Ms2ScanWithSpecificMass.PrecursorMassToMatch`, described in an earlier revision of this summary, was **removed** during review: it changed meaning depending on the search being run, which is exactly the ambiguity the observation/interpretation split above was introduced to eliminate.
- Centroid-mass path for isotopically unresolved (>35 kDa) species is plumbed but dormant pending an unresolved charge-determination algorithm (future work); nothing in production sets that state, so the resolved path is exactly what the validation above measures.

---

## Review follow-up: does Classic search really select on the apex? (`baf74ba`)

**Question raised in review:** `ClassicSearchEngine.cs:149` passes `specificBioPolymer.MonoisotopicMass` into `GetAcceptableScans`, which reads as though most-abundant mode is still selecting candidates on the monoisotopic mass.

**It is not.** That value is the *input to* the theory-side apex conversion, not a bypass of it. Both ends of the comparison are apexes:

| step | file:line | what happens |
| --- | --- | --- |
| 1 | `ClassicSearchEngine.cs:149` | `GetAcceptableScans(mono, SearchMode)` — engine hands over the **monoisotopic** mass |
| 2 | `ClassicSearchEngine.cs:253` | forwards it to `SearchMode.GetAllowedPrecursorMassIntervalsFromTheoreticalMass(mono)` |
| 3 | `MostAbundantMassDiffAcceptor.cs:120-128` | **converts here:** windows are centred on `mono + averagineApexOffset(mono)` — the **apex** — at `apex + k·spacing` |
| 4 | `ClassicSearchEngine.cs:258, :268, :276` | compares those windows against `MyScanPrecursorMasses` |
| 5 | `ClassicSearchEngine.cs:46` | ...which was built from `GetPrecursorMassForSearch(...)` — the **observed apex** |

Passing an apex at `:149` would in fact be the bug: the acceptor would then add the offset a second time and every window would sit ~5 Da high at 9.5 kDa. The engine is deliberately kept ignorant of apexes and the acceptor owns the conversion, which is why a single call site serves both search modes unchanged.

`ClassicSearchEngine` contains **no other mass comparison** — every occurrence of `PrecursorMass` / `MonoisotopicMass` / `MyScanPrecursorMasses` / `SearchMode` in that file is on lines 21, 29, 46, 55, 149, 251, 253, 258, 268, 276. There is no second path where the monoisotopic mass could re-enter.

### The reason this couldn't be settled by reading: it wasn't tested

Fair catch. Every test in `MostAbundantMassTest` exercised the acceptor, the extension methods, or `GetMs2Scans` **in isolation** — nothing drove `ClassicSearchEngine` and asserted what it actually selected. That coverage gap is now closed.

**`ClassicSearch_SelectsScansByApex_NotMonoisotopic`** runs the real engine over a ~9.6 kDa peptide with one scan carrying a deliberately **wrong monoisotopic mass** (+3.5 Da — far outside tolerance, and not a whole number of neutrons so it cannot be rescued by a notch) and a **correct apex**:

- monoisotopic search → `null`; it cannot reach the scan
- most-abundant search → **finds the peptide, notch 0, correct base sequence**
- **control** — same setup with a *correct* monoisotopic mass → the monoisotopic search finds it

The control is what makes this conclusive: it rules out protease, score cutoff, or fragment matching as the cause of the `null`. The only variable that changed the outcome was which mass was correct. Theoretical fragments are supplied as neutral-mass envelopes so fragment matching is guaranteed and the test isolates precursor selection.

Fixture: **22/22 passing.**

### Corroborating argument

If selection were silently monoisotopic while the scan array is sorted and binary-searched by apex (`:46`, `:276`), then at 9.5 kDa every window would sit ~5 Da from every scan mass and the search would return **essentially zero IDs** — not the +16% reported above. A monoisotopic fallback here is not a subtle bug; it is a total-failure mode.

### Naming note (pre-existing, not introduced here)

The underlying confusion is a real asymmetry in the base signature `MassDiffAcceptor.Accepts(scanPrecursorMass, peptideMass)` (`MassDiffAcceptor.cs:23`): the **second** parameter is always monoisotopic, but the **first** silently becomes the apex in most-abundant mode. Happy to rename it to something like `observedMassToMatch` in a follow-up if that would help readers.

### Same proof for G-PTM-D, and negative controls (`511fd47a`)

The G-PTM-D side had the identical coverage gap: `GptmdRunsInMostAbundantMode` is a smoke test (`DoesNotThrow` + files written), and the dot acceptor and `MatchesCandidateMass` were only tested in isolation. Nothing drove `GptmdEngine` and asserted that mod *discovery* goes through the apex.

**`Gptmd_DiscoversModByApex_NotMonoisotopic`** — ~9.6 kDa peptide, four N sites, a 21.981943 Da mod on N (far from any whole number of neutrons, so an accepted match cannot be an isotope artifact in disguise). The scan carries a wrong monoisotopic mass (+3.5 Da) and a correct apex *for the modified candidate*:

- monoisotopic G-PTM-D → **0 mods**
- most-abundant G-PTM-D → **mod discovered at all 4 N sites, correct mod mass**
- control (correct monoisotopic mass) → monoisotopic G-PTM-D finds the same mod at the same 4 sites

Fragment scoring is deliberately uninformative (single-peak spectrum, no filters) so every legal site scores identically and the test isolates precursor-driven mod selection — the only thing most-abundant mode changes here.

**Negative controls** were added to both tests. The positive assertions show most-abundant mode accepts what it should; on their own they do not show it can *reject*, which matters because the mode widens acceptance to a ±2-neutron notch set. Each negative control reuses the same helper and engine and changes only the observed apex, so it cannot pass vacuously:

- **ClassicSearch** — apex moved 7.5 Da (beyond the ±2.007 Da notch set, and not a whole number of neutrons, so no `k` rescues it) → no PSM
- **G-PTM-D** — same 7.5 Da apex displacement → no mod discovered
- **G-PTM-D** — apex correct but the only candidate mod's mass wrong by 7.5 Da → no mod discovered, confirming the ±k windows do not swallow the mod mass

`MostAbundantMassTest` + `GptmdEngineTest`: **50/50 passing.**

### Warn on the silent monoisotopic fallback (`c38957c0`)

`GetPrecursorMassForSearch` falls back to the monoisotopic mass for any scan whose `PrecursorMostAbundantMass` is 0 — any precursor with no deconvoluted envelope. The fallback is correct and deliberate (a most-abundant search cannot match on a peak that was never observed), but it was **silent**: a run in which most scans take it degrades to a monoisotopic search while still reporting itself as most-abundant, and the only symptom is a disappointing identification count.

`PrecursorMassExtensions.GetMonoisotopicFallbackWarning(scans, commonParameters, dataFileName)` now reports how many scans of how many will be matched monoisotopically, returning `null` when there is nothing to say — monoisotopic mode, an empty scan set, or a set where every precursor has an observed apex. It is a pure function over the scan array rather than a warning raised inside `GetPrecursorMassForSearch`, which is called per scan in hot loops (including `ClassicSearchEngine`'s constructor) and would emit one message per scan.

Wired into the three places that build the sorted scan array and therefore know the whole set — `SearchTask`, `GPTMDTask`, `CalibrationTask` — each reporting through the existing `MetaMorpheusTask.Warn`.

Two tests: `MonoisotopicFallbackWarning_ReportsOnlyWhenScansLackAnApex` (correct count/percentage/file name; silent in monoisotopic mode; silent when every scan has an apex; null/empty input safe) and `SearchTask_EmitsMonoisotopicFallbackWarning_WhenNoApexObserved`, which subscribes to `WarnHandler` and runs a real search with precursor deconvolution off so every scan falls back — proving the warning is *emitted*, not merely computable.

**Full suite: 1573 passed, 0 failed, 1 skipped.**


Alexander-Sol (dismissed):

MetaMorpheus/EngineLayer/PrecursorSearchModes/MostAbundantMassDiffAcceptor.cs
```diff
@@ -0,0 +1,126 @@
...
+        {
+            // Indexed (ModernSearch) path: convert the observed most-abundant mass back to candidate
+            // monoisotopic windows by subtracting the offset evaluated near the observed mass, for each
+            // apex-offset notch. Top-down uses the exact theory-driven ClassicSearch path; this
+            // observed-side conversion carries a small near-boundary ambiguity and is only exercised by
+            // indexed bottom-up search, which this mode does not target.
+            double monoApprox = observedMostAbundantMass - ApexOffset(observedMostAbundantMass);
+            foreach (int k in ApexOffsetsInNeutrons)
+            {
+                double mass = monoApprox - k * Constants.C13MinusC12;
```
nbollis: hardcoded constant will fail for RNA and is present in several locations. Consider passing in the deconvolution parameters expected isotopic spacing.

MetaMorpheus/EngineLayer/PrecursorSearchModes/MostAbundantDotMassDiffAcceptor.cs
nbollis, line 1: hardcoded constant will fail for RNA and is present in several locations. Consider passing in the deconvolution parameters expected isotopic spacing.

MetaMorpheus/TaskLayer/GPTMDTask/GPTMDParameters.cs
nbollis, line 1: Why is the GPTMD Filters change necessary? I am fairly sure this change will invalidate old tomls due to the property name change

MetaMorpheus/EngineLayer/Ms2ScanWithSpecificMass.cs
```diff
@@ -11,11 +11,17 @@ public class Ms2ScanWithSpecificMass
...
             double precursorDeconvolutionScore = 0)
         {
             PrecursorMonoisotopicPeakMz = precursorMonoisotopicPeakMz;
             PrecursorCharge = precursorCharge;
             PrecursorMass = PrecursorMonoisotopicPeakMz.ToMass(precursorCharge);
+            // Mass used for candidate selection by the MassDiffAcceptor. Defaults to the monoisotopic
+            // PrecursorMass; in most-abundant mode the caller passes the envelope's most-abundant (or,
+            // for unresolved species, average) observed mass. PrecursorMass itself is left unchanged
+            // because it still drives fragment-bin math and precursor mass-error reporting.
+            PrecursorMassToMatch = precursorMassToMatch ?? PrecursorMass;
```
nbollis: Why have a second property that defaults to the first? The comments say that monoisotopic mass is used for mass error calculations and fragment-bin math. I am not sure how precursor mass plays into fragment-bin matching, but wouldn't the mass error calculation being performed upon the same mass in which we are using to search make the most sense? I am pliable on this.

MetaMorpheus/EngineLayer/PrecursorSearchModes/MostAbundantMassDiffAcceptor.cs
trishorts: Done — replaced the hardcoded `Constants.C13MinusC12` with an `expectedIsotopeSpacing` constructor argument, threaded from the existing `DeconvolutionParameters.ExpectedIsotopeSpacing` at every construction site (Calibration / GPTMD / Search). It defaults to `C13MinusC12` and is already varied for decoy runs (~0.9444 Da) and non-carbon-dominated polymers, so this removes the carbon-only assumption without new plumbing. (commit 659fd683e)

MetaMorpheus/EngineLayer/PrecursorSearchModes/MostAbundantDotMassDiffAcceptor.cs
trishorts: Done — same fix applied to the Dot acceptor: `expectedIsotopeSpacing` sourced from `PrecursorDeconvolutionParameters.ExpectedIsotopeSpacing`; no more hardcoded neutron spacing. (commit 659fd683e)

MetaMorpheus/TaskLayer/GPTMDTask/GPTMDParameters.cs
trishorts: Agreed this is out of scope for the most-abundant work — the GPTMD filter-serialization feature (`GptmdFilterTypes` / `GetActiveFilters` / `CreateFilter` / `[TomlIgnore]`) is unrelated, so I reverted it out of this PR (commit 659fd683e) and will resubmit it as its own PR.

One clarification: there is no property *rename* — `GptmdFilters` keeps its name. The only toml-affecting change was `[TomlIgnore]` on a `List<IGptmdFilter>` (interface instances) that never round-tripped through Nett anyway, so old tomls are unaffected.

MetaMorpheus/EngineLayer/Ms2ScanWithSpecificMass.cs
trishorts: Good point. I kept the two-property design — `PrecursorMass` stays monoisotopic so fragment-bin math and the existing `Mass Diff` columns remain monoisotopic-vs-monoisotopic (comparable across search modes) — but I acted on the underlying instinct and added an optional **`Most Abundant Mass Diff (ppm)`** column to the psmtsv (commit cca25b9b1) that reports the observed most-abundant peak vs the candidate's theoretical averagine-apex mass, i.e. the error on the mass we actually searched.

It's gated by the same data-driven optional-column idiom as the 1/K0 and collision-energy columns (only present when a run used most-abundant mode, so normal output is unchanged), and it's write-only / MM-only — mzLib's psmtsv reader maps known headers by name and ignores unknown columns, so no mzLib change is needed. Selection and reporting share the same `AveragineApexOffset` helper.

Read together, the two columns separate the error sources: monoisotopic `Mass Diff` off but most-abundant tight ⇒ deconvolution mis-assigned the monoisotopic peak (rescued by most-abundant); most-abundant off by ~N ⇒ an apex misprediction the notch tolerated.

MetaMorpheus/EngineLayer/SpectrumMatch/PsmTsvWriter.cs
```diff
@@ -19,6 +19,14 @@ public static class PsmTsvWriter
         internal const double ToleranceForDoubleResolutionF5 = 1e-6;
         internal const double ToleranceForDoubleResolutionF2 = 1e-3;
 
+        /// <summary>
+        /// Optional psmtsv column header for the most-abundant-mode precursor mass error. Written only
+        /// when a run used most-abundant selection (see <see cref="SpectralMatch.MostAbundantMassErrorPpm"/>).
+        /// A write-only column: mzLib's psmtsv reader maps known headers by name and ignores unknown
+        /// columns, so this is backward-compatible without a mzLib change.
+        /// </summary>
+        internal const string MostAbundantMassDiffPpmHeader = "Most Abundant Mass Diff (ppm)";
```
nbollis: Careful. We keep all of these headers in Mzlib SpectrumMatchTsvHeader to ensure that they are usable for reading and writing. 

I am fine with it for this pull request IF there is an accompanying mzlib PR that enables the reading.

Alexander-Sol (approved):

Alexander-Sol (dismissed):

RayMSMS (dismissed):

MetaMorpheus/EngineLayer/Calibration/DataPointAquisitionResults.cs
```diff
@@ -35,7 +36,25 @@ public DataPointAquisitionResults(
...
+            // the raw difference is used unchanged - baseline calibration behavior is preserved exactly.
+            bool stripNeutronOffsets = precursorMassMatchMode == PrecursorMassMatchMode.MostAbundant;
+            var precursorErrors = psms.Select(p =>
+            {
+                double deltaDa = p.ScanPrecursorMass - p.BioPolymerWithSetModsMonoisotopicMass.Value;
+                if (stripNeutronOffsets)
+                {
+                    // AwayFromZero so the nearest-neutron assignment at an exact half-neutron boundary is
+                    // parity-independent (default banker's rounding would snap to the even integer).
+                    deltaDa -= Math.Round(deltaDa / Constants.C13MinusC12, MidpointRounding.AwayFromZero) * Constants.C13MinusC12;
```
nbollis: C13-C12 will likely cause issues. I believe the inverse of this calculation was performed in mzlib using the expected isotope spacing

MetaMorpheus/EngineLayer/Ms2ScanWithSpecificMass.cs
```diff
@@ -43,6 +49,21 @@ public Ms2ScanWithSpecificMass(MsDataScan mzLibScan, double precursorMonoisotopi
...
+        /// <para>
+        /// This is deliberately a separate property rather than an overwrite of <see cref="PrecursorMass"/>:
+        /// candidate selection is the only step that should key off the most-abundant peak. Fragment-bin
+        /// math and the reported precursor mass error must stay on the monoisotopic mass so that (a) the
+        /// error is the monoisotopic-vs-monoisotopic value comparable across search modes, and (b) it is
+        /// not conflated with the ±k-neutron apex offset the acceptor already models per-notch. Collapsing
+        /// the two would report error against a peak that is intentionally N neutrons off the monoisotopic.
+        /// </para>
+        /// </summary>
+        public double PrecursorMassToMatch { get; }
```
nbollis: This is a major change and is extremely concerning. Adding this property and changing where we are accessing the precursor mass has potential for huge ramifications unless every call site is updated diligently. 

In addition, this adds a confusing set of properties to the class. It would be very easy for somebody to make a mistake by assuming Ms2scanwithspecificmass.PrecursorMass is the precursor mass that we care about. This would only be correct in the monoisotopic search, but not the most abundant.

MetaMorpheus/EngineLayer/SpectralMatch.cs
```diff
@@ -97,6 +100,16 @@ protected SpectralMatch(IBioPolymerWithSetMods peptide, int notch, double score,
         /// </summary>
         public double PrecursorScanDeconvolutionScore { get; }
         public double ScanPrecursorMass { get; }
+        /// <summary>
+        /// The observed precursor mass used for candidate selection: equals <see cref="ScanPrecursorMass"/>
+        /// (monoisotopic) in the default mode, or the envelope's most-abundant (or average) observed mass
+        /// in most-abundant mode. Used to report <see cref="MostAbundantMassErrorPpm"/>.
+        /// </summary>
+        public double ScanPrecursorMassToMatch { get; }
```
nbollis: This is a major change and is extremely concerning. Adding this property and changing where we are accessing the precursor mass has potential for huge ramifications unless every call site is updated diligently. 

In addition, this adds a confusing set of properties to the class. It would be very easy for somebody to make a mistake by assuming SpectralMatch.PrecursorMass is the precursor mass that we care about. This would only be correct in the monoisotopic search, but not the most abundant.

nbollis (commented):
My beef with this implementation is that the Ms2ScanWithSpecificMass and SpectralMatch representations depend upon the state of the search engine when they should be stateless for the search type being performed. The addition of these new properties introduces many potential bugs and sources of error. 

For example, the localization engine calculates the mass difference based upon psm.ScanPrecursorMass and peptide.MonoisotpicMass. Now the localization engine will be wrong if we do a most abundant search because it will be comparing the scan most abundant and the peptide monoisotopic. 

I am not sure what the solution is, but this has potential to cause a lot of bugs and confusion. Maybe something like having mono and avg mass as separate fields then the code that needs it will pick the correct one? Maybe a helper method that is GetPrecursorMass(bool isMostAbundant) in both ms2scanwithspecificmass and SpectralMatch. Maybe it's an extension method instead?

MetaMorpheus/EngineLayer/Ms2ScanWithSpecificMass.cs
trishorts: Agreed, and fixed — `PrecursorMassToMatch` is gone (commit 4102843).

You're right that a property whose meaning depends on the search type is a bug factory. `Ms2ScanWithSpecificMass` now carries only **observations**, populated identically in every search:

- `PrecursorMass` — the deconvoluted monoisotopic mass. **Always.** It never means anything else, in any mode, so the failure you describe ("somebody assumes `PrecursorMass` is the precursor mass we care about... only correct in the monoisotopic search") can't happen anymore.
- `PrecursorMostAbundantMass` — the observed neutral mass of the envelope's tallest isotopologue, or 0 if no envelope was deconvoluted. Also always recorded: it's what the detector saw, not a search decision.

`GetMs2Scans` no longer branches on the search mode at all.

The knowledge of *which* observation a search matches on moved to the object that actually defines the search type — the `MassDiffAcceptor` (`MatchesMostAbundantPeak`) — and is reached through an extension method, which is your third suggestion:

```csharp
scan.GetPrecursorMassForSearch(massDiffAcceptor)
psm.GetPrecursorMassForSearch(massDiffAcceptor)
```

Classic/Modern search and all three task-level scan sorts go through it, so the sort key and the search key can't drift apart. And because a most-abundant acceptor falls back to the monoisotopic mass when no apex was observed, the scan-header/pre-deconvoluted paths behave sanely without a special case.

MetaMorpheus/EngineLayer/SpectralMatch.cs
trishorts: Same fix — `ScanPrecursorMassToMatch` is gone (commit 4102843).

`SpectralMatch` mirrors the scan: `ScanPrecursorMass` is always the monoisotopic mass, and `ScanPrecursorMostAbundantMass` is the observed apex (0 if no envelope). Neither is redefined by the search type, and the acceptor decides which one a given search matched on via `psm.GetPrecursorMassForSearch(massDiffAcceptor)`.

`IsMostAbundantMode` stays, but only to gate the optional `Most Abundant Mass Diff (ppm)` psmtsv column — it's search configuration recorded on the PSM, like `DigestionParams` and `RunnerUpScore` already are, and it can't be mistaken for a mass.

MetaMorpheus/EngineLayer/Calibration/DataPointAquisitionResults.cs
trishorts: Good catch — removed (commit 0e4ba37).

`Constants.C13MinusC12` is gone from this path entirely. The precursor error is now computed as

```csharp
double observedMass = p.GetObservedMonoisotopicMass(theoreticalMass, massDiffAcceptor);
```

and `MassDiffAcceptor.ToObservedMonoisotopicMass()` removes the isotope offset **using the acceptor's own `ExpectedIsotopeSpacing`**, which came from `DeconvolutionParameters` at construction — so it's correct for RNA, for decoy runs (~0.9444 Da), and for any non-carbon-dominated polymer, exactly as with the acceptors you flagged last round. The rounding hack is gone with it; the acceptor re-derives the offset it actually allowed.

`DataPointAcquisitionEngine` takes the acceptor the calibration PSMs were found with, and `CalibrationTask` passes the search mode it already had in scope.

trishorts:
This was the right call, and it was worse than you thought. Restructured across three commits (`4102843`, `0e4ba37`, `a9b0449`) — full suite green, 1568/1568.

## 1. The representations are now stateless w.r.t. search type

`PrecursorMassToMatch` / `ScanPrecursorMassToMatch` are **deleted**. `Ms2ScanWithSpecificMass` and `SpectralMatch` carry only observations, populated identically in every search:

| property | meaning |
|---|---|
| `PrecursorMass` | the deconvoluted monoisotopic mass — **always**, in every mode |
| `PrecursorMostAbundantMass` | the observed apex of the envelope (0 if none was deconvoluted) |

`PrecursorMass` never changes meaning, so it can't be silently wrong. The search-type knowledge lives on the `MassDiffAcceptor` — the object that *is* the search type — and is reached by extension method, your suggestion #3:

```csharp
scan.GetPrecursorMassForSearch(massDiffAcceptor)
psm.GetPrecursorMassForSearch(massDiffAcceptor)
```

`GetMs2Scans` no longer branches on the mode at all. Classic/Modern search and all three scan sorts go through the extension, so the sort key and the search key can't diverge.

## 2. Your localization example — inverted mechanism, real bug

Strictly, the mechanism you describe doesn't occur: `ScanPrecursorMass` stays monoisotopic in most-abundant mode, so `LocalizationEngine` **is** comparing scan-monoisotopic to peptide-monoisotopic.

But the conclusion is correct, for a sharper reason. Most-abundant matching deliberately **accepts candidates whose deconvoluted monoisotopic peak is off by ±1–2 neutrons** — that's the entire point of the apex notch set, and those PSMs are exactly the ones this feature exists to rescue. For them, `ScanPrecursorMass - peptide.MonoisotopicMass ≈ k · 1.00335 Da`. That is an isotope-assignment artifact, **not chemistry** — and three places were reading it as chemistry:

- **`LocalizationEngine:58`** — localized a phantom ~1–2 Da modification onto a residue.
- **`GptmdEngine:104`** — fed the same bogus difference into mod-candidate matching. **This one is a bug we shipped in this PR, and you found it.** Worse, `MostAbundantDotMassDiffAcceptor` carried a remark asserting the opposite — *"downstream G-PTM-D mod assignment recomputes the modification from the monoisotopic precursor mass, so it is unaffected by this acceptor."* That claim was the bug. Corrected in code and comment.
- **`DataPointAquisitionResults`** — counted it as m/z drift (the 1940 ppm runaway tolerance). This was the only one we'd found, and we'd patched it in place with the hardcoded `C13MinusC12` you flagged above — i.e. we hit this bug class once, fixed the symptom, and missed that it was systemic. Your review is what surfaced that.

All three now go through one accessor:

```csharp
public virtual double MassDiffAcceptor.ToObservedMonoisotopicMass(double observedPrecursorMass, double theoreticalMonoisotopicMass)
```

which undoes exactly the offsets the acceptor itself allowed (averagine apex offset + the k-neutron apex misprediction), using the isotope spacing it took from `DeconvolutionParameters`. **It is the identity on every monoisotopic acceptor**, so default-mode behaviour is unchanged bit-for-bit — the calibration/localization/G-PTM-D diffs are all "route through the acceptor," not "special-case most-abundant."

One design note: I did *not* use `GetPrecursorMass(bool isMostAbundant)`. A bool at the call site is the same coupling in a thinner disguise — the caller still has to know which search it's in. Asking the acceptor means there's exactly one source of truth. It also turned out to be load-bearing for G-PTM-D: the Dot acceptor's notch encodes only the mass *shift* (all k share a notch, to preserve per-shift FDR grouping), so k can't be read back from the notch — the acceptor has to re-derive the (shift, k) pair itself, and only the acceptor knows its shift set.

## 3. Tests

New coverage for the observation/selection split, for `ToObservedMonoisotopicMass` across k = −2..+2 (asserting the mass difference collapses to ~0 rather than k·1.00335, and that a monoisotopic acceptor is the identity), and for the scan no longer changing shape with the mode.

The old calibration test was also replaced: it never exercised the path it claimed to — it built PSMs with no observed apex, so the "neutron stripping" it asserted was running on a fallback. It now builds a real apex observation plus a k-neutron misprediction, with a control asserting the same PSMs show a >50 ppm IQR when read *without* the acceptor, so the test proves the correction (not the fixture) is what tightens the spread.

MetaMorpheus/EngineLayer/PrecursorSearchModes/PrecursorMassExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// <see cref="Ms2ScanWithSpecificMass"/> and <see cref="SpectralMatch"/> store only OBSERVATIONS —
+    /// the deconvoluted monoisotopic mass and the most-abundant (tallest) isotopologue mass. Neither
+    /// property changes meaning with the search being run, so reading
+    /// <see cref="Ms2ScanWithSpecificMass.PrecursorMass"/> always gives the monoisotopic mass and can
+    /// never silently return something else. The knowledge of WHICH observation a given search keys off
+    /// belongs to the <see cref="MassDiffAcceptor"/>, which is the object that defines the search type;
+    /// these extensions are the one place that asks it.
+    /// </para>
+    /// </summary>
+    public static class PrecursorMassExtensions
```
nbollis, line 31: It would be great if these could operate upon booleans and/or the new enum you created that is stored in common parameters. I think this would allow you to get away with not adding the mass diff accepter to all of the engines. 

You could then pass the common params enum into the extension method to get the correct mass

nbollis (commented):
Love the direction this is going in. I beleive in most cases where the mass diff acceptor was added to the engine parameters (for the purpose of switching masses) we only need the enum in common params which is passed into all engines. Therefor if we make the extension methods operate upon the enum, we will not need the mass diff acceptor in the engines. 

In addition, you should do a search for all .PrecurorMasses in the codebase for both Ms2scanWithSpecificMass as well as SpectralMatch. There are likely numerous locations to make the switch yet.

MetaMorpheus/EngineLayer/PrecursorSearchModes/PrecursorMassExtensions.cs
trishorts: Done — the extensions now take `CommonParameters` and read `PrecursorMassMatchMode` (commit `fe221b0`).

You were right that this pays for itself: `MassDiffAcceptor` is back to **untouched** (the `MatchesMostAbundantPeak` and `ToObservedMonoisotopicMass` virtuals I added are deleted), and the acceptor parameter is gone from `DataPointAcquisitionEngine`, `LocalizationEngine` and `GptmdEngine` — all three already had `CommonParameters`, so those signatures return to exactly what they were.

Three methods, and which one you want depends on the question you're asking:

```csharp
scan.GetPrecursorMassForSearch(commonParameters)       // which mass do we SELECT candidates with
psm.GetObservedMonoisotopicMass(theoretical, params)   // read a mass DIFFERENCE without the isotope offset
psm.MatchesCandidateMass(candidate, tol, params)       // does the observation support this candidate
```

**One place the enum alone genuinely isn't enough, and I want you to sanity-check my reasoning: G-PTM-D.**

`GetObservedMonoisotopicMass` recovers the k-neutron apex misprediction by rounding the residual to the nearest isotopologue. That is exact for calibration and localization *because* most-abundant mode always searches with the zero-centred apex acceptor — `GetMassDiffAcceptor` overrides the `MassDiffAcceptorType` in that mode — so the residual is k neutrons plus a within-tolerance ppm error and nothing else.

G-PTM-D breaks that invariant: it searches with **discrete mass shifts**, so the residual also contains an *unknown modification mass*. Round it and the modification gets folded into k and destroyed (a phospho at +79.966 would come back as ~-0.06 Da). The mod mass and the isotope offset land in the same residual and cannot be separated after the fact.

So G-PTM-D doesn't correct the precursor once — `GetPossibleMods` tests **each candidate mass** (peptide + modification) against the observation apex-to-apex via `MatchesCandidateMass`. That's the same acceptance rule as `MostAbundantDotMassDiffAcceptor.Accepts`, just expressed over the deconvolution parameters, so it still needs no acceptor. Its combo recursion now accumulates a candidate-mass *offset* rather than subtracting from the target mass — equivalent arithmetic, but it leaves the observed mass untouched so the apex test always sees what was actually measured.

New test pins it: a phosphorylated proteoform observed one neutron off its apex is still matched to phospho, and still **not** matched to oxidation.

`MostAbundantMassDiffAcceptor.DefaultMaxApexOffsetNeutrons` is now shared with the extensions, so the acceptor and the read-back can't drift apart on the k range.

trishorts:
Both asks done — `fe221b0` (production) and `92e13e0` (tests). Full suite green, **1569/1569**.

## 1. Extensions operate on the enum; acceptor out of the engines

`MassDiffAcceptor` is **back to untouched**, and the acceptor parameter is removed from `DataPointAcquisitionEngine`, `LocalizationEngine` and `GptmdEngine`. Details and the one exception (G-PTM-D, which searches with discrete mass shifts, so the isotope offset can't be recovered by rounding) are in the thread reply above — worth a look, since that's the only place I didn't do exactly what you asked.

## 2. The audit — you were right, there were more

Swept every `Ms2ScanWithSpecificMass.PrecursorMass` and `SpectralMatch.ScanPrecursorMass` consumer. **Three more sites were reading the wrong mass:**

| site | what was wrong |
|---|---|
| **`NonSpecificEnzymeSearchEngine:88, :128`** | Selected candidates off the raw monoisotopic mass — so non-specific search **silently ignored most-abundant mode entirely**. |
| **`BinTreeStructure:19, :97`** | Builds the mass-shift histogram from `ScanPrecursorMass - peptideMonoisotopic`. Same defect as localization: it would have grown **spurious ~1 and ~2 Da bins that look like real modifications**. |
| **PSM ranking tie-breakers** (`CalibrationTask`, `NonSpecificEnzymeSearchEngine` ×2) | Rank equal-scoring matches by precursor closeness. An isotope offset is not "closeness" — left in, a *correct* match with a 1-neutron apex misprediction ranks behind an inferior one. |

**Deliberately left on the monoisotopic mass** — flagging these explicitly so you can push back rather than have it be silent:

- **Fragment-bin / complementary-ion math** (`ModernSearchEngine`, `MetaMorpheusEngine.MatchFragmentIons`). The monoisotopic mass is what that math *means*; the apex mass would be wrong there.
- **All reporting and writing** (`PsmTsvWriter`, `PepXMLWriter`, `WriteXlFile`, MetaDraw). The psmtsv `Mass Diff` column stays monoisotopic-vs-monoisotopic so it remains comparable across search modes, with the optional `Most Abundant Mass Diff (ppm)` column reported alongside it. Read together they separate the two error sources.
- **Crosslink / glyco search.** They build their own acceptors and don't expose this mode. Worth noting: if a user sets `PrecursorMassMatchMode = MostAbundant` in a toml and runs an XL or glyco task, it is currently a **silent no-op**. Happy to make that an explicit warning (or wire it up) if you'd rather — say the word and I'll do it in this PR.

## 3. Tests

`MatchesCandidateMass_FindsModification_DespiteApexMisprediction` is the new one worth reading — it's the case a rounding correction cannot handle. `GetObservedMonoisotopicMass_RemovesApexAndNeutronOffsets` (k = −2..+2) now also asserts that read **raw**, the same PSM shows a whole-neutron mass difference, so the phantom modification is pinned explicitly rather than just being absent.

MetaMorpheus/EngineLayer/PrecursorSearchModes/PrecursorMassExtensions.cs
```diff
@@ -1,58 +1,139 @@
...
+        /// apex-to-apex within a ±k isotopologue apex tolerance in most-abundant mode.
+        /// <para>
+        /// This is the form a search carrying discrete mass shifts needs (G-PTM-D): each candidate mass
+        /// (peptide + modification) is tested on its own, so the modification mass can never be confused
+        /// with the k-neutron apex misprediction. It mirrors the acceptance test in
+        /// <see cref="MostAbundantDotMassDiffAcceptor.Accepts"/> — both call the same apex model — but needs
+        /// only the deconvolution parameters, not the acceptor's shift set.
+        /// </para>
+        /// </summary>
+        public static bool MatchesCandidateMass(this SpectralMatch psm, double candidateMonoisotopicMass, Tolerance tolerance, CommonParameters commonParameters)
```
nbollis, line 110: Why is this method needed if we have the special mass diff accepters?

MetaMorpheus/EngineLayer/SpectrumMatch/PsmTsvWriter.cs
```diff
@@ -192,7 +192,7 @@ internal static void AddBasicMatchData(Dictionary<string, string> s, SpectralMat
             s[SpectrumMatchFromTsvHeader.Notch] = psm == null ? " " : Resolve(psm.BestMatchingBioPolymersWithSetMods.Select(p => p.Notch / MassDiffAcceptor.NotchScalar)).ResolvedString;
         }
 
-        internal static void AddPeptideSequenceData(Dictionary<string, string> s, SpectralMatch sm, IReadOnlyDictionary<string, int> ModsToWritePruned)
+        internal static void AddPeptideSequenceData(Dictionary<string, string> s, SpectralMatch sm, IReadOnlyDictionary<string, int> ModsToWritePruned, bool includeMostAbundantColumn = false)
         {
             bool pepWithModsIsNull = sm == null || sm.BestMatchingBioPolymersWithSetMods == null || !sm.BestMatchingBioPolymersWithSetMods.Any();
```
nbollis, line 197: We should likely also write out the MostAbundantMass as a new column in addition to the mass error, but I will not hold up the PR on this

nbollis (commented):
One question and one suggestion

MetaMorpheus/EngineLayer/PrecursorSearchModes/PrecursorMassExtensions.cs
trishorts: Fair question — they *do* compute the same test, so let me justify why the method exists rather than the acceptor.

`MatchesCandidateMass` and `MostAbundantDotMassDiffAcceptor.Accepts` are the same apex-to-apex arithmetic (candidate apex = mono + averagine offset; accept if the observed apex is within tolerance of `apex + k·spacing` for k in ±`DefaultMaxApexOffsetNeutrons`). It is deliberately not a second implementation — both go through the same `AveragineApexOffset` model and share `DefaultMaxApexOffsetNeutrons`.

The reason it's a method on the extensions and not just a call to `.Accepts` is your **own round-2 request**: take the `MassDiffAcceptor` back out of the engines and drive everything off `CommonParameters`. `GptmdEngine.GetPossibleMods` is inside that boundary. So the choice was:

1. thread the Dot acceptor back into `GetPossibleMods` purely so it can call `.Accepts` — which puts back exactly the coupling you asked me to remove, in the one engine where I couldn't use the rounding correction, or
2. express the same acceptance test over `CommonParameters` — which is `MatchesCandidateMass`.

I went with (2) so no engine holds an acceptor. If you'd rather GPTMD just take the `MostAbundantDotMassDiffAcceptor` and call `.Accepts` directly, I'm happy to switch — it's a wash on correctness, and it would delete `MatchesCandidateMass`. The tradeoff is purely "GPTMD carries an acceptor again" vs. "the apex test lives in one more small method." Your call; tell me which you prefer.

(One structural note in favour of keeping it off the acceptor: `Accepts` folds the candidate's shifts into its own shift set, whereas `GetPossibleMods` already enumerates each fully-formed candidate mass and just needs a yes/no on it. `MatchesCandidateMass` fits that caller exactly; routing through `Accepts` would mean constructing a Dot acceptor whose shift set is the single candidate, which is a stranger shape than the plain per-candidate test.)

MetaMorpheus/EngineLayer/SpectrumMatch/PsmTsvWriter.cs
trishorts: Agreed, and I've teed up the mzLib half so we can do it right (owned by mzLib, not MM-local):

**smith-chem-wisc/mzLib#1097** — adds the optional `Precursor Most Abundant Mass` psmtsv column to the reader/header. Same recipe as #1086 (which added `Most Abundant Mass Diff (ppm)`): optional, mapped by name, reads back null when absent, so old files are unaffected. The one difference from the error column is that this is an observation of the precursor envelope, not a per-hypothesis value — so it's one number per row and copies straight through disambiguation rather than being `|`-split.

Once #1097 releases, the MetaMorpheus side is the same pattern we used for the error column: `PsmTsvWriter` writes `SpectralMatch.ScanPrecursorMostAbundantMass` keyed by the mzLib header const, gated by the same `includeMostAbundantColumn` flag so normal output is unchanged. Since you said you won't hold up this PR on it, I'll add the MM writer change here once #1097 is merged/released, or as an immediate follow-up PR — whichever you prefer for merge sequencing.

MetaMorpheus/EngineLayer/PrecursorSearchModes/PrecursorMassExtensions.cs
nbollis: I guess I am confused why GPTMD specifically needed this correction but it is not needed elsewhere

MetaMorpheus/EngineLayer/SpectrumMatch/PsmTsvWriter.cs
trishorts: Done in `008982f4d`. The observed apex now writes as its own column, **Precursor Most Abundant Mass**, right next to Most Abundant Mass Diff (ppm) and under the same `includeMostAbundantColumn` gate. It's `ScanPrecursorMostAbundantMass` (0 when no envelope resolved) written as a precursor observation, independent of the peptide hypothesis.

Per your earlier note that these headers live in mzLib's `SpectrumMatchFromTsvHeader`, the const + reader/round-trip went in mzLib #1097 (`fc2889cc`, released 1.0.583), and this PR bumps mzLib 1.0.582 → 1.0.583 to consume it — so the column reads back through mzLib's TSV reader rather than being write-only. Added a writer test asserting the header/row stay column-aligned and the cell carries the observed apex. Filtered suite (MostAbundant/PsmTsv/TsvReader) 41/41.

trishorts:
## Why G-PTM-D handles the isotope offset differently from search and calibration

A question came up about why the most-abundant isotope-offset correction takes a different form in G-PTM-D than in the search/calibration/localization paths. Recording the answer here.

### The shared problem

In most-abundant mode we deliberately accept precursor candidates whose deconvoluted **monoisotopic** peak is off by `k` neutrons (±1–2 Da) from the true monoisotopic — that acceptance is the entire point of the mode. But `ScanPrecursorMass` stays monoisotopic, so any consumer reading

```
ScanPrecursorMass − peptideMonoisotopicMass
```

sees a **whole-isotopologue offset, not chemistry**. Three engines were reading it as chemistry: calibration (as m/z drift — the 1940 ppm runaway), localization (a phantom ~1–2 Da mod on a residue), and G-PTM-D (a bogus mod-candidate difference).

### Why calibration / search / localization can just round

Most-abundant mode always searches with the **zero-centred apex acceptor** (`GetMassDiffAcceptor` overrides `MassDiffAcceptorType`). Under that acceptor the residual can only ever be `k neutrons + a within-tolerance ppm error, and nothing else`. So these paths recover the truth by **rounding to the nearest neutron**, subtracting `k`, and returning to a clean mono-vs-mono comparison — lossless, and the identity for every monoisotopic acceptor (default mode is unchanged bit-for-bit).

### Why G-PTM-D cannot round

G-PTM-D does not use the apex acceptor — it searches over **discrete modification mass shifts (the notches)**. So its residual is:

```
k neutrons  +  an UNKNOWN modification mass
```

Rounding would **fold the modification into `k` and destroy it** (a real +79.966 phospho would round away). The neutron offset and the modification mass are entangled and rounding cannot separate them.

So G-PTM-D inverts the test. Rather than correcting the observed mass and then asking "what mod is this," `GetPossibleMods` tests **each candidate (peptide + modification) individually** against the observation, apex-to-apex, via `MatchesCandidateMass` — the same acceptance rule as `MostAbundantDotMassDiffAcceptor.Accepts`, expressed over the deconvolution parameters. The combo recursion accumulates a candidate-mass **offset** instead of subtracting from the target mass, so the **observed mass is left untouched** and the apex test always sees what was really measured. Covered by `MatchesCandidateMass_FindsModification_DespiteApexMisprediction`: a proteoform observed one neutron off its apex still matches phospho and still does not match oxidation.

### One-liner

Calibration/search know the leftover mass *is* just neutrons, so they round it away. G-PTM-D's leftover is "neutrons **plus** an unknown mod (a notch)," which rounding can't disentangle — so G-PTM-D tests every candidate mass against the raw observation instead of correcting the observation first.

Related detail: the Dot acceptor re-derives the `(shift, k)` pair rather than reading `k` back from the notch, because all `k` for a given mass shift share **one notch** (to preserve per-shift FDR grouping), so the notch alone does not carry the neutron count.

nbollis:
What is meant by rounding neutron masses? I believe we multiple a double by a fixed amount that represents more significant figures than mass spec resolution to produce the notch integer. 



"In most-abundant mode we deliberately accept precursor candidates whose deconvoluted monoisotopic peak is off by k neutrons (±1–2 Da) from the true monoisotopic" 

Can you clarify what you mean by using the monoisotopic for a most abundant search? The statement also appears to contradict another: 

"Most-abundant mode always searches with the zero-centred apex acceptor"




IfI read this one right, "So these paths recover the truth by rounding to the nearest neutron, subtracting k, and returning to a clean mono-vs-mono comparison," the way the code works is by taking the most abundant mass from the our deconvolution, and converts it to monoisotopic allowing for k neutron shift. How is that different than a mono search with missed monos? 


The issues with

ScanPrecursorMass − peptideMonoisotopicMass

are valid. However I believe the solution is flawed. The manuscript states, " We instead compare the observed most abundant isotopic peak mass against each candidate's theoretical most abundant mass," and that does not appear to be the case

nbollis (approved):

Alexander-Sol (approved):

RayMSMS (approved):

PR #2675, opened 2026-06-19
feat(output): ProForma column in top-down results

trishorts:
Adds a **ProForma 2.0** column to top-down results — in the psmtsv output (immediately after **Full Sequence**) and in the MetaDraw PSM grid.

## What changed

**`EngineLayer/SpectrumMatch/PsmTsvWriter.cs`** — in `AddPeptideSequenceData`, emit `SpectrumMatchFromTsvHeader.ProForma` right after `FullSequence`, gated on `GlobalVariables.AnalyteType == AnalyteType.Proteoform`. The value is `IBioPolymerWithSetMods.ToProFormaString()`, resolved across ambiguous matches with the same `Resolve(...)` idiom as the neighbouring sequence columns.

- The single analyte-type gate keeps the header (built with a null match) and the data rows in sync, so the column appears as a unit or not at all.
- Both `AllPSMs.psmtsv` and `AllProteoforms.psmtsv` write through `GetTabSeparatedHeader` / `SpectralMatch.ToString`, so a top-down run gets the column in both files; bottom-up and oligo runs get it in neither.

**`GUI/MetaDraw/MetaDraw.xaml` + `.xaml.cs`** — add a `ProForma` column to the PSM grid, also immediately after Full Sequence, so the string is visible without opening the tsv. It binds to `SpectrumMatchFromTsv.ProForma`, which mzLib reads via `GetOptionalValue`, so pre-ProForma files load fine.

The column ships `Visibility="Collapsed"` and is revealed after load only when the loaded results actually carry a ProForma string — bottom-up and legacy files show no empty column. Visibility is keyed off the loaded data rather than `GlobalVariables.AnalyteType`, because that global defaults to `Peptide` and is never set by standalone MetaDraw; gating on it would have hidden the column for genuine top-down results.

ProForma is a top-down-only feature and is added automatically for top-down runs — deliberately no toml setting and no GUI option to control it.

## Tests (`Test/PsmTsvWriterTests.cs`)

- `ProForma_TopDownSearch_ColumnFollowsFullSequenceAndCarriesProFormaString` — asserts the column sits immediately after Full Sequence in both header and data row, and carries the proteoform's ProForma string including the modification descriptor.
- `ProForma_BottomUpSearch_ColumnAbsent` — asserts the column is absent from the header for bottom-up.
- Both restore `GlobalVariables.AnalyteType` in a `finally`.

Patch coverage hits both `if` branches and both ternary outcomes; `codecov/patch` passes.

## Dependency — satisfied

This builds on the ProForma reader/writer from **smith-chem-wisc/mzLib#1072**, merged 2026-07-15 and shipped in **mzLib 1.0.583**, which master already references. No package bump is needed in this PR, and CI is green across all jobs.


pcruzparri (dismissed):

nbollis (approved):

MetaMorpheus/EngineLayer/SpectrumMatch/PsmTsvWriter.cs
```diff
@@ -200,6 +201,10 @@ internal static void AddPeptideSequenceData(Dictionary<string, string> s, Spectr
 
             s[SpectrumMatchFromTsvHeader.BaseSequence] = pepWithModsIsNull ? " " : sm.BaseSequence ?? Resolve(pepWithModsIsNull ? null : pepsWithMods.Select(b => b.BaseSequence)).ResolvedString;
             s[SpectrumMatchFromTsvHeader.FullSequence] = pepWithModsIsNull ? " " : sm.FullSequence != null ? sm.FullSequence : Resolve(pepWithModsIsNull ? null : pepsWithMods.Select(b => b.FullSequence)).ResolvedString;
+            // ProForma 2.0 string, top-down only, immediately after Full Sequence. Gated on the run's
+            // analyte type so the header (sm == null) and the data rows include/exclude the column together.
+            if (GlobalVariables.AnalyteType == AnalyteType.Proteoform)
+                s[SpectrumMatchFromTsvHeader.ProForma] = pepWithModsIsNull ? " " : Resolve(pepsWithMods.Select(b => b.ToProFormaString())).ResolvedString;
```
acesnik, line 207: Worth being deliberate rather than incidental, since this changes a produced output: inserting here shifts every column after Full Sequence by one for top-down psmtsv.

Name-based readers are unaffected — mzLib reads this through `GetOptionalValue`, and anything using pandas reads by header — but a consumer with hardcoded column indices, an `awk` or `cut` script, breaks on top-down files. Appending at the end would avoid that and lose the adjacency that makes the column readable.

I think readability is the right trade for a new top-down-only column and I am not asking for it to change. Just worth having on the record as a choice.

MetaMorpheus/Test/PsmTsvWriterTests.cs
```diff
@@ -517,6 +519,204 @@ public static void TestAddMatchScoreData_ToStringIntegration_NotchAmbiguity()
...
+                int fullSeqIndex = headerSplits.IndexOf(SpectrumMatchFromTsvHeader.FullSequence);
+                int proFormaIndex = headerSplits.IndexOf(SpectrumMatchFromTsvHeader.ProForma);
+
+                // Column is present and sits immediately after Full Sequence.
+                Assert.That(proFormaIndex, Is.EqualTo(fullSeqIndex + 1));
+
+                string[] rowSplits = psm.ToString(new Dictionary<string, int>()).Split('\t');
+
+                // Header and data row stay in sync - the invariant the single analyte-type gate exists to protect.
+                Assert.That(rowSplits.Length, Is.EqualTo(headerSplits.Length));
```
acesnik, line 566: This is the assertion that actually defends the change. A header/row field-count mismatch is how a tsv silently corrupts every column after the insertion point, and it would sail past a test that only checked the new value was present.

Also good that the absent-column case is parameterised over `Peptide` and `Oligo` rather than bottom-up alone, and that both tests are `[NonParallelizable]` with the reason written down and the global restored in `finally`.

acesnik (approved):
🤖 Approving. Four lines in the writer, and the tests defend the right invariant.

Two things I checked rather than took from the description.

**Header and rows cannot desynchronise.** `GetTabSeparatedHeader` calls `DataDictionary(null, null)` and takes `.Keys`, so the header comes off the same `AddPeptideSequenceData` path as the data rows — one `AnalyteType` gate governs both.

**`ToProFormaString()` cannot crash a run.** I went and read smith-chem-wisc/mzLib#1072, since a throw here would surface as a top-down search failing while writing results. The syntax is written by TopDownProteomics' own `ProFormaWriter` rather than hand-rolled, and `ProFormaConverter.BuildDescriptor` falls back to a name-keyed descriptor for a modification it cannot map to an ontology, so an unrecognised mod degrades to `[ModName]` instead of throwing. The exceptions in that converter are all on the read side, which this does not use.

Nice call keying the MetaDraw column's visibility off whether the loaded results carry a ProForma string rather than off `GlobalVariables.AnalyteType` — standalone MetaDraw never sets that global, so the obvious version of that check would have hidden the column for genuine top-down results.

zhuoxinshi (approved):

PR #2676, opened 2026-06-25
Fix a Crash

nbollis:
something weird happened with a collaborators search. We have not finished troubleshooting why it occurred, but this will prevent it from happening in the future.


pcruzparri (approved):

Alexander-Sol (approved):

trishorts:
The null guard here closes the crash for `GetMs2Scans`, but the sibling `GetMs2ScansWrapByScanNum` consumes the same `_GetMs2Scans(...)` array and still uses the pre-fix pattern (`scansWithPrecursors.SelectMany(p => p.SelectMany(...))` and the indexed `scansWithPrecursors[i].Any()` loop). If a null element can appear from `_GetMs2Scans` — which is the premise of this fix — that method will throw the same `NullReferenceException` on the MsAlign / scan-number-wrapped path.

Recommend applying the same `p != null` / `p.Count > 0` guards there so the crash class is fully closed rather than partially.

Separately (non-blocking): since the root cause is still unknown, consider having the guard emit a `Warn(...)` (or a once-per-run count) when it drops a null/empty partition, so the anomaly stays observable for diagnosis instead of being silently masked.

PR #2679, opened 2026-06-29
Two New MetaDraw Plots: Histogram of Ambiguity Levels and Histogram of Notch

nbollis:
image


pcruzparri (approved):

MetaMorpheus/Test/MetaDraw/MetaDrawTest.cs
```diff
@@ -2295,11 +2295,788 @@ bool HasColor(Bitmap bmp, OxyColor color)
...
+                    new ObservableCollection<SpectrumMatchFromTsv>(allPsms),
+                    psmsByFile,
+                    parameters);
+
+                Assert.That(plot.Model.Axes.OfType<PlotCategoryAxis>()
+                    .Any(a => a.Key == "GroupAxis" && a.Title == "Precursor Charge"),
+                    Is.True, "GroupAxis should be created when grouping is enabled for the ambiguity plot");
+                Assert.That(plot.PlotData.All(r => r.ContainsKey(parameters.GroupingProperty)), Is.True,
+                    "Every PlotData row should carry the grouping property");
+                Assert.That(plot.PlotData.Select(r => r[parameters.GroupingProperty]).Distinct().Count(), Is.GreaterThanOrEqualTo(1));
```
trishorts, line 2740: **[Major] AmbiguityLevel grouped-by-charge test never actually groups across charges**

`TestPlotModelStat_AmbiguityLevel_GroupedByPrecursorCharge` carries the comment "Two PSMs (charge 2) and two PSMs (charge 3)… forces `GetCategoryDictForGroup` case 13", but `WriteAmbiguityLevelPsmTsv` hardcodes the Precursor Charge column to the literal `"2"` for every row. All four PSMs therefore share charge 2, so the grouping pipeline only ever produces a **single group**. This assertion is satisfied by one group and never verifies that distinct charges produce distinct groups:

```csharp
Assert.That(plot.PlotData.Select(r => r[parameters.GroupingProperty]).Distinct().Count(), Is.GreaterThanOrEqualTo(1));
```

A regression that mis-buckets PSMs into the wrong charge group, or collapses multiple groups into one, would not be detected.

**Suggested fix:** give `WriteAmbiguityLevelPsmTsv` a way to emit varied Precursor Charge values (e.g. accept a charge per row, or alternate 2/3), then assert the distinct grouping-key count is exactly 2 rather than `>= 1`. The same pattern applies to `TestPlotModelStat_Notch_GroupedByPrecursorCharge` (Minor).

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code) — automated review finding</sub>

Alexander-Sol (approved):

RayMSMS (approved):

PR #2681, opened 2026-07-09
test(gui): make UniProt download test offline; gate live call behind [Explicit] canary

trishorts:
## Why

`GuiFunctionsTest.UniprotHtmlQueryTest` streamed the *Bacillus phage phi29* proteome (UP000001207) from `rest.uniprot.org` on every run and asserted an exact protein count. That made the test — and therefore **CI on every open mzLib PR** (the mzLib `integration` job clones MetaMorpheus master and runs this suite) — depend on a live third-party endpoint.

It is currently red everywhere: UniProt's `/stream` endpoint returns

```
HTTP/1.1 200
Error encountered when streaming data. Please try again later.
```

so zero proteins parse and `Assert.That(reader.Count == listCount)` fails (`Expected: True / But was: False`). Even when UniProt is healthy, the hardcoded counts drift whenever the reference proteome is revised.

## What changed

Network dependency removed from CI; live path preserved as an opt-in canary (same idea as the `[Category("Koina")]` network tests):

- **`TestGetUniProtHtmlQueryString`** — offline; asserts the exact REST URL built for all 11 reviewed/isoform/format/compressed combinations. This is the actual GUI production code (`DownloadUniProtDatabaseFunctions.GetUniProtHtmlQueryString`) under test, with no network.
- **`LoadUniProtProteomeFromLocalFile_GeneratesTargetsAndDecoys`** — offline "local mock"; loads committed UniProtKB-format fixtures (`Test/DatabaseTests/UniProtMock/phi29_mock_reviewed.{xml,fasta}`) through the same `ProteinDbLoader` path the download uses and confirms targets + reverse decoys are generated (count == 2 × entries).
- **`UniProtDownloadCanary`** — the real live REST download, marked `[Test, Explicit]` + `[Category("UniProt")]` so it **never runs in an unfiltered `dotnet test`** (stays out of CI) and only fires when a developer selects it (`dotnet test --filter "FullyQualifiedName~UniProtDownloadCanary"`). It asserts `proteins parsed > 0` rather than an exact count, so a UniProt content revision can't turn it into a false alarm — its job is to confirm the live pipeline still works.

`TestGetUniprotFilename` is unchanged.

## Verification

- Unfiltered run of the fixture: **26 passed, 0 failed, 0 skipped**; the 2 `[Explicit]` canary cases are correctly excluded.
- Selecting the canary explicitly runs it; it currently fails *because UniProt is down right now* — i.e. it does exactly what a canary should.


pcruzparri (approved):

zhuoxinshi (approved):

nbollis (approved):

MetaMorpheus/Test/GuiTests/GuiFunctionsTest.cs
```diff
@@ -44,72 +44,105 @@ public static void TestGetUniprotFilename(string proteomeID, bool reviewed, bool
...
 
             File.Delete(filePath);
 
-            if (testName.Equals("11.fasta")) // only triggers for last test cases
-            {
-                Console.WriteLine("Finished with Uniprot HTML query test.");
-            }
-
-            NUnit.Framework.Assert.That(reader.Count == listCount);
+            Assert.That(reader.Count, Is.GreaterThan(0),
```
nbollis, line 144: Will this fail once uniprot is back online?

trishorts: Good question — it's actually the reverse. `UniProtDownloadCanary` is `[Explicit]`, so it never runs in an unfiltered `dotnet test` (neither MM's `Test.yml` nor the mzLib `integration` job passes a `--filter`), which is what keeps it out of CI. When you *do* run it explicitly against a healthy UniProt, `GetAsync` returns the proteome and `reader.Count > 0` passes. It only fails if the live endpoint is genuinely down or has changed its URL/format — i.e. exactly when we'd want the heads-up. The assertion is `> 0` rather than an exact count on purpose, so a UniProt content revision can't turn it into a false alarm. It "fails" today only because UniProt's `/stream` is currently returning an HTTP-200 error body.

Follow-up (per chat with Michael): we'll generalize this into a shared `[Category("ExternalService")]` label + a classify-or-skip helper so these live tests run in a non-blocking job and `Assert.Ignore` (Skip) on a real outage instead of failing — "we tried, UniProt was down, don't worry" — while still failing loudly on an actual contract break.

PR #2684, opened 2026-07-10
MetaDraw: Add Score and Intensity coloration modes to BioPolymerCoverage View

nbollis:
Human Explanation. 

Add two new color mapping types to the protein coverage overlay. 

By Intensity
image

By Intensity, log scaled
image

By Score
image

## Motivation

The BioPolymer Coverage view previously colored peptide rectangles only by coverage type (unique/shared/tandem) or file origin. This adds two new coloration modes to visualize intensity and score distributions directly on the coverage map.

## Summary

Adds two new `ColorResultsBy` modes — **PrecursorIntensity** and **Score** — that color each peptide rectangle using a 20-bin viridis scale normalized to the min/max of the current protein group's filtered results.

### Changes

**`BioPolymerCoverageMapViewModel.cs`:**
- Added `PrecursorIntensity` and `Score` to `ColorResultsBy` enum
- Added 20-color static `ViridisColors` array (purple to blue to teal to green to yellow)
- Added `GetIntensityBrush(normalizedValue)` helper for bin mapping
- Pre-computes min/max/range per group before the drawing loop
- Replaces `GetColorBrush(res)` with intensity-aware ternary for the new modes
- Draws a gradient bar in the legend showing the metric name and actual min/max values
- `CreateLegendItems` returns empty for intensity modes (gradient bar replaces text legend)
- `headerBlockHeight` accounts for gradient bar height to prevent overlap

**Test files:**
- Added `SetScore(double)` to `DummySpectralmatch` for test flexibility
- 14 new tests covering `ViridisColors` initialization, `GetIntensityBrush` mapping (bounds, mid, clamping), ColorBy setter, `CreateLegendItems` early return, and integration plotting with various intensity scenarios (nulls, equal values, varying scores)

## Testing

- All 16 test methods pass
- Categories: unit (brush mapping), integration (CoverageDrawing produced), edge case (all equal values, null PrecursorIntensity, clamping)
- Uses existing `DummySpectralmatch` pattern with reflection for PrecursorIntensity backing field access

## Migration

No migration steps. New modes appear automatically in the "Color By" dropdown via `Enum.GetValues`. No XAML or configuration changes needed.


zhuoxinshi (dismissed):

RayMSMS (dismissed):

trishorts (commented):
Automated code review - 2 finding(s) posted as inline comments.

MetaMorpheus/GuiFunctions/MetaDraw/BioPolymerCoverage/BioPolymerCoverageMapViewModel.cs
```diff
@@ -352,7 +357,7 @@ private void Redraw()
                         roundRight = true;
```
trishorts, line 357: **Major** - DrawNumericLegend assumes GradientBrushes is non-null and non-empty

**Problem:**
`DrawNumericLegend` unconditionally dereferences `colorMapper.GradientBrushes!` (null-forgiving) and immediately computes `double bandW = barWidth / brushes.Count`. It is reached whenever `colorMap.IsNumeric` is true, which depends only on the selected `ColorBy` mode, not on whether any results exist. When the user selects PrecursorIntensity/Score and the current group's `filteredResults` is empty (e.g. the active MetaDraw filter rejects every match), `colorMap.Prepare(filteredResults, …)` runs against an empty list. If `Prepare` leaves `GradientBrushes` null (or empty) in that case, the `!` dereference throws a `NullReferenceException`, and an empty brush list makes `bandW` a divide-by-zero (`Infinity`). The exception propagates out of `Redraw` and aborts the entire coverage render, not just the legend.

**Impact:**
Switching to an intensity/score coloration mode while the current protein group has zero filter-passing results can crash the coverage drawing render instead of showing an empty/gray plot.

**Suggested fix:**
Guard `DrawNumericLegend` (and the `if (colorMap.IsNumeric)` call site) against a null/empty `GradientBrushes`, returning a zero-height block when there is nothing to draw; alternatively guarantee in `Prepare` that numeric mappers always populate a non-empty `GradientBrushes` even for empty input.

**Affected test scenarios:**
- Redraw - ColorBy=PrecursorIntensity with all results filtered out (empty filteredResults) - no exception, coverage image still produced
- Redraw - ColorBy=Score with empty filteredResults - gradient legend omitted or drawn empty, no divide-by-zero

MetaMorpheus/GuiFunctions/MetaDraw/BioPolymerCoverage/ColorMapping/PrecursorIntensityColorMapper.cs
```diff
@@ -0,0 +1,13 @@
+namespace GuiFunctions.MetaDraw;
+
+public sealed class PrecursorIntensityColorMapper : NumericBioPolymerCoverageColorMapper
```
trishorts, line 3: **Major** - PrecursorIntensity → Score fallback is never wired up; all-null intensity renders gray

**Problem:**
The PR explicitly claims "all-null `PrecursorIntensity` falls back to `Score` for normalization." The base class `NumericBioPolymerCoverageColorMapper` contains the full machinery for this: `protected virtual NumericBioPolymerCoverageColorMapper? GetFallbackMapper() => null;` and the `if (usable.Count == 0) { var fallback = GetFallbackMapper(); if (fallback is not null) {...} }` block in `Prepare`. However, `PrecursorIntensityColorMapper` overrides only `ColorBy`, `DisplayName`, `DefaultUseLogScale`, and `GetNumericValue` — it never overrides `GetFallbackMapper()`. Therefore `GetFallbackMapper()` always returns `null`, the fallback branch is dead, and when all `PrecursorIntensity` values are null (or all ≤ 0 in the default log-scale mode), `usable.Count == 0` → `transformed.Count == 0` → `scale = new BioPolymerCoverageColorScale(0, 0, gradient)`. Every rectangle then fails `IsRenderable` (null value) or normalizes to 0 and the view shows no meaningful intensity coloring.

**Impact:**
The advertised graceful-degradation behavior does not exist. On the common real-world case of a protein group whose matches have no precursor intensity populated, the PrecursorIntensity mode produces an all-gray/degenerate coverage map instead of falling back to Score coloring. The entire fallback code path in the base class is unreachable dead code.

**Suggested fix:**
Override `GetFallbackMapper()` in `PrecursorIntensityColorMapper` to return a `ScoreColorMapper` instance so the base `Prepare` logic can engage the fallback when no usable intensity values exist. Add a regression test asserting the fallback actually activates.

**Affected test scenarios:**
- PrecursorIntensityColorMapper.Prepare - all matches have null PrecursorIntensity, Score populated - GetBrush returns non-gray brushes derived from Score, and GradientLegendTitle reflects "Score"
- PrecursorIntensityColorMapper.Prepare - log scale on, all intensities ≤ 0 - fallback to Score engages rather than degenerate (0,0) scale

trishorts:
**Major** — `ResetSettings` does not restore the new `BioPolymerCoverageGradientType`

_(Posting as a conversation comment because the relevant line, `MetaDrawSettings.cs:289`, is not inside this PR's changed hunks and so can't be anchored inline.)_

**Problem:**
This PR adds a new mutable global static, `BioPolymerCoverageGradientType`, initialized to `ColorGradientType.Viridis`. `ResetSettings()` — whose XML doc explicitly states it is *"Used to reset the settings to their default value, particularly needed for unit testing"* — resets dozens of individual settings and calls `SetDefaultColors()`, but never assigns `BioPolymerCoverageGradientType`. `SetDefaultColors()` / `SetDefaultBioPolymerCoverageColors()` also leave it untouched. As a result, once any code path (or test) mutates `BioPolymerCoverageGradientType`, `ResetSettings()` cannot bring it back to the default.

**Impact:**
Global gradient state leaks across unit tests that run in the same process. A test that sets a non-Viridis gradient will leave the static dirty, so a later test asserting default coloration behavior can fail or pass nondeterministically depending on execution order. In the running application, invoking "reset to defaults" silently leaves the gradient on whatever the user last chose.

**Suggested fix:**
Add an explicit reset of `BioPolymerCoverageGradientType` to `ColorGradientType.Viridis` inside `ResetSettings()` (or inside `SetDefaultBioPolymerCoverageColors()` so it participates in `SetDefaultColors()`), matching how every other new customizable setting is handled.

nbollis:
Fixed. Added `BioPolymerCoverageGradientType = ColorGradientType.Viridis;` in `MetaDrawSettings.ResetSettings()`. Added `ResetSettings_RestoresDefaultBioPolymerCoverageGradientType` to lock the behavior in — it sets a non-Viridis gradient, calls `ResetSettings`, and asserts the value returned to Viridis. The test restores the original value in a `finally` block to keep the global static from leaking into other tests.

MetaMorpheus/GuiFunctions/MetaDraw/BioPolymerCoverage/ColorMapping/PrecursorIntensityColorMapper.cs
nbollis: Good catch — that fallback path was dead code, and switching the mapping on the fly is confusing UX. Per discussion, I removed the entire `GetFallbackMapper` machinery: the base class no longer has the hook, `PrecursorIntensityColorMapper` no longer references it, and the in-source README claim of a Score fallback is gone. When PrecursorIntensity has no usable values, the mapper now produces a (0,0) scale and `GetBrush` returns gray — the user sees that they need to switch to Score themselves, rather than having the display silently change the underlying metric.

Replaced the old fallback test with `NumericMapper_AllValuesNull_ProducesZeroRangeScale`, which asserts the new contract.

MetaMorpheus/GuiFunctions/MetaDraw/BioPolymerCoverage/BioPolymerCoverageMapViewModel.cs
nbollis: Fixed. `DrawNumericLegend` now early-returns with a 0-height block when `GradientBrushes` is null or empty, so the null-forgiving dereference and the divide-by-zero are both gone. Added `DrawNumericLegend_EmptyFilteredResults_DoesNotCrash` (PrecursorIntensity with null intensities on all results) to lock in the behavior.

trishorts:
Thanks for the quick turnaround on all three.

Heads up though — the branch head is still `108dd7c`, the same commit I reviewed, so none of these fixes (or the three new tests) are actually pushed yet. I verified against HEAD:

- **#1 `DrawNumericLegend`** — still has the unguarded `colorMapper.GradientBrushes!` deref and `barWidth / brushes.Count`; no early-return, and `DrawNumericLegend_EmptyFilteredResults_DoesNotCrash` isn't in the branch.
- **#2 fallback machinery** — `GetFallbackMapper()` and the `if (usable.Count == 0)` fallback branch are still present in `NumericBioPolymerCoverageColorMapper.cs`; `NumericMapper_AllValuesNull_ProducesZeroRangeScale` isn't there yet.
- **#3 `ResetSettings`** — no `BioPolymerCoverageGradientType` assignment in `ResetSettings()`, and `ResetSettings_RestoresDefaultBioPolymerCoverageGradientType` isn't in the branch.

Looks like the commits just haven't been pushed. Once they land I'll re-review. Also 👍 on the design call for #2 — removing the fallback rather than wiring it up resolves the dead-code concern.

Alexander-Sol (approved):

RayMSMS (approved):

PR #2685, opened 2026-07-12
Added support for bruker qTOF files

### LINKED ISSUE #2722: Bruker .d validity is re-derived in two places and has drifted from mzLib: .tsf folders are rejected
MetaMorpheus decides whether a `.d` folder is readable using its own hand-written marker-file check, in two independent copies. mzLib already owns that classification, and the copies have drifted from it — most visibly for `analysis.tsf` (non-TIMS timsTOF/flex data), which MetaMorpheus rejects and mzLib reads.

This is pre-existing on master. It was found while reviewing #2685 and is explicitly **not** caused by that PR — #2685 only widened the check to accept `analysis.baf`, faithfully following the existing pattern.

## The divergence

mzLib classifies a `.d` folder in one place, `Readers/Util/SupportedFileTypes.cs:97-104`:

```csharp
case ".d":
    if(!Directory.Exists(filePath)) throw new FileNotFoundException();
    var fileList = Directory.GetFiles(filePath).Select(p => Path.GetFileName(p));
    if (fileList.Any(file => file == "analysis.baf"))
        return SupportedFileType.BrukerD;
    if (fileList.Any(file => file == "analysis.tdf" || file == "analysis.tsf"))
        return SupportedFileType.BrukerTimsTof;
    throw new MzLibException("Bruker file type not recognized");
```

MetaMorpheus re-derives this twice, and neither copy knows about `.tsf`:

| Copy | Marker set |
|---|---|
| `CMD/CommandLineSettings.cs:267` `IsValidBrukerDirectory` | `analysis.tdf` **and** `analysis.tdf_bin`, or `analysis.baf` |
| `GUI/MainWindow.xaml.cs:1711` `IsValidTimsTofDirectory` | `analysis.tdf` **and** `analysis.tdf_bin` |

So a `.d` folder containing `analysis.tsf` + `analysis.tsf_bin` is rejected by both entry points — CMD silently (the directory crawl at `CommandLineSettings.cs:215` has no `else`), the GUI with `"... is not a valid Bruker data file."` — even though mzLib routes it to `TimsTofFileReader`, which has a real TSF path: `tsf_open` at `TimsTofFileReader.cs:152`, `LoadAllDataTsf` at `:410`, and a dedicated `TsfFrameProxy`.

**Scope caveat, so this doesn't promise more than it delivers:** mzLib's TSF support is currently DMRM-only — `TimsTofFileReader.cs:425` throws `"Only DMRM data is supported for .tsf files"` for other scan modes. Fixing the gate makes DMRM `.tsf` data reachable; it does not make all TSF data work.

Three smaller disagreements come from the same root cause:

- **`analysis.tdf_bin`** — mzLib's *classifier* doesn't require it; the pair check lives one layer down at `TimsTofFileReader.cs:76`. MetaMorpheus enforces it at the gate instead. MetaMorpheus's is arguably the better UX (a clear rejection rather than a `FileNotFoundException` from the reader), but the requirement now lives in two layers that can move independently.
- **Case sensitivity** — mzLib compares filenames with `==` (ordinal); `File.Exists` is case-insensitive on NTFS. A folder holding `Analysis.baf` passes the MetaMorpheus gate and then throws `MzLibException("Bruker file type not recognized")` from mzLib.
- **Trailing separator** — mzLib keys on `Path.GetExtension`, which yields `""` for `"...\foo.d\"` and falls through to `"File type not supported"`. The MetaMorpheus helpers take a directory path and `Path.Combine`, so they are insensitive to it.

## Why it will keep drifting

`Test/Test.csproj` references only `CMD`, `EngineLayer`, `GuiFunctions` and `TaskLayer` — **not** the WPF `GUI` project. `MainWindow.xaml.cs` and `MetaDraw.xaml.cs` are therefore unreachable by the test suite, so only the CMD copy has any coverage and nothing can catch the two copies diverging.

## Suggested fix

The gate wants to be a non-throwing question, and mzLib has no such API today (there is no `TryParseFileType`; searched, zero matches). So this is two PRs, not one:

1. **mzLib** — add `TryParseFileType(this string path, out SupportedFileType type)` beside `ParseFileType` in `Readers/Util/SupportedFileTypes.cs`, and collapse the `.tdf`/`.tsf` + `_bin` pair requirement into one layer instead of the current split between `SupportedFileTypes.cs:102` and `TimsTofFileReader.cs:76`.
2. **MetaMorpheus** — delete both copies and call it, so CMD and the GUI agree with each other and with the reader that will actually open the file.

Worth doing in that order, with each PR naming the other, so the new mzLib API doesn't read as dead code on its own.

Alexander-Sol:
This PR will need a mzLib release before it's ready to go in. 

Previously, we only supported bruker timsTOF files. Now, we support bruker files produced by qTOFs. The only real change is in how we check and validate file paths for this type of data.

Closes #2722 — the Bruker `.d` validity check is no longer re-derived per entry point; `EngineLayer/Util/BrukerDataDirectory.cs` is now the single source of truth and delegates classification to mzLib's `ParseFileType`, which is what brings `.tsf`/`.tsf_bin` in.


trishorts (dismissed):
Approving — the qTOF `.d` support is a clean, well-scoped extension of the existing timsTOF path handling, and the decoding itself lives in mzLib. No blocking issues. A few minor comments to consider (none required for merge):

**1. `IsValidBrukerDirectory` validation asymmetry (CMD/CommandLineSettings.cs).** timsTOF requires both `analysis.tdf` and `analysis.tdf_bin`, but qTOF passes on `analysis.baf` alone. A truncated/interrupted `.d` (lone `analysis.baf`) clears early validation and only fails deep in the reader. Consider checking a minimal companion-file set (e.g. `analysis.baf_idx`/`analysis.sqlite`) to mirror the two-file timsTOF check — or a brief comment noting the single-file check is intentional.

**2. `.baf` in `AcceptedSpectraFormats` (EngineLayer/GlobalVariables.cs).** The input unit for Bruker is the `.d` directory; adding bare `.baf` makes the interior `analysis.baf` a first-class loadable input. A bare `analysis.baf` path (or a `*.baf` glob) now passes validation but may fail downstream, and a `.d` enumeration could discover the same run twice. Worth confirming the intended unit and validating consistently.

**3. GUI `*.baf` dialog filter (GUI/MainWindow.xaml.cs).** Windows' legacy prefix-matching means `*.baf` also lists `analysis.baf_idx`/`analysis.baf_xtr`. Selecting one falls through to `default` → "Unrecognized file type: .baf_idx" rather than loading the dataset. Redirecting in-`.d` siblings to the parent `.d` (as `case ".baf"` already does) or a friendlier default message would improve discoverability.

**4. Test fidelity/coverage (Test/TestCmd.cs).** The new BAF tests cover only the positive path with a single hand-written `analysis.baf`. Consider (a) a negative case mirroring `TestNestedDotDFolderDiscovery` (a `.d` lacking `analysis.baf` is rejected), and (b) building the fixture from the full qTOF marker set (or pointing at the committed `centroid_1x_MS1_4x_autoMS2.d`) so the test tracks how real `.d` folders validate.

Also: three trailing-whitespace-only edits crept into `CommandLineSettings.cs` and `MainWindow.xaml.cs` alongside the functional renames — minor churn, easy to drop.

(Sequencing reminder, not a defect: needs mzLib 1.0.583 released before it can build/merge, as noted.)

MetaMorpheus/CMD/CommandLineSettings.cs
```diff
@@ -209,12 +209,13 @@ private static void FindSpectraFilesRecursive(string path, List<string> spectraF
             {
                 if (GlobalVariables.AcceptedSpectraFormats.Contains(GlobalVariables.GetFileExtension(path).ToLowerInvariant()))
                 {
-                    // If a .tdf or .tdf_bin file is found, check if the parent directory is a valid .d folder and add that instead
-                    if (new[] { ".tdf", ".tdf_bin" }.Contains(GlobalVariables.GetFileExtension(path).ToLowerInvariant()))
+                    // If a Bruker inner file is found (.tdf/.tdf_bin for timsTOF, .baf for regular TOF),
+                    // check if the parent directory is a valid .d folder and add that instead
+                    if (new[] { ".tdf", ".tdf_bin", ".baf" }.Contains(GlobalVariables.GetFileExtension(path).ToLowerInvariant()))
```
trishorts, line 214: The inner-file correction branch here (the `new[] { ".tdf", ".tdf_bin", ".baf" }.Contains(...)` block, lines ~214-218) currently has no test coverage — patch coverage flags 214, 217, and 218 as missed.

Two things worth considering:

1. **Lines 214/217 are coverable** — I've left two small NUnit tests on `TestCmd.cs` that exercise them (verified passing against this branch).

2. **Line 218 (`path = parentDirectory.FullName;`) looks unreachable on this recursion path.** To reach a *file* via `FindSpectraFilesRecursive`, its parent directory must have been recursed into — and for a `.d` folder that only happens when `IsValidBrukerDirectory` returned **false** at the folder-level check (lines ~237-245; a *valid* `.d` is added and `return`s before its contents are ever enumerated). Line 217 then re-checks that same directory, so `IsValidBrukerDirectory(parentDirectory.FullName)` is false again and line 218 never runs. The genuine "user passed the inner file directly" correction is already handled by the separate loop at lines 104-117 (covered by `TestTdfFilePathCorrectedToParentDotDDirectory`). Consider removing the redundant 216-218 block, or if it's kept as defensive symmetry, a one-line comment noting it's unreachable-by-recursion would help the next reader.

MetaMorpheus/Test/TestCmd.cs
```diff
@@ -259,5 +259,105 @@ public static void TestTdfBinFilePathCorrectedToParentDotDDirectory()
                 }
             }
         }
+
+        /// <summary>
+        /// Regression test that a regular (non-tims) Bruker TOF .d folder, identified by an
+        /// analysis.baf file, is discovered as a spectra file when its parent directory is provided.
+        /// </summary>
+        [Test]
+        public static void TestBrukerBafDotDFolderDiscovery()
```
trishorts, line 268: These new BAF tests cover folder discovery and the lines 104-117 path-correction, but nothing drives `FindSpectraFilesRecursive` *into* a Bruker inner file, so the inner-file branch in `CommandLineSettings.cs` (lines ~214-218) stays uncovered. Here are two tests that exercise it — both **verified passing** against this branch. (Note: they cover lines 214 and 217; line 218 stays uncovered because it appears unreachable via recursion — see my comment on `CommandLineSettings.cs`.)

```csharp
/// <summary>
/// When a Bruker inner file (e.g. analysis.baf) sits loose in a directory that is NOT a .d folder,
/// recursive discovery should still find it via the accepted-format check. Exercises the inner-file
/// branch of FindSpectraFilesRecursive where the parent directory is not a .d folder (no path correction).
/// </summary>
[Test]
public static void TestBrukerBafFileDiscoveredThroughDirectoryRecursion()
{
    string tempFolder = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestBafLooseRecursion");

    try
    {
        if (Directory.Exists(tempFolder))
        {
            Directory.Delete(tempFolder, true);
        }
        Directory.CreateDirectory(tempFolder);

        // A bare analysis.baf not wrapped in a .d directory
        File.WriteAllText(Path.Combine(tempFolder, "analysis.baf"), "fake baf content");

        var settings = new CommandLineSettings
        {
            _spectra = new[] { tempFolder },
            _tasks = new[] { Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\Task1-SearchTaskconfig.toml") },
            _databases = new[] { Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\gapdh.fasta") }
        };

        settings.ValidateCommandLineSettings();

        // The loose .baf is discovered; parent is not a .d folder, so no path correction occurs
        Assert.That(settings.Spectra.Any(s => s.EndsWith("analysis.baf")), Is.True,
            "Loose analysis.baf should be discovered through directory recursion");
    }
    finally
    {
        if (Directory.Exists(tempFolder))
        {
            Directory.Delete(tempFolder, true);
        }
    }
}

/// <summary>
/// A .d folder missing a required marker file (analysis.tdf without analysis.tdf_bin) is invalid,
/// so recursion descends into it and encounters the inner .tdf file. Because the parent .d is invalid,
/// no folder-correction occurs and the inner file is added as-is. Exercises the IsValidBrukerDirectory
/// check inside the inner-file branch of FindSpectraFilesRecursive.
/// </summary>
[Test]
public static void TestBrukerInnerFileInsideInvalidDotDIsNotFolderCorrected()
{
    string tempFolder = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestInvalidDotDInnerFile");

    try
    {
        if (Directory.Exists(tempFolder))
        {
            Directory.Delete(tempFolder, true);
        }

        // Invalid timsTOF .d: has analysis.tdf but is missing analysis.tdf_bin
        string invalidDotD = Path.Combine(tempFolder, "incomplete.d");
        Directory.CreateDirectory(invalidDotD);
        File.WriteAllText(Path.Combine(invalidDotD, "analysis.tdf"), "fake tdf content");

        var settings = new CommandLineSettings
        {
            _spectra = new[] { tempFolder },
            _tasks = new[] { Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\Task1-SearchTaskconfig.toml") },
            _databases = new[] { Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\gapdh.fasta") }
        };

        settings.ValidateCommandLineSettings();

        // Parent .d is invalid, so the inner .tdf is kept as-is (not corrected to the .d folder)
        Assert.That(settings.Spectra.Any(s => s.EndsWith("analysis.tdf")), Is.True,
            "Inner .tdf inside an invalid .d folder should be added without folder correction");
        Assert.That(settings.Spectra.Any(s => s.EndsWith("incomplete.d")), Is.False,
            "Invalid .d folder should not be added as a spectra folder");
    }
    finally
    {
        if (Directory.Exists(tempFolder))
        {
            Directory.Delete(tempFolder, true);
        }
    }
}
```

trishorts (commented):
A couple of follow-ups on the new Bruker qTOF path-validation code, focused on test coverage of the `FindSpectraFilesRecursive` inner-file branch (flagged by patch coverage). Non-blocking — the approval stands.

nbollis (dismissed):

nbollis:
@copilot please fix the merge conflicts in this pull request.

nbollis:
@copilot please fix the merge conflicts in this pull request.

nbollis (dismissed):

MetaMorpheus/EngineLayer/GlobalVariables.cs
```diff
@@ -76,6 +76,7 @@ public static class GlobalVariables
         public static void SetUpGlobalVariables()
         {
             AcceptedDatabaseFormats = new List<string> { ".fasta", ".fa", ".xml", ".msp", ".msl" };
+            AcceptedSpectraFormats = new List<string> { ".raw", ".mzml", ".mgf", ".msalign", ".tdf", ".tdf_bin", ".baf", ".d" };
             AcceptedSpectraFormats = new List<string> { ".raw", ".mzml", ".mgf", ".msalign", ".tdf", ".tdf_bin", ".d" };
```
trishorts: **`.baf` never actually reaches `AcceptedSpectraFormats` — line 79 is overwritten by line 80.**

Line 80 is master's older list, re-added by an earlier `Merge branch 'master'` on this branch rather than merged into line 79. Master has exactly one assignment to this property, so this is a merge artifact rather than a pattern anyone chose. The second assignment wins, and `.baf` is dropped at runtime:

```
.raw, .mzml, .mgf, .msalign, .tdf, .tdf_bin, .d
```

What that costs:

| Path | Effect |
|---|---|
| CMD directory crawl (`CommandLineSettings.cs:215`) | **Silent.** There is no `else`, so a qTOF `.d` is dropped with no message at any verbosity |
| CMD explicit `.baf` path (`:164`) | throws `Unrecognized spectra file format: .baf` |
| MetaDraw (`MetaDraw.xaml.cs:115`) | gates this PR's own new `.baf` branch at `:117-121` — currently unreachable |
| MetaDraw file dialog (`:424`) | the filter is derived from this list, so `*.baf` is never offered |
| MainWindow GUI | unaffected — the filter at `:515` and `case ".baf":` at `:1828` are hardcoded, and both are correct |

To show this rather than assert it, I ran the exact `.baf` analogue of your own passing `TestInnerBrukerFilesFoundByRecursionCollapseToParentDotDDirectory` — same shape, `analysis.baf` in place of `analysis.tdf`. It discovers **0** spectra files. Deleting line 80 makes the same probe find the `.d`, with nothing else changed.

The suite stays green because all four Bruker `TestCmd` tests reach the `.d` through the folder-level check, which never consults this list — and `Test.csproj` does not reference the `GUI` project, so neither GUI file is reachable by any test.

**Fix:** delete line 80.

Worth adding the guard as well, since the house convention two lines from your own code would have caught this — `MsDataFileTest.cs:304` already pins `.d`, and `MslDatabaseAcceptanceTests.cs:21` does the same for databases:

```csharp
Assert.That(GlobalVariables.AcceptedSpectraFormats, Does.Contain(".baf"));
```

MetaMorpheus/CMD/CommandLineSettings.cs
trishorts: You were right and I was wrong here — this line is reachable. `TestInnerBrukerFilesFoundByRecursionCollapseToParentDotDDirectory` shows how: a trailing separator on the `.d` path means the folder-level check does not match, so the walk descends into the folder and reaches `analysis.tdf` as an individual file, and this parent check then fires. Covered now, so resolving.

One note for a future cleanup rather than this PR: the correction loop at lines 104-117 runs over everything in `Spectra`, including files found by recursion, so it would reach the same result on its own. Redundant rather than wrong.

MetaMorpheus/Test/TestCmd.cs
trishorts: Addressed — the equivalent tests landed as `TestInnerBrukerFilesFoundByRecursionCollapseToParentDotDDirectory` and `TestInnerBrukerFileInInvalidDotDFolderIsNotCollapsed`, and they improve on what I sketched: the trailing-separator setup is a more realistic trigger than a loose marker file. Resolving.

acesnik (commented):
:robot: Mostly a note to save you re-checking something: @trishorts's `.baf` observation from 14 August still applies at `5f919ea0`. `GlobalVariables.cs` carries two consecutive assignments to `AcceptedSpectraFormats`, on lines 79 and 80, and the second one — which omits `.baf` — wins. So the qTOF path still drops out of the CMD directory crawl, and the new MetaDraw `.baf` branch stays unreachable. The only commit since his comment is the 15 August merge from master, which doesn't touch that file, so I think this is simply still outstanding rather than something you've already dealt with.

It looks like a one-line deletion. His suggested guard seems worth taking alongside it — `Assert.That(GlobalVariables.AcceptedSpectraFormats, Does.Contain(".baf"));` — because nothing in the suite as it stands would catch a regression here: the Bruker `TestCmd` tests reach the `.d` through the folder-level check, which never consults this list, and `Test.csproj` doesn't reference `GUI`.

For what it's worth the rest of it reads well, and qTOF support is a genuinely useful thing to have — it would be a shame for the whole change to sit on one line. I'm not the maintainer of record here, so please take this as a data point rather than a gate; nbollis has already approved.

nbollis (approved):

MetaMorpheus/EngineLayer/GlobalVariables.cs
trishorts: Fixed — resolving.

`AcceptedSpectraFormats` now has exactly one assignment repo-wide (`GlobalVariables.cs:82`), and it carries `.baf` along with `.tsf`/`.tsf_bin`:

```csharp
AcceptedSpectraFormats = new List<string> { ".raw", ".mzml", ".mgf", ".msalign", ".baf", ".tdf", ".tdf_bin", ".tsf", ".tsf_bin", ".d" };
```

The duplicate second assignment that was silently winning is gone, so every consumer that reads this list — the CMD directory crawl, MetaDraw's gate and its file-dialog filter — now sees `.baf`. The added comment about keeping the list lower-case is a good touch; that was the other half of the trap.

trishorts (approved):
Approving. Verified locally at `1ad2c22`: Test project builds clean (0 errors) and the 42 Bruker/timsTOF/qTOF tests pass.

Collapsing the four hand-rolled `.d` validity checks into `BrukerDataDirectory`, delegating classification to mzLib's `ParseFileType`, is the right call — that divergence was #2722, now linked to close on merge.

Two incidental fixes worth naming since neither is in the PR description:

- the MetaDraw drag-drop check was `Regex.IsMatch(path, ".d$")` — unescaped dot, so it matched any folder ending in *any character* followed by `d`. Now a real `EndsWith(".d")`.
- the GUI file-dialog filter is derived from `AcceptedSpectraFormats` instead of a hardcoded duplicate, so a newly supported format can't silently go missing from the picker.

Nit, no action needed: `GlobalVariables.cs` line 1 lost its UTF-8 BOM.

Still gated on the mzLib release you mentioned; 1.0.585 is pinned.

RayMSMS (approved):

PR #2688, opened 2026-07-16
Don't fail the search when Chronologer's native libraries are missing

Alexander-Sol:
If the TorchSharp native libraries aren't present, constructing the Chronologer retention time predictor throws a DllNotFoundException that escapes FdrAnalysisEngine.GetRTPredictor. 

Native libraries are sometimes missed by the installer for unknown reasons, this can cause crashes that come at the end of searches, which kills a multi-hour run and loses all results. 

GetRTPredictor already treats null as a supported outcome (it returns null for non-standard searches and for unrecognized predictor names), and PepAnalysisEngine falls back to SSRCalc3 when it's handed a null predictor. So catch the DllNotFoundException, warn the user that PEP will be less accurate, and return null.


MetaMorpheus/EngineLayer/MetaMorpheusEngine.cs
```diff
@@ -359,6 +359,15 @@ protected void Warn(string v)
             WarnHandler?.Invoke(this, new StringEventArgs(v, NestedIds));
         }
 
+        /// <summary>
+        /// Static counterpart to <see cref="Warn(string)"/>, for warnings raised from static contexts,
+        /// where no engine instance (and therefore no nested id) is available
+        /// </summary>
+        protected static void WarnStatic(string v)
+        {
+            WarnHandler?.Invoke(null, new StringEventArgs(v, null));
```
trishorts, line 368: `WarnStatic` is new in this PR and currently has no test coverage. Here's a small NUnit test that exercises it — it verifies the static-warning contract the Chronologer fallback relies on (message delivered to `WarnHandler` subscribers, with a **null** `NestedIDs` and null sender). Drop it into `Test/AnalysisEngineTest.cs` (verified passing locally, Release/coverlet):

```csharp
/// <summary>
/// WarnStatic is the static counterpart to Warn(string), used where no engine instance (and therefore
/// no nested id) is available. It must deliver the message to subscribers with a null NestedIDs.
/// </summary>
[Test]
public static void WarnStatic_WithSubscriber_DeliversMessageAndNullNestedIds()
{
    MethodInfo warnStatic = typeof(MetaMorpheusEngine).GetMethod("WarnStatic",
        BindingFlags.NonPublic | BindingFlags.Static);
    Assert.That(warnStatic, Is.Not.Null, "Expected protected static WarnStatic method.");

    StringEventArgs received = null;
    object senderSeen = new object();
    EventHandler<StringEventArgs> handler = (o, e) => { received = e; senderSeen = o; };
    MetaMorpheusEngine.WarnHandler += handler;
    try
    {
        warnStatic.Invoke(null, new object[] { "native library warning" });

        Assert.That(received, Is.Not.Null);
        Assert.That(received.S, Is.EqualTo("native library warning"));
        Assert.That(received.NestedIDs, Is.Null, "Static warnings carry no nested id.");
        Assert.That(senderSeen, Is.Null, "Static warnings have no engine instance as sender.");
    }
    finally
    {
        MetaMorpheusEngine.WarnHandler -= handler;
    }
}
```

Requires `using System.Reflection;` in the test file.

Note on the sibling `GetChronologer` `DllNotFoundException` catch — I tried to add a regression test for it but it can't be unit-tested as written: the `catch` only fires when `new ChronologerRetentionTimePredictor()` throws (TorchSharp native libs present in CI, so it never does), and .NET 8 forbids swapping the `static readonly _chronologerInstance` `Lazy<>` via reflection (`FieldAccessException: Cannot set initonly static field`). Covering that path would need a small production seam (e.g. an injectable predictor factory / `internal` setter).

MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
```diff
@@ -423,13 +423,34 @@ private static IRetentionTimePredictor GetRTPredictor(string searchType, List<(s
...
+        /// PEP falls back to the default retention time predictor. A missing predictor should cost PEP accuracy,
+        /// not the entire search.
+        /// </summary>
+        private static IRetentionTimePredictor GetChronologer()
+        {
+            try
+            {
+                return _chronologerInstance.Value;
+            }
+            catch (DllNotFoundException e)
```
trishorts, line 445: I don't think this `catch (DllNotFoundException)` will actually fire in the case it's meant for. `_chronologerInstance.Value` constructs `Chronologer`, and its instance field initializers (`torch.nn.Embedding(...)`, `torch.nn.Conv1d(...)`, etc.) are the first calls into TorchSharp. That first touch runs `TorchSharp.torch`'s static initializer, and when the native libraries are missing the CLR wraps the `DllNotFoundException` in a `TypeInitializationException` — the classic:

```
System.TypeInitializationException: The type initializer for 'TorchSharp.torch' threw an exception.
 ---> System.DllNotFoundException: Unable to load DLL 'LibTorchSharp' or one of its dependencies...
```

A `catch (DllNotFoundException)` won't match that wrapper, so the exception escapes `GetChronologer` (the `Lazy` caches and re-throws it) and we still crash the search at the FDR/PEP stage — exactly what this PR is meant to prevent.

Suggest broadening the catch to unwrap it, e.g.:

```csharp
catch (Exception e) when (e is DllNotFoundException || e.InnerException is DllNotFoundException)
```

and pull the message from the inner exception when present. Worth confirming against the actual failure you reproduced — if it surfaced as a bare `DllNotFoundException` then the narrow catch is fine, but the `Lazy` + static-initializer path makes the wrapped form the likely one in the field.

MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
```diff
@@ -423,13 +423,34 @@ private static IRetentionTimePredictor GetRTPredictor(string searchType, List<(s
...
             };
         }
 
+        /// <summary>
+        /// Chronologer runs on TorchSharp, whose native libraries aren't present in every installation.
+        /// When they're missing, we return null (as GetRTPredictor already does for unrecognized predictors) so that
+        /// PEP falls back to the default retention time predictor. A missing predictor should cost PEP accuracy,
+        /// not the entire search.
+        /// </summary>
+        private static IRetentionTimePredictor GetChronologer()
```
trishorts, line 439: **Test coverage (critical path):** this catch→`null` fallback is the whole deliverable of the PR, but nothing exercises it, so a future refactor of `GetChronologer`/`GetRTPredictor` could silently reintroduce the escaping exception. The path can't be forced today (private static method + a private static `Lazy` with no injection point, and CI has the native libs present), so covering it needs a tiny seam. `EngineLayer` already has `[assembly: InternalsVisibleTo("Test")]`, so make the construction swappable and the method `internal`:

```csharp
// FdrAnalysisEngine.cs — test seam so a unit test can simulate the missing-native-lib failure
internal static Func<IRetentionTimePredictor> ChronologerFactory { get; set; } =
    () => _chronologerInstance.Value;

internal static IRetentionTimePredictor GetChronologer()   // internal so Test can drive it
{
    try
    {
        return ChronologerFactory();
    }
    catch (Exception e) when (e is DllNotFoundException || e.InnerException is DllNotFoundException)
    {
        WarnStatic("Chronologer retention time prediction was skipped ...");
        return null;
    }
}
```

Then this NUnit test (e.g. in `Test/FdrTest.cs`) — no reflection needed thanks to `InternalsVisibleTo`:

```csharp
[Test]
[TestCase(false)] // bare DllNotFoundException
[TestCase(true)]  // wrapped in TypeInitializationException — the real TorchSharp shape
public static void GetChronologer_NativeLibrariesMissing_ReturnsNull_DoesNotThrow(bool wrapped)
{
    var original = FdrAnalysisEngine.ChronologerFactory;
    try
    {
        FdrAnalysisEngine.ChronologerFactory = () =>
        {
            var dll = new DllNotFoundException("Unable to load DLL 'LibTorchSharp'");
            throw wrapped ? new TypeInitializationException("TorchSharp.torch", dll) : (Exception)dll;
        };

        IRetentionTimePredictor result = null;
        Assert.DoesNotThrow(() => result = FdrAnalysisEngine.GetChronologer());
        Assert.That(result, Is.Null);
    }
    finally
    {
        FdrAnalysisEngine.ChronologerFactory = original;   // don't leak the seam into other tests
    }
}
```

Heads-up: the `[TestCase(true)]` (wrapped) case will **fail against the current narrow `catch (DllNotFoundException)`** — that's intentional, it's the same gap I flagged on the catch line above. Broaden the catch as suggested there and both cases go green, and the fallback is pinned against regression.

MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
```diff
@@ -423,13 +423,34 @@ private static IRetentionTimePredictor GetRTPredictor(string searchType, List<(s
...
+        {
+            try
+            {
+                return _chronologerInstance.Value;
+            }
+            catch (DllNotFoundException e)
+            {
+                WarnStatic("Chronologer retention time prediction was skipped because its native libraries could not be loaded (" + e.Message.Trim() +
```
trishorts, line 447: **Minor:** because `_chronologerInstance` is a `Lazy<>` in `ExecutionAndPublication` mode, a factory exception is cached and re-thrown on *every* subsequent `.Value` access. So once the native libs are missing, every `GetChronologer()` call re-enters this catch and warns again. `GetRTPredictor` is called inside the `foreach (var proteasePsms in psmsGroupedByProtease)` loop in `DoFalseDiscoveryRateAnalysis`, so a multi-protease search surfaces the identical "native libraries could not be loaded" warning once per protease group rather than once.

Functionally harmless (the `null` fallback still works each time), but it reads in the log like several distinct failures. Could guard with a static "already warned" flag set inside the catch so the message is surfaced a single time per process regardless of how many times `GetChronologer` is invoked.

trishorts:
Thanks @Alexander-Sol — the design here is right: reusing the existing `null`-predictor → SSRCalc3 fallback so a missing native lib costs PEP accuracy instead of the whole search is exactly the correct instinct, and it's nicely scoped. I left a few inline comments; one is worth a look before merge:

**Should fix**
- The `catch (DllNotFoundException)` probably won't fire in the case it's meant for. `Chronologer`'s field initializers touch TorchSharp first, which runs `TorchSharp.torch`'s static initializer, so a missing native lib comes back as a `TypeInitializationException` wrapping the `DllNotFoundException` — the narrow catch misses it and the search still crashes. Suggested broadening it to unwrap the inner exception. (This is the one that actually threatens the PR's goal.)

**Nice to have**
- The warning re-emits once per protease group, because the `Lazy` caches the thrown exception and every `GetChronologer()` call re-enters the catch. A static "already warned" flag would keep it to one message per run.
- No test pins the fallback path (the PR's whole deliverable). I sketched a small `internal` seam + a parameterized NUnit test in a comment — the wrapped-exception case doubles as a regression test for the catch-broadening above.

None of it is blocking beyond the first item. Happy to help wire up the test if useful.

MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
Alexander-Sol: This code didn't work for me (both tests failed) but I refactored to use reflection to access/set the _chronologerInstance directly

nbollis (dismissed):

RayMSMS (approved):

trishorts (approved):

PR #2689, opened 2026-07-16
Install TorchSharp/libtorch .dll depenencies so Chronologer RT prediction works

Alexander-Sol:
Product.wxs uses DoNotHarvest, so every shipped file must be listed by hand. The managed TorchSharp.dll was listed but its native dependencies never were, so any MSI-installed MetaMorpheus threw System.DllNotFoundException: Unable to load DLL 'LibTorchSharp' when Chronologer retention time prediction ran during PEP. The installed CMD.deps.json already resolves these to runtimes/win-x64/native.

Adds the 12 missing natives to the WinNativeMathComponents group: LibTorchSharp, torch_cpu, torch, torch_global_deps, c10, fbgemm, asmjit, uv, libiomp5md and libiompstubs5md for the libtorch stack, plus libSkiaSharp (the native pair for the already-shipped managed SkiaSharp.dll) and sni.dll.

Tradeoff: this grows the MSI by roughly 250 MB, almost all of it torch_cpu.dll at ~238 MB. **This is a 10x increase in the size of the installer**.

### DIFF
--- a/MetaMorpheus/MetaMorpheusSetup/Product.wxs
+++ b/MetaMorpheus/MetaMorpheusSetup/Product.wxs
@@ -499,6 +499,46 @@
 	  <Component Id="src_SQLite.Interop.dll" Guid="A918BB3C-0720-469E-900B-27FA93B7BB86">
 		<File Id="src_SQLite.Interop.dll" Name="SQLite.Interop.dll" Source="$(var.GUI_TargetDir)runtimes\win-x64\native\SQLite.Interop.dll" />
 	  </Component>
+      <!--TorchSharp/libtorch natives required by Chronologer retention time prediction during PEP.-->
+      <!--The managed TorchSharp.dll P/Invokes LibTorchSharp, which in turn pulls in the libtorch stack.-->
+      <!--CMD.deps.json probes runtimes\win-x64\native, so these must land here or PEP throws DllNotFoundException.-->
+      <Component Id="src_LibTorchSharp.dll" Guid="9BA2F494-3C28-4276-A8FC-B14399E4BDE7">
+        <File Id="src_LibTorchSharp.dll" Name="LibTorchSharp.dll" Source="$(var.GUI_TargetDir)runtimes\win-x64\native\LibTorchSharp.dll" />
+      </Component>
+      <Component Id="src_torch_cpu.dll" Guid="757479C9-7B0B-4873-802B-0D2BC9AA9379">
+        <File Id="src_torch_cpu.dll" Name="torch_cpu.dll" Source="$(var.GUI_TargetDir)runtimes\win-x64\native\torch_cpu.dll" />
+      </Component>
+      <Component Id="src_torch.dll" Guid="C1C621CB-87FD-41B8-A367-D540C8EB5CF6">
+        <File Id="src_torch.dll" Name="torch.dll" Source="$(var.GUI_TargetDir)runtimes\win-x64\native\torch.dll" />
+      </Component>
+      <Component Id="src_torch_global_deps.dll" Guid="F077F9E2-D56E-460E-BAAA-3762E4B0C37A">
+        <File Id="src_torch_global_deps.dll" Name="torch_global_deps.dll" Source="$(var.GUI_TargetDir)runtimes\win-x64\native\torch_global_deps.dll" />
+      </Component>
+      <Component Id="src_c10.dll" Guid="BDA4715C-9201-4ECA-91FD-DD717D80FEDC">
+        <File Id="src_c10.dll" Name="c10.dll" Source="$(var.GUI_TargetDir)runtimes\win-x64\native\c10.dll" />
+      </Component>
+      <Component Id="src_fbgemm.dll" Guid="B93DD689-FB38-4E76-A32F-35D9220530ED">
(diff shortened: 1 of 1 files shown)


nbollis (approved):

MetaMorpheus/MetaMorpheusSetup/Product.wxs
```diff
@@ -506,6 +506,46 @@
 	  <Component Id="src_SQLite.Interop.dll" Guid="A918BB3C-0720-469E-900B-27FA93B7BB86">
 		<File Id="src_SQLite.Interop.dll" Name="SQLite.Interop.dll" Source="$(var.GUI_TargetDir)runtimes\win-x64\native\SQLite.Interop.dll" />
 	  </Component>
+      <!--TorchSharp/libtorch natives required by Chronologer retention time prediction during PEP.-->
+      <!--The managed TorchSharp.dll P/Invokes LibTorchSharp, which in turn pulls in the libtorch stack.-->
+      <!--CMD.deps.json probes runtimes\win-x64\native, so these must land here or PEP throws DllNotFoundException.-->
+      <Component Id="src_LibTorchSharp.dll" Guid="9BA2F494-3C28-4276-A8FC-B14399E4BDE7">
```
trishorts, line 512: This is the right fix, and the set looks complete for the current TorchSharp / libtorch-cpu pin — these 10 are exactly the standard win-x64 libtorch native set plus `LibTorchSharp`.

The thing worth flagging is the future: with `DoNotHarvest=True` this list is hand-maintained, and the failure mode is asymmetric. WiX errors at build if a *listed* `Source` goes missing (safe — you'd notice), but it silently ships an incomplete MSI if an upstream bump *adds* a required native — nothing references the new file, the build stays green, and users hit the exact `DllNotFoundException` this PR is fixing. That's the same silent-regression trap, one version bump later.

A lightweight guard would close it, e.g. either:
- a small build/CI check that diffs the filenames under `$(GUI_TargetDir)runtimes\win-x64\native\` against the components listed here and fails on any unlisted native, or
- harvesting just this one folder (heat) so the component set is generated from the build output instead of transcribed by hand.

Not blocking — the current closure is correct — just so a TorchSharp bump can't quietly reintroduce the crash.

MetaMorpheus/MetaMorpheusSetup/Product.wxs
```diff
@@ -506,6 +506,46 @@
 	  <Component Id="src_SQLite.Interop.dll" Guid="A918BB3C-0720-469E-900B-27FA93B7BB86">
 		<File Id="src_SQLite.Interop.dll" Name="SQLite.Interop.dll" Source="$(var.GUI_TargetDir)runtimes\win-x64\native\SQLite.Interop.dll" />
 	  </Component>
+      <!--TorchSharp/libtorch natives required by Chronologer retention time prediction during PEP.-->
+      <!--The managed TorchSharp.dll P/Invokes LibTorchSharp, which in turn pulls in the libtorch stack.-->
+      <!--CMD.deps.json probes runtimes\win-x64\native, so these must land here or PEP throws DllNotFoundException.-->
```
trishorts, line 511: **Minor:** this comment justifies the location via `CMD.deps.json`, but the GUI is the primary distribution and PEP/Chronologer runs from there too. I think the GUI is already covered — the `Source` here is `$(GUI_TargetDir)runtimes\win-x64\native\`, so these natives come straight from the GUI's own build output, and both `CMD.exe` and `MetaMorpheusGUI.exe` install into `MetaMorpheusInstallFolder` and share this one `runtimes\win-x64\native` subfolder, so `MetaMorpheusGUI.deps.json` should resolve them the same way (same mechanism the existing `SQLite.Interop.dll` already relies on).

Could you confirm a GUI-launched search actually loads the natives, and maybe broaden the comment to "CMD/GUI deps.json"? Just want the primary path explicitly nailed down since that's the whole point of the PR.

trishorts:
Thanks @Alexander-Sol — this is the right companion to #2688: #2688 makes a missing native degrade gracefully, this makes the installer actually ship them so it doesn't have to. I checked the mechanics and they hold up: the components land in `MetaMorpheus\runtimes\win-x64\native\` (the `NativeMathDllFolder_win_x64_native` directory tree), which is exactly where `deps.json` probes; `WinNativeMathComponents` is referenced by `ProductFeature` so they'll actually install; all 12 Component/File Ids are unique; and the 10 libtorch DLLs are the complete standard win-x64 CPU set + `LibTorchSharp`.

Two things, neither blocking:

- **Future drift (the one worth acting on).** With `DoNotHarvest=True` this list is hand-maintained, and the failure mode is asymmetric — a *removed* upstream native breaks the build (safe), but an *added* one ships nothing and silently reintroduces the `DllNotFoundException` a version bump later. Left a note suggesting a small build-time check (diff the `runtimes\win-x64\native` folder against the listed components) or harvesting just that folder.
- **GUI path.** The justifying comment cites `CMD.deps.json`, but the GUI is the primary distribution. I think it's already covered (the natives are sourced from the GUI's own output and both exes share the folder), but worth a quick confirm that a GUI-launched search loads them.

`sni.dll` / `libSkiaSharp.dll` naming is self-verifying — WiX fails the build if those `Source` paths are wrong, so no worry there.

Since it's a WiX-only change there's nothing to unit-test; the real proof is a clean-VM install + a Chronologer RT-prediction run through PEP. Approve once that smoke test is green.

PR #2690, opened 2026-07-16
Library validation now runs recursively, checks libraries in the "runtimes" folder

Alexander-Sol:
The validate_installer_contents job compares build output to the install directory using `Get-ChildItem -Path $folder -Filter *.dll`, which is not recursive. It therefore only ever inspected top-level DLLs, and everything under runtimes/ was invisible to the comparison.

That blind spot let the MSI ship without the TorchSharp/libtorch natives (LibTorchSharp.dll, torch_cpu.dll, torch.dll, c10.dll, ...) and without libSkiaSharp.dll. These are loaded through DllImport at runtime, so an MSI-installed MetaMorpheus throws DllNotFoundException during Chronologer RT prediction while every CI job stayed green.

This adds a second comparison step covering runtimes/win-x64/native, following the existing step's structure: collect from the same build output folders, diff against the install directory in both directions, support whitelists for known-acceptable gaps, and exit 1 with a clear message.

The existing top-level DLL check is untouched; this widens coverage rather than replacing it.

NOTE: this check is expected to FAIL on this branch alone, because the natives genuinely are missing from the installer - that is the check doing its job. The Product.wxs fix that adds them is being developed in parallel on fix/installer-torchsharp-natives. This should merge after, or together with, that branch. Please do not weaken this check to obtain a green build.


.github/workflows/InstallRunAndArtifact.yml
```diff
@@ -273,3 +273,101 @@ jobs:
...
+
+          )
+
+          # Define whitelist for extra natives (install-only)
+          $whitelistedExtraNatives = @(
+
+          )
+
+          # Define which folders to check in the repo (same build outputs as the DLL check above)
+          $foldersToCheck = @(
```
trishorts, line 304: **Minor (maintainability):** this `$foldersToCheck` array is an exact copy of the one in the preceding "Compare built DLLs" step. Each `run:` block is its own PowerShell process so they genuinely can't share a variable — but that leaves two independently-maintained copies of the build-output folder list. If a project is later added/renamed or a TFM moniker changes and only the top-level step gets updated, this native check would silently stop scanning that project's `runtimes\win-x64\native` and still report green, re-opening the blind spot this PR closes.

Not blocking. A "keep in sync with the DLL step above" comment on both arrays would help, or hoist the list into a job-level env var written once in an earlier step so both consume one source.

.github/workflows/InstallRunAndArtifact.yml
```diff
@@ -273,3 +273,101 @@ jobs:
...
+          $buildNatives = @()
+          $buildNativeLocations = @{}
+
+          foreach ($folder in $foldersToCheck) {
+              $nativeFolder = Join-Path $folder $nativeSubPath
+              if (Test-Path $nativeFolder) {
+                  Get-ChildItem -Path $nativeFolder -Filter *.dll | ForEach-Object {
+                      $nativeName = $_.Name
+                      $buildNatives += $nativeName
+                      $buildNativeLocations[$nativeName] = $nativeFolder
```
trishorts, line 322: **Very minor (diagnostic only):** since the same native (e.g. `torch_cpu.dll`) lands in several projects' output, this unconditional assignment leaves `$buildNativeLocations[$name]` pointing at whichever folder was iterated last, so the "`- $name (from $folder)`" line on a failure can name an arbitrary source folder. Doesn't affect pass/fail — purely the diagnostic message. Fine to leave as-is (it mirrors the existing DLL step), but wording it as "representative source" or joining all matching folders would avoid a confusing trace when the check does fire.

trishorts:
Nice — this is the automated drift guard I was hoping for on the #2689 side: it fails CI the moment the installer's `runtimes\win-x64\native` set diverges from what the build actually produces, so a future TorchSharp bump can't silently drop a native and reintroduce the `DllNotFoundException`. The implementation mirrors the existing top-level DLL check closely and even improves on it with the `if ($buildNatives.Count -eq 0) { exit 1 }` guard against a silent pass if the build layout ever moves. Scoping to win-x64 name comparison (rather than a recursive relative-path diff) is the right call, and your comment explaining why — the linux/osx folders are deliberately renamed in `Product.wxs`, so a recursive path diff would report false mismatches — is exactly right.

Two minor things, neither blocking:

- **Duplicated `$foldersToCheck`.** The build-output folder list is now copied verbatim from the DLL step above. Separate `run:` processes can't share it, but the two copies can drift — if a project is added/renamed in only the top-level step, this native check silently narrows while still going green. A "keep in sync" comment on both, or hoisting the list into a job-level env var, closes that.
- **`$buildNativeLocations` last-write-wins** — diagnostic only; the "(from <folder>)" line on a failure can name an arbitrary source folder when a native appears in several outputs. Cosmetic, and it mirrors the existing step, so entirely optional.

The red CI on this branch is the check working as designed (natives genuinely missing until #2689 lands) — not a defect. Approve from me once it's paired with #2689 so the gate goes green for the right reason.

trishorts (dismissed):
Approving — this is a solid drift guard and the implementation is correct (name-only, non-recursive, mirrors + improves the existing DLL check). The two notes I left inline are minor and non-blocking: the duplicated `$foldersToCheck` list (can drift from the DLL step above) and the last-write-wins `$buildNativeLocations` diagnostic (cosmetic). Good to merge once paired with #2689 so the gate goes green for the right reason.

nbollis (approved):

RayMSMS (approved):

PR #2691, opened 2026-07-21
Fixed Typo: "Indicies" --> "Indices"

Alexander-Sol:
Mission critical patch right here.

### DIFF
--- a/MetaMorpheus/EngineLayer/FdrAnalysis/PEPAnalysisEngine.cs
+++ b/MetaMorpheus/EngineLayer/FdrAnalysis/PEPAnalysisEngine.cs
@@ -408,7 +408,7 @@ namespace EngineLayer
 
                         int ambigousPeptidesRemovedinThread = 0;
 
-                        List<int> indiciesOfPeptidesToRemove = new List<int>();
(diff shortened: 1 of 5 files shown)


nbollis (dismissed):

trishorts (dismissed):
@
Approved.

Verified the rename is safe: `GetIndicesOfPeptidesToRemove` is `public static`, but its only call sites repo-wide are `PEPAnalysisEngine.cs:433` and `FdrTest.cs:518` — both updated here. Everything else renamed is a method-local, and nothing renamed is referenced by string or reflection, so there is no breakage a compiler wouldn't catch.

Two non-blocking leftovers if you want them in a follow-up — a `Indicies` grep still hits this file:
- `PEPAnalysisEngine.cs:664` — the method was renamed but its own parameter is still `indiciesOfPeptidesToRemove`.
- `PEPAnalysisEngine.cs:661` — the `<summary>` still reads "find the indicies of...".
@

trishorts (approved):

nbollis (approved):

RayMSMS (approved):

PR #2695, opened 2026-07-28
align GUI most-abundant precursor mass flow

nbollis:
## Motivation
This change aligns the GUI precursor mass shift flow with the search-layer mass-difference acceptor model, so most-abundant precursor matching is selected explicitly through mass-diff acceptor choices instead of being split across deconvolution state and special-case search behavior.

## Summary of Changes
- move precursor mass match mode ownership out of DeconHostViewModel and into MassDifferenceAcceptorSelectionViewModel
- add explicit MostAbundant_* MassDiffAcceptorType values and wire SearchTask.GetMassDiffAcceptor(...) to return the correct most-abundant acceptors by type
- split GUI mass-diff choices into monoisotopic and most-abundant collections, add mode toggle radio buttons, and update Search/GPTMD/Calibration task windows to persist checkbox state correctly
- add/repair tests for the new enum extension, view-model mode switching behavior, collection partitioning, most-abundant labels/missed-mono counts, and search acceptor construction

## Testing Notes
- dotnet test MetaMorpheus/Test/Test.csproj --configuration Release --no-restore --filter MostAbundantMassTest|MassDiffAcceptorVmTest|MassDiffAcceptorTypeTests
- dotnet test MetaMorpheus/Test/Test.csproj --configuration Release --no-restore --filter DeconHostViewModelTest
- verified the previous outdated MostAbundantMassTest.SearchTask_GetMassDiffAcceptor_ReturnsMostAbundantWhenModeSet expectation against the new implementation and replaced it with explicit MostAbundant_* acceptor coverage

## Migration Steps
- no external migration required
- GUI users will now choose monoisotopic vs most-abundant precursor matching through the mass-diff acceptor control rather than deconvolution view-model state


MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
```diff
@@ -94,6 +89,15 @@ public static MassDiffAcceptor GetMassDiffAcceptor(Tolerance precursorMassTolera
                         },
                         precursorMassTolerance);
 
+                case MassDiffAcceptorType.MostAbundant_Exact:
```
trishorts, line 92: **Critical** — `GetNumNotches` (`SearchTask.cs:545`) was not extended alongside `GetMassDiffAcceptor`. Its switch still covers only the eight legacy types and falls through to:

```csharp
default: throw new MetaMorpheusException("Unknown mass difference acceptor type");
```

`RunSpecific` calls it at `SearchTask.cs:503` — **after** the per-file search loop has finished. So a search configured with any of the three new `MostAbundant_*` types runs the entire engine over every spectra file and *then* throws, before `PostSearchAnalysisParameters` is populated. No results are written, a long search loses all of its compute, and the message gives the user no indication that the acceptor selection is the cause.

Counts consistent with the `maxApexOffsetNeutrons` argument passed just above: `MostAbundant_Exact` → 1, `MostAbundant_PlusMinusOne` → 3, `MostAbundant_PlusMinusTwo` → 5.

Better still: delete the parallel hardcoded switch and read `NumNotches` off the acceptor `GetMassDiffAcceptor` already returns. The two can then never drift again, and it removes a second place where the same knowledge lives.

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
```diff
@@ -40,14 +40,9 @@ public static MassDiffAcceptor GetMassDiffAcceptor(Tolerance precursorMassTolera
...
         {
-            // Most-abundant mode matches candidates by their theoretical most-abundant isotopic peak,
-            // which directly solves the off-by-N problem. It replaces (rather than augments) the
-            // missed-monoisotopic notches, so it overrides the MassDiffAcceptorType selection.
-            if (precursorMassMatchMode == PrecursorMassMatchMode.MostAbundant)
-            {
-                return new MostAbundantMassDiffAcceptor("mostAbundant", precursorMassTolerance, averagineModel ?? new Averagine(),
-                    expectedIsotopeSpacing: expectedIsotopeSpacing);
-            }
+            averagineModel ??= GlobalVariables.AnalyteType == AnalyteType.Oligo
```
trishorts, line 43: **Critical** — after this change `precursorMassMatchMode` is still a parameter (`:39`) and is still passed by `RunSpecific` (`:234`), but nothing in the body reads it any more. The *observed* side is untouched: `GetMs2Scans(...).OrderBy(b => b.GetPrecursorMassForSearch(combinedParams))` still derives its mass from `CommonParameters.PrecursorMassMatchMode`, and the comment just above it still states the invariant — *"ClassicSearchEngine binary-searches that same mass, so the array and the search key must use the same quantity."*

So the theory side now comes from `SearchParameters.MassDiffAcceptorType` while the observed side comes from `CommonParameters.PrecursorMassMatchMode`, and nothing forces them to agree.

Any existing `.toml` or CMD run that pairs a legacy acceptor type with `PrecursorMassMatchMode = "MostAbundant"` — say `MassDiffAcceptorType = "ThreeMM"` — now builds a monoisotopic `DotMassDiffAcceptor` and hands it apex masses. On a 35 kDa proteoform that is roughly a 20 Da offset: PSMs are lost or assigned to the wrong notch, with no exception and no warning. `GetMonoisotopicFallbackWarning` also keys off the old mode, so the one diagnostic that might have surfaced this is suppressed in exactly the mismatched case.

The GUI migrates on open (the view model auto-switches `SelectedType`), so this is invisible from the window and only bites TOML/CLI users — which is where most existing most-abundant configurations live.

Two workable fixes: derive `PrecursorMassMatchMode` from `massDiffAcceptorType.IsMostAbundant()` where `combinedParams` is built, so the sort key, the fallback warning and the acceptor all follow one setting; or validate the pairing at task start and fail fast with a message naming both. Either way the invariant in that retained comment should be asserted in code rather than left to caller discipline.

MetaMorpheus/GUI/Views/MassDiffAcceptorControl.xaml
```diff
@@ -42,11 +44,54 @@
...
+            <StackPanel Grid.Column="0" Grid.Row="0"  Margin="0" >
+
+                <StackPanel Orientation="Horizontal" Margin="0,0,0,4">
+                    <RadioButton Content="Monoisotopic Mass" IsChecked="{Binding UseMonoisotopicMass, Mode=TwoWay}" GroupName="PrecursorMassMatchMode" Margin="0,0,8,0" />
+                    <RadioButton Content="Most Abundant Mass" IsChecked="{Binding UseMostAbundantMass, Mode=TwoWay}" GroupName="PrecursorMassMatchMode" />
+                </StackPanel>
+
+                <!-- Mono Mass Diff Acceptor Types -->
+                <ListBox ItemsSource="{Binding MonoMassDiffAcceptorTypes}"
+                             SelectedItem="{Binding SelectedType, Mode=TwoWay}"
```
trishorts, line 56: **Critical** — both ListBoxes bind `SelectedItem="{Binding SelectedType, Mode=TwoWay}"` while their item sources are disjoint by construction (`MonoMassDiffAcceptorTypes` here, `MostAbundantMassDiffAcceptorTypes` at `:90`). Only `Visibility` differs — but a `Collapsed` ListBox is still loaded, still has a live `ItemsSource`, and still evaluates its `SelectedItem` binding.

When `SelectedType` holds a value that is not present in a given list, `Selector` coerces its own selection to `null`, and because the binding is `TwoWay` that `null` is written straight back into `SelectedType`. Since the two collections are disjoint, one of the two ListBoxes is *always* in that state.

Two consequences:

1. `SelectedType` can be nulled at control load and on every mode toggle — the radio buttons show nothing selected, and the task is saved with a null/default acceptor rather than the one the user clicked.
2. The `SelectedType` setter dereferences `_selectedType.Type` with no null guard, so that write-back can also throw a `NullReferenceException` from a view-model setter on the UI thread, taking the window down and losing everything entered in the dialog.

Options: expose `SelectedMonoType` / `SelectedMostAbundantType` separately with a computed read-only `SelectedType` returning whichever matches the current mode; or keep a single ListBox over one `ActiveMassDiffAcceptorTypes` collection that the view model swaps on mode change. If two lists must stay, the inactive one needs `Mode=OneWay` (and/or a nulled `ItemsSource` while hidden) so it cannot write back — and the setter wants a null guard regardless.

trishorts (commented):
Ran a local review pass over this branch. Three critical issues below — the first one crashes every search that uses the new acceptor types, so it's worth confirming before anything else. Happy to talk through any of them.

MetaMorpheus/GUI/Views/MassDiffAcceptorControl.xaml
nbollis: Agreed — implemented the single-list option. The two disjoint ListBoxes are now one ListBox bound to a new ActiveMassDiffAcceptorTypes collection … the inactive collection is never bound, so there is no live SelectedItem binding left to coerce a null back into SelectedType. The SelectedType setter also now ignores a null write … Added ActiveMassDiffAcceptorTypes_FollowsUseMostAbundantMass and SelectedType_NullAssignment_IsIgnored.

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
nbollis: Agreed — deleted the hardcoded switch rather than extending it … numNotches is now captured from massDiffAcceptor.NumNotches … a most-abundant search no longer runs the entire engine over every file and then throws … MassDiffAceptorTest now asserts NumNotches 1/3/5

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
nbollis: Fixed using your first option. SearchTask.RunSpecific now derives the observed-side setting from the theory side … so a TOML/CLI run pairing a legacy acceptor with PrecursorMassMatchMode = "MostAbundant" is normalized to a consistent monoisotopic search instead of silently searching apex masses against monoisotopic theory … SearchTask_EmitsMonoisotopicFallbackWarning_WhenNoApexObserved now sets MassDiffAcceptorType = MostAbundant_Exact

acesnik (commented):
:robot: This is a nice cleanup. Having the precursor match mode live alongside the mass-diff acceptor, rather than split between deconvolution state and a special case buried at the top of `GetMassDiffAcceptor`, makes the whole flow much easier to follow — and I particularly like that `numNotches` now comes off the acceptor you actually constructed instead of a parallel switch that had to be kept in step by hand. The old `GetNumNotches` would have needed three new cases for this PR and thrown if anyone missed one, so that's a genuine class of bug retired rather than just moved. Appending the new enum members at the end instead of slotting them in alphabetically was the right instinct too: existing TOMLs keep round-tripping.

Two things I noticed while reading, both small. `GetMassDiffAcceptor` no longer reads its `precursorMassMatchMode` parameter now that the `MostAbundant` branch is gone, but the parameter is still in the signature, so a caller who passes it gets something other than what they asked for with nothing to explain why — `MostAbundantMassTest.cs:112` is the live example. Dropping it from the signature would hand the audit to the compiler, and since nearly all of the ~30 call sites just take the default, it should point at only the handful that meant it. The other is `SearchTask.cs:238`, where assigning `combinedParams.PrecursorMassMatchMode` unconditionally quietly overrides anyone whose TOML sets `PrecursorMassMatchMode = MostAbundant` — the only way to ask for most-abundant matching before this PR — so those runs come back monoisotopic with nothing in the log to say so. Either a warning when the two disagree, or a straight migration onto the new `MostAbundant_*` values, would close that off.

Neither is a blocker from where I'm sitting; happy to see this land once they're sorted.

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
trishorts: Verified fixed at `9da3bfd6`. `GetNumNotches` was deleted rather than extended, and `numNotches` now comes off the acceptor that was actually constructed (`SearchTask.cs:246`), so the two can no longer drift apart. `SearchTaskTest.cs:55-61` pins 1/3/5 for the three new types. Resolving.

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
trishorts: Verified fixed at `9da3bfd6`. `RunSpecific` now derives `combinedParams.PrecursorMassMatchMode` from `SearchParameters.MassDiffAcceptorType.IsMostAbundant()` (`SearchTask.cs:238-240`), so the sort key, the fallback warning and the acceptor all follow one setting and the invariant in that retained comment is asserted in code. Resolving.

The TOML half of this is still open, but as acesnik's point rather than mine: normalizing an existing `PrecursorMassMatchMode = MostAbundant` run *down* to monoisotopic silently drops the search it asked for. Tracking that in his review thread instead of reopening here.

MetaMorpheus/GUI/Views/MassDiffAcceptorControl.xaml
trishorts: Verified fixed at `9da3bfd6`. The two disjoint ListBoxes are now one, bound to `ActiveMassDiffAcceptorTypes` (`MassDiffAcceptorControl.xaml:53-54`), so the inactive collection has no live `SelectedItem` binding left to coerce a null back into `SelectedType`. The setter also ignores a null write (`MassDifferenceAcceptorSelectionViewModel.cs:165-168`), which is what covers the ordering inside the `PrecursorMassMatchMode` setter — `SelectedType` is reassigned before `ActiveMassDiffAcceptorTypes` is re-raised, so the ListBox does briefly hold an item that is not in its current source. Pinned by `MassDiffAcceptorVmTest.cs:496` and `:510`. Resolving.

trishorts:
@nbollis @acesnik — I put the two follow-ups from Anthony's 2026-08-15 review into a PR against this branch rather than pushing to it: **https://github.com/nbollis/MetaMorpheus/pull/12** (`trishorts:fix/mm2695-review-followups` → `nbollis:MostAbundantUI`). Nic, merge it if you like it or rewrite it your own way — one commit on top of `9da3bfd6`, applies clean.

1. **`GetMassDiffAcceptor` no longer takes `PrecursorMassMatchMode`.** Handing the audit to the compiler turned up only one call site that passed it — `SearchTask.cs` itself. The three in `MostAbundantMassTest.cs` use the named `averagineModel:` argument and skip it, so it's a one-line call-site change rather than a sweep.

2. **A legacy `PrecursorMassMatchMode = MostAbundant` run migrates instead of silently downgrading**, and warns. Worth flagging the target: the old override omitted `maxApexOffsetNeutrons`, so it took `MostAbundantMassDiffAcceptor.DefaultMaxApexOffsetNeutrons` = **2**, which makes `MostAbundant_PlusMinusTwo` the behavior-preserving choice. `MostAbundant_Exact` is the one the name suggests and would have quietly narrowed every existing most-abundant toml from 5 notches to 1. There's a test pinning the equivalence so the target has to move if that default ever does.

One gap left open deliberately: `RunTask` writes the settings toml before it calls `RunSpecific`, so that file still records the original pairing and the warning is the only trace of the migration. Closing that means restructuring `RunTask`, which seemed like more than this review warranted here.

Note there are no checks on PR 12 — a PR between two forks doesn't run this repo's workflows. Locally the solution builds clean and 127 tests pass across the search/acceptor/GUI-view-model fixtures, and the migration test was verified red before the fix. Real CI runs here once it's merged into `MostAbundantUI`.

My three threads above are verified against `9da3bfd6` and resolved — nothing outstanding from my side.

nbollis:
Merged in teh changes based upon these recommendations

RayMSMS (dismissed):

zhuoxinshi (dismissed):

acesnik (approved):

PR #2700, opened 2026-08-03
Survive a missing UniProt proteome catalogue at startup

trishorts:
> **Project of origin:** none — standalone (pairs with mzLib #1126, UniProt)

Companion to **smith-chem-wisc/mzLib#1126**, and useful on its own.

## The problem

`GlobalVariables.LoadAvailableProteomes` read the proteome catalogue with no guard:

```csharp
AvailableUniProtProteomes = ProteinDbRetriever.UniprotProteomesList(
    Path.Combine(DataDir, @"Proteomes", @"availableUniProtProteomes.txt.gz"));
```

`SetUpGlobalVariables` is called from the `MainWindow` constructor and from `Program.Main` **outside any try/catch**, so anything that stops that file being read takes the whole program down at launch rather than costing one GUI feature.

That was survivable while mzLib answered a missing file with `null`. It is not once it throws — which is what mzLib#1126 does, stopping `ProteinDbRetriever` from reporting every failure (bad ID, unsupported format, UniProt outage, missing file) as the same `null`.

The file is shipped beside the executable, but `DataDir` does not always resolve there:

- a Program Files install resolves it to `%LOCALAPPDATA%\MetaMorpheus`, which an upgrade can leave incomplete;
- `--customDataDir` pointed at a folder that **already exists** skips `CopyFilesRecursively` entirely (`SetUpDataDirectory` only copies when the directory is absent), so `Proteomes\` is simply not there.

## The fix

The catalogue is a convenience for the "Download UniProt Database" window, not a prerequisite for searching. Failing to read it now degrades to an empty dictionary and a console message.

**Empty rather than null is deliberate, and fixes a second latent bug.** `DownloadUniProtDatabaseWindow` enumerates this property and calls `FirstOrDefault` on it with no null check:

```csharp
foreach (var item in GlobalVariables.AvailableUniProtProteomes)   // :44
GlobalVariables.AvailableUniProtProteomes.FirstOrDefault(...)     // :145
```

so the old `null` was already a `NullReferenceException` waiting for someone to open that window.

The `catch` is deliberately broad: absent, wrong extension, truncated, locked and unreadable all have the same consequence for an optional catalogue, and none is worth refusing to start over.

## Merge order

This is **forward- and backward-compatible** — it handles both the current `null` and the post-#1126 throw — so it can merge before or after the mzLib release, and before the version bump. Merging it first is safest.

## Tests

`TestStartUpSurvivesAMissingProteomeCatalogue` hides the `Proteomes` directory, asserts `SetUpGlobalVariables()` does not throw, and asserts the property is non-null and empty. Verified it **fails without the guard** (`Expected: not null / But was: null`) and passes with it. `TestProteomeCatalogueIsReadWhenPresent` covers the normal path. All 9 tests in the fixture pass.


MetaMorpheus/Test/GlobalVariablesTest.cs
```diff
@@ -8,6 +8,67 @@ namespace Test
...
+        /// it may simply be absent. This hides it to reproduce that.
+        /// </summary>
+        [Test]
+        public static void TestStartUpSurvivesAMissingProteomeCatalogue()
+        {
+            string proteomesDirectory = Path.Combine(GlobalVariables.DataDir, "Proteomes");
+            string hiddenDirectory = proteomesDirectory + ".hidden-for-test";
+
+            if (Directory.Exists(hiddenDirectory))
+                Directory.Delete(hiddenDirectory, true);
```
acesnik: This recovery path can destroy the thing it's protecting. If a previous run died between the `Move` below and the `finally` — a crash, a cancelled run, a killed test host — then `Proteomes` is absent and `Proteomes.hidden-for-test` holds the only copy of the real directory. This run then deletes it, sees `wasPresent == false`, and restores nothing.

It's recoverable with a rebuild so it isn't serious, but the fix is about as cheap as fixes get: if the hidden directory exists and the real one doesn't, move it back instead of deleting it.

MetaMorpheus/Test/GlobalVariablesTest.cs
```diff
@@ -8,6 +8,67 @@ namespace Test
...
+        /// searching, so MetaMorpheus must still start without it. SetUpGlobalVariables runs in the
+        /// MainWindow constructor and in Program.Main outside any try/catch, so an exception escaping
+        /// LoadAvailableProteomes is an unhandled crash at launch — and mzLib now reports a missing
+        /// catalogue by throwing rather than returning null (smith-chem-wisc/mzLib#1126).
+        ///
+        /// The file is shipped beside the executable, but DataDir can resolve to %LOCALAPPDATA%\MetaMorpheus
+        /// for a Program Files install, or to a --customDataDir the user already created, and in those cases
+        /// it may simply be absent. This hides it to reproduce that.
+        /// </summary>
+        [Test]
```
acesnik, line 22: Consider `[NonParallelizable]` here. This test mutates process-wide state twice over — it renames a directory every other test can see, and it re-runs `SetUpGlobalVariables()`, which resets every global in the process. That's exactly the situation the four markers in `PsmTsvWriterTests.cs` are annotated for (`:530`, "mutates the process-wide GlobalVariables.AnalyteType").

To be fair, I couldn't find an assembly-level `[Parallelizable]` or a `.runsettings`, and NUnit defaults to no parallelism, so this isn't flaky today. It's cheap insurance and house style rather than a bug.

MetaMorpheus/Test/GlobalVariablesTest.cs
```diff
@@ -8,6 +8,67 @@ namespace Test
...
+            }
+        }
+
+        /// <summary>With the catalogue present it is read, and the human proteome is in it.</summary>
+        [Test]
+        public static void TestProteomeCatalogueIsReadWhenPresent()
+        {
+            Assert.That(GlobalVariables.AvailableUniProtProteomes, Is.Not.Null);
+
+            if (Directory.Exists(Path.Combine(GlobalVariables.DataDir, "Proteomes")))
```
acesnik: This guard makes the happy-path test pass vacuously whenever the directory is missing — including if the sibling test above leaves it that way. Since the whole point of this one is "the catalogue is read when it's present", I'd rather it assert the directory exists and fail loudly if the fixture is broken, so a green run actually means something.

MetaMorpheus/EngineLayer/GlobalVariables.cs
```diff
@@ -532,9 +532,44 @@ private static void LoadDissociationTypes()
...
+                // smith-chem-wisc/mzLib#1126); either way the result is the same empty dictionary here.
+                AvailableUniProtProteomes = ProteinDbRetriever.UniprotProteomesList(proteomeListPath)
+                                            ?? new Dictionary<string, string>();
+            }
+            catch (Exception e)
+            {
+                // Deliberately broad: every way reading a local file can fail — absent, wrong extension,
+                // truncated, locked, unreadable — has the same consequence for this optional catalogue, and
+                // none of them is worth refusing to start over.
+                Console.WriteLine($"Could not read the list of available UniProt proteomes from '{proteomeListPath}'. " +
```
acesnik, line 577: Minor: in the WPF build there's no console attached, so this message goes nowhere and a GUI user gets an empty proteome list with no explanation. The degraded state is fairly self-evident in that window, so I'm not sure it's worth wiring into a real warning channel — just noting that the diagnostic only reaches CLI users.

Also cosmetic, one level up: `AvailableUniProtProteomes` is assigned an empty dictionary at the top of the method and then again via the `??`, so one of the two is redundant.

acesnik (dismissed):
:robot: Good fix, and the crash path is real — I checked that `LoadAvailableProteomes` was the only unguarded reader in `SetUpGlobalVariables` that mzLib#1126 converts from a `null` return into a throw, and `SetUpGlobalVariables` really does run from both the `MainWindow` constructor and `Program.Main` with nothing catching above it. Degrading to an empty dictionary rather than `null` is the right call, and it does quietly close the latent `NullReferenceException` in `DownloadUniProtDatabaseWindow` at `:44` and `:145`. Writing it to tolerate both the old `null` and the new throw so merge order doesn't matter is the right instinct too.

Approving — my notes are all on the tests rather than the fix, and none of them blocks. The first is the one I'd actually like you to take.

MetaMorpheus/Test/GlobalVariablesTest.cs
trishorts: Taken, and thanks — you're right, and it's worse than "recoverable with a rebuild": the run that destroys the catalogue still passes green, so nothing tells you it happened.

I staged that state deliberately (moved `Proteomes` to `Proteomes.hidden-for-test`, left no real directory) and ran the old test against it. Both directories were gone afterwards and the test reported Passed. With the fix the catalogue comes back and the test still passes.

Fixed as you suggested — if the hidden directory exists and the real one doesn't, it's moved back rather than deleted.

MetaMorpheus/Test/GlobalVariablesTest.cs
trishorts: Added `[NonParallelizable]` to both tests in the fixture — the new one for the reasons you give, and `TestProteomeCatalogueIsReadWhenPresent` because it reads the global the other one deliberately empties, so they must not overlap.

Agreed it isn't flaky today; taking it as house style before someone turns parallelism on.

MetaMorpheus/Test/GlobalVariablesTest.cs
trishorts: Agreed — a conditional assertion is a test that reports success for not running. It now asserts the directory exists, with a message naming the fixture problem, before checking the contents.

That's safe here: `availableUniProtProteomes.txt.gz` is copied to the output directory with `CopyToOutputDirectory=Always` (EngineLayer.csproj:112) and `TestCustomDataDirectory` puts `DataDir` back when it's done, so the only ways the directory can be missing are the two you'd want to hear about — a broken build output, or the sibling test above failing to restore it.

MetaMorpheus/EngineLayer/GlobalVariables.cs
trishorts: The redundant assignment is gone — the empty dictionary is now created only in the `catch`, where it's the actual degraded value, and the `??` covers the pre-#1126 `null` return.

On the console message reaching nobody in the WPF build: I went looking for a channel to use instead and didn't find one worth using. The only startup diagnostic both front-ends drain is `GlobalVariables.ErrorsReadingMods`, and it isn't a general warning channel — the GUI drains it once in the `MainWindow` constructor (`MainWindow.xaml.cs:87`), but CMD only drains it inside the loop over `.xml` databases (`Program.cs:187`), so a CLI run with no XML database never prints it at all. Putting a proteome-catalogue failure in a list named for modification-reading errors would also break the counting assertions in `MyTaskTest` that treat its length as a mod-error count.

So it stays on the console, which at least reaches CLI users, and the GUI user sees an empty list in the one window that consumes it. If a real startup-warning channel ever shows up, this is a one-line move.

acesnik (approved):
🤖 All four taken. The first one I verified end to end: I staged the crashed-run state — catalogue present only under `Proteomes.hidden-for-test`, no real directory — and ran both versions. The old test reported Passed and left both directories gone; the current one restores the catalogue and cleans up the hidden copy. Matches your description exactly, including the green-while-destroying part.

The other three check out: `[NonParallelizable]` on both, the happy-path directory check is a real assertion with a message rather than a condition, and the double assignment is gone with the `??` left covering the pre-#1126 `null`. The fixture is 7/9 on macOS arm64 — the two failures are `TestCustomDataDirectory` and `TestCustomFileExtensionGetter`, both pre-existing Windows path literals and untouched here.

Agreed on leaving the console message where it is. `ErrorsReadingMods` isn't a general warning channel, and CMD never drains it without an XML database.

Issue #2503, opened 2025-04-29
Conda install of metamorpheus fails


erikj540:
**Describing the issue**

Hello. Thanks for your help in advance!

I am on a Mac (with M3 processors) going through the ["Test Conda Installation..."](https://github.com/smith-chem-wisc/MetaMorpheus/wiki/Getting-Started#test-conda-installation-linux-macos-windows) instructions. 
I'm using conda v24.11.3. 
The `conda install -c conda-forge metamorpheus` fails with this output
```
$ conda install -c conda-forge metamorpheus
Channels:
 - conda-forge
 - bioconda
 - defaults
Platform: osx-arm64
Collecting package metadata (repodata.json): done
Solving environment: failed

PackagesNotFoundError: The following packages are not available from current channels:

  - metamorpheus

Current channels:

  - https://conda.anaconda.org/conda-forge
  - https://conda.anaconda.org/bioconda
  - defaults

To search for alternate channels that may provide the conda package you're
looking for, navigate to

    https://anaconda.org

and use the search bar at the top of the page.
```

I tried downloading the first `osx_*` `.conda` file from [here](https://anaconda.org/conda-forge/metamorpheus/files) and installing via `conda install metamorpheus-1.0.8-h6e16a3a_0.conda`. The install appears to work. But then running `metamorpheus -h" produces this error:
```
metamorpheus -h
Unhandled exception. System.IO.FileLoadException: Could not load file or assembly 'CMD, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'.
/Users/erjo3868/miniforge3/envs/metamorpheus/bin/metamorpheus: line 3:  1669 Abort trap: 6           dotnet $DOTNET_TOOLS/metamorpheus/CMD.dll $@
```

Do you know why the `conda install` seems to be broken? Thank you!

trishorts:
hmmmmm. not my area of expertise. i'm gonna reach out to someone that might be able to help. thanks for your patience.

acesnik:
Hi there, I'm seeing a similar error on my side. I think it has something to do with `dotnet` not metamorpheus. Assuming you have also installed .NET 8 without conda (https://dotnet.microsoft.com/en-us/download/dotnet), could you double-check whether this errors out? 
`/usr/local/share/dotnet/dotnet $DOTNET_TOOLS/metamorpheus/CMD.dll -h`

This command works for me presently on mac, but the conda dotnet fails.

[...]

Issue #2504, opened 2025-04-29
.NET installation failure on Mac


erikj540:
I am on a Mac with M3 processors and trying to install MetaMorpheus. 
I wasn't able to do it via `conda` which this [issue tracks](https://github.com/smith-chem-wisc/MetaMorpheus/issues/2503).
I tried to install via .NET and am getting this error which ChatGPT could not fix 😄 .
```
$ dotnet CMD.dll -h
Unhandled exception. System.IO.FileLoadException: Could not load file or assembly 'CMD, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'.
[1]    18276 abort      dotnet CMD.dll -h
```

I get the same error when I try to run the specific commands in the instructions [here](https://github.com/smith-chem-wisc/MetaMorpheus/wiki/Getting-Started#test-installation-via-net-core-dll---linux-macos-windows). 

My `dotnet` version is `8.0.401`.

Thanks for your help!

acesnik:
🤖 Status update.

**Same root cause as #2503, and it is fixed.** Assemblies built with `PlatformTarget=x64` carry AMD64 in their PE header, which .NET on Apple Silicon refuses to load — hence `Could not load file or assembly 'CMD'`. Reproduced verbatim by running the shipped release under a native arm64 .NET 8 runtime, and confirmed to be architecture rather than versioning: the same assembly loads under an x64 runtime. Two layers, both merged: #2701 (MetaMorpheus) and smith-chem-wisc/mzLib#1127 (the mzLib package shipped x64-stamped assemblies too).

**Still needs a release.** The `dotnet CMD.dll` route in the wiki instructions uses the released zip, and every release so far — 1.1.7 included — predates the fix. It needs an mzLib release, then a MetaMorpheus release built against it. MetaMorpheus master carries an arm64 smoke-test check that currently records exactly this remaining gap, failing on `Omics, Version=1.0.583.0` until the mzLib bump lands.

This issue covers the direct `dotnet` route and should close at that MetaMorpheus release. #2503 tracks the conda-forge package, which additionally needs an `osx-arm64` feedstock build (conda-forge/metamorpheus-feedstock#33), so it will close later.

**Related:** #2702 tracks the remaining x64-only paths (Bruker, timsTOF, Thermo `.raw`, IsoDec), which will not work on arm64 even after this ships.

acesnik:
[...]

PR #2701, opened 2026-08-03
Build CMD, EngineLayer and TaskLayer architecture-neutral so the CLI runs on arm64

acesnik:
🤖 Addresses the MetaMorpheus half of #2503 and #2504. Both issues report the same error, on Apple Silicon:

```
System.IO.FileLoadException: Could not load file or assembly
'CMD, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'.
```

## Root cause

`CMD`, `EngineLayer` and `TaskLayer` each set `<PlatformTarget>x64</PlatformTarget>`. That stamps AMD64 into the assembly's PE header, and CoreCLR on arm64 refuses to load such an assembly — the loader surfaces it as the `FileLoadException` above.

Reproduced by running the shipped `MetaMorpheus_CommandLine` release under a native arm64 .NET 8 runtime on an M-series Mac, giving the reporters' error verbatim. Confirmed to be architecture rather than versioning: the same assembly loads under an x64 runtime and fails under arm64, while `AssemblyName.GetAssemblyName` reports the expected version in both cases.

`file` makes the difference visible:

| | header |
|---|---|
| `PlatformTarget=x64` | `PE32+ executable (DLL) x86-64 Mono/.Net assembly` |
| `PlatformTarget` unset | `PE32 executable (DLL) Intel 80386 Mono/.Net assembly` (MSIL, architecture-neutral) |

## The change

Remove `PlatformTarget` from those three projects.

Two things worth knowing for review:

- **It must be left unset, not set to `AnyCPU`.** `Microsoft.ML`'s `_CheckForUnsupportedPlatformTarget` target errors on any explicit `PlatformTarget` that isn't `x64`/`x86`, but deliberately exempts an *empty* `PlatformTarget` for portable .NET Core apps. Setting `AnyCPU` fails the build with "Microsoft.ML currently supports 'x64' and 'x86' processor architectures".
- **x64 users are unaffected.** On .NET Core an architecture-neutral app still runs 64-bit on a 64-bit runtime. Verified by running the resulting build under an x64 runtime.

`GUI` and `Test` still set `PlatformTarget=x64` and are deliberately left alone — they are Windows-only and feed the x64 WiX installer. `GuiFunctions` sets it for `Release` only (there is no `Debug` group), and is likewise untouched. Note `GUI/GUI_h05aw1sq_wpftmp.csproj` is an accidentally-committed WPF temp project that also sets x64; pre-existing and out of scope here, but it will show up if you grep to verify the above.

## This is necessary but not sufficient on its own

The mzLib NuGet package also ships x64-stamped assemblies, because `mzLib.nuspec` packages them out of `bin/x64/Release/`. With only this change, an arm64 run gets past `CMD` and into MetaMorpheus code, then fails on `Omics, Version=1.0.583.0` with the same error shape.

To confirm the diagnosis is complete, I rebuilt mzLib's assemblies architecture-neutral locally and dropped them in. On native arm64 that combination gives:

- `CMD.dll --test` completes the full Calibration → GPTMD → Search pipeline, exit code 0, 89 target PSMs and 47 protein groups at 1% FDR
- the conda recipe's own smoke test, `metamorpheus -g -o .`, succeeds

So closing #2503/#2504 fully also needs an mzLib release built architecture-neutral, and then an `osx-arm64` build in the conda-forge feedstock (which currently has none, for this reason).

Note the remaining x64 files in a published output — `baf2sql_c`, `timsdata`, `isodeclib`, `isogenmass`, `msvcp110`, `svml_dispmd`, `vcomp110`, `ThermoFisher.CommonCore.MassPrecisionEstimator` — are *native* Windows/x64 libraries. They do not block startup or an mzML search, but Bruker/Thermo/IsoDec paths specifically would still not work on arm64.

## Verification

- `dotnet build MetaMorpheus/MetaMorpheus.sln -c UbuntuMac -p:EnableWindowsTargeting=true` — succeeds, 0 errors
- `dotnet publish MetaMorpheus/CMD/CMD.csproj -c Release` — produces `PE32` MSIL assemblies with no command-line override
- resulting build runs correctly under both an x64 and an arm64 .NET 8 runtime

**CI coverage (corrected).** An earlier version of this description claimed existing CI would cover the change. It would not, and that is why the bug survived: `build.yml`'s `macos-latest` leg is arm64 and is green *today* with `PlatformTarget=x64` in place, because it only runs `dotnet build` — cross-compiling an x64-stamped IL assembly on an arm64 host is legal, and the stamp only bites at load time. Every job that executes MetaMorpheus is `windows-latest`.

So this PR adds two steps to `build.yml`:

- **`Assert CLI assemblies are architecture-neutral`** — required, green today. Reads the PE header of the three assemblies via `file` on the Unix legs and fails if any is `x86-64`, naming the project. Guards the regression class without depending on the mzLib fix. Verified in both directions.
- **`Smoke test on arm64`** — executes `CMD.dll --test` on the macos leg, `continue-on-error: true` with a `TODO(#2503)` to become required once an architecture-neutral mzLib is consumed.

The Windows test suite has not been run locally for this change (it needs `Microsoft.WindowsDesktop.App`), but `Test.csproj` keeps `PlatformTarget=x64` and its Thermo `.raw` fixtures are exercised on `windows-latest`, so CI covers the mixed x64-host/neutral-dependency combination.

### DIFF
--- a/MetaMorpheus/CMD/CMD.csproj
+++ b/MetaMorpheus/CMD/CMD.csproj
@@ -10,13 +10,14 @@
     <ApplicationIcon></ApplicationIcon>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
-    <PlatformTarget>x64</PlatformTarget>
-  </PropertyGroup>
-
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
-    <PlatformTarget>x64</PlatformTarget>
-  </PropertyGroup>
+  <!-- PlatformTarget is deliberately left unset so the assembly is architecture-neutral (MSIL) and can be
+       loaded by an arm64 .NET runtime as well as an x64 one. Setting it to x64 stamps AMD64 into the PE
+       header, and CoreCLR on Apple Silicon then refuses the assembly with
+       "FileLoadException: Could not load file or assembly 'CMD'" (issues #2503, #2504).
+       Note that Microsoft.ML's _CheckForUnsupportedPlatformTarget guard rejects an explicit 'AnyCPU' but
+       explicitly permits an empty PlatformTarget for portable .NET Core apps, so this must stay unset
+       rather than being changed to AnyCPU. On .NET Core an architecture-neutral app still runs 64-bit on a
+       64-bit runtime, so nothing is lost on x64. -->
 
   <ItemGroup>
     <PackageReference Include="CommandLineParser" Version="2.9.1" />
--- a/MetaMorpheus/EngineLayer/EngineLayer.csproj
+++ b/MetaMorpheus/EngineLayer/EngineLayer.csproj
@@ -8,13 +8,7 @@
     <DebugSymbols>true</DebugSymbols>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
-    <PlatformTarget>x64</PlatformTarget>
-  </PropertyGroup>
-
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
-    <PlatformTarget>x64</PlatformTarget>
-  </PropertyGroup>
+  <!-- Left architecture-neutral (no PlatformTarget) so the CLI can run on arm64; see CMD.csproj. -->
 
   <ItemGroup>
     <None Remove="Mods\RnaMods.txt" />
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,9 +5,15 @@ name: Build
 
 on:
   push:
-    branches: [ "master" ] 
+    branches: [ "master" ]
   pull_request:
     branches: [ "master" ]
+
+# Mirrors InstallRunAndArtifact.yml so the SDK version and target framework are declared in one place
+# per workflow rather than repeated inline.
+env:
+  DOTNET_VERSION: 8.0.204
+  DOTNET_FRAMEWORK: net8.0
 
 jobs:
   build:
@@ -30,8 +36,46 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@v4
         with:
-          dotnet-version: 8.0.204
+          dotnet-version: ${{ env.DOTNET_VERSION }}
       - name: Restore dependencies
         run: dotnet restore ./MetaMorpheus/MetaMorpheus.sln
       - name: Build
         run: dotnet build --no-restore ./MetaMorpheus/MetaMorpheus.sln --configuration ${{ matrix.configuration }}
+
+      # Guards issues #2503/#2504. An assembly built with PlatformTarget=x64 has AMD64 in its PE header and
+      # an arm64 .NET runtime refuses to load it. Building on an arm64 host does NOT catch this - cross
+      # compiling an x64-stamped IL assembly is legal and the stamp only bites at load time - so assert on
+      # the header directly. Architecture-neutral IL reads as "PE32 ... Intel 80386"; x64-stamped reads as
+      # "PE32+ ... x86-64". Runs on the Unix legs only, where `file` is available; IL is host-independent,
+      # so checking on one platform is sufficient.
+      - name: Assert CLI assemblies are architecture-neutral
+        if: runner.os != 'Windows'
+        run: |
+          set -euo pipefail
+          status=0
+          for project in CMD EngineLayer TaskLayer; do
+            asm="$project/bin/Release/${{ env.DOTNET_FRAMEWORK }}/$project.dll"
+            path="./MetaMorpheus/$asm"
+            if [ ! -f "$path" ]; then
+              echo "::error::expected assembly not found: $path"
+              status=1
+              continue
+            fi
+            desc=$(file -b "$path")
+            if echo "$desc" | grep -q 'x86-64'; then
+              echo "::error file=$asm::architecture-specific assembly ($desc). Remove PlatformTarget from this project - an x64 stamp cannot be loaded on arm64 (issues #2503, #2504). Note PlatformTarget must be left unset rather than set to AnyCPU, which Microsoft.ML rejects at build time."
+              status=1
+            else
+              echo "ok: $asm -> $desc"
+            fi
+          done
+          exit $status
+
+      # Non-blocking for now: this still fails on 'Omics, Version=...' because the mzLib package ships
+      # x64-stamped assemblies too (see smith-chem-wisc/mzLib#1127). Once an architecture-neutral mzLib
+      # release is consumed, drop continue-on-error so arm64 load regressions become a hard failure.
+      # TODO(#2503): make this step required after the mzLib bump lands.
+      - name: Smoke test on arm64 (non-blocking until mzLib ships architecture-neutral)
+        if: matrix.os == 'macos-latest'
+        continue-on-error: true
+        run: dotnet ./MetaMorpheus/CMD/bin/Release/${{ env.DOTNET_FRAMEWORK }}/CMD.dll --test -o "${{ runner.temp }}/arm64-smoke"
--- a/MetaMorpheus/TaskLayer/TaskLayer.csproj
+++ b/MetaMorpheus/TaskLayer/TaskLayer.csproj
@@ -7,13 +7,7 @@
     <DebugSymbols>true</DebugSymbols>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
-    <PlatformTarget>x64</PlatformTarget>
-  </PropertyGroup>
-
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
-    <PlatformTarget>x64</PlatformTarget>
-  </PropertyGroup>
+  <!-- Left architecture-neutral (no PlatformTarget) so the CLI can run on arm64; see CMD.csproj. -->
 
   <ItemGroup>
     <PackageReference Include="MathNet.Numerics" Version="5.0.0" />


trishorts:
looking at this now

acesnik:
Hi @trishorts! There'll be an accompanying mzLib PR for this one

acesnik:
🤖 Companion PR for the mzLib half: smith-chem-wisc/mzLib#1127. Both are required to close #2503/#2504; neither is sufficient alone.

With the two combined, verified on a native arm64 .NET 8 runtime: `CMD.dll --test` completes the full Calibration → GPTMD → Search pipeline with exit code 0 and results identical to the x64 baseline (89 target PSMs, 60 peptides, 47 protein groups at 1% FDR, 482 precursors).

trishorts (commented):
Nice diagnosis — the PE-header table and the `AnyCPU` vs unset distinction make this easy to follow, and I verified the mechanics hold: the `.sln` already maps all six managed projects to `Any CPU`, nothing overrides `PlatformTarget`, and no x64 project is *upstream* of CMD/EngineLayer/TaskLayer, so the fix can't be silently undone.

I also checked the four things I most expected to break, and none of them do — worth recording so review doesn't stall on them:

- **Output paths are unchanged.** `OutputPath` derives from `$(Platform)`, not `$(PlatformTarget)`, and the sln pins `Any CPU`, so output is already `bin/Release/net8.0/`. There are zero `bin\x64` strings anywhere in the repo.
- **Installer is unaffected.** `Product.wxs` sources via `$(var.CMD.TargetDir)`/`$(var.GUI.TargetDir)`, never literal paths, and GUI (which supplies ~380 of the ~389 listed files, including `EngineLayer.dll` and `TaskLayer.dll`) keeps x64 regardless.
- **No CI arguments need changing.** No workflow passes `-p:Platform`, `--runtime`, `-r` or `-a`; no project declares a `RuntimeIdentifier`.
- **32-bit is not a practical risk.** `Product.wxs:17` gates on `VersionNT64` and installs to `ProgramFiles64Folder`, and the apphost RID comes from the SDK rather than `PlatformTarget`.

Two substantive comments inline. Neither blocks the mechanical change — the second is explicitly fine as follow-up — but the first contradicts a claim in the description, so I'd like it resolved before merge.

Two minor things I'm *not* commenting on inline: `GuiFunctions.csproj` sets x64 for Release only (no `Debug` group), so "GuiFunctions deliberately sets x64" is only half true today; and `GUI/GUI_h05aw1sq_wpftmp.csproj` is an accidentally-committed WPF temp project that also sets x64 and will trip up anyone grep-verifying the claim. Both are pre-existing and out of scope here.

MetaMorpheus/CMD/CMD.csproj
```diff
@@ -10,13 +10,14 @@
...
-    <PlatformTarget>x64</PlatformTarget>
-  </PropertyGroup>
-
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
-    <PlatformTarget>x64</PlatformTarget>
-  </PropertyGroup>
+  <!-- PlatformTarget is deliberately left unset so the assembly is architecture-neutral (MSIL) and can be
+       loaded by an arm64 .NET runtime as well as an x64 one. Setting it to x64 stamps AMD64 into the PE
+       header, and CoreCLR on Apple Silicon then refuses the assembly with
+       "FileLoadException: Could not load file or assembly 'CMD'" (issues #2503, #2504).
```
trishorts, line 16: **Major — "CI should cover it" is not true here, and that's worth fixing rather than assuming.**

The description closes with: *"this is a build-configuration-only change with no source edits, so CI should cover it."* It won't, and the reason is the same reason the bug survived this long.

`build.yml` already runs a `macos-latest` leg — which has been Apple Silicon (arm64) since the macos-14 migration — and it is **green today, with `PlatformTarget=x64` still in place**. That's not a contradiction: it only calls `dotnet build`. Cross-compiling an x64-stamped IL assembly on an arm64 host is perfectly legal; the AMD64 stamp only matters at **load** time. Every workflow that actually *executes* MetaMorpheus is `windows-latest` (`Test.yml:12,47`; `InstallRunAndArtifact.yml:16,74,133,165`).

So, concretely:
- no job would have caught #2503/#2504,
- no job will demonstrate that this PR fixes them,
- and nothing stops a future `PlatformTarget=x64` from being reintroduced and silently re-breaking arm64.

The shape that fits this repo is an execute step on the macos leg of `build.yml`, mirroring the Windows smoke test at `InstallRunAndArtifact.yml:112`:

```yaml
- name: Smoke test (arm64 load check)
  if: matrix.os == 'macos-latest'
  run: dotnet ./MetaMorpheus/CMD/bin/Release/net8.0/CMD.dll --test
```

One important caveat, which I think argues for adding it **non-blocking now** rather than deferring: as your own "necessary but not sufficient" section says, this will still fail on `Omics, Version=1.0.583.0` until mzLib also ships architecture-neutral. So either mark it `continue-on-error: true` with a TODO to make it required once the mzLib release lands, or add it in the follow-up mzLib-consumption PR. What I'd rather avoid is merging with no arm64 execution coverage at all and no tracked plan to add it — that's exactly the gap that let this ship broken.

If a hosted runner isn't wanted for the runtime cost, even `dotnet CMD.dll --help` (or anything that forces assembly load) would be enough to catch the regression class, and is seconds rather than a full Calibration→GPTMD→Search pipeline.

MetaMorpheus/EngineLayer/EngineLayer.csproj
```diff
@@ -8,13 +8,7 @@
...
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
-    <PlatformTarget>x64</PlatformTarget>
-  </PropertyGroup>
-
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
-    <PlatformTarget>x64</PlatformTarget>
-  </PropertyGroup>
+  <!-- Left architecture-neutral (no PlatformTarget) so the CLI can run on arm64; see CMD.csproj. -->
```
trishorts, line 11: **Major — this trades one loud failure for several quiet ones, and there is no guard anywhere in the codebase to catch them.**

To be clear up front: I'm not asking you to fix this *in this PR*, and I don't think it should block the merge. But it decides whether arm64 users get a good experience or a baffling one, so I'd like it tracked rather than left implicit.

Today on arm64 the process dies at startup with `FileLoadException` — brutal, but unambiguous. After this change the process starts, an mzML search with Classic deconvolution genuinely works, and the remaining x64 dependencies fail only when reached. The description names those files, but not what hitting them actually looks like:

| Path | Trigger | Failure |
|---|---|---|
| **IsoDec** — mzLib `IsoDecAlgorithm.cs:67` `[DllImport("isodeclib.dll")]`, reached via `EngineLayer/Ms2ScanWithSpecificMass.cs:96` | `DeconvolutionType = "IsoDecDeconvolution"` in the TOML | `DllNotFoundException` raised **inside the `Parallel.For` over MS2 scans** → surfaces as an `AggregateException` partway through a run, attributed to scan processing rather than to the platform. This is the worst of the set. |
| **Thermo `.raw`** | any `.raw` input, via `TaskLayer/MyFileManager.cs:50` | `ThermoFisher.CommonCore.MassPrecisionEstimator.dll` is itself an **x64-marked managed** assembly (Machine=AMD64, ILONLY) → `BadImageFormatException`. Unfixable from our side; it's a third-party binary. |
| **Bruker `.baf` / timsTOF** | any `.d`/`.tdf` input, same dispatch | fails at file open — at least this one fails early and consistently. |
| **PEP / TorchSharp** | PEP-based post-search analysis | `CMD.deps.json` probes `runtimes/win-x64/native`, which has no arm64 counterpart. `Product.wxs:511` already documents this probing behaviour. |

The reason none of these degrade gracefully: **MetaMorpheus has no architecture branching at all.** `RuntimeInformation.ProcessArchitecture` and `Environment.Is64BitProcess` appear zero times in this repo, and `ProcessArchitecture` appears zero times in mzLib. The x64 assumption was expressed *entirely* through `PlatformTarget`, which is precisely why removing it is both correct and completely unguarded.

What I'd like, in rough order of preference:

1. A single guard at the point of use — a `RuntimeInformation.ProcessArchitecture != Architecture.X64` check on the IsoDec and Thermo/Bruker entry points yielding something like *"IsoDec deconvolution and Thermo/Bruker file reading require x64; on this platform use mzML input with Classic deconvolution."* Note this would be the repo's first architecture branch, so there's no precedent to copy — worth a short design discussion rather than being smuggled into a build-config PR.
2. Failing that, a tracked follow-up issue plus an explicit line in the release notes / README stating **arm64 = mzML + Classic deconvolution only**.

Mitigating context, so this isn't read as heavier than it is: these paths are *already* broken on the Linux Docker image this repo publishes (`Dockerfile:2`, alpine runtime — mzLib ships no `.so` for `isodeclib`/`timsdata`/`baf2sql_c`). This PR doesn't create a new class of failure; it extends an existing unguarded one to a platform where users will now get far enough to trip over it.

trishorts:
> Hi @trishorts! There'll be an accompanying mzLib PR for this one

fyi, we have a non public pyMzLib right now that is cross OS and works. https://github.com/smith-chem-wisc/pyMzLib

MetaMorpheus/CMD/CMD.csproj
acesnik: 🤖 You're right, and the claim was simply wrong — I've removed it from the description. I confirmed your reasoning: `build.yml`'s macos leg only runs `dotnet build`, and every job that actually *executes* MetaMorpheus is `windows-latest` (`Test.yml:12`, `InstallRunAndArtifact.yml:16,74,133,165`), so nothing would have caught #2503/#2504 or shown that this PR fixes them.

Added in 1e4aea4e, two steps deliberately different in kind:

**`Assert CLI assemblies are architecture-neutral`** — required, and green today. It reads the PE header of CMD/EngineLayer/TaskLayer via `file` on the Unix legs and fails if any reads as `x86-64`, naming the offending project and noting that `PlatformTarget` must be left *unset* rather than set to `AnyCPU`. This guards the exact regression class without waiting on the mzLib release. Verified both directions: passes on this branch, and fails with per-file `::error` annotations when the three projects are rebuilt with `PlatformTarget=x64`.

**`Smoke test on arm64`** — your suggestion, with `continue-on-error: true` and a `TODO(#2503)` to make it required once an architecture-neutral mzLib is consumed. Added now rather than deferred so the gap is tracked in the workflow rather than in a comment thread.

Also hoisted `DOTNET_VERSION`/`DOTNET_FRAMEWORK` into a workflow-level `env:` block mirroring `InstallRunAndArtifact.yml`, since the new steps would otherwise have repeated the TFM four times.

MetaMorpheus/EngineLayer/EngineLayer.csproj
acesnik: 🤖 Agreed, and tracked as #2702 rather than changed here.

I verified your findings independently rather than taking them on trust: `ProcessArchitecture`/`OSArchitecture`/`Is64BitProcess` appear **zero times** in both MetaMorpheus and mzLib; the IsoDec import is at `mzLib/MassSpectrometry/Deconvolution/Algorithms/IsoDecAlgorithm.cs:67`; and `ThermoFisher.CommonCore.MassPrecisionEstimator.dll` is indeed an x64-marked *managed* assembly — it was the only managed x64 binary left over after I swapped in architecture-neutral mzLib assemblies, which is what made it stand out. Your Docker point holds too (`Dockerfile:2` is alpine), and I've included it in the issue as mitigating context: this PR extends an existing unguarded failure rather than creating a new class of one.

One note on where a fix belongs: the `DllImport`s and the native packaging are all in mzLib, so the guards would live there rather than here. smith-chem-wisc/mzLib#1127 gates the *packaging* side but deliberately adds no runtime guard, since as you say that would be the repo's first architecture branch and deserves its own discussion.

Also taking your two minor points, since one of them was a factual error in my description: `GuiFunctions.csproj` sets x64 for `Release` only, with no `Debug` group, so "GuiFunctions deliberately sets x64" was half true as written — corrected. And `GUI/GUI_h05aw1sq_wpftmp.csproj` is indeed tracked in git; I've left it alone as pre-existing, but noted it in the description so anyone grep-verifying the claim isn't tripped up.

trishorts (approved):
Approving. Both comments are resolved, and I re-checked the CI guard rather than reading it off the commit message.

**Comment 1 — the "CI should cover it" gap.** `1e4aea4e` closes it with two steps that are deliberately different in kind, which is the right split:

The required `Assert CLI assemblies are architecture-neutral` step guards the regression class directly by reading the PE header, without waiting on the mzLib release. I confirmed it genuinely ran and passed on the macos-latest leg — step 6, not skipped, not a no-op — and the error path is well built: per-file `::error file=` annotations, and the message names both the fix and the trap that `PlatformTarget` must be left *unset* rather than set to `AnyCPU`, which is exactly the detail someone would otherwise get wrong.

The non-blocking arm64 smoke test with `TODO(#2503)` is the right call for the interim. Adding it now with `continue-on-error` rather than deferring it means the gap lives in the workflow where it will be seen, instead of in a resolved comment thread. Worth dropping `continue-on-error` in the same PR that bumps to the architecture-neutral mzLib, so the two changes are reviewable together.

**One observation on the assertion, not a request.** It hardcodes `bin/Release/`, but the legs it runs on build `--configuration UbuntuMac`, and `CMD.csproj:7` declares only `Debug;Release`. It resolves correctly because the solution maps `UbuntuMac|Any CPU` → `Release|Any CPU` for CMD, EngineLayer and TaskLayer — so the path is right, but by way of a mapping that isn't visible from the workflow.

I checked the failure mode before deciding whether to raise it, and it's fail-loud: if that mapping ever changed, the step reports `expected assembly not found` and goes red rather than silently passing. That's the correct behaviour and it's why this is a note rather than a finding. Worth being aware of it, though — it's the same class of invisible sln-mapping coupling that turned out to be the blocking bug in smith-chem-wisc/mzLib#1127, and it's now load-bearing in both repos with nothing recording it in either.

**Comment 2 — the unguarded x64 paths.** Correctly tracked as #2702 rather than fixed here; a build-configuration PR is the wrong place for what would be the repo's first architecture branch. Agreed the guards belong in mzLib, since the `DllImport`s and the native packaging are both upstream.

Thanks for independently re-running the claims instead of taking them on trust, and for correcting the two description errors — the half-true `GuiFunctions.csproj` line and the committed `GUI_h05aw1sq_wpftmp.csproj`. Both were minor enough that I'd left them out of the inline comments, so picking them up anyway is appreciated.

Merge order unchanged: smith-chem-wisc/mzLib#1127 merges and cuts a release first, then this one consumes it. Until then this PR is necessary but not sufficient — the CLI will still fail on `Omics, Version=1.0.583.0`, which is precisely what the smoke test is documenting.

acesnik:
🤖 Thank you — this is an exceptional standard of review, and both PRs are materially better for it.

What stands out most is that you verified by building rather than by reading, in both repos, and that it wasn't a formality. Catching the `Development` `Build.0` asymmetry in smith-chem-wisc/mzLib#1127 required constructing the state and observing it, because the diff genuinely doesn't show it. That one would have shipped, and it would have failed at the worst possible point — after merge, on the first tag, between Test and Push.

The judgment on display is rarer than the findings themselves. Checking the failure mode *before* deciding whether the hardcoded `bin/Release/` path was a finding or a note is the distinction that separates a useful review from a noisy one, and you drew it correctly: it is fail-loud, `expected assembly not found` turns the leg red, so it warrants awareness rather than a change. The same goes for confirming the assertion actually executed on the macos leg as step 6, rather than trusting that a step named "Assert" asserts anything.

And the observation that ties the two together is the most valuable thing here: it's the same class of invisible sln-mapping coupling in both repos — `UbuntuMac|Any CPU` → `Release|Any CPU` on this side, `Release|x64` lacking `Build.0` on the other — now load-bearing in both, with nothing recording it in either. Recognising a shared class from two dissimilar instances is a harder thing than finding either one, and it's the part I'd have missed.

Taking both forward-looking points:

- I'll drop `continue-on-error` in the same PR that bumps to the architecture-neutral mzLib, so the two changes are reviewable together as you suggest.
- Happy to record the `UbuntuMac` → `Release|Any CPU` mapping in `build.yml` next to the path it justifies, so the coupling is written down on this side too rather than left to follow-up B's mzLib-side equivalent. Say the word and I'll add it here.

Merge order understood — mzLib#1127 releases first, then this consumes it. #2702 is open for the unguarded paths.

Alexander-Sol (approved):

RayMSMS (approved):

PR #2703, opened 2026-08-03
Discarded PRs: solving the bug for the loss of GlycanIon when N-glycan is used as a variable modification.

RayMSMS:
## Problem

Setting an N-glycan as a variable modification while running an O-glycan or N/O-glycan search caused a `NullReferenceException` in `GlycoPeptides.GetGlycanYIons`, called from `GlycoSearchEngine.CreateGsm`.

## Root Cause

N-glycans loaded into `GlobalVariables.AllModsKnownDictionary`—the known-modification list used to populate variable-modification selections—were loaded without their Y-ion fragments generated (`ToGenerateIons: false`).

Previously, ion generation was performed only for the dedicated N-glycan search path through `GlycanBox.GlobalNGlycans`.

When an N-glycan was instead selected as a variable modification, it was attached directly to peptides during digestion. `CreateGsm` then attempted to generate Y ions from a null ion list, resulting in the exception.

## Fix

In the `GlycoSearchEngine` constructor, after the existing glycan-database setup, the code now identifies N-glycans referenced by the selected variable modifications and generates their Y-ion fragments when they have not already been generated.

## Testing

Added regression tests covering the following cases:

- N-glycan selected as a variable modification during an O-glycan search.
- N-glycan selected as a variable modification during an N/O-glycan search.
- Successful generation of glycan Y ions without a `NullReferenceException`.

## Impact

This change prevents crashes when an N-glycan is used as a variable modification outside the dedicated N-glycan search mode. Existing dedicated N-glycan search behavior is unchanged.


MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
```diff
@@ -94,6 +94,22 @@ public GlycoSearchEngine(List<GlycoSpectralMatch>[] globalCsms, Ms2ScanWithSpeci
                 //DecoyGlycans = Glycan.BuildTargetDecoyGlycans(NGlycans);
             }
 
+            // Ensure any N-glycan selected as a variable modification has its Y-ion fragments
+            // Generated. Glycans in GlobalVariables.AllModsKnownDictionary are loaded without
+            // ions (ToGenerateIons: false) for performance reasons, since that dictionary mainly
+            // serves the "known mods" list. When such a glycan is used as a variable mod in
+            // O-glycan/NO-glycan search, we need its Ions populated to build Y-ions in CreateGsm.
+            foreach (var (modType, idWithMotif) in commonParameters.ListOfModsVariable)
```
trishorts: **[Major] The same NRE is still reachable through fixed mods (and DB/GPTMD-annotated mods).**

This loop only walks `ListOfModsVariable`, but `CreateGsm` inspects `peptideWithMod.AllModsOneIsNterminus`, which carries **both** fixed and variable mods. `MetaMorpheusTask.cs:776-778` builds `fixedModifications` from the identical source with the identical shape:

```csharp
fixedModifications = GlobalVariables.AllModsKnown
    .Where(b => CommonParameters.ListOfModsFixed.Contains((b.ModificationType, b.IdWithMotif)))
    .ToList();
```

and the GUI offers the same mod tree for fixed mods (`GlycoSearchTaskWindow.xaml.cs:411-414`). So selecting the N-glycan as a *fixed* mod instead of a variable one reproduces the exact crash this PR fixes, one checkbox away. Database-annotated / GPTMD-written glycan mods reach `AllModsOneIsNterminus` the same way (`DatabaseLoadingEngine.cs:180`, with `localizableModificationTypes = AllModTypesKnown`, so nothing filters glycans out).

**Suggestion â€” fix it one layer up instead.** The root cause is `GlobalVariables.cs:505`:

```csharp
var nGlycans = GlycanDatabase.LoadGlycan(path, false, false);   // ToGenerateIons: false
```

Hydrating there (or inside that loop, exactly as `LoadTxtGlycan` already does at `GlobalVariables.cs:576`) closes the variable-mod, fixed-mod, and database-annotated routes at once, runs **once** at startup on a single thread, and lets this constructor block be dropped entirely â€” which also resolves the shared-state concern below. The tradeoff worth measuring is startup cost for the full `NGlycan.gdb`; if that turns out to matter, it argues for hydrating lazily inside `GlycoPeptides.GetGlycanYIons` (`GlycoPeptides.cs:78`), which is the single chokepoint every route passes through.

trishorts (commented):
Reviewed with a bug-hunting pass plus a prior-art/conventions sweep of the glyco subsystem. Nice, well-explained fix for a real crash â€” the diagnosis in the PR description is correct and the inline comments in the tests explaining *why* `FindOGlycan` â†’ `CreateGsm` is the only reachable path are genuinely useful; I verified that claim and it holds (`CreateGsm` has exactly one caller, `FindOGlycan` at `:508`; `FindSingle` never touches a glycan).

Two things I want to flag before this goes in, and one cleanup:

1. **The fix is incomplete** â€” an N-glycan chosen as a **fixed** mod (or annotated in the database / added by GPTMD) reaches the same `CreateGsm` line with null `Ions` and still crashes. Handling it at the load site (`GlobalVariables.cs:505`) closes every route at once and would let the constructor block go away entirely.
2. **The object ends up half-hydrated** â€” `Modification.NeutralLosses` is computed from `Ions` inside the `Glycan` constructor, so assigning `.Ions` afterwards leaves it empty. The crash is fixed but the underlying "this glycan behaves differently from a database glycan" problem isn't.

Plus: the added 4.4 MB mzML is byte-identical to one already in `GlycoTestData/N_O_glycoWithFileSpecific/` â€” please point the tests at that one.

Things I checked and am **not** asking you to change: `OGlycanCompositionCombinationChildIons` on an N-glycan is correct-by-precedent (master does the same at `GlycanDatabase.cs:199/209`), `CopyToOutputDirectory=Always` matches every other glyco mzML, and `Assert.DoesNotThrow` is a legitimate regression check here since `RunTask` rethrows unconditionally (`MetaMorpheusTask.cs:699`).

MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
```diff
@@ -94,6 +94,22 @@ public GlycoSearchEngine(List<GlycoSpectralMatch>[] globalCsms, Ms2ScanWithSpeci
...
+            {
+                if (modType == "N-linked glycosylation" &&
+                    GlobalVariables.AllModsKnownDictionary.TryGetValue(idWithMotif, out var mod) &&
+                    mod is Glycan variableNGlycan &&
+                    (variableNGlycan.Ions == null || variableNGlycan.Ions.Count == 0))
+                {
+                    variableNGlycan.Ions = GlycanDatabase.OGlycanCompositionCombinationChildIons(variableNGlycan.Kind);
```
trishorts: **[Major] Assigning `Ions` after construction leaves the glycan half-hydrated â€” `NeutralLosses` is never rebuilt.**

`Modification.NeutralLosses` is derived from `Ions` *inside the `Glycan` constructor* (`Glycan.cs:58-68`), behind an `if (Ions != null)` guard, using an N-glycan-specific filter:

```csharp
else if (type == GlycanType.N_glycan)
{
    ModificationType = "N-linked glycosylation";
    if (Ions != null)
    {
        List<double> lossMasses = Ions.Where(p => p.IonMass < 57000000)
            .Select(p => (double)p.LossIonMass / 1E5).OrderBy(p => p).ToList();
        neutralLosses.Add(DissociationType.HCD, lossMasses);   // + CID, EThcD
    }
}
```

These glycans were constructed with `ToGenerateIons: false`, so `Ions` was null at construction and `NeutralLosses` came out **empty**. Setting `.Ions` here does not recompute it. The result is that a variable-mod N-glycan still fragments differently from a database-path N-glycan of the same composition â€” anything going through mzLib's `Modification.NeutralLosses` (e.g. `GlycoPeptides.GenerateGlycopeptide` â†’ `PeptideWithSetModifications.Fragment`) still sees zero losses. The Y-ion crash is fixed; the fragmentation difference it was masking is not.

Rebuilding the object (`new Glycan(kind, motif, GlycanType.N_glycan)` with ions supplied) rather than patching one property would keep the two in sync. A shared `GlycanDatabase.EnsureIons(Glycan)` that sets **both** and is called from `LoadKindGlycan`, `LoadTxtGlycan`, and here would stop this drifting again.

---

*Separately, on the generator choice â€” not a defect, but worth a comment in the code:* using `OGlycanCompositionCombinationChildIons` on an N-glycan looks wrong at first read, but it is exactly what master already does for N-glycans at `GlycanDatabase.cs:199` and `:209` (the `ToGenerateIons: true` **N-glycan** branch) and at `GlobalVariables.cs:576`. Since the shipped `NGlycan.gdb` is composition-format and routes to `LoadKindGlycan`, this produces the *same* `Ions` the real N-glycan search builds. Worth a one-line comment naming that precedent so the `OGlycanâ€¦` prefix does not read as a bug. Two caveats for the record: a **structure**-format N-glycan database routes to `Struct2Glycan` instead, which adds the 83.03819 cross-ring ion (`Glycan.cs:385`) that the composition generator can never produce â€” so those two paths do diverge; and `NGlycanCompositionFragments` (`GlycanDatabase.cs:304`), the one generator that enforces the trimannosyl core, is deliberately dead per its own comment. Please **don't** switch to it here.

MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
```diff
@@ -94,6 +94,22 @@ public GlycoSearchEngine(List<GlycoSpectralMatch>[] globalCsms, Ms2ScanWithSpeci
...
             }
 
+            // Ensure any N-glycan selected as a variable modification has its Y-ion fragments
+            // Generated. Glycans in GlobalVariables.AllModsKnownDictionary are loaded without
+            // ions (ToGenerateIons: false) for performance reasons, since that dictionary mainly
+            // serves the "known mods" list. When such a glycan is used as a variable mod in
+            // O-glycan/NO-glycan search, we need its Ions populated to build Y-ions in CreateGsm.
+            foreach (var (modType, idWithMotif) in commonParameters.ListOfModsVariable)
+            {
+                if (modType == "N-linked glycosylation" &&
```
trishorts: **[Major] This mutates a process-global object from an engine constructor, and the mutated field is read under `Parallel.ForEach`.**

`AllModsKnownDictionary` is populate-at-startup / read-only-thereafter â€” declared `private set` (`GlobalVariables.cs:65`), filled once in `SetUpGlobalVariables()`, and every other reference in the codebase is a read. I could not find another place where engine or task code mutates an object *contained* in it; the four runtime writers to `GlobalVariables.*` all assign scalars.

Three specific concerns:

1. **Lifetime.** `GlycoSearchEngine` is constructed once per spectra file per partition (`GlycoSearchTask.cs:134-135`, inside the file loop at `:96` and the partition loop at `:115`), so this runs repeatedly; and the mutation is never undone, so it outlives the task in the GUI, which does not restart between runs.
2. **Concurrency.** `.Ions` is enumerated by worker threads â€” `GlycoSearchEngine.cs:121` `Parallel.ForEach(threads, ...)` â†’ `CreateGsm` `:313` â†’ `GlycoPeptides.cs:78` `foreach (var ion in glycan.Ions)`. Replacing the `List<GlycanIon>` reference on a shared instance while another thread enumerates it is the classic torn-read/`InvalidOperationException` hazard. The repo's convention for shared mutable statics is an explicit lock (`FragmentationReanalysisViewModel.cs:28,263`, which even has a dedicated race-condition regression test) or `Lazy<T>` (`FdrAnalysisEngine.cs:15`) â€” the glyco subsystem is the part that skips it, so the surrounding code isn't guarding this for you.
3. **It works by object-identity coincidence.** The search resolves its mods from `GlobalVariables.AllModsKnown` (`MetaMorpheusTask.cs:773`), *not* from `AllModsKnownDictionary`. This only fixes anything because `GlobalVariables.cs:496-500`/`:510-512` add the **same** `Glycan` instance to both collections. If those ever stop sharing instances, this silently stops working with no test failing.

In fairness, this constructor already writes `GlycanBox.GlobalOGlycans`/`GlobalNGlycans` (`:65`, `:78-88`), so it is not a new sin in this class â€” but it is worth not adding to. Hydrating in `GlobalVariables.LoadGlycans` (see my comment above) makes all three points go away.

MetaMorpheus/Test/Test.csproj
```diff
@@ -89,6 +89,9 @@
     <None Update="GlycoTestData\2019_09_16_StcEmix_35trig_EThcD25_rep1_4565.mgf">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Update="GlycoTestData\2019_09_16_StcEmix_35trig_EThcD25_rep1_4999_5500.mzML">
```
trishorts: **[Major] This 4.4 MB mzML is byte-identical to one already in the repo â€” please reuse it rather than adding a second copy.**

`MetaMorpheus/Test/GlycoTestData/N_O_glycoWithFileSpecific/2019_09_16_StcEmix_35trig_EThcD25_rep1_4999_5500.mzML` already exists and has the same name. I verified the contents match exactly (identical MD5 after normalizing CRLF/LF; both 4,633,018 bytes of content), so this adds ~4.4 MB to the repo for a file it already has.

Pointing both new tests at the existing path lets you drop the new file and this `.csproj` entry:

```csharp
string raw = Path.Combine(TestContext.CurrentContext.TestDirectory,
    @"GlycoTestData\N_O_glycoWithFileSpecific\2019_09_16_StcEmix_35trig_EThcD25_rep1_4999_5500.mzML");
```

(That path is already registered with `CopyToOutputDirectory` in this file, so no new entry is needed.) For what it's worth, `Always` is the right copy mode â€” every mzML under `GlycoTestData` uses it â€” so that part matches convention; it's only the duplication that's the problem.

MetaMorpheus/Test/TestNOGlyco.cs
```diff
@@ -342,5 +342,80 @@ public static void TestGetModPos()
...
+                new EverythingRunnerEngine(new List<(string, MetaMorpheusTask)> { ("Task", glycoSearchTask) },
+                    new List<string> { raw }, new List<DbForTask> { db }, outputFolder).Run();
+            });
+
+            // Confirm the fix populated Ions on the shared Glycan instances used as a variable mods.
+            foreach (var (modType, nGlycanIdWithMotif) in nGlycanVariableModsFromNGlycan_ForNoSearch)
+            {
+                var nGlycanMod = GlobalVariables.AllModsKnownDictionary[nGlycanIdWithMotif] as Glycan;
+                Assert.That(nGlycanMod, Is.Not.Null);
+                Assert.That(nGlycanMod.Ions, Is.Not.Null.And.Not.Empty);
```
trishorts: **[Major] These assertions can pass without the fix having run in this test.**

Two problems, both from the fact that the thing being asserted on is process-global and never reset:

1. **Order dependence.** `TestOGlyco.OGlycoSearch_WithNGlycanAsVariableMod_DoesNotCrash` asserts on the *same two* global mods (`"N1 on Nxs"` / `"N1 on Nxt"`). Whichever test runs second finds `Ions` already populated by the first, so this assertion passes regardless of whether the fix executed in *that* test. Only the `Assert.DoesNotThrow` is really carrying the regression, and only for whichever test runs first.
2. **Nothing checks the ions are *right*.** `Is.Not.Null.And.Not.Empty` cannot distinguish correct N-glycan Y-ions from an arbitrary non-empty list. Since the PR's stated goal is "successful generation of glycan Y ions", the assertion that would actually pin the behaviour is an equality check against the path the dedicated N-glycan search uses for the same composition:

```csharp
var expected = GlycanDatabase.LoadGlycan(nGlycanDbPath, ToGenerateIons: true, isOGlycan: false)
    .First(g => g.IdWithMotif == nGlycanIdWithMotif).Ions;
Assert.That(nGlycanMod.Ions.Select(i => i.IonMass), Is.EquivalentTo(expected.Select(i => i.IonMass)));
```

That version would also have surfaced the `NeutralLosses` gap I noted on the engine file, since the two objects would visibly disagree.

Your own comment above concedes the hits are FDR-filtered so this is really a smoke test â€” that's fine and worth keeping, but it's worth saying so in the test name (`..._DoesNotCrash` already does, nicely) and adding one assertion that fails if the ion content regresses.

Also: this test and `TestOGlyco.OGlycoSearch_WithNGlycanAsVariableMod_DoesNotCrash` are ~85 near-identical lines in two files, differing only in `GlycoSearchType` and the `NGlycanDatabasefile` line. A single `[TestCase(GlycoSearchType.OGlycanSearch)] [TestCase(GlycoSearchType.N_O_GlycanSearch)]` parameterized test would halve the maintenance and make the intended difference obvious.

MetaMorpheus/Test/TestNOGlyco.cs
```diff
@@ -342,5 +342,80 @@ public static void TestGetModPos()
...
+            // can hit the crash. So we need a genuinely co-N/O-glycosylated candidate -- N1 on an
+            // Nxs/Nxt site plus an O-glycan on a separate S/T site -- for this test to mean anything.
+            //
+            // In this test, there are two N- and O- co-glycosylated peptides found in the data, However,
+            // the peptides are filtered out by the FDR, so the test is only checking that the search completes without crashing.
+
+            string outputFolder = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TESTGlycoData_OWithNVarMod");
+            Directory.CreateDirectory(outputFolder);
+
+            GlobalVariables.NGlycanDatabasePaths.Add(
```
trishorts: **[Minor] Global-state setup without the idempotent guard, and never restored.**

`NGlycanDatabasePaths` accumulates for the rest of the test session. Five existing glyco tests do exactly this (`TestNOGlyco.cs:55,107,151,188`, `TestOGlyco.cs:164`), so this matches local convention and I won't push hard â€” but the better form is already in this repo, in `GlycoSearchEngineTest.cs:33-34`:

```csharp
if (!GlobalVariables.NGlycanDatabasePaths.Contains(nglycanPath))
    GlobalVariables.NGlycanDatabasePaths.Add(nglycanPath);
```

Worth adopting here since both new tests add paths, and combined with the never-reset `Ions` it's what makes these two tests order-dependent.

MetaMorpheus/Test/TestNOGlyco.cs
```diff
@@ -342,5 +342,80 @@ public static void TestGetModPos()
...
+
+            // Confirm the fix populated Ions on the shared Glycan instances used as a variable mods.
+            foreach (var (modType, nGlycanIdWithMotif) in nGlycanVariableModsFromNGlycan_ForNoSearch)
+            {
+                var nGlycanMod = GlobalVariables.AllModsKnownDictionary[nGlycanIdWithMotif] as Glycan;
+                Assert.That(nGlycanMod, Is.Not.Null);
+                Assert.That(nGlycanMod.Ions, Is.Not.Null.And.Not.Empty);
+            }
+
+            Directory.Delete(outputFolder, true);
```
trishorts: **[Minor] Cleanup won't run if an assertion fails, and the folder name is shared with the other new test.**

Both new tests use `TESTGlycoData_OWithNVarMod`. With a bare end-of-body `Directory.Delete`, a failing assertion short-circuits it and leaves the folder on disk for the other test â€” which calls `Directory.CreateDirectory` rather than clearing first, so it inherits stale output. The `try`/`finally` form is already used in this repo, in the file this PR's sibling test lives beside (`GlycoSearchEngineTest.cs:126-132`):

```csharp
finally
{
    if (Directory.Exists(outputFolder)) Directory.Delete(outputFolder, true);
}
```

Distinct folder names per test would also do it.

MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
trishorts: Verified fixed. Moving the hydration out of the `GlycoSearchEngine` constructor and up into `GlycoSearchTask.RunSpecific` (`GlycoSearchTask.cs:51`) resolves all three of my concerns at once: it now runs **once** per task rather than once per spectra file per partition, and it completes before indexing and before any `Parallel.ForEach`, so the `.Ions` list can no longer be swapped while worker threads enumerate it in `GetGlycanYIons`. The `if (Ions != null && Ions.Count > 0) return;` guard in `RegenerateIons` also makes it idempotent, which is the right belt-and-braces.

The comment you added about `AllModsKnownDictionary` and `AllModsKnown` sharing the same `Glycan` instances is exactly the thing that needed writing down - that identity relationship is what makes this work, and it was previously implicit.

MetaMorpheus/Test/TestNOGlyco.cs
trishorts: Verified fixed - the `if (!GlobalVariables.NGlycanDatabasePaths.Contains(...))` guard is in, and you went further than I suggested by also removing the path again in the `finally`, so the test no longer leaks into the rest of the session at all.

MetaMorpheus/Test/TestNOGlyco.cs
trishorts: Verified fixed - the body is wrapped in `try`/`finally` so cleanup survives a failed assertion, and the output folder name is now derived per case (`TESTGlycoData_{NO|O}With{Fixed|Var}NMod`), so the four `[TestCase]` runs no longer share a directory or inherit each other's stale output.

MetaMorpheus/Test/Test.csproj
trishorts: Verified fixed - the duplicate mzML is gone from the diff and both tests now point at the existing `GlycoTestData/N_O_glycoWithFileSpecific/...` copy, so the ~4.4 MB is no longer added to the repo. Thanks for chasing that down.

MetaMorpheus/Test/TestNOGlyco.cs
trishorts: Verified fixed, and more thoroughly than I asked for. Closing this out. Specifically:

- **Order dependence is gone.** The test now asserts `Ions` is null *before* the search as an explicit setup precondition, and resets `g.Ions = null` in the `finally`. So each case proves the fix ran in that case, rather than inheriting hydration from whichever test happened to run first.
- **The ions are actually checked.** `Is.EqualTo(expectedNGlycan.Ions.Select(ion => ion.IonMass))` against a fresh `LoadGlycan(..., ToGenerateIons: true, ...)` is exactly the equality check that pins the behaviour, and it is applied to the instances in *both* `AllModsKnownDictionary` and `AllModsKnown`.
- **The duplication is gone.** The four `[TestCase]`s collapse the two near-identical ~85-line tests and make the fixed-vs-variable / O-vs-N_O axes explicit.

The long comment explaining why `FindSingle` vs `FindOGlycan` matters - and therefore why the test needs a genuinely co-N/O-glycosylated candidate and the one-composition `NGlycan_fakeN-glycan.gdb` - is the kind of thing that stops someone "simplifying" this test later and silently gutting it. Worth keeping.

MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
trishorts: Verified fixed. I checked the claim in the new `RegenerateIons` comment rather than taking it on faith, and it holds: `LoadKindGlycan` builds every composition-format glycan through `new Glycan(kind, motif, type)`, which chains to `this(null, GetMass(kind), kind, null, false, motif, type)` (`Glycan.cs:98-99`), so `Ions` is null *at construction* on both the `ToGenerateIons: true` and `false` branches - `.Ions` is only assigned afterwards at `GlycanDatabase.cs:199/209`. That means `NeutralLosses` comes out empty on the dedicated N-glycan path too, so not regenerating it here really does match existing behavior and there is no fragmentation divergence to fix. Good call documenting the reasoning in the code.

(For the record, the structure-format path via `Struct2Glycan` *does* construct with ions and so does get neutral losses - a pre-existing asymmetry, not something this PR introduces or needs to address.)

MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
trishorts: Thanks - the **fixed-mod** half of this is done, and the move up to `GlycoSearchTask` is the right shape. Leaving the thread open for the rest of it, because re-reviewing turned up a concrete, shipped-data reproduction of the same crash that the current fix does not cover.

### 1. N-linked mods that are not `Glycan` instances

`ReloadGlycanInModList` only hydrates entries that satisfy `mod is Glycan`. But the crash site selects by **string** and then downcasts (`GlycoSearchEngine.cs:309-313`):

```csharp
if (peptideWithMod.AllModsOneIsNterminus.Any(p => p.Value.ModificationType == "N-linked glycosylation"))
{
    Glycan nGlycan = peptideWithMod.AllModsOneIsNterminus
        .First(p => p.Value.ModificationType == "N-linked glycosylation").Value as Glycan;
    fragmentsForEachGlycoPeptide.AddRange(GlycoPeptides.GetGlycanYIons(theScan.PrecursorMass, nGlycan));
}
```

So a modification carrying that `ModificationType` string that is a plain `Modification` rather than a `Glycan` is skipped by the new hydration loop, makes `as Glycan` yield `null`, and then NREs at `GlycoPeptides.cs:77` on `glycan.Mass` - the same exception this PR exists to remove.

**This is reachable with the database we ship.** `LoadTxtGlycan` adds the entry as a plain `Modification` whenever the composition fails to parse (`GlobalVariables.cs:566-569`):

```csharp
var kind = GlycanDatabase.String2Kind(glycoMod.OriginalId);
if (kind.Sum(p => p) == 0)
{
    _AllModsKnown.Add(glycoMod);   // plain Modification, NOT a Glycan
    continue;
}
```

and `String2Kind` returns an all-zero kind for any ID with no parenthesised count, because the loop is `while (i < x.Length - 1)` and a bare token splits into a single element (`GlycanDatabase.cs:222-234`). In `Mods/glyco.txt` two N-linked entries are exactly that shape:

```
ID   Hex        MT   N-linked glycosylation
ID   dHex       MT   N-linked glycosylation
```

Ticking either **`Hex`** or **`dHex`** as a variable mod in an O-glycan search should still crash on this branch. Worth adding as a fifth `[TestCase]`, since it is a one-line addition to the test you already built and it fails today.

### 2. Database- and GPTMD-annotated glycan mods

Still open from my original comment. Mods annotated in the protein XML resolve against the same unhydrated instances - `DatabaseLoadingEngine.cs:180` passes `GlobalVariables.AllModsKnown` into `ProteinDbLoader.LoadProteinXML` - and they appear in neither `ListOfModsVariable` nor `ListOfModsFixed`, so `ReloadGlycanInModList` never sees them. Note also that the call sits at `GlycoSearchTask.cs:51`, *before* `DatabaseLoadingEngine` runs, so it could not cover this route even if the predicate were widened.

### Suggestion

Both routes close at the consumption point rather than the hydration point, which is also where they cannot drift apart again. Pattern-match instead of `as`, and hydrate lazily there:

```csharp
var nGlycans = peptideWithMod.AllModsOneIsNterminus.Values.OfType<Glycan>()
    .Where(g => g.ModificationType == "N-linked glycosylation");
foreach (var nGlycan in nGlycans)
{
    nGlycan.RegenerateIons();   // already idempotent and cheap
    fragmentsForEachGlycoPeptide.AddRange(GlycoPeptides.GetGlycanYIons(theScan.PrecursorMass, nGlycan));
}
```

`OfType<Glycan>()` makes the non-`Glycan` case a no-op instead of a crash, and calling `RegenerateIons()` at the point of use covers the variable-mod, fixed-mod, database-annotated, and GPTMD routes with one guard. If you would rather not call it from inside the parallel region, the alternative is to keep `ReloadGlycanInModList` for the mod-list routes and add just the `OfType<Glycan>()` guard here so the remaining cases degrade to "no Y ions" rather than an NRE - that alone would be enough to close this thread.

Issue #1125, opened 2018-04-18
Viewing/using/validating MetaMorpheus search results in other programs


rmillikin:
We want to ensure that MetaMorpheus users can view results (mzID, pepXML, mzML formats in particular) in other programs. I'll make a list here of programs here that we'd like to support. Feel free to add/suggest more.

It's basically the same as issue #52 but I'd like to have it a bit more organized.

mzML:
- [ ] Batmass
- [ ] Proteowizard MSConvert

mzID:
- [ ] Skyline
- [ ] Scaffold
- [ ] HUPO mzidentML (I think this is more for validating than viewing)
- [ ] PRIDE (I think this is more for validating than viewing)
- [ ] Proteowizard idconvert

pepXML:
- [ ] proxyl (crosslink search)

lonelu:
pep.xml now works for TPP. Since TPP doesn't support MetaMorpheus, I had to pretend 'Tandom' is the search engine of MetaMorpheus' output.

rmillikin:
We definitely don't want to say that our search results came from Tandom. That will cause a lot of confusion. Can we contact the TPP people and ask them to support MetaMorpheus as a search engine?

Issue #1114, opened 2018-04-15
Microsoft Visual C++ Redistributable should be included in the installer


rmillikin:
This was included in the .exe which was giving several people (Zach and Lei) "weird issues". We discontinued the .exe and now the C++ redistributable is not being included in the installation process.

Are those problems with the .exe related to the C++ redistributable stuff?

rmillikin:
I don't think this is needed anymore with RawFileReader

PR #2704, opened 2026-08-05
mzLib 1.0.584: ETD and ECD no longer produce y ions

trishorts:
Consumes [mzLib #1114](https://github.com/smith-chem-wisc/mzLib/pull/1114), which removed `ProductType.y` from the ETD and ECD product sets (they are now exactly `{ c, zDot }`) and dropped the ETD/ECD arm of `GetWaterAndAmmoniaLossProductTypesFromDissociation`. `EThcD` is deliberately unchanged.

That change turns four MetaMorpheus tests red in mzLib's `integration` job, which builds MM master against mzLib master. This PR bumps the pin and fixes them.

## Dependency — resolved

This PR previously could not go green on its own: it pins mzLib 1.0.584, and the newest published package was 1.0.583, which sat 9 commits behind #1114. Every CI job failed at restore with `NU1102: Unable to find package mzLib with version (>= 1.0.584)` — nothing in the diff was ever the cause.

**mzLib 1.0.584 has now been released** from `smith-chem-wisc/mzLib@master` (`e3220d6a`), which carries #1114 along with #1127 (arm64 packaging), #1126 (UniProt retrieval) and #1125 (FlashLFQ scratch dir). MM's arm64 counterpart #2701 is already on master, so the pin is consistent in both directions.

## Verification

Before the release was cut, the exact 1.0.584 payload was built and packed locally from `e3220d6a` and MetaMorpheus was restored, built and tested against it, with this branch merged up to master:

- full solution builds `Release`, 0 errors;
- full suite under CI's own filter (`Category!=ExternalService`): **1723 passed, 1 skipped, 0 failed** (1724 total);
- the glyco and crosslink fixtures this PR re-baselines (`TestOGlyco`, `TestNOGlyco`, `XLTest`, `GlycoSearchEngineTest`): 100/100 passed.

Master has since merged three glyco fixes (#2696 localization-scan selection, #2698 glycan-database paths, #2699 precondition message). Those are merged in here and the re-baselined numbers below still hold unchanged against them.

## Re-baselined expectations

Originally measured against a local mzLib build of `f6b0f0d1` (#1114's merge commit, chosen deliberately *before* #1127 so the numbers isolate the y-ion removal), and re-confirmed against the released 1.0.584. The search score reproduces the CI failure to 15 digits (`20.207304008910487`).

| test | assertion | was | now |
|---|---|---|---|
| `NandO_GlycoSearch_onlyO_Run2` | `Psms.First().Score` | 23.2 | **20.2** |
| `OGlycoTest_FragmentIons2` | `fragments_etd.Count` | 22 | **15** |
| `TestMixedMs2Ms2` | `ChildMatchedFragmentIons` | 21 | **15** |
| `TestMixedMs2Ms2` | `BetaPeptide.ChildMatchedFragmentIons` | 25 | **18** |
| `TestMixedMs2Ms2` | `ChildScanMatchedIons` (TSV round-trip) | 21 | **15** |
| `TestMixedMs2Ms2` | `BetaPeptideChildScanMatchedIons` (TSV round-trip) | 25 | **18** |

Only three of these were in the reported failure list — NUnit stops a test at its first failed assertion, so the rest were queued behind them. `TestMixedMs2Ms2` asserts its child-scan counts twice, once on the CSM and again after the TSV round-trip; both copies had to move.

`OGlycoTest_FragmentIons2`'s 15 is the one number that is analytically derivable rather than empirical: `TVYLGASK` is 8 residues, giving 7 c + 8 zDot (the extra being the full-length z•), and the 7 y ions are gone.

The direction of every change is correct rather than regressive — ETD genuinely does not produce y ions, so the glyco and crosslink ETD paths had been matching and scoring ions that cannot exist. These are re-baselines, not bug reports.

## Removed: `GlycoPeptides.GetFragmentHash` and `OGlycoTest_FragmentIonsHash`

`GetFragmentHash` had a latent bug that #1114 exposed:

```csharp
var len = products.Count / 3;   // "ETD emits c, y and z-dot"
...
newFragments[len * 3 - j + 2] += glycanMass;
```

`len` is meant to be *ions per series*, and the `/ 3` hardcoded the assumption that ETD emits three of them. With `{ c, zDot }` a 20-product ETD set gives `len = 6` instead of 10, which **skips the c-ion loop entirely** (its guard is `while (j <= len + 1)`, and for a glycan at position 9, `9 <= 7` is false, so the glycan mass is never added to any c ion) and **writes the z-ion shift into the wrong slots**, indexing as though the array were 30 long when it is 20. Its test's `overlap == 14` / `overlap2 == 23` would therefore have been re-baselined onto meaningless output.

Rather than repair arithmetic nothing depends on, both are deleted: **`GetFragmentHash` has no production callers.** Its only references anywhere in the repo were the three calls inside its own test. The Release build passes with the method gone, which confirms it.

The two HCD assertions in that test (`fragments_hcd.Count() == 20`, `fragmentsMod_hcd.Count() == 61`) are *not* about the hash — they pin that the default `Fragment` method emits glycan neutral-loss ions for an O-glycopeptide, which is still true and still worth covering. They are kept, and the trimmed test is renamed `OGlycoTest_ModifiedFragmentsIncludeNeutralLosses` to match what it now actually asserts.

## A comment that became false

```csharp
//ETD didn't change y ions.
```

That comment sat above one of the deleted hash assertions and was a claim about ETD fragmentation that #1114 falsifies. It went with the deletion rather than being left standing.

## Assertion style

The counts above were written `Assert.That(x == n)`, whose failure message is `Expected: True / But was: False` — it does not report the actual value, which is precisely what you need when a number moves upstream. The ones whose numbers this PR is already changing are converted to `Assert.That(x, Is.EqualTo(n))`. Scoped strictly to lines the diff already owns; no file-wide sweep.

## Not addressed (follow-ups)

- **The N-glyco ETD path also loses its y ions** (`GlycoSearchEngine.cs:547-549`) but no test exercises N-glyco under ETD — the N-glyco tomls are all EThcD/HCD. Silent behaviour change with no coverage.
- `GlycoPeptides.cs:234` and `:251` filter `ProductType.y` out of the ETD product list. Those are now no-ops. Harmless, and left in place so the intent survives if y ever returns.
- `CrosslinkSearchEngine.DissociationTypeGenerateSameTypeOfIons` treats ETD and ECD as equivalent. That stays correct only because #1114 changed both rows identically — worth knowing if they ever diverge.

### DIFF
--- a/MetaMorpheus/CMD/CMD.csproj
+++ b/MetaMorpheus/CMD/CMD.csproj
@@ -24,7 +24,7 @@
     <PackageReference Include="Microsoft.ML.CpuMath" Version="3.0.1" />
     <PackageReference Include="Microsoft.ML.FastTree" Version="3.0.1" />
     <PackageReference Include="Microsoft.NETCore.App" Version="2.2.8" />
-    <PackageReference Include="mzLib" Version="1.0.583" />
+    <PackageReference Include="mzLib" Version="1.0.584" />
     <PackageReference Include="Nett" Version="0.15.0" />
     <PackageReference Include="SQLite.Interop.dll" Version="1.0.103" />
     <PackageReference Include="System.Data.SQLite" Version="1.0.118" />
--- a/MetaMorpheus/EngineLayer/EngineLayer.csproj
+++ b/MetaMorpheus/EngineLayer/EngineLayer.csproj
@@ -29,7 +29,7 @@
     <PackageReference Include="Microsoft.ML.CpuMath" Version="3.0.1" />
     <PackageReference Include="Microsoft.ML.FastTree" Version="3.0.1" />
     <PackageReference Include="Microsoft.NETCore.App" Version="2.2.8" />
-    <PackageReference Include="mzLib" Version="1.0.583" />
+    <PackageReference Include="mzLib" Version="1.0.584" />
     <PackageReference Include="NETStandard.Library" Version="2.0.3" />
     <PackageReference Include="Nett" Version="0.15.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
--- a/MetaMorpheus/EngineLayer/GlycoSearch/GlycoPeptides.cs
+++ b/MetaMorpheus/EngineLayer/GlycoSearch/GlycoPeptides.cs
@@ -379,47 +379,6 @@ namespace EngineLayer.GlycoSearch
 
 
         /// <summary>
-        /// Generate the new fragment list, we add the glycan mass to the c ions and z ions from the peptide fragment list
-        /// </summary>
-        /// <param name="products"></param>
-        /// <param name="keyValuePair"></param>
-        /// <param name="OGlycanBoxes"></param>
-        /// <param name="FragmentBinsPerDalton"></param>
-        /// <returns></returns>
-        public static int[] GetFragmentHash(List<Product> products, Tuple<int, int[]> keyValuePair, GlycanBox[] OGlycanBoxes, int FragmentBinsPerDalton)
-        {
-            double[] newFragments = products.OrderBy(p=>p.ProductType).ThenBy(p=>p.FragmentNumber).Select(p => p.NeutralMass).ToArray(); // store the fragment mass in the order of c1, c2, c3, y1, y2, y3, z1, z2, z3
-            var len = products.Count / 3;
-            if (keyValuePair.Item2!=null)
-            {
-                for (int i = 0; i < keyValuePair.Item2.Length; i++) // we want to add the glycan mass to the c ions and z ions that contain the glycan.
-                {                                                   // y ions didn't change in EThcD for O-glyco, so we just need to deal with c ions and z ions.
-                    var j = keyValuePair.Item2[i];
-                    while (j <= len + 1) // for c ions
-                    {
-                        newFragments[j - 2] += (double)GlycanBox.GlobalOGlycans[OGlycanBoxes[keyValuePair.Item1].ModIds[i]].Mass/1E5;
-                        j++;
-                    }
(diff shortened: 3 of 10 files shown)


trishorts:
Marking this **draft** until mzLib 1.0.584 is published — the red CI is the blocker described above, not a defect in the diff.

Every job fails at the Restore step, before anything compiles or any test runs:

```
error NU1102: Unable to find package mzLib with version (>= 1.0.584)
  - Found 508 version(s) in nuget.org [ Nearest version: 1.0.583 ]
```

### To unblock

Push tag `1.0.584` on `smith-chem-wisc/mzLib@master` (currently `b92fc65d`). The release workflow builds, tests, packs and pushes to nuget.org; then this PR's CI can be re-run and should go green.

### One thing to know before tagging

1.0.584 would be the **first release cut through the new Any-CPU path**. 1.0.583 was tagged at `fc2889cc`, which predates #1127 (arm64). Post-#1127 the pieces are consistent —

- `release.yml` builds with `/p:Platform="Any CPU"`, so output lands in `bin\Release`
- `mzLib.nuspec` now reads from `bin\Release` rather than `bin\x64\Release`
- the `Development` project has **no** `.Build.0` under either `Release|x64` or `Release|Any CPU`, so the solution-wide `dotnet test` in the release job won't try to build dev-only fixtures

— but that path has not actually been exercised by a release yet, so the release run is worth watching rather than assuming.

### Verification already done

The content of this PR was validated locally against a build of mzLib at `f6b0f0d1` (#1114's merge commit): **1719 passed, 0 failed, 1 skipped**, `Category!=ExternalService`. The re-baselined search score reproduces the original CI failure to 15 digits.

RayMSMS (approved):

pcruzparri (approved):

zhuoxinshi (approved):

PR #2706, opened 2026-08-07
Fix(glyco): solving the bug for the loss of GlycanIon when N-glycan is used as a variable modification.

RayMSMS:
## Problem

Setting an N-glycan as a **fixed** or **variable** modification while running an O-glycan or N/O-glycan search caused a `NullReferenceException` in `GlycoPeptides.GetGlycanYIons`, called from `GlycoSearchEngine.CreateGsm`.

## Root Cause

N-glycans loaded into `GlobalVariables.AllModsKnown` (the known-modification list that populates fixed/variable mod selections) are loaded without their Y-ion fragments generated (`ToGenerateIons: false`). Ion generation was only performed on the dedicated N-glycan search path.

When an N-glycan is instead selected as a fixed or variable mod, it is attached to peptides during digestion and reaches `CreateGsm`, which then tried to build Y ions from a `null` ion list — throwing the exception. The old consumption-site code also cast with `as Glycan` after selecting by `ModificationType` string, so a non-`Glycan` modification carrying that type string (e.g. a database/GPTMD annotation) would cast to `null` and hit the same crash.

## Fix

At the consumption site in `CreateGsm`, N-linked glycan mods are now selected with `OfType<Glycan>()` and only contribute Y ions when they actually have them:

- `OfType<Glycan>()` filters out any non-`Glycan` modification carrying the `"N-linked glycosylation"` type string, so it can no longer cast to `null` and crash.
- If the glycan has no Y ions (an N-glycan used as a plain fixed/variable mod), its Y ions are simply skipped rather than generated on the fly.

This is a deliberate, minimal fix: N-glycan-as-a-mod is **not** scored on its Y ions in this mode. Correct N-glycan identification is the job of the dedicated **NO-search**, which is where users should go for that.

## Testing

- `GlycoSearch_WithNGlycanAsFixedOrVariableMod_DoesNotCrash` — parameterized integration smoke test over {fixed, variable} × {O-search, NO-search}. A pre-search assertion confirms the N-glycan mods start with `null` Ions (the state that used to crash), so the crash-free run is a meaningful regression check. Results-level assertions aren't made because the matched N/O co-glycopeptides are FDR-filtered out.

## Impact

Prevents the crash when an N-glycan is used as a fixed or variable modification, including from database/GPTMD annotations. N-glycan Y ions are not scored in this mode by design — dedicated N-glycan / NO-search behavior is unchanged.

### DIFF
--- a/MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
+++ b/MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
@@ -305,12 +305,18 @@ namespace EngineLayer.GlycoSearch
 
             var fragmentsForEachGlycoPeptide = GlycoPeptides.OGlyGetTheoreticalFragments(CommonParameters.DissociationType, CommonParameters.CustomIons, peptide, peptideWithMod);
 
-            // if the peptide contains N-glycan, we need to add the Y ion from N-glycan.
-            if (peptideWithMod.AllModsOneIsNterminus.Any(p=>p.Value.ModificationType == "N-linked glycosylation"))
+            // If the peptide contains N-glycan, we need to add the Y ion from N-glycan.
+            // At first, we only consider the mod is N-linked glycosylation and as a glycan object.
+            var nGlycans = peptideWithMod.AllModsOneIsNterminus.Values
+                .OfType<Glycan>()
+                .Where(g => g.ModificationType == "N-linked glycosylation");
+            foreach (var nGlycan in nGlycans) 
             {
-                Glycan nGlycan = peptideWithMod.AllModsOneIsNterminus.First(p =>
-                    p.Value.ModificationType == "N-linked glycosylation").Value as Glycan;
-                fragmentsForEachGlycoPeptide.AddRange(GlycoPeptides.GetGlycanYIons(theScan.PrecursorMass, nGlycan));
+                if (nGlycan.Ions != null) // guard against null Ions property
+                {
+                    fragmentsForEachGlycoPeptide.AddRange(GlycoPeptides.GetGlycanYIons(theScan.PrecursorMass, nGlycan));
+                }
+                // else: no ions → skip (dummy score); users should use NO-search for correct N-glycan ID
             }
 
             var matchedIons = MatchFragmentIons(theScan, fragmentsForEachGlycoPeptide, CommonParameters);
--- a/MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
+++ b/MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
@@ -310,7 +310,7 @@ namespace EngineLayer.GlycoSearch
             // excluded; correct N-glycan ID is the NO-search's job.
             var nGlycan = peptideWithMod.AllModsOneIsNterminus.Values
                 .OfType<Glycan>()
-                .FirstOrDefault(g => g.ModificationType == "N-linked glycosylation" && g.HasIons);
+                .FirstOrDefault(g => g.Type == GlycanType.N_glycan && g.HasIons);
             if (nGlycan != null)
             {
                 fragmentsForEachGlycoPeptide.AddRange(GlycoPeptides.GetGlycanYIons(theScan.PrecursorMass, nGlycan));
--- a/MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
+++ b/MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
@@ -305,18 +305,16 @@ namespace EngineLayer.GlycoSearch
 
             var fragmentsForEachGlycoPeptide = GlycoPeptides.OGlyGetTheoreticalFragments(CommonParameters.DissociationType, CommonParameters.CustomIons, peptide, peptideWithMod);
 
-            // If the peptide contains N-glycan, we need to add the Y ion from N-glycan.
-            // At first, we only consider the mod is N-linked glycosylation and as a glycan object.
-            var nGlycans = peptideWithMod.AllModsOneIsNterminus.Values
+            // Only Glycan-typed N-linked mods with generated Y ions contribute Y ions here.
+            // Non-Glycan annotations (DB/GPTMD) and mod-loaded glycans without ions are intentionally
+            // excluded; correct N-glycan ID is the NO-search's job.
+            var nGlycan = peptideWithMod.AllModsOneIsNterminus.Values
                 .OfType<Glycan>()
-                .Where(g => g.ModificationType == "N-linked glycosylation");
-            foreach (var nGlycan in nGlycans) 
+                .FirstOrDefault(g => g.ModificationType == "N-linked glycosylation" && g.HasIons);
+            // else: no ions → skip (dummy score); users should use NO-search for correct N-glycan ID
+            if (nGlycan != null)
             {
-                if (nGlycan.Ions != null) // guard against null Ions property
-                {
-                    fragmentsForEachGlycoPeptide.AddRange(GlycoPeptides.GetGlycanYIons(theScan.PrecursorMass, nGlycan));
-                }
-                // else: no ions → skip (dummy score); users should use NO-search for correct N-glycan ID
+                fragmentsForEachGlycoPeptide.AddRange(GlycoPeptides.GetGlycanYIons(theScan.PrecursorMass, nGlycan));
             }
 
             var matchedIons = MatchFragmentIons(theScan, fragmentsForEachGlycoPeptide, CommonParameters);
--- a/MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
+++ b/MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
@@ -311,7 +311,6 @@ namespace EngineLayer.GlycoSearch
             var nGlycan = peptideWithMod.AllModsOneIsNterminus.Values
                 .OfType<Glycan>()
                 .FirstOrDefault(g => g.ModificationType == "N-linked glycosylation" && g.HasIons);
-            // else: no ions → skip (dummy score); users should use NO-search for correct N-glycan ID
             if (nGlycan != null)
             {
                 fragmentsForEachGlycoPeptide.AddRange(GlycoPeptides.GetGlycanYIons(theScan.PrecursorMass, nGlycan));
--- a/MetaMorpheus/Test/TestNOGlyco.cs
+++ b/MetaMorpheus/Test/TestNOGlyco.cs
@@ -368,11 +368,13 @@ namespace Test
             Directory.CreateDirectory(outputFolder);
 
             string fakeNGlycanDbPath = null;
-            var nGlycanModsFromNGlycan_ForNoSearch = new List<(string, string)>
-        {
-            ("N-linked glycosylation", "N1 on Nxs"),
-            ("N-linked glycosylation", "N1 on Nxt"),
-    };
+            bool addedFakeDbPath = false;
+            var nGlycanModsUnderTest = new List<(string, string)>
+            {
+                ("N-linked glycosylation", "N1 on Nxs"),
+                ("N-linked glycosylation", "N1 on Nxt"),
+            };
+            var noMods = new List<(string, string)>();
 
             try
             {
@@ -387,26 +389,26 @@ namespace Test
                     if (!GlobalVariables.NGlycanDatabasePaths.Contains(fakeNGlycanDbPath))
                     {
                         GlobalVariables.NGlycanDatabasePaths.Add(fakeNGlycanDbPath);
+                        addedFakeDbPath = true;
                     }
                 }
-                Assert.That(nGlycanModsFromNGlycan_ForNoSearch.All(mod => GlobalVariables.AllModsKnown.Any(m => m.IdWithMotif == mod.Item2)), Is.True,
+                Assert.That(nGlycanModsUnderTest.All(mod => GlobalVariables.AllModsKnown.Any(m => m.IdWithMotif == mod.Item2)), Is.True,
                     "Test setup assumption failed: expected N-glycans mod not found in AllModsKnown.");
 
                 // Confirm the "N1" mods start with null Ions — the exact state that used to NRE in
                 // GetGlycanYIons. This makes the DoesNotThrow below a real regression check.
-                foreach (var (_, nGlycanIdWithMotif) in nGlycanModsFromNGlycan_ForNoSearch)
+                foreach (var (modType, idWithMotif) in nGlycanModsUnderTest)
                 {
-                    if (GlobalVariables.AllModsKnown.First(mod => mod.IdWithMotif == nGlycanIdWithMotif) is Glycan g)
-                    {
-                        Assert.That(g.Ions, Is.Null, $"Test setup assumption failed: expected Glycan {nGlycanIdWithMotif} to have null Ions before the search.");
-                    }
+                    var mod = GlobalVariables.AllModsKnown.Single(m => m.ModificationType == modType && m.IdWithMotif == idWithMotif);
+                    Assert.That(mod, Is.InstanceOf<Glycan>(), $"setup: {idWithMotif} is not a Glycan.");
+                    Assert.That(((Glycan)mod).Ions, Is.Null, $"setup: expected {idWithMotif} null Ions before search.");
                 }
 
                 var commonParameters = new CommonParameters(
                     dissociationType: DissociationType.HCD,
                     ms2childScanDissociationType: DissociationType.EThcD,
-                    listOfModsFixed: asFixedMod ? nGlycanModsFromNGlycan_ForNoSearch : null,
-                    listOfModsVariable: asFixedMod ? null : nGlycanModsFromNGlycan_ForNoSearch);
+                    listOfModsFixed: asFixedMod ? nGlycanModsUnderTest : noMods,
+                    listOfModsVariable: asFixedMod ? noMods : nGlycanModsUnderTest);
 
                 var glycoSearchTask = new GlycoSearchTask
                 {
@@ -425,11 +427,6 @@ namespace Test
 
                 // Before the fix, this call would throw a NullReferenceException from GetGlycanYIons
                 // once a peptide candidate carrying the N-glycan fixed/variable mod reached CreateGsm.
-                //Assert.DoesNotThrow(() =>
-                //{
-                //    new EverythingRunnerEngine(new List<(string, MetaMorpheusTask)> { ("Task", glycoSearchTask) },
-                //        new List<string> { raw }, new List<DbForTask> { db }, outputFolder).Run();
-                //});
                 new EverythingRunnerEngine(new List<(string, MetaMorpheusTask)> { ("Task", glycoSearchTask) },
                         new List<string> { raw }, new List<DbForTask> { db }, outputFolder).Run();
 
@@ -440,9 +437,8 @@ namespace Test
                 {
                     Directory.Delete(outputFolder, true);
                 }
-
-                if (fakeNGlycanDbPath != null) // only remove if it was added
-                    GlobalVariables.NGlycanDatabasePaths.Remove(fakeNGlycanDbPath);
+                // clean up the fake N-glycan database path from GlobalVariables if it was added
+                if (addedFakeDbPath) GlobalVariables.NGlycanDatabasePaths.Remove(fakeNGlycanDbPath);
             }
         }
     }
--- a/MetaMorpheus/Test/TestNOGlyco.cs
+++ b/MetaMorpheus/Test/TestNOGlyco.cs
@@ -396,7 +396,7 @@ namespace Test
                     "Test setup assumption failed: expected N-glycans mod not found in AllModsKnown.");
 
                 // Confirm the "N1" mods start with null Ions — the exact state that used to NRE in
-                // GetGlycanYIons. This makes the DoesNotThrow below a real regression check.
+                // GetGlycanYIons. This makes the unwrapped run below a real regression check.
                 foreach (var (modType, idWithMotif) in nGlycanModsUnderTest)
                 {
                     var mod = GlobalVariables.AllModsKnown.Single(m => m.ModificationType == modType && m.IdWithMotif == idWithMotif);
--- a/MetaMorpheus/Test/TestNOGlyco.cs
+++ b/MetaMorpheus/Test/TestNOGlyco.cs
@@ -360,9 +360,9 @@ namespace Test
             // here is a crash-free run; there is nothing to check on the results (the matched N/O
             // co-glycopeptides are FDR-filtered out anyway).
             //
-            // The pre-search assertion below confirms the "N1" mods start with null Ions — i.e. the
-            // exact state that used to trigger the crash — so DoesNotThrow is meaningfully exercising
-            // the fixed path, not a trivially-safe one.
+            // The pre-search assertion below confirms the "N1" mods start with null Ions — the exact
+            // state that used to trigger the crash — so the unwrapped run below is a real regression
+            // check: any thrown exception fails the test directly (and surfaces the real stack trace).
             string outputFolder = Path.Combine(TestContext.CurrentContext.TestDirectory,
                 $@"TESTGlycoData_{(isNOSearch ? "NO" : "O")}With{(asFixedMod ? "Fixed" : "Var")}NMod");
             Directory.CreateDirectory(outputFolder);
--- a/MetaMorpheus/Test/TestNOGlyco.cs
+++ b/MetaMorpheus/Test/TestNOGlyco.cs
@@ -342,5 +342,108 @@ namespace Test
             Assert.That(site_case3[4] == "Nxt" && site_case3[6] == "T");
             Assert.That(site_case3[14] == "S");
         }
+
+        [Test]
+        [TestCase(true, false)]  // O-search, N-glycan as fixed mod
+        [TestCase(false, false)] // O-search, N-glycan as variable mod
+        [TestCase(true, true)]   // N_O-search, N-glycan as fixed mod
+        [TestCase(false, true)]  // N_O-search, N-glycan as variable mod
+        public static void GlycoSearch_WithNGlycanAsFixedOrVariableMod_DoesNotCrash(bool asFixedMod, bool isNOSearch)
+        {
+            // Integration smoke test: the search pipeline must complete WITHOUT CRASHING when an
+            // N-glycan is used as a fixed or variable modification and reaches GetGlycanYIons.
+            // This is the crash (NullReferenceException) this PR fixes.
+            //
+            // N-glycan-as-a-mod is not hydrated in this mode: if the glycan has no Y ions, the
+            // consumption site skips them (dummy score) rather than generating them. Users should
+            // use NO-search for correct N-glycan identification. So the only assertion that matters
+            // here is a crash-free run; there is nothing to check on the results (the matched N/O
+            // co-glycopeptides are FDR-filtered out anyway).
+            //
+            // The pre-search assertion below confirms the "N1" mods start with null Ions — i.e. the
+            // exact state that used to trigger the crash — so DoesNotThrow is meaningfully exercising
+            // the fixed path, not a trivially-safe one.
+            string outputFolder = Path.Combine(TestContext.CurrentContext.TestDirectory,
+                $@"TESTGlycoData_{(isNOSearch ? "NO" : "O")}With{(asFixedMod ? "Fixed" : "Var")}NMod");
+            Directory.CreateDirectory(outputFolder);
+
+            string fakeNGlycanDbPath = null;
+            var nGlycanModsFromNGlycan_ForNoSearch = new List<(string, string)>
+        {
+            ("N-linked glycosylation", "N1 on Nxs"),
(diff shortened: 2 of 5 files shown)


MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
```diff
@@ -305,12 +305,18 @@ private GlycoSpectralMatch CreateGsm(Ms2ScanWithSpecificMass theScan, int scanIn
...
             var fragmentsForEachGlycoPeptide = GlycoPeptides.OGlyGetTheoreticalFragments(CommonParameters.DissociationType, CommonParameters.CustomIons, peptide, peptideWithMod);
 
-            // if the peptide contains N-glycan, we need to add the Y ion from N-glycan.
-            if (peptideWithMod.AllModsOneIsNterminus.Any(p=>p.Value.ModificationType == "N-linked glycosylation"))
+            // If the peptide contains N-glycan, we need to add the Y ion from N-glycan.
+            // At first, we only consider the mod is N-linked glycosylation and as a glycan object.
+            var nGlycans = peptideWithMod.AllModsOneIsNterminus.Values
+                .OfType<Glycan>()
+                .Where(g => g.ModificationType == "N-linked glycosylation");
+            foreach (var nGlycan in nGlycans)
```
trishorts: **Undeclared behaviour change: `First(...)` became a loop over *all* matching glycans.**

The base code contributed exactly one Y-ion series no matter how many N-glycan mods the peptide carried:

```csharp
if (AllModsOneIsNterminus.Any(p => p.Value.ModificationType == "N-linked glycosylation"))
{
    Glycan nGlycan = AllModsOneIsNterminus.First(...).Value as Glycan;
    fragmentsForEachGlycoPeptide.AddRange(GetGlycanYIons(theScan.PrecursorMass, nGlycan));
}
```

The new `foreach` calls `AddRange` once **per** matching glycan. This matters because `GetGlycanYIons(precursorMass, glycan)` builds a *whole-precursor* series (`GlycoPeptides.cs:77` computes Y0 as `precursorMass - glycan.Mass`), not a per-glycan partial series. A peptide with two N-sequons and an N-glycan **fixed** mod gets two N-glycan mods attached during digestion -- an entirely reachable configuration in these searches -- and `fragmentsForEachGlycoPeptide` then receives two overlapping whole-precursor Y-ion series. That inflates the theoretical fragment list feeding `MatchFragmentIons`, and duplicate matched ions flow into `CalculatePeptideScore`/`GlycanScore`.

This isn't mentioned in the title, body, or commits, which describe the change as the `as Glycan` cast fix plus the null-`Ions` guard.

**Suggested fix** -- keep the fix strictly crash-only by restoring single-glycan semantics:

```csharp
var nGlycan = peptideWithMod.AllModsOneIsNterminus.Values
    .OfType<Glycan>()
    .FirstOrDefault(g => g.Type == GlycanType.N_glycan && g.HasIons);

if (nGlycan != null)
{
    fragmentsForEachGlycoPeptide.AddRange(GlycoPeptides.GetGlycanYIons(theScan.PrecursorMass, nGlycan));
}
```

This also restores the base code's short-circuit on the overwhelming majority of candidates (no N-glycan mod at all), which the two-stage `OfType`/`Where` pipeline currently gives up on a `Parallel.ForEach` hot path (`GlycoSearchEngine.cs:121`).

If multiple Y-ion series *is* the intent, that's a legitimate design choice -- but please say so in the PR body and confirm `GetGlycanYIons` composes correctly under repeated application against the same precursor mass, because today it does not obviously do so.

**How to test it:** a direct unit test is much cheaper than another `EverythingRunnerEngine` run. Build a `PeptideWithSetModifications` carrying two `Glycan` N-linked mods with populated `Ions` (use `GlycanDatabase.LoadGlycan(path, ToGenerateIons: true, IsOGlycan: false)` to get hydrated glycans), call the selection block's logic, and assert the theoretical `ProductType.M` ion set has the expected count with no duplicate `NeutralMass` values. Add the single-glycan case alongside it asserting the fragment list is identical to the pre-PR baseline.

MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
```diff
@@ -305,12 +305,18 @@ private GlycoSpectralMatch CreateGsm(Ms2ScanWithSpecificMass theScan, int scanIn
...
+            // At first, we only consider the mod is N-linked glycosylation and as a glycan object.
+            var nGlycans = peptideWithMod.AllModsOneIsNterminus.Values
+                .OfType<Glycan>()
+                .Where(g => g.ModificationType == "N-linked glycosylation");
+            foreach (var nGlycan in nGlycans) 
             {
-                Glycan nGlycan = peptideWithMod.AllModsOneIsNterminus.First(p =>
-                    p.Value.ModificationType == "N-linked glycosylation").Value as Glycan;
-                fragmentsForEachGlycoPeptide.AddRange(GlycoPeptides.GetGlycanYIons(theScan.PrecursorMass, nGlycan));
+                if (nGlycan.Ions != null) // guard against null Ions property
```
trishorts: **The guard is in the right spirit but the wrong place -- and it leaves a second identical crash reachable.**

Two observations:

**(a) `!= null` tests the reference, not the property you describe.** The PR body says the intent is to contribute Y ions only "when they actually have them," and the trailing comment says "no ions -> skip." A `Glycan` whose `Ions` list is non-null but *empty* passes this guard. That case is harmless today (`GetGlycanYIons` just returns the lone Y0 ion) but the condition doesn't match the documented contract.

**(b) The unguarded dereference is still there, for the next caller.** `GlycoPeptides.GetGlycanYIons` has two overloads and *both* iterate `glycan.Ions` with no null check -- `GlycoPeptides.cs:78` and `GlycoPeptides.cs:91`. This PR guards one of the two call sites of the first overload. The `PeptideWithSetModifications` overload (`GlycoPeptides.cs:86`) and `MatchBestGlycan` (`GlycoPeptides.cs:99`) have no callers today, so nothing crashes right now -- but the moment either is wired up against a mod-loaded glycan, the same NRE returns and needs its own call-site guard.

Worth noting the codebase already answers this question elsewhere: `Glycan`'s own constructor null-guards `Ions` **twice** (`Glycan.cs:49` and `Glycan.cs:62`) before building neutral losses. The established idiom is that a glycan with no generated ions is the *type's* business, not each caller's.

**Suggested fix** -- add a property beside `Ions` (`Glycan.cs:138`):

```csharp
/// <summary>
/// True when this glycan's Y-ion fragments have been generated. Glycans loaded for the
/// fixed/variable-mod list (GlobalVariables.LoadGlycans, ToGenerateIons: false) have none.
/// </summary>
public bool HasIons => Ions != null && Ions.Count > 0;
```

then make both overloads null-safe at the source, e.g. in `GetGlycanYIons(double precursorMass, Glycan glycan)`:

```csharp
List<Product> YIons = new List<Product>();
YIons.Add(new Product(ProductType.M, FragmentationTerminus.Both,
    precursorMass - (double)glycan.Mass / 1E5, 0, 0, (double)glycan.Mass / 1E5)); // Y0
if (!glycan.HasIons)
{
    return YIons; // N-glycan used as a plain fixed/variable mod: no Y ions were generated.
                  // Not scored here by design; correct N-glycan ID is the NO-search's job.
}
foreach (var ion in glycan.Ions) { ... }
```

Note this is a small behavioural difference from the current diff: it still contributes the **Y0** ion (bare peptide, whole glycan lost), which is computed from `glycan.Mass` alone and so is always available. That's strictly more information than skipping the glycan entirely, and it's the honest fragment set for "glycan attached, substructure unknown." If you prefer to keep the skip semantics exactly as they are, `if (nGlycan.HasIons)` at this call site still reads better than an inline null check.

I realise `GetGlycanYIons` is outside the three-file change set -- but it's the function named in this PR's own title and root-cause section, so hardening it there is the payload rather than scope creep, and it costs two lines.

**How to test it:** a unit test straight on the function, which is far cheaper than the integration run and pins the contract where it belongs:

```csharp
[Test]
public static void GetGlycanYIons_GlycanWithNoIons_ReturnsY0Only()
{
    var kind = GlycanDatabase.String2Kind("HexNAc(2)Hex(5)");
    var glycan = new Glycan(kind, "Nxs", GlycanType.N_glycan); // Ions never assigned -> null
    Assert.That(glycan.HasIons, Is.False);

    var yIons = GlycoPeptides.GetGlycanYIons(2000.0, glycan);

    Assert.That(yIons.Count, Is.EqualTo(1));
    Assert.That(yIons[0].ProductType, Is.EqualTo(ProductType.M));
}
```
Add the `Ions = new List<GlycanIon>()` (empty, non-null) variant as a second case -- that's the one the current `!= null` check lets through.

MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
```diff
@@ -305,12 +305,18 @@ private GlycoSpectralMatch CreateGsm(Ms2ScanWithSpecificMass theScan, int scanIn
 
             var fragmentsForEachGlycoPeptide = GlycoPeptides.OGlyGetTheoreticalFragments(CommonParameters.DissociationType, CommonParameters.CustomIons, peptide, peptideWithMod);
 
-            // if the peptide contains N-glycan, we need to add the Y ion from N-glycan.
-            if (peptideWithMod.AllModsOneIsNterminus.Any(p=>p.Value.ModificationType == "N-linked glycosylation"))
+            // If the peptide contains N-glycan, we need to add the Y ion from N-glycan.
+            // At first, we only consider the mod is N-linked glycosylation and as a glycan object.
+            var nGlycans = peptideWithMod.AllModsOneIsNterminus.Values
+                .OfType<Glycan>()
+                .Where(g => g.ModificationType == "N-linked glycosylation");
```
trishorts: **Two smaller points on the selection itself.**

**(a) The `OfType<Glycan>()` half of the fix is undocumented.** It converts a crash into a *silent drop*: a plain `Modification` carrying `ModificationType == "N-linked glycosylation"` -- exactly the database/GPTMD annotation case the PR body calls out -- is now invisible to this block. That is unambiguously better than an NRE, but there are now **two distinct** skip paths (wrong runtime type; no `Ions`) and only the second has a comment. A future reader can't tell whether the type filter is a deliberate contract or an accident awaiting follow-up.

Consider replacing the "At first, we only consider..." narration on line 309 with the contract itself: only `Glycan`-typed N-linked modifications carrying Y-ion fragments contribute Y ions here, everything else is intentionally excluded, and correct N-glycan identification is the dedicated NO-search's job.

**(b) After `OfType<Glycan>()` you have a typed discriminator available, so the string compare is redundant.** `Glycan` sets `ModificationType` *from* `Type` in its constructor (`Glycan.cs:48` and `Glycan.cs:61`), and `Type` is public (`Glycan.cs:118`). Once the sequence is narrowed to `Glycan`, `g.Type == GlycanType.N_glycan` is the equivalent check and won't silently disable all N-glycan Y-ion generation if the type string is ever renamed upstream:

```csharp
var nGlycans = peptideWithMod.AllModsOneIsNterminus.Values
    .OfType<Glycan>()
    .Where(g => g.Type == GlycanType.N_glycan);
```

(To be clear, this is different from "hoist the magic string to a constant" -- the point is that a typed enum already exists on the object you've just narrowed to.)

**How to test it:** add a `[TestCase]` that supplies a plain `Modification` with `ModificationType = "N-linked glycosylation"` (not a `Glycan`) through `listOfModsVariable`, and assert the search completes without a `NullReferenceException` and that no `ProductType.M` ions are attributed to it. Right now nothing distinguishes `OfType<Glycan>()` from the old `Where(...).Select(... as Glycan)` form -- reverting it would leave the whole suite green, which makes half of the stated fix untested.

MetaMorpheus/Test/TestNOGlyco.cs
```diff
@@ -342,5 +342,108 @@ public static void TestGetModPos()
...
+                    }
+                }
+                Assert.That(nGlycanModsFromNGlycan_ForNoSearch.All(mod => GlobalVariables.AllModsKnown.Any(m => m.IdWithMotif == mod.Item2)), Is.True,
+                    "Test setup assumption failed: expected N-glycans mod not found in AllModsKnown.");
+
+                // Confirm the "N1" mods start with null Ions — the exact state that used to NRE in
+                // GetGlycanYIons. This makes the DoesNotThrow below a real regression check.
+                foreach (var (_, nGlycanIdWithMotif) in nGlycanModsFromNGlycan_ForNoSearch)
+                {
+                    if (GlobalVariables.AllModsKnown.First(mod => mod.IdWithMotif == nGlycanIdWithMotif) is Glycan g)
```
trishorts: **This precondition can execute zero assertions and still pass -- which removes the only thing making this test a regression check.**

```csharp
if (GlobalVariables.AllModsKnown.First(mod => mod.IdWithMotif == nGlycanIdWithMotif) is Glycan g)
{
    Assert.That(g.Ions, Is.Null, ...);
}
```

If the matched entry is a plain `Modification` rather than a `Glycan` -- precisely the case the PR body says can occur -- the `if` falls through, the loop body does nothing, and all four `TestCase`s pass. The method comment above claims this block "confirms the `N1` mods start with null Ions ... so `DoesNotThrow` is meaningfully exercising the fixed path, not a trivially-safe one," but the code doesn't enforce that; it merely checks it when convenient.

**Suggested fix** -- assert the type instead of branching on it, so every iteration produces at least one executed assertion:

```csharp
foreach (var (_, nGlycanIdWithMotif) in nGlycanModsUnderTest)
{
    var mod = GlobalVariables.AllModsKnown.First(m => m.IdWithMotif == nGlycanIdWithMotif);
    Assert.That(mod, Is.InstanceOf<Glycan>(),
        $"Test setup assumption failed: {nGlycanIdWithMotif} is not a Glycan.");
    Assert.That(((Glycan)mod).Ions, Is.Null,
        $"Test setup assumption failed: expected Glycan {nGlycanIdWithMotif} to have null Ions before the search.");
}
```

**How to test the test:** temporarily change the looked-up `IdWithMotif` to one you know resolves to a non-`Glycan` modification and confirm the test now fails with the `Is.InstanceOf<Glycan>` message rather than passing silently. That one-minute check is what proves the guard is load-bearing.

Related: this precondition reads a mutable field on process-wide static `GlobalVariables.AllModsKnown` instances shared with every other test in the assembly, so it's order-dependent by construction. Worth a one-line comment saying so, since anything that later hydrates those `Ions` (including the load-site alternative discussed in the PR body) flips this to a confusing "setup assumption failed" rather than a clear signal.

MetaMorpheus/Test/TestNOGlyco.cs
```diff
@@ -342,5 +342,108 @@ public static void TestGetModPos()
...
+                        MaximumOGlycanAllowed = 1,
+                    }
+                };
+
+                DbForTask db = new(Path.Combine(TestContext.CurrentContext.TestDirectory, @"GlycoTestData/FiveMucinFasta.fasta"), false);
+                string raw = Path.Combine(TestContext.CurrentContext.TestDirectory, @"GlycoTestData/N_O_glycoWithFileSpecific/2019_09_16_StcEmix_35trig_EThcD25_rep1_4999_5500.mzML");
+
+                // Before the fix, this call would throw a NullReferenceException from GetGlycanYIons
+                // once a peptide candidate carrying the N-glycan fixed/variable mod reached CreateGsm.
+                //Assert.DoesNotThrow(() =>
```
trishorts: **Leftover commented-out assertion block, and two comments that still describe it.**

Lines 428-432 carry a dead `//Assert.DoesNotThrow(() => { ... new EverythingRunnerEngine(...).Run(); });` immediately above the live, uncommented `new EverythingRunnerEngine(...).Run();`. Both the method-level comment ("so `DoesNotThrow` is meaningfully exercising the fixed path") and the inline comment above the dead block reference an assertion form the file no longer has.

**Suggested fix:** delete lines 428-432 and reword the two comments to describe what the test actually does -- an unwrapped run, where a thrown exception fails the test directly. (That's the better form anyway: NUnit reports the real stack trace instead of `DoesNotThrow`'s wrapper.) Also worth renaming `nGlycanModsFromNGlycan_ForNoSearch` (line 371): the name asserts these mods come from `NGlycan_ForNoSearch.gdb`, but the test never registers or loads that file -- the NO-search cases register `NGlycan_fakeN-glycan.gdb` and the O-search cases register nothing. They actually resolve out of whatever `GlobalVariables.AllModsKnown` holds at startup, and the current name sends a debugger to the wrong file.

**How to test it:** no behavioural change, so the four existing `TestCase`s passing unchanged is the whole check.

MetaMorpheus/Test/TestNOGlyco.cs
```diff
@@ -342,5 +342,108 @@ public static void TestGetModPos()
...
+
+            }
+            finally
+            {
+                if (Directory.Exists(outputFolder)) // clean up the output folder after the test
+                {
+                    Directory.Delete(outputFolder, true);
+                }
+
+                if (fakeNGlycanDbPath != null) // only remove if it was added
```
trishorts: **The cleanup can delete another test's global registration.**

The guard here tests whether the *variable was assigned*, not whether *this test performed the `Add`*. `fakeNGlycanDbPath` is set unconditionally at line 386, while the `Add` at line 389 is conditional on the path being absent (line 387). So when the entry was already present -- registered by another fixture, or by an earlier `TestCase` of this very method -- this `finally` still removes it, mutating shared static `GlobalVariables.NGlycanDatabasePaths` state the test doesn't own. That produces order-dependent failures in whichever test registered it first, which is unpleasant to diagnose in a full-assembly run.

**Suggested fix** -- track the actual `Add`:

```csharp
bool addedFakeDbPath = false;
...
if (!GlobalVariables.NGlycanDatabasePaths.Contains(fakeNGlycanDbPath))
{
    GlobalVariables.NGlycanDatabasePaths.Add(fakeNGlycanDbPath);
    addedFakeDbPath = true;
}
...
finally
{
    if (addedFakeDbPath)
        GlobalVariables.NGlycanDatabasePaths.Remove(fakeNGlycanDbPath);
}
```

**How to test it:** run the four `TestCase`s and assert `GlobalVariables.NGlycanDatabasePaths` has the same contents before and after -- or simply pre-add the fake path in a throwaway `[Test]` that runs first and confirm it survives. Since the two NO-search cases both hit this path, the second one already depends on the first one's cleanup being correct.

MetaMorpheus/Test/TestNOGlyco.cs
```diff
@@ -342,5 +342,108 @@ public static void TestGetModPos()
...
+                    if (GlobalVariables.AllModsKnown.First(mod => mod.IdWithMotif == nGlycanIdWithMotif) is Glycan g)
+                    {
+                        Assert.That(g.Ions, Is.Null, $"Test setup assumption failed: expected Glycan {nGlycanIdWithMotif} to have null Ions before the search.");
+                    }
+                }
+
+                var commonParameters = new CommonParameters(
+                    dissociationType: DissociationType.HCD,
+                    ms2childScanDissociationType: DissociationType.EThcD,
+                    listOfModsFixed: asFixedMod ? nGlycanModsFromNGlycan_ForNoSearch : null,
```
trishorts: **`null` here doesn't mean "no mods" -- `CommonParameters` substitutes its built-in defaults.**

`CommonParameters.cs:132-133`:

```csharp
ListOfModsVariable = listOfModsVariable ?? new List<(string, string)> { ("Common Variable", "Oxidation on M") };
ListOfModsFixed    = listOfModsFixed    ?? new List<(string, string)> { ("Common Fixed", "Carbamidomethyl on C"), ("Common Fixed", "Carbamidomethyl on U") };
```

So the `asFixedMod: true` arm runs with default **variable** mods active (Oxidation on M), and the `asFixedMod: false` arm runs with default **fixed** mods active (Carbamidomethyl). The two arms therefore differ in more than the one dimension `asFixedMod` names, and the search spaces aren't comparable -- a failure in one arm but not the other couldn't be attributed to the fixed/variable distinction.

**Suggested fix** -- pass an explicit empty list for the unused slot so only the N-glycan placement varies:

```csharp
var noMods = new List<(string, string)>();
var commonParameters = new CommonParameters(
    dissociationType: DissociationType.HCD,
    ms2childScanDissociationType: DissociationType.EThcD,
    listOfModsFixed:    asFixedMod ? nGlycanModsUnderTest : noMods,
    listOfModsVariable: asFixedMod ? noMods : nGlycanModsUnderTest);
```

If leaving the defaults in place is deliberate (e.g. you want a realistic search), a one-line comment saying so is enough -- the problem is only that it currently reads as "no other mods."

**How to test it:** assert on the constructed parameters before running, e.g. `Assert.That(commonParameters.ListOfModsVariable, Is.Empty)` in the fixed arm. That pins the intent and would have caught the `??` substitution.

MetaMorpheus/Test/TestNOGlyco.cs
```diff
@@ -342,5 +342,108 @@ public static void TestGetModPos()
...
+                    // any N-glycan to come from the "N1" mod, so candidates route through FindOGlycan ->
+                    // CreateGsm (the GetGlycanYIons path this test guards) instead of being explained by
+                    // the N-glycan search.
+                    fakeNGlycanDbPath = Path.Combine(TestContext.CurrentContext.TestDirectory, "GlycoTestData", "NGlycan_fakeN-glycan.gdb");
+                    if (!GlobalVariables.NGlycanDatabasePaths.Contains(fakeNGlycanDbPath))
+                    {
+                        GlobalVariables.NGlycanDatabasePaths.Add(fakeNGlycanDbPath);
+                    }
+                }
+                Assert.That(nGlycanModsFromNGlycan_ForNoSearch.All(mod => GlobalVariables.AllModsKnown.Any(m => m.IdWithMotif == mod.Item2)), Is.True,
```
trishorts: **This setup assertion checks only half of the tuple it's validating.**

`nGlycanModsUnderTest` is a list of `(ModificationType, IdWithMotif)` pairs, and `CommonParameters` resolves fixed/variable mods by **both** components. The assertion only checks `m.IdWithMotif == mod.Item2`; `Item1` (`"N-linked glycosylation"`) is never validated, and nothing later in the test verifies that the N-glycan mod was actually attached to any candidate peptide during the run.

That matters because if the type string doesn't match what `AllModsKnown` carries, the mod is silently dropped during digestion, no peptide ever reaches `CreateGsm` carrying an N-glycan, and all four `TestCase`s pass **without entering the code path this PR fixes**. The test would go green for the wrong reason.

**Suggested fix** -- assert on the full tuple, and reuse the matched instance for the null-`Ions` check below so setup and search agree on which modification is under test:

```csharp
foreach (var (modType, idWithMotif) in nGlycanModsUnderTest)
{
    var match = GlobalVariables.AllModsKnown.SingleOrDefault(
        m => m.ModificationType == modType && m.IdWithMotif == idWithMotif);
    Assert.That(match, Is.Not.Null,
        $"Test setup assumption failed: no known mod with type '{modType}' and id '{idWithMotif}'.");
    Assert.That(match, Is.InstanceOf<Glycan>(), ...);
    Assert.That(((Glycan)match).Ions, Is.Null, ...);
}
```

**How to test it:** flip the type string to something bogus (`"N-linked glycosylationX"`) and confirm the test fails at setup with your message, rather than running a vacuous search and passing. Ideally also add a positive assertion that at least one PSM in the results carries the N-glycan mod, so the test proves the guarded path was reached rather than merely assuming it -- if FDR filtering makes that impossible at results level, a counter or a `CreateGsm`-level unit test would give you the same proof more cheaply.

trishorts (commented):
## Automated review + code-oracle pass (MetaMorpheus PR #2706)

Thanks for chasing this down -- the diagnosis in the PR body is correct and I was able to confirm every step of it against the source. `GlobalVariables.cs:505` does load N-glycans with `ToGenerateIons: false`, `GlycanDatabase.cs:197-210` gates every `Ions` assignment on that flag, and `GlycoSearchEngine.cs:72/81` loads with `true` -- so the dedicated N-glycan path is genuinely unaffected and `CreateGsm` was the one exposed consumption site. The null-`Ions` guard is sound in substance.

Comments below are grouped into three code-side and five test-side items; each carries a suggested fix and a suggested test. Two things I'd treat as blocking:

1. **`First(...)` -> `foreach` is an undeclared behaviour change** (comment on line 313). A peptide with two N-sequons plus an N-glycan fixed mod now gets two whole-precursor Y-ion series appended instead of one. Nothing in the PR describes this and nothing tests it.
2. **The new test's only meaningful precondition can execute zero assertions** (comment on line 399). The `is Glycan g` pattern falls through silently, so the test can pass without exercising the fixed path at all.

The rest are smaller: guard placement, an undocumented second skip path, and four test-hygiene items around shared static state and `null`-means-default.

### Two notes that are explicitly *not* change requests

**Nothing here belongs upstream in mzLib.** I checked, since `Glycan` derives from mzLib's `Omics.Modifications.Modification`. mzLib has no glycan domain at all -- no `Glycan`, `GlycanIon`, `GlycanDatabase`, or `ToGenerateIons`; the only glyco code upstream is result-*reading* (`Readers/.../GlycoPsmFromTsv.cs`) and a digestion-level filter (`Protein.GetGlycoPeptides`). Fixing this in MetaMorpheus is correct, and no counterpart PR or package bump is needed.

**The "skip rather than hydrate" decision is defensible, and it'd be worth saying *why* in the PR body.** A generator already exists -- `GlycanDatabase.NGlycanCompositionFragments(byte[] kind)` (`GlycanDatabase.cs:304`) -- so hydrating would have been a one-liner, and a reviewer who finds it will ask why you didn't. Its own header comment answers that: *"The fragment generation is not as good as structure based method. So it is better to use a structure based N-Glycan database ... we don't use it now."* Citing that turns a judgement call into a documented one.

One genuine open question for whoever owns the glyco layer (**not** something to fix in this PR): `LoadTxtGlycan` -- the *other* loader feeding `AllModsKnown` -- **does** hydrate `Ions` unconditionally (`GlobalVariables.cs:576` and `:581`). So the same N-glycan composition behaves differently depending on whether it arrived via `glyco.txt` or via a `.gdb` file: one gets Y-ion scoring, the other silently doesn't. This PR makes the `.gdb` branch safe, which is the right call for a bug fix, but the inconsistency between the two loaders is worth a follow-up issue.

<sub>Findings produced by a local automated review pipeline plus a prior-art/conventions pass over the mzLib and MetaMorpheus clones; every claim above was verified by reading the current source, and the file:line references are to `master` as of this review. Reviewed and edited by a human before posting.</sub>

MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
RayMSMS: Added HasIons and used it at the call site (FirstOrDefault(g => g.Type == GlycanType.N_glycan && g.HasIons)), so an unhydrated glycan never reaches GetGlycanYIons on this path. I'd prefer to leave GetGlycanYIons's internals out of this crash-fix PR: the second overload and MatchBestGlycan have no callers today, so hardening them isn't exercised by anything here. Happy to file a follow-up issue to make both overloads null-safe at the source (returning Y0) if you'd like that tracked separately

trishorts (approved):
**Re-reviewed at `9f189e1`. All eight of my earlier comments are addressed — approving.**

Confirmed one by one against the diff:

- **L313** — `foreach` reverted to `FirstOrDefault`, restoring the single Y-series behaviour master had at `GlycoSearchEngine.cs:309-312`. Correct.
- **L312** — the narration is replaced with an actual contract statement naming both skip paths. Exactly what I asked for.
- **L315(a)** — `HasIons => Ions != null && Ions.Count > 0` closes the non-null-but-empty case, so the property now matches the documented intent.
- **L399** — `Single(...)` plus unconditional `Is.InstanceOf<Glycan>()` / `Ions Is.Null`. The precondition can no longer pass vacuously.
- **L428** — dead `Assert.DoesNotThrow` block removed and both comments reworded to describe the unwrapped run.
- **L444** — `addedFakeDbPath` is set only inside the `Add` branch, so the `finally` no longer removes a registration this test didn't make.
- **L408** — `noMods` in both arms; the two arms now differ only in the dimension `asFixedMod` names.
- **L392** — both tuple components are now validated via `Single(m => m.ModificationType == modType && m.IdWithMotif == idWithMotif)`, and the matched instance is reused for the null-`Ions` check. That was the substance of the comment.

Two things worth recording, both of which I traced rather than assumed:

**The unwrapped run is a genuine regression check.** `MetaMorpheusTask.RunTask` catches broadly at `MetaMorpheusTask.cs:683` to write `results.txt`, but it rethrows at line 699, and `EverythingRunnerEngine.Run()` has no catch around `RunTask`. So an NRE from `GetGlycanYIons` propagates out and fails the test directly. Dropping the `DoesNotThrow` wrapper was the right call — it also surfaces the real stack trace.

**On `g.Type == GlycanType.N_glycan` replacing the `ModificationType` string comparison.** This is safe: the `Glycan` constructor assigns `ModificationType` *from* `type` (`Glycan.cs:48` and `:61`), and those are its only assignments, so the two predicates are equivalent by construction. A glycan that took the constructor's `O_glycan` default would carry `"O-linked glycosylation"` and would have failed the old string test too. No behaviour change at this site.

**Re: `GetGlycanYIons` hardening** — agreed, and out of scope for a crash-fix PR. Both overloads live in a file this PR doesn't touch, and neither the second overload nor `MatchBestGlycan` has a caller today. Worth a follow-up issue if you want it tracked, but I don't consider it owed here.

I've resolved all eight review threads. Nice, tight fix — thanks for working through the detail on it.

zhuoxinshi (approved):

nbollis (approved):

PR #2709, opened 2026-08-12
Making the MetaMorpheus GUI usable on macOS and Linux

acesnik:
🤖

## Making the MetaMorpheus GUI usable on macOS and Linux

MetaMorpheus's engine is cross-platform and the CLI runs anywhere, but the GUI is WPF and
therefore Windows-only. Users on macOS and Linux can run searches, but only by hand-writing
TOML — there is no way to build a task list, adjust settings, or see progress.

This adds an Avalonia front end that runs on all three platforms. It is deliberately additive: the
WPF GUI remains the Windows experience, and `GuiFunctions` is untouched. Deleting the two new
directories reverts the front end itself.

It is no longer confined to those two directories, so the blast radius in full — `CommonParameters`
is load-bearing enough that it should not have to be found in the diff:

| file | diff | why |
|---|---|---|
| `EngineLayer/CommonParameters.cs` | +46 −7 | `CloneWithNewValues`, so a front end editing a subset of the settings cannot silently reset the rest; `CloneWithNewDissociationType` now delegates to it |
| `EngineLayer/DatabaseLoading/DbForTask.cs` | +9 | `LooksLikeContaminant`, the shared contaminant filename heuristic |
| `CMD/Program.cs` | +1 −12 | drops its private copy of that heuristic |
| `GUI/ForDisplayingInDataGrids/ProteinDbForDataGrid.cs` | +1 −5 | drops its inline copy |
| `Test/CommonParametersCloneAndContaminantTests.cs` | +218 | behaviour pins for both of the above |
| `MetaMorpheus.sln` | +16 | the two new projects |
| `.github/workflows/Test.yml`, `build.yml` | +34, +8 −2 | build and test them on all three OSes |

An earlier version of this paragraph said `GUI` and `GuiFunctions` were both untouched. That was true
when written; routing the two existing callers through the shared heuristic changed it for `GUI`.
`GuiFunctions` still is.

**Opened as a draft for a direction decision rather than for merge.** That decision has now been
given (@trishorts, converge rather than coexist), and the review's line-level findings are addressed —
see the update below.

### What works

- Building a task list, adding spectra and databases, running `EverythingRunnerEngine`,
  live progress and log output
- Settings dialogs for all six task types (Search, Calibrate, GPTMD, XLSearch, GlycoSearch, Average),
  each reachable from its own button
- Fixed and variable modification selection over the whole modification set, grouped and filterable,
  plus GPTMD's `ListOfModsGptmd` on the same widget
[...]

Issue #1142, opened 2018-04-26
file type confusion in load


trishorts:
Here is output from spritz where we get an unrecognized file type error. I know this happens for other files, where I think that it shouldn't. There should be a way to isolate the final .### and identify it as the file type and ignore all the previous "."s 

SRR1025675_1-trimmedAligned.sortedByCoord.outProcessed.out.fixedQuals.split.snpEffAnnotated.protein.fasta

rmillikin:
This is how Path.GetExtension() works; it isolates the text after the last "." and uses this as the extension. I decided to not use that function because it messed up .xml.gz (it thought the extension was .gz). So what it does now it is assumes that whatever is after the first dot in the filename is the extension. There may be a more complex/robust way of handling that but I think the easy answer is just to remove the dots in the filename except the one before the extension

trishorts:
I'd like to have it handle the more complex scenerio. Anything ending in .gz should be examined again to see which bin it goes to (either raw file or database). That way we can use compressed for everything. This is not urgent, but I would like it to get done. Early job for the next undergrad.

rmillikin:
see #1164

Issue #1141, opened 2018-04-26
warning: non-specific enzyme designed for classic/modern search not non-specific search


trishorts:
seems counter-intuitive to me.

zrolfs:
"Please use 'SingleN' or 'SingleC'" 
Them's the rules.
I could make non-specific protease default to SingleN or SingleC when doing a non-specific search mode?

trishorts:
I guess I don't understand non-specific search, or at least I don't understand why it is named that if it is not actually a non-specific search. Also don't know why you'd specify an enzyme or something like it (SingleN or SingleC) for a non-specific search. while I'm at it, what do SingleN and SingleC mean? I think more real-time support for these options by way of tool tips, or more accurate and obvious naming would help.

zrolfs:
Let me know if that's coherent or if something needs more elucidation.

trishorts:
Seems like a "Proteoform Fragment Search" might be a better name, since you are looking for a single cleavage in a proteoform either from the N or C terminus. I could be wrong, but I think that in common usage, non-specific search is used more for ID of peptides, which are often internal proteoform fragments produced enzymatically by, for example, the proteasome.

zrolfs:
Perhaps I was not clear... We are looking for peptides. The output of this search is the same as if you selected classic/modern+non-specific protease. The cleavage of the proteins is simply the algorithm that is a means to obtain the peptide ids. They do not need to be on the termini of the protein.

trishorts:
even more confusing. What does picking singlen or singlec do then? and why bother with that? and why not allow picking non-specific enzyme for non-specific search?

zrolfs:
I will allow non-specific protease to by used in the GUI be defaulting to SingleN or SingleC behind the scenes. That should prevent this confusion.

zrolfs:
I understand your frustration.
""SingleN" or "SingleC", which cleaves only once per residue and uses either N- or C-terminal ions, respectively."
These proteases cleave each residue only once, leaving either the C or N terminus fixed. These are needed for the algorithm, as explained above. Using the non-specific protease cleaves every residue hundreds of times, where the gains made by the algorithm would be mute. The choice between SingleN and SingleC does not particularly matter, as both methods give similar results. The user does not need to know about these proteases; they can be left in the background.


If you would prefer, I can wipe Non-Specific Search Mode from the GUI until further revision.

zrolfs:
[...]

PR #2710, opened 2026-08-12
Move from .NET 8 to .NET 10

acesnik:
🤖

## Move from .NET 8 to .NET 10

.NET 8 leaves support in **November 2026**. .NET 10 is the current LTS, supported to November 2028.

All seven projects move to `net10.0` / `net10.0-windows`. No package versions change, and nothing
needed a code change to compile.

### What changed

- The seven project files, plus `global.json`
- CI: `setup-dotnet` versions, `DOTNET_FRAMEWORK`, the hardcoded framework folder in the release
  zip step and in the two Docker publish steps
- `Dockerfile`: `dotnet/runtime:8.0-alpine` → `10.0-alpine`
- README: the .NET 8 requirement pointed at four pinned SDK 8.0.401 installer links. It now points
  at the .NET 10 download page, which covers all platforms and does not go stale on every patch.

`GUI_h05aw1sq_wpftmp.csproj` is deliberately untouched — it is a WPF build temp file that happens to
be tracked, not a project.

### One .NET version per workflow

`build.yml` and `InstallRunAndArtifact.yml` already declared `DOTNET_VERSION` and `DOTNET_FRAMEWORK`
and used them throughout. `BuildReleaseArtifact.yml`, `DockerUpload.yml` and `Test.yml` hardcoded the
version in five places between them; they now declare the same two variables and reference them.
Changing .NET version is one edit per workflow plus `global.json`.

CI asks for `10.0.x` rather than a fixed feature band, so it picks up new SDK releases without a
bump. `global.json` keeps a concrete version because there it acts as a floor, and
`rollForward: latestFeature` carries it to whatever CI installs.

### Verification

On macOS/arm64 with SDK 10.0.400:

- the solution builds under `UbuntuMac` with 0 errors
- `Test.csproj` compiles with `EnableWindowsTargeting=true`
- `CMD.dll --test` runs Calibrate → GPTMD → Search to completion

That last run gives **89 PSMs and 47 proteins at 1% FDR — the same as master on .NET 8**, and
`AllPSMs.psmtsv`, `AllPeptides.psmtsv` and `AllQuantifiedPeptides.tsv` are byte-identical between the
two runs. Both runs produced the same 25 output files. So this is a framework move with no visible
change to results.

Not verified here: the test suite, which is `net10.0-windows` and cannot run on macOS. CI covers it.

### iText 8.0.5 → 9.7.0 (folded in from #2711)

Moving to .NET 10 turns on transitive package auditing, which surfaced four advisories. One is ours
to fix: iText 8.0.5 pulls `BouncyCastle.Cryptography` 2.2.1, carrying three moderate advisories
(Marvin timing side-channel, infinite loop on a crafted signature, CPU exhaustion parsing
certificates). 8.0.5 is the last 8.x, so the upstream fix is only in 9.x, which uses 2.6.2.

I opened #2711 pinning BouncyCastle 2.6.2 as a workaround, then tested the major upgrade instead of
assuming it was risky — and it isn't. GuiFunctions and Test both build against 9.7.0 with **no
source changes**. MetaDraw uses a narrow, stable part of the API (`iText.IO.Image`,
`iText.Kernel.Pdf`, `iText.Layout`, `iText.StyledXmlParser.Jsoup`), and PDF export is exercised by
`MetaDrawTest` and `ChimeraAnalysisTabViewModelTests`, so CI covers the runtime behaviour a compile
cannot. #2711 is closed in favour of this.

The other three advisories — `System.Drawing.Common` (critical), `System.Data.SqlClient` (high) and
`OpenMcdf` (moderate) — all come from mzLib and are handled in smith-chem-wisc/mzLib#1142. None of
them are new; .NET 10 only made them visible.

### Installer

`System.Drawing.Common` and `System.CodeDom` are no longer copied to the GUI's output, because
`net10.0-windows` gets them from the Windows Desktop shared framework and the SDK prunes the
packages (NU1510). Plain `net10.0` does not, so CMD still ships both, and the installer now sources
`System.Drawing.Common` from there. What lands on disk is unchanged.

The installer check also now fails when one of its build folders is missing rather than silently
skipping it — `GuiFunctions` had been skipped for exactly that reason, since it targets
`net<n>.0-windows` but the list looked under `net<n>.0`.

### Merge order

This should merge **before** the matching mzLib PR. mzLib's CI builds MetaMorpheus master against
its freshly built package, and a `net8.0` MetaMorpheus cannot reference a `net10.0` mzLib — so the
mzLib PR's integration job stays red until this one lands. Nothing here depends on mzLib moving:
a `net10.0` app consumes the existing `net8.0` mzLib 1.0.584 package fine, which is exactly what the
verification above did.


trishorts:
This is needed and overdue — .NET 8 goes out of support on 2026-11-10, so this has to happen this quarter regardless of anything else. The workflow consolidation is a real improvement on its own: hoisting `DOTNET_VERSION`/`DOTNET_FRAMEWORK` into `env:` in the three workflows that were hard-coding versions, fixing `InstallRunAndArtifact.yml` to look for `GuiFunctions` under `net<n>.0-windows` rather than `net<n>.0`, and making a missing build folder a hard failure rather than a silent skip. Dropping the ten `System.*` `PackageReference`s from `GUI.csproj` is also right — the Windows Desktop shared framework supplies them on `net10.0-windows` and the SDK prunes them.

Four things, one of which I'd genuinely like changed before merge.

## 1. The `global.json` floor will lock out contributors — please consider `10.0.100`

```diff
-    "version": "8.0.204",
+    "version": "10.0.400",
     "rollForward": "latestFeature"
```

`10.0.400` is the **current tip** — per the official release metadata, `latest-sdk` for the 10.0 channel is exactly `10.0.400`. And `latestFeature` rolls *forward* to a higher feature band but never *back*, so this requires an SDK from the 4xx band or newer.

Every other published .NET 10 SDK is therefore excluded: 10.0.101, 10.0.102, 10.0.103, 10.0.200, 10.0.201, 10.0.202, 10.0.203, 10.0.300, 10.0.301, 10.0.302. A contributor on any of those — including someone who installed a recent SDK a few weeks ago — gets a hard `A compatible .NET SDK was not found` on `dotnet restore`, before anything else runs.

Concretely: the machine I'm writing this from has `10.0.300` as its newest .NET 10 SDK and is blocked by this line. It also means CI is pinned to whatever the runner images ship, which will drift.

**`"version": "10.0.100"` with the same `"rollForward": "latestFeature"` is equally correct and much friendlier** — it still guarantees .NET 10, still rolls forward to whatever the runner or the developer has, and accepts every published 10.0 SDK. Unless there's something specific in the 4xx band this depends on, that seems strictly better.

## 2. iText 8.0.5 → 9.7.0 is a major-version bump riding inside a TFM migration

```diff
-    <PackageReference Include="itext7" Version="8.0.5" />
-    <PackageReference Include="itext7.bouncy-castle-adapter" Version="8.0.5" />
+    <PackageReference Include="itext7" Version="9.7.0" />
+    <PackageReference Include="itext7.bouncy-castle-adapter" Version="9.7.0" />
```

The security rationale is sound — the BouncyCastle advisories against 2.2.1 are real — and I gather #2711 was the standalone version, closed in favour of folding it in here.

The concern is coupling, not the bump itself. It compiles with no source changes, but iText 9 is a major version and the runtime PDF-export behaviour is only as covered as `MetaDrawTest` and the MetaDraw view-model tests happen to be. If a PDF regression shows up three weeks from now, the bisect lands on "the .NET 10 migration" and someone spends a day in the wrong place.

Either re-split it (it was already a standalone PR), or — if it stays — call it out explicitly in the release notes and have someone export a PDF from MetaDraw once by hand before this merges. A one-line note in the PR description that this is a deliberate major-version bump would help whoever reads the log later.

## 3. Heads-up: this will break #2709's restore, and there's no merge conflict to warn anyone

Not this PR's problem to fix, but worth knowing so it isn't a mystery when it happens.

#2709 adds `GUI.Avalonia` (`net8.0`) with a `ProjectReference` to `TaskLayer`, plus rows for it in `MetaMorpheus.sln`. Once `TaskLayer` becomes `net10.0`, that reference is **NU1201** — a hard error at restore, and because the projects are in the solution it breaks the *solution-wide* restore in `build.yml`, both `Test.yml` jobs, `InstallRunAndArtifact.yml` and `DockerUpload.yml`, on every platform leg.

I checked with a real three-way merge against the shared base: your `Test.yml` hunks and #2709's are disjoint, and this PR doesn't touch `MetaMorpheus.sln` at all — so **the merge succeeds cleanly and the breakage appears with nothing pointing at the cause.** The leftover `dotnet-version: 8.0.204` inside #2709's new `avalonia-gui-tests` job also survives the merge while every sibling job moves to `${{ env.DOTNET_VERSION }}`.

The fix is four lines on #2709's side and I've written it up there — flagging it here only so the merge order is a conscious choice.

## 4. Merge order with mzLib#1141, and one smaller thing

**This should go first.** I checked and this PR keeps `mzLib` at `1.0.584`, which is the current latest published version — the line in the diff is context, not a change — so **#2710 depends on nothing from #1141**. The reverse isn't true: mzLib's `integration` job clones MetaMorpheus master and builds it against the freshly built package, so a `net10.0` mzLib can't be validated against a `net8.0` MetaMorpheus and #1141's integration leg stays red until this lands. Worth saying out loud so nobody blocks this one waiting on the other.

Smaller: `Product.wxs` re-sourcing `System.Drawing.Common` from `$(var.CMD_TargetDir)` instead of `$(var.GUI_TargetDir)` follows correctly from the package pruning, but "the MSI payload is unchanged" is asserted rather than demonstrated. Since the installer is the one artifact users actually receive, worth having someone build it once and diff the file list before this merges.

trishorts:
Follow-up to my earlier comment — I've now reviewed this properly against a worktree at `094b934b`, including building it and running a real search on both runtimes. Summary: **the migration is correct and the science is provably unchanged.** Most of what follows is small, plus one thing that needs to happen before this merges and isn't in either repo.

First, a correction to my own earlier comment: I said .NET 10 "turns on transitive package auditing … which may add warnings of their own." **The auditing half is wrong.** A controlled experiment — two throwaway projects identical but for TFM — shows NU1902/1903/1904 fire identically on `net8.0` with a modern SDK. Those warnings are pre-existing, not migration artifacts. The pruning half (NU1510) is right, but MetaMorpheus has **zero** NU1510; they're all on the mzLib side.

---

## The science is unchanged — measured, not assumed

Real Classic bottom-up search, identical inputs on both sides (`Task-SearchTaskconfig_NoMods.toml`, a Velos mzML snip, `human_rev_GPTMDpruned.xml`). Four runs — two per runtime, so same-runtime determinism was controlled first. .NET 8.0.27 (SDK 8.0.418) vs .NET 10.0.8. 133 target PSMs / 101 peptides / 105 protein groups at q ≤ 0.01.

| File | Result |
|---|---|
| `AllPSMs.psmtsv` | **byte-identical** |
| `AllPeptides.psmtsv` | **byte-identical** |
| `AllProteinGroups.tsv` | **byte-identical** |
| `AllPSMs_FormattedForPercolator.tab` | **byte-identical** |
| `Task Settings\...config.toml` | **byte-identical** |
| `.mzID` | identical after stripping `creationDate` |

So no float drift, no unstable-sort reordering, no `double.ToString()` formatting change, same q-values, same FDR boundary, same identification counts. Targeted suites (TestToml, FdrTest, PsmTsvWriterTests, SearchEngineTests, SearchModesTest, PostSearchAnalysisTaskTests, PeptideSpectralMatchTest) are **136/136 on both**, with identical passing sets. This is well-founded rather than lucky — the diff touches **zero `.cs` files**.

**The gap worth naming: ML.NET/PEP never trained.** Every row carries the `PEP=0` / `PEP_QValue=2` placeholders, because a snip file has too few PSMs. `Microsoft.ML` 3.0.1 is `netstandard2.0`-only and predates .NET 9 and 10, and FlashLFQ does a `Model.Save`/`Model.Load` round-trip. **Please run the PEP tests and FlashLFQ's MBR tests before merging.** Failure would be loud rather than silent, but it gates FDR. And please **don't** bump to ML.NET 4.x as part of this — FastTree defaults changed between majors, which would inject exactly the kind of silent result change this migration otherwise avoids.

---

## Before this merges: mzLib master and every open mzLib PR will go red

This is the one thing that needs action outside this PR, and it isn't obvious.

mzLib's `integration` job builds MetaMorpheus as well as mzLib. `setup-dotnet` installs into a user-local dotnet root, and the SDK resolver then sees **only what that step installed** — the runner's preinstalled SDKs are invisible. So once this PR raises `global.json` to a .NET 10 floor, every mzLib branch still asking for `8.0.x` dies at SDK resolution before reaching restore:

| mzLib branch asks | MM master floor | integration |
|---|---|---|
| `8.0.x` | `8.0.204` (today) | green |
| `10.0.x` | `8.0.204` | **red — mzLib#1141 right now** |
| `8.0.x` | `10.0.400` (after this) | **red — mzLib master + all open PRs** |
| `10.0.x` | `10.0.400` | green |

Both off-diagonal cells are red, so no merge ordering avoids it, and multi-targeting mzLib wouldn't help — it's SDK resolution, not TFM. I've opened smith-chem-wisc/mzLib#1143, a 3-line change giving that job both bands. **Merging #1143 first makes this PR safe to merge whenever you like.**

---

## `global.json: 10.0.400` — please consider `10.0.100`

I raised this before; it's now confirmed empirically and it's harsher than I described. It isn't a warning or a fallback — `dotnet --version` in the repo exits **155**:

```
A compatible .NET SDK was not found.
Requested SDK version: 10.0.400
global.json file: ...\global.json
Installed SDKs:
```

Note `Installed SDKs:` prints **empty** — zero candidates satisfied the constraint. **No `dotnet` command works in the repo at all**, including `restore`. `latestFeature` rolls forward only *within* the 4xx band.

Two things sharpen it: `10.0.400` shipped **2026-08-11**, one day before this PR, and GitHub's runner images preinstall at most **10.0.302**. CI is green only because `setup-dotnet` fetches 10.0.400 on demand — which is exactly what hides the breakage from you. Changing the floor to `10.0.100` (same `rollForward`) selects 10.0.300 locally and restores cleanly; CI behaviour is identical.

---

## `System.Drawing.Common`: your availability claim is right, but the MSI now ships a 2019 binary

I expected to find the removal unsafe and was wrong — it's fine. `Microsoft.WindowsDesktop.App.Ref\10.0.8\FrameworkList.xml:28` lists it with `Profile="WindowsForms;WPF"`, both GUI projects set `UseWPF`, and all actual usage is in `GuiFunctions`, which never had the package.

The real finding is downstream. `Product.wxs` now sources the DLL from `$(var.CMD_TargetDir)`, and CMD's copy is **package 4.7.0 / assemblyVersion 4.0.0.1 / fileVersion 4.6.26919.2** — a .NET Core 2.x-era binary from 2019. The MSI previously installed 8.0.0.0. That's a silent ~6-year downgrade of a shipped file. It arrives via `Microsoft.NETCore.App 2.2.8` → `ConfigurationManager 4.7.0` → `Security.Permissions` → `Windows.Extensions` → `System.Drawing.Common 4.7.0`.

Cleanest fix: **drop the `src_System.Drawing.Common.dll` component** and add the DLL to `whitelistedMissingDlls` in `InstallRunAndArtifact.yml:201-213` — right beside `System.Windows.Extensions.dll` and `System.Security.Permissions.dll`, which are already whitelisted and come from the *same* 4.7.0 chain.

## The root cause of that, and an easy win

`<PackageReference Include="Microsoft.NETCore.App" Version="2.2.8" />` survives in **four** projects — `CMD.csproj:27`, `EngineLayer.csproj:25`, `TaskLayer.csproj:17`, `GUI.csproj:65` — plus `NETStandard.Library 2.0.3` in `EngineLayer.csproj:27`.

These are out-of-support .NET Core 2.2-era metapackages, meaningless since .NET Core 3.0, and they're the feeder of the fossil 4.7.0 graph above. They're precisely the class of reference your own NU1510-based cleanup rationale covers. **Removing them would likely make the `System.Drawing.Common` question disappear entirely.**

## Two tracked files the migration missed

- **`MetaMorpheus/GUI/GUI_h05aw1sq_wpftmp.csproj`** — tracked, not gitignored, absent from the 8-project solution, still `net8.0-windows`, and pinning `mzlib 1.0.574` and `itext 8.0.5` alongside ~200 absolute `C:\Users\Nic\...` paths. It's the only remaining `net8.0` string in any csproj. A .NET 10 + iText 9 PR that leaves a net8.0/iText-8 project file behind is confusing at best — `git rm` it here.
- **`MetaMorpheus/Dockerfile_WindowsInstaller:6`** — still `runtime:8.0-nanoserver-1809`. Worth knowing: **that tag doesn't exist in MCR, and neither does a `10.0-nanoserver-1809`** (only `-ltsc2022`/`-ltsc2025`). The file is already dead on master and can't be mechanically bumped. Delete rather than fix.

## The MSI installs no runtime and doesn't check for one

`Bootstrapper/bootstrapper.wxs:8-10` chains only the MSI — no `ExePackage`, no `NetFx` extension; its only search is for `ucrtbase.dll`. `Product.wxs:15,17` gates on Windows version and 64-bit only, and the app is framework-dependent.

Pre-existing, but this PR makes it **universal**: every current user has .NET 8, not 10, so the first launch after upgrading is an apphost error dialog for everyone rather than for the occasional fresh machine. Worth a release-note warning at minimum, and a prerequisite `ExePackage` as a follow-up issue.

## iText 9.7.0 — I withdraw the "check it by hand" ask; it's verified

I asked for a manual PDF export. Done better: `MetaDrawTest.TestMetaDrawOutputFormats` was run against iText 9.7.0 and **passes, producing a real PDF**. BouncyCastle moves 2.2.1 → **2.6.2** transitively, with no NU1605 and no version-conflict warnings, and the shipped assembly set is unchanged (the same 13 `itext.*` DLLs). There are eight PDF-asserting tests — worth citing one in the PR body so the next reader doesn't have to wonder.

## Verified good

- **All 155 `Product.wxs` source paths resolve** against the real .NET 10 output — and the reverse direction is clean too: every built DLL is either installed or on the whitelist, and no whitelist entry is dead.
- **This PR fixes a real latent bug** at `InstallRunAndArtifact.yml:227`: `GuiFunctions` was being looked for under `$DOTNET_FRAMEWORK` rather than `$DOTNET_FRAMEWORK-windows`, so nothing it produced had *ever* been validated. The hard-fail at `:234-243` stops that recurring. Nice catch, and easy to miss in a TFM diff.
- Zero `#if NET*`, zero `NETCOREAPP`, zero TFM-conditioned MSBuild, zero conditional `PackageReference` — the only `#if` is `#if DEBUG`, the only `RuntimeInformation` use is an OS check. Nothing stale.
- The arm64 `PlatformTarget` guards were correctly left alone, and the PE-header assertion path resolves for real (green ubuntu/macOS legs).
- `MetaMorpheusSetup.wixproj` needed no change — `DoNotHarvest=True` plus `$(var.<Name>_TargetDir)` tracks the TFM automatically.

## Minor

- `global.json` (`10.0.400`) and the workflows (`10.0.x`) are now two sources of truth where master had one exact string — mild tension with the PR's own "one version variable" goal.
- `README.md:36` and `:60` still say ".NET Core" (retired at .NET 5); `:33` says "only the Runtime" where the WPF GUI needs the **Desktop** Runtime.

## One systemic thing worth knowing about .NET 9+

Float→int conversions are now IEEE-saturating. Measured: `(int)double.NaN` was `-2147483648`, is now `0`; `(int)1e30` was `-2147483648`, is now `2147483647`.

It showed up harmlessly in my test — the written pruned database changed `mass="-2147483648"` to `mass="0"` for 436 of 9,159 sequences, all of which contain an ambiguous residue (X/B/Z/J/U) whose mass is `NaN`. I confirmed it doesn't propagate: re-running the search against each database gave byte-identical output. But **it will change the bytes of any stored pruned-database regression baseline**, so it's worth knowing before someone reports a "failure".

The live version of this is in mzLib, not here — `ClassicDeconvolutionAlgorithm.cs:77,81` and `MzSpectrum.cs:264` do `charge = (int)Math.Floor(1 / deltaMass)`, and a duplicated m/z makes `deltaMass` exactly zero. I've raised it on mzLib#1141.

nbollis (dismissed):

MetaMorpheus/EngineLayer/EngineLayer.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFramework>net10.0</TargetFramework>
```
trishorts, line 4: On the metapackage rebuttal — the references themselves are at `:25` and `:27`; this is the nearest line in the diff I can attach to.

I tested it, and **removing them does not break the build.** `GuiFunctions` doesn't *use* `System.Runtime.InteropServices.WindowsRuntime` — it carries a single **unused** `using` for it at `GuiFunctions/MzLibExtensions.cs:13`. No type from that namespace appears anywhere in the file: no `WindowsRuntimeMarshal`, no `WindowsRuntimeBufferExtensions`, no `AsBuffer`, no `IBuffer`. It compiles today only because a 2019 metapackage happens to supply the namespace — the same shape of defect as the stray `using System.Data.Entity.Core` you spotted in the DIA test.

So I removed all five references — `Microsoft.NETCore.App 2.2.8` from `CMD.csproj:27`, `EngineLayer.csproj:25`, `TaskLayer.csproj:17`, `GUI.csproj:65`, plus `NETStandard.Library 2.0.3` from `EngineLayer.csproj:27` — along with that one `using`, and built the solution:

```
Build succeeded.
    0 Error(s)
```

Then I compared the shipped output, because "it builds" and "it ships the same thing" are different claims. Every file under `bin\Debug` for CMD, GUI, GuiFunctions, EngineLayer and TaskLayer, by name and size:

| | files |
|---|---|
| this branch as-is | 462 |
| five references removed | 462 |
| differences | **none — identical** |

A six-line deletion that changes no output. It's cleanup rather than a defect, so splitting it into its own PR is entirely reasonable — I'd just rather see it split out than recorded as impossible, since the next person to try will hit the same unused `using` and reach the same wrong conclusion.

MetaMorpheus/GUI/GUI.csproj
```diff
@@ -72,18 +72,8 @@
     <PackageReference Include="SharpLearning.GradientBoost" Version="0.31.8" />
     <PackageReference Include="SharpLearning.Metrics" Version="0.31.8" />
     <PackageReference Include="SQLite.Interop.dll" Version="1.0.103" />
-    <PackageReference Include="System.Buffers" Version="4.5.1" />
-    <PackageReference Include="System.CodeDom" Version="8.0.0" />
-    <PackageReference Include="System.Collections.Immutable" Version="8.0.0" />
     <PackageReference Include="System.Data.SQLite" Version="1.0.118" />
```
trishorts: **Your diagnosis of the 4.7.0 chain is right and mine was wrong** — worth saying plainly. `dotnet nuget why` confirms it end to end, with no path through `Microsoft.NETCore.App` anywhere in the graph:

```
mzLib 1.0.584
└── System.Data.SQLite 1.0.118
    └── System.Data.SQLite.EF6 1.0.118
        └── EntityFramework 6.4.4
            └── System.Configuration.ConfigurationManager 4.7.0
                └── System.Security.Permissions 4.7.0
                    └── System.Windows.Extensions 4.7.0
                        └── System.Drawing.Common 4.7.0
```

You're also right that it can't be cut from this repo yet: mzLib#1142 fixed it at source — master's nuspec now declares only `System.Data.SQLite.Core` — but 1.0.584 was published before that merged, so the released package still carries the metapackage.

One thing that changes the follow-up, though: **MetaMorpheus is a second, independent path.** This line, plus `CMD.csproj:31`, `EngineLayer.csproj:31` and `TaskLayer.csproj:22`, reference the full `System.Data.SQLite` metapackage *directly* — `dotnet nuget why` lists them as their own top-level branches, not underneath mzLib. So bumping mzLib alone will **not** clear the chain, nor the NU1904 **critical** advisory on `System.Drawing.Common` 4.7.0 that rides along with it.

They do look redundant: `System.Data.SQLite.Core` is already referenced on the next line in all four projects, and nothing in the repo uses EF6 — the only reference to it is that stray unused `using` in the DIA test. I dropped all four and built: `0 Error(s)`. The advisories stayed, exactly as you'd predict, because mzLib 1.0.584 still pulls them.

So the complete fix is "bump mzLib past #1142 **and** drop these four, in the same pass". Worth recording wherever the mzLib bump is tracked, or it will look like the bump simply didn't work.

.github/workflows/InstallRunAndArtifact.yml
```diff
@@ -201,6 +201,10 @@ jobs:
           $whitelistedMissingDlls = @(
             "System.Configuration.ConfigurationManager.dll",
             "System.Security.Cryptography.ProtectedData.dll",
+            # Same EntityFramework 4.7.0 chain as the two below, and it reaches the build only
+            # through mzLib's nuspec. Installing it would ship a 2019 binary in place of the
+            # 8.0.0.0 the MSI used to carry; smith-chem-wisc/mzLib#1142 removes it at source.
+            "System.Drawing.Common.dll",
```
trishorts, line 207: This is the right call, and naming mzLib#1142 in the comment is what makes it maintainable rather than mysterious a year from now. I checked the part that mattered: `Validate Installer Contents` and `Search with Installed` are both green, so the installed CLI runs a real search without this DLL. Nothing in `CMD` references `System.Drawing` at all — every use is in `GuiFunctions`, which is `net10.0-windows` and gets the type from the Windows Desktop framework.

One loose end attached to this same mzLib#1142 dependency. The commit message says *"Also the stray `using System.Data.Entity.Core` in a DIA test"* was removed — it wasn't. It's still at `MetaMorpheus/Test/DIATests/TestPfPairAndPfGroup.cs:10`, and `git log -S` shows nothing on this branch touched it. Nothing in that file uses a type from the namespace, so it's dead; it compiles only because `EntityFramework` is in the graph.

Which makes it a latent build break timed to exactly the event this comment points at: the moment mzLib ships #1142 and the EF6 chain leaves, `Test` stops compiling. I deleted the line and the solution builds with `0 Error(s)`. Much cheaper to take now, as a one-line edit, than to have it surface as a mystery failure during the mzLib bump — especially since the commit message will have everyone believing it's already gone.

trishorts (commented):
Re-reviewed at `9ae5827b` against a rebuilt worktree — full solution build, a restore-graph trace, and two controlled experiments on the parts that were answered rather than applied. **Everything I asked for that you applied is correct, and one of your two rebuttals is right in a way that means I was wrong.** The other doesn't survive testing; that plus one undone item from your own commit message are inline.

## Closed

**1. `global.json` → `10.0.100` — verified on the machine that was blocked.** `dotnet --version` in the repo now returns `10.0.300`, exit 0. Before, it was exit **155** with an empty candidate list and no working `dotnet` command at all. `latestFeature` from a `100` floor picks up the 3xx band exactly as expected. This was the one I most wanted changed and it's the cheapest possible fix.

**2. `System.Drawing.Common` — better than what I suggested.** Dropping the component and whitelisting the DLL is what I asked for; putting the *reason* in the whitelist comment, with the mzLib issue number, is the part I didn't think to ask for and is why this won't be a puzzle later. Verified downstream rather than assumed: nothing in `CMD` touches `System.Drawing`, all usage is in `GuiFunctions` (`net10.0-windows`, served by the Windows Desktop framework), and **`Validate Installer Contents` and `Search with Installed` are both green** — so the installed CLI runs a real search without the DLL present. The 2019 binary is no longer shipped and the MSI is no longer silently downgrading a file.

**3. Both dead tracked files gone.** `GUI_h05aw1sq_wpftmp.csproj` and `Dockerfile_WindowsInstaller` deleted. That also removes the last `net8.0` string in any csproj.

**4. README.** The Desktop-Runtime-vs-Runtime distinction is now explicit, and the retired ".NET Core" wording is gone from both spots.

**5. The ML.NET / PEP ask — satisfied, and I'll withdraw it.** `Test on Windows` is green at 21m, and `FdrTest.cs:225`/`:290` construct a real `PepAnalysisEngine`, so the `Model.Save`/`Model.Load` round-trip on .NET 10 genuinely executed rather than being skipped. (The FlashLFQ MBR half of that ask was misdirected — those tests live in mzLib, and its suite is green.)

**6. iText.** The PR body now names `MetaDrawTest` and `ChimeraAnalysisTabViewModelTests` as the runtime coverage, which is what I asked for. Combined with `TestMetaDrawOutputFormats` producing a real PDF, I have nothing left here.

**7. Merge order — resolved and no longer a constraint.** mzLib#1143 merged at 17:29, so mzLib's `integration` job now installs both SDK bands. This can merge whenever you like without reddening mzLib master or the open mzLib PRs.

## Where I was wrong

My dependency chain for the fossil `4.7.0` graph was incorrect — I traced it to `Microsoft.NETCore.App 2.2.8` and it does not go through there at all. Your `EntityFramework` diagnosis is right, and so is the conclusion that it can't be cut from this repo until mzLib#1142 ships. Details and the full `dotnet nuget why` output inline on `GUI.csproj`, along with one wrinkle that changes the follow-up: MetaMorpheus carries its own direct path to the same chain, so the mzLib bump alone won't clear it.

## Still open from the earlier review

**The MSI installs no runtime and doesn't check for one.** Not mentioned in the response, and I don't think it should block — it's pre-existing and the fix is a prerequisite `ExePackage`, which is its own piece of work. But this PR is what makes it universal: every existing user has .NET 8, so the first launch after upgrading is an apphost dialog rather than a rare fresh-machine problem. A release-note warning is the minimum, and a follow-up issue for the bootstrapper would be worth filing while the context is fresh.

**A small one while you're in there:** `*_wpftmp.csproj` isn't in `.gitignore`. The WPF markup pass regenerates one on every GUI build (mine made `GUI_tkmgiois_wpftmp.csproj`) and normally deletes it again — which is exactly how the tracked one got committed by accident in the first place. A single ignore line stops the file you just deleted from coming back.

---

Not approving yet, and only because of the two one-line items inline: with nbollis's approval already in, mine would make this mergeable, and I'd rather the stray `using` land first — it's a build break scheduled for the mzLib bump that's part of this same migration, and the commit message currently says it's already fixed. Flip that and drop `System.Data.SQLite` where you like, and I'll approve.

MetaMorpheus/EngineLayer/EngineLayer.csproj
trishorts: Closing this — removed in `d086f2840`: all five references plus the unused `using`. Verified at that head, build is 0 errors.

Keeping it in its own commit so it can be dropped if you'd rather it were a separate PR was the right instinct, and it means the migration stays reviewable on its own terms either way.

MetaMorpheus/GUI/GUI.csproj
trishorts: Closing this — the four direct `System.Data.SQLite` references are gone in `54d11b5d4`, with `.Core` retained.

The predicted effect is confirmed. `dotnet nuget why` on `CMD` now shows every remaining path to `System.Drawing.Common` 4.7.0 running through `mzLib 1.0.584 → System.Data.SQLite → System.Data.SQLite.EF6 → EntityFramework`, and on `GUI` it reports *"does not have a dependency on System.Drawing.Common"* at all — pruned outright, since `net10.0-windows` gets the type from the Windows Desktop framework.

So the NU1904 critical advisory is down from four independent paths to one project and one path, and it clears with the mzLib bump exactly as you said it would. Both halves were needed.

.github/workflows/InstallRunAndArtifact.yml
trishorts: Closing this — removed in `54d11b5d4`, so the latent break is gone well before the mzLib bump could trigger it.

Thanks for gitignoring `*_wpftmp.csproj` in the same commit; that closes the loop on how the tracked one got committed in the first place, which is the part that would otherwise have recurred.

trishorts (dismissed):
Re-reviewed at `d086f2840`. **All three inline findings are fixed; I've replied on each thread and resolved them.** Approving.

## What closed

**The stray `using System.Data.Entity.Core`** (`54d11b5d4`) — gone, so the latent `Test` build break is removed before the mzLib bump could trigger it. `*_wpftmp.csproj` is now gitignored in the same commit, which closes the loop on how the tracked one arrived.

**The four redundant `System.Data.SQLite` metapackage references** (`54d11b5d4`) — dropped, `.Core` retained. The predicted effect is confirmed rather than assumed:

- `CMD` (plain `net10.0`): every remaining path to `System.Drawing.Common` 4.7.0 now runs through `mzLib 1.0.584 → System.Data.SQLite → System.Data.SQLite.EF6 → EntityFramework`. Four independent paths became one.
- `GUI` (`net10.0-windows`): `dotnet nuget why` reports *"does not have a dependency on System.Drawing.Common"* — pruned outright, because the Windows Desktop framework supplies the type.

So the NU1904 **critical** advisory is confined to one project and one path, and it clears with the mzLib bump. That is exactly what you said would happen.

**The 2019 metapackages and the unused `WindowsRuntime` using** (`d086f2840`) — all five references plus the one `using`, build clean at 0 errors. Isolating this in its own commit so it can be dropped if you'd rather it were a separate PR was the right call; the migration stays reviewable either way.

## The correction, on the record

My original diagnosis of the 4.7.0 chain was wrong — I traced it to `Microsoft.NETCore.App 2.2.8` and it never went through there. Your `EntityFramework` diagnosis was right. The piece I added was that MetaMorpheus carried its own four direct paths to the same chain, so the mzLib bump alone would not have cleared it. **Both halves were needed, and both are now done** — which is why the graph above collapses to a single mzLib-only path instead of merely shrinking.

## Still open, not blocking

**The MSI installs no runtime and doesn't check for one.** Unchanged, and I don't think it should hold this PR — it's pre-existing and the fix is a prerequisite `ExePackage`, which is its own work. But this PR is what makes it universal: every current user has .NET 8, so the first launch after upgrading is an apphost dialog rather than a rare fresh-machine problem. A release-note warning is the minimum, and a follow-up issue for the bootstrapper is worth filing while the reasoning is fresh.

## CI

`Create Installer Artifact`, `Validate Installer Contents`, `Search with Built`, `Search with Installed`, the three cross-platform legs and the external-service leg are all green. `Test on Windows` was still running when I wrote this; it's the required check, so merge waits on it regardless.

---

Approving. Merge order is unchanged and this is the one that goes first: **#2710 → mzLib#1141** (approved) **→ the mzLib release tag.** mzLib#1143 already landed, so nothing on the mzLib side is blocking you.

MetaMorpheus/EngineLayer/EngineLayer.csproj
acesnik: 🤖 Your test beats my rebuttal, and the way I got there is the instructive part: I removed the references, the build broke, and I concluded `GuiFunctions` needed `System.Runtime.InteropServices.WindowsRuntime` — without checking what was actually consuming it, which was one unused `using` and nothing else. That's exactly the defect I'd flagged in the DIA test, so I walked into the thing I'd just pointed at.

All five references and the `using` are gone in `d086f284`, and the solution, `Test` and `GuiFunctions` all build. I've kept it on this branch rather than splitting it out, since your objection was to it being *recorded as impossible* — and the wrong record lives here, in this thread and in my commit message, so this is where correcting it counts. Happy to lift the commit out if you'd rather it travelled separately.

MetaMorpheus/GUI/GUI.csproj
acesnik: 🤖 Thanks for tracing it properly, and the second path you spotted was the more valuable half. All the direct references are dropped, and the result is better than either of us predicted: `dotnet nuget why` now reports GUI and GuiFunctions have no dependency on `System.Drawing.Common` at all. What remains enters solely through mzLib 1.0.584 → `System.Data.SQLite` → EF6 → … → 4.7.0, reaching EngineLayer, TaskLayer, CMD and Test transitively. So the critical advisory genuinely clears with the mzLib bump, with no third path behind it — and #2713 is already the right home for the "bump mzLib **and** drop these together" note.

One correction to make plainly: I did this here without noticing you had already opened #2712 for it, which we'd looked at earlier the same day. Yours was the more complete change — it also caught the fifth reference in `MetaMorpheusSetup.wixproj`, which my first pass missed. I've since added that in `7c5a1522`, so this branch now covers #2712 entirely and that PR can be closed as superseded rather than both of us deleting the same five lines. Your call if you'd sooner keep yours and I'll drop mine.

.github/workflows/InstallRunAndArtifact.yml
acesnik: 🤖 The loose end was real, and my commit message was the worst part of it: claiming the `using` was removed when nothing had touched it is more misleading than not mentioning it, since anyone reading the history would have trusted it. Gone now, and `Test` builds. You were right that it was far cheaper as a one-line edit than as a mystery failure midway through the mzLib bump.

acesnik:
🤖 Both blocking items are in, plus the two extras.

The stray `using` is removed and `Test` builds. Every `System.Data.SQLite` metapackage reference is gone — all five, including the `MetaMorpheusSetup.wixproj` one my first pass missed — which clears the 4.7.0 chain out of GUI and GuiFunctions entirely and leaves mzLib 1.0.584 as its only remaining route. The 2019 metapackages and the unused `using` went too, in their own commit. And `*_wpftmp.csproj` is now ignored, so the file I deleted can't come back the way it originally arrived. Solution, `Test` and `GuiFunctions` all build clean.

**This branch now supersedes #2712.** I wrote that change without noticing you'd already opened a PR for it earlier the same day, which was careless of me — yours also caught the wixproj reference that I'd missed. Rather than have two PRs racing to delete the same five lines, this one now covers all of it, so #2712 can close as superseded. Say the word if you'd prefer the reverse and I'll drop mine instead.

On the MSI runtime gap — filed as #2714 rather than fixed here. Your framing was what convinced me it's worth recording now: the chain in `bootstrapper.wxs` is a single `MsiPackage` with no `ExePackage` and no runtime `LaunchCondition` anywhere, so this PR turns a rare fresh-machine problem into one that every upgrading user meets, since their existing .NET 8 won't satisfy it. The issue covers both the bootstrapper work and the release-note warning, and notes the latter is the cheap half that shouldn't wait for the former. The conda path is unaffected, since the feedstock already declares `dotnet-runtime`.

MetaMorpheus/MetaMorpheusSetup/Product.wxs
```diff
@@ -358,9 +358,6 @@
 	  	<File Id="src_System.Data.SQLite.EF6.dll" Name="System.Data.SQLite.EF6.dll" Source="$(var.GUI_TargetDir)System.Data.SQLite.EF6.dll" />
```
trishorts, line 358: Not a defect in this PR — but this line is a tripwire set to go off on the *next* one, and it's cheapest to note while the reasoning is loaded.

`Product.wxs` installs three files that exist in `GUI_TargetDir` only because of the EF6 chain:

- `:340` `EntityFramework.dll`
- `:343` `EntityFramework.SqlServer.dll`
- `:358` `System.Data.SQLite.EF6.dll` — this line

All three come from `mzLib 1.0.584 → System.Data.SQLite → System.Data.SQLite.EF6 → EntityFramework 6.4.4`, and they're still in `bin\Release\net10.0-windows\` at this head. **mzLib master has already cut that chain** — #1142 changed the nuspec from `System.Data.SQLite` to `System.Data.SQLite.Core` in both dependency groups:

```
-    <dependency id="System.Data.SQLite" version="1.0.118" />
+    <dependency id="System.Data.SQLite.Core" version="1.0.118" />
```

So the moment MetaMorpheus bumps to an mzLib release past #1142, those three files stop being copied into `GUI_TargetDir`, and because `Product.wxs` hard-references them by source path, the **installer build fails outright** — a WiX file-not-found, not a warning. The MSI is also the artifact least likely to be built locally before pushing, so it would surface as a red `Create Installer Artifact` on an otherwise innocuous version bump.

Worth deleting these three components **in the same change as the mzLib bump**, not before (they're needed while 1.0.584 is still referenced) and not after.

This is the same shape as the `System.Drawing.Common` component you removed just above: an installer entry whose existence depends on a transitive package, with nothing tying the two together. It's also the same class of cross-repo ordering trap as #1127 — a change in one repo silently invalidating something in the other with no textual conflict to warn anyone.

trishorts (commented):
Two follow-ups from a deeper pass over the dependency graph. **Neither changes my approval** — one is a one-line miss in this PR's own cleanup, the other is a tripwire for the next PR rather than a defect in this one.

## A fifth `System.Data.SQLite` reference the sweep missed

`54d11b5d4` removed the metapackage from CMD, EngineLayer, GUI and TaskLayer. There is a fifth:

```
MetaMorpheus/MetaMorpheusSetup/MetaMorpheusSetup.wixproj:57
    <PackageReference Include="System.Data.SQLite" Version="1.0.118" />
```

with `System.Data.SQLite.Core` already sitting beside it on `:58`, exactly like the four you fixed. A repo-wide sweep for `Include="System.Data.SQLite"` now returns this line and nothing else, so removing it finishes the job. `MetaMorpheusSetup` is in the solution, so it restores on every solution build and will keep the advisory in the graph — and in any Dependabot or SCA report — from here.

While in that file: `:56` references `SQLite.Interop.dll` version `1.0.103`, which contributes nothing. That package ships a bare DLL at the package root with no `lib/` or `runtimes/` folder, so NuGet takes no assets from it; the real interop binaries arrive through `System.Data.SQLite.Core`'s `runtimes/` tree. It's noise rather than a bug, but it's noise that reads as load-bearing.

## Correction to my own earlier comment

I said dropping the four direct `.Core` references would be a clean follow-up. That needs a caveat I hadn't checked: `GUI.csproj:23-26` sets `ContentSQLiteInteropFiles` / `CopySQLiteInteropFiles` / `CleanSQLiteInteropFiles` / `CollectSQLiteInteropFiles`, and those properties are consumed by targets that NuGet imports from `build/` assets of **direct** references only. `Product.wxs:504` then hard-codes `runtimes\win-x64\native\SQLite.Interop.dll` out of `GUI_TargetDir`. So the `.Core` references may be doing something the metapackage ones weren't, and that wants verifying before anyone removes them — don't treat it as the same trivial deletion.

The inline note on `Product.wxs` is the more important of the two.

trishorts (approved):
Re-approving at `7c5a1522b` — my previous approval was dismissed by the push.

**The fifth reference is gone.** `git grep 'Include="System.Data.SQLite"'` across the whole branch now returns nothing, and `.Core` is retained in all five files. That closes the metapackage cleanup completely.

## On your two questions

**#2712 is genuinely superseded — close it.** Your branch is a strict superset: I opened #2712 for the four csproj references and it never had the `MetaMorpheusSetup.wixproj` one, which you've now caught. Two of us deleting the same five lines in different PRs is the worse outcome. I'll close it as superseded and point it here.

**Leave the metapackage commit on this branch.** Your reasoning is right and I'd rather not churn it: my objection was never to the change living here, it was to it being recorded as impossible — and the wrong record is in this thread and in a commit message on this branch, so this is where correcting it counts. Splitting it out now would separate the correction from the thing it corrects.

## What the record should say

Worth being explicit, since three of us have now been wrong about this chain in one direction or another:

- The fossil 4.7.0 graph originates in **`EntityFramework`**, reached via `System.Data.SQLite → System.Data.SQLite.EF6`. Not `Microsoft.NETCore.App 2.2.8` — that was my error.
- MetaMorpheus had **five** independent direct paths to it, now all removed. That was the half I added.
- What remains enters **only** through `mzLib 1.0.584`, and mzLib master has already cut it (#1142 switched the nuspec to `System.Data.SQLite.Core` in both groups). So it clears on the mzLib bump, with nothing behind it.

The `Product.wxs` note is the one thing still outstanding, and it's a tripwire for the bump rather than anything wrong here — #2713 is the right home for it, as you say.

## CI

`Test on Windows` and `Create Installer Artifact` are re-running on this head; the earlier head had every leg green including `Validate Installer Contents` and `Search with Installed`, and `7c5a1522b` only deletes a `PackageReference`. Merge waits on the required check regardless.

---

Order unchanged: **this → mzLib#1141 → the mzLib release tag.**

MetaMorpheus/MetaMorpheusSetup/Product.wxs
trishorts: Correcting my own comment upward — I undercounted. It's **four** components, not three, and the blast radius reaches the whitelist too.

The fourth is `:352`, `System.Data.SqlClient.dll`, and it has the same single origin:

```
mzLib 1.0.584 -> System.Data.SQLite 1.0.118 -> System.Data.SQLite.EF6 1.0.118
             -> EntityFramework 6.4.4 -> System.Data.SqlClient 4.8.1
```

So the full set that loses its source when mzLib ships #1142: `:340` `EntityFramework.dll`, `:343` `EntityFramework.SqlServer.dll`, `:352` `System.Data.SqlClient.dll`, `:358` `System.Data.SQLite.EF6.dll`.

And **six of the twelve `whitelistedMissingDlls` entries go dead in the same moment** — every one of these is on the EF chain rather than the iText one: `System.Configuration.ConfigurationManager.dll`, `System.Security.Cryptography.ProtectedData.dll`, `System.Drawing.Common.dll`, `System.Security.Permissions.dll`, `System.Windows.Extensions.dll`, and `Microsoft.Win32.SystemEvents.dll` (which is itself a child of `System.Drawing.Common`). The remaining six `Microsoft.Extensions.*` entries come from iText 9.7.0 and are unaffected.

Dead whitelist lines are harmless where a missing `File/@Source` is fatal, so the four components are the urgent half — but it's worth removing both in that change, since a whitelist that lists things which can no longer appear is how the list stops being trustworthy.

Small related nit while I'm here: the new comment above `"System.Drawing.Common.dll"` says *"the same EntityFramework 4.7.0 chain as the two below"*. It's five other entries, not two — two sit above it (`ConfigurationManager`, `ProtectedData`) and three below (`Security.Permissions`, `Windows.Extensions`, `Win32.SystemEvents`). Not worth a push on its own; worth folding in if you touch the file again.

acesnik: 🤖 Thanks — the fourth one is the load-bearing correction, since `System.Data.SqlClient.dll` at `:352` is the entry I'd have been most likely to miss when doing the bump; it doesn't read as EF-derived from the filename.

The timing constraint is the part worth pinning: before the bump breaks the installer now, after breaks it in the window between, so all four deletions have to be in the same commit as the version change. I've recorded the full set on #2713 — the four components, the six dead whitelist entries, and the `.Core` caveat, which is the one that would have bitten someone treating them as the same trivial deletion.

Also taking the nit: my comment above `"System.Drawing.Common.dll"` says "the two below" when it's five entries, two above and three below. Since this head is approved and a push would dismiss it, I've noted that on #2713 to fold in rather than pushing for a comment fix.

nbollis (approved):
Lgtm. These updates usually break integration somewhere. Once this is merged, I will update our jenkins scripts.

Alexander-Sol (approved):

PR #2715, opened 2026-08-12
Stop the manuscript prose repeating itself

### LINKED ISSUE #2428: Repeated analysis duplicates auto generated manuscript prose
**Describing the issue**

Run 1
image

Run 2
image

Run 3
image

acesnik:
🤖 Closes #2428.

MetaMorpheus writes your methods paragraph for you, and until now it got a little more insistent about the search settings every time you re-ran an analysis. Once in run 1, twice in run 2, three times in run 3 — the screenshots in the issue are a nice little staircase.

`ProseCreatedWhileRunning` is an instance `StringBuilder`, and the GUI hands the *same* task objects back for each run (`MainWindow.xaml.cs:965` passes `PreRunTasks[i].metaMorpheusTask` straight through, and that collection only empties when you delete tasks). So the buffer was never emptied, and each run happily appended to the last one's paragraph. The CLI builds fresh tasks every time, which is why it never picked up the habit.

One line in `RunTask`, right next to the `FileSpecificParameters` reset that was already doing this job for its own state. Reviewers who'd like a regression test attached, say the word.


MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -630,6 +630,10 @@ public MyTaskResults RunTask(string output_folder, List<DbForTask> currentProtei
 
             FileSpecificParameters = new List<(string FileName, CommonParameters Parameters)>();
 
+            // The GUI re-runs the same task objects, so this survives between runs and would otherwise
+            // append a second copy of every sentence to AutoGeneratedManuscriptProse.txt.
+            ProseCreatedWhileRunning.Clear();
```
trishorts, line 636: Worth attaching the regression test you offered — and it's close to free, because a test in the suite **already reproduces the bug and doesn't notice**.

`Test/MyTaskTest.cs:591` `TestFileOutput` builds one `SearchTask` and runs it through two `EverythingRunnerEngine`s on the same instance (`:603-607`). On master, `MultipleMassSpectraFileOutput\AutoGeneratedManuscriptProse.txt` therefore contains the settings paragraph twice today; the test only compares filename sets (`:621`, `:631`), so it passes either way. Every other prose assertion in the suite is an existence check too (`AveragingTests.cs:66/107/148/220`, `OutputTest.cs:123`, `TestOGlyco.cs:1596`) — nothing reads the content.

Dropping this in after the multi-file assertion makes it red on master, green here, at zero extra runtime:

```csharp
var prose = File.ReadAllText(Path.Combine(thisTaskOutputFolder, "MultipleMassSpectraFileOutput", "AutoGeneratedManuscriptProse.txt"));
Assert.That(Regex.Matches(prose, "The following search settings were used: ").Count, Is.EqualTo(1));
```

`"The following search settings were used: "` is unique to `SearchTask.cs:178`. The point is asserting an exact count — non-empty or no-exception wouldn't catch the 1/2/3 staircase in #2428.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -630,6 +630,10 @@ public MyTaskResults RunTask(string output_folder, List<DbForTask> currentProtei
 
             FileSpecificParameters = new List<(string FileName, CommonParameters Parameters)>();
 
+            // The GUI re-runs the same task objects, so this survives between runs and would otherwise
+            // append a second copy of every sentence to AutoGeneratedManuscriptProse.txt.
```
trishorts, line 635: The comment is accurate but undersells the bug: this isn't only a *re-run* problem. `PreRunTask` holds the task by reference (`GUI/ForDisplayingInDataGrids/PreRunTask.cs:9`, and `Clone()` is a `MemberwiseClone`), so Ctrl+C/Ctrl+V (`MainWindow.xaml.cs:1145-1158`) and Ctrl+D (`:1168`) put the **same** `MetaMorpheusTask` instance in the list twice, and `EverythingRunnerEngine.cs:99` then calls `RunTask` on it twice within a single click of Run All Tasks. Same staircase, no second run required.

Suggest widening it to something like "the GUI re-runs the same task objects, and the same task object can appear twice in one run list", and stating the invariant outright — `ProseCreatedWhileRunning` must be empty at the start of every `RunTask` — so a future refactor reads the line as load-bearing rather than incidental. Non-blocking.

(Also for the record: `Clear()` isn't a style choice here — the field is `readonly` at line 181, so reassignment wouldn't compile. Right call; the adjacent `FileSpecificParameters` line reassigns only because it's a settable property.)

trishorts (commented):
Local review pass (engine + two grounded investigators against `upstream/master @ dac54917f`). **No blockers** — the fix is correct and I verified the load-bearing claim by hand rather than taking it on trust: every one of the `ProseCreatedWhileRunning.Append` sites in the repo sits inside `RunSpecific` or something it calls (`SearchTask.cs:178`, `GPTMDTask.cs:67`, `XLSearchTask.cs:68`, `GlycoSearchTask.cs:63`, and `CalibrationTask.WriteProse:390` via `Initialize:314`), and the only read is `file.Write(...)` at `MetaMorpheusTask.cs:707`. So the reset at the top of `RunTask` discards nothing that belongs to the current run, and it lands before the `try`, which also stops a run that throws from leaking its orphaned prose into the next successful run's file.

Patch coverage is 100% (1/1 line, Codecov API), project 93.28% flat. Two non-blocking notes inline.

Also worth knowing, from the mzLib side: `Readers\InternalResults\MetaMorpheusProseFile.cs` *parses* `AutoGeneratedManuscriptProse.txt` and scrapes the `Spectra files:` / `Databases:` blocks by first-index. A doubled paragraph doesn't just look silly — it silently feeds that reader the wrong lists. Second reason this is worth fixing.

trishorts (approved):
Approving. The reset is correct by audit — every `ProseCreatedWhileRunning.Append` in the repo is downstream of it, the only read is at `MetaMorpheusTask.cs:707`, and sitting before the `try` means a failed run no longer leaks orphaned prose into the next one. No-op for the CLI, corrective for the GUI. Patch coverage 100%.

The two inline notes are non-blocking: the regression test you offered (`MyTaskTest.TestFileOutput` already reuses the instance and would go red on master with a two-line count assert) and widening the comment to cover Ctrl+V/Ctrl+D duplication within a single run. Happy either way on whether they land here or as a follow-up.

nbollis (approved):

zhuoxinshi (approved):

PR #2717, opened 2026-08-12
Treat CID-cleavable crosslinkers as LowCID-cleavable (merge after #2736)

### LINKED ISSUE #1797: LowCID should be included in crosslinker dissociation type if CID is included
Many crosslinkers fragment w/ CID. If they are annotated as such, LowCID should also be included in that list

### LINKED ISSUE #2716: Custom crosslinkers silently lose EThcD when saved and reloaded
🤖 Noticed while working on #1797.

A custom crosslinker whose cleave-dissociation list contains EThcD loses that entry when it is saved and reloaded, silently.

`Crosslinker.DissociationTypes2String` serialises via `d.ToString()`, so `DissociationType.EThcD` is written as the string `"EThcD"`:

```csharp
x += d.ToString() + "|";
```

But `GetCleaveDissociationTypes` matches on `"ETHCD"`:

```csharp
case "ETHCD":
    cleaveDissociationTypes.Add(DissociationType.EThcD);
    break;
```

C# `switch` on strings is case-sensitive, so `"EThcD"` falls through to `default: break;` and is dropped without a warning.

This is reachable from the GUI rather than being merely latent: `GUI/Views/CustomCrosslinkerWindow.xaml.cs:66` writes new crosslinkers with `newCrosslinker.ToString(true)`, which goes through `DissociationTypes2String`. So a user who defines an EThcD-cleavable crosslinker gets a definition that no longer reports EThcD after a round-trip, and `CrosslinkedPeptides.cs:18` then localizes the intact crosslinker mass instead of the cleave masses for those scans.

The shipped `EngineLayer/Data/Crosslinkers.tsv` only uses `CID|HCD`, so the built-in crosslinkers are unaffected — this is custom-crosslinker only.

Two possible fixes: accept both spellings in the parser (`case "EThcD":` alongside `case "ETHCD":`), or make the comparison case-insensitive so the parser and serialiser can't drift again. The second is more robust, and would also cover any future enum member whose casing doesn't match the tsv convention.

acesnik:
🤖 Closes #1797 and #2716.

The issue asks that crosslinkers annotated as CID-cleavable in `Crosslinkers.tsv` also be treated as cleavable under LowCID, since low-resolution CID is still CID. It has been open since 2019 with a disagreement left unresolved.

## What the change does

`Crosslinker.IsCleavedBy(dissociationType)` replaces the inline `Cleavable && CleaveDissociationTypes.Contains(...)` test at `CrosslinkedPeptides.cs:18`, and treats LowCID as satisfying a CID declaration. The equivalence runs one way only: declaring LowCID does not imply CID, and an uncleavable crosslinker stays uncleavable.

That gate governs four call sites in `CrosslinkSearchEngine`:

| Site | Dissociation type used | Affected |
|---|---|---|
| `:519`, `:573` | `CommonParameters.DissociationType` — parent-scan alpha/beta localization | yes, this is what #1797 asks for |
| `:760` | `MS2ChildScanDissociationType` — MS2 child | yes |
| `:767` | `MS3ChildScanDissociationType` — MS3 fallback | incidentally |

An earlier version of this description claimed the change fixed localization of cleaved MS3 signature ions. That was wrong, and @trishorts caught it. The primary cleaved-MS3 branch at `CrosslinkSearchEngine.cs:733` builds its fragments with `Fragment(...)` inline and never calls `XlGetTheoreticalFragments`, so it consults only the `Cleavable` bool and was never affected by this gate.

So the real scope is narrower than the issue title suggests: it changes results for searches configured with LowCID as the main or MS2-child dissociation type. That is the case lonelu was sceptical anyone runs, and the scepticism is fair.

Line numbers throughout are as of the `master` merge on this branch. They shifted +9 from the numbers quoted earlier in the review thread, when #2742 and #2759 retyped the fragment and peptide indexes.

## Output change

For a CID-declared cleavable crosslinker fragmented under LowCID, the localized signature ions change from the intact crosslinker to the two cleaved stubs. `TestTheoreticalFragmentsCleavableCrosslinkLowCid` measures it on `PEPTIDE` (799.35996 Da) with a test crosslinker of `TotalMass` 1000, `CleaveMassShort` 15, `CleaveMassLong` 25 and a 10 000 Da partner peptide. Before the change, LowCID produced a single `ProductType.M` ion at **11799.35996** — peptide plus intact crosslinker plus the whole partner peptide. After, it produces **two**, at **814.35996** and **824.35996**, matching CID exactly. The mass moved by 10 985 Da, which is the partner peptide plus the uncleaved linker.

Undeclared types are unaffected: under ETD the same crosslinker still yields the single intact 11799.35996 ion.

`TestMs2Ms3`'s alpha child-ion count also moves 1 → 2, but that one is *not* evidence the change improves anything. Scan 4 is beta's signature ion; when alpha is scored it fails to match alpha's stub masses and falls through to the `:767` fallback, which annotates it against alpha regardless. That extra match is coincidental. The fallback's habit of annotating a child scan against a peptide whose stub masses it did not match is pre-existing, and is now #2826, with a fix proposed in #2827. Measuring it there confirms the reading above: alpha was matching exactly one ion apiece against beta's two signature spectra.

## Case-insensitive parsing

`GetCleaveDissociationTypes` matched on upper case, while `ToString` writes the enum's own casing. The custom-crosslinker GUI upper-cases what the user typed, so the two sources disagreed and entries were dropped silently. The new `LowCID` case was unreachable from both directions as originally written. Matching on `ToUpperInvariant()` fixes it, and fixes the identical `EThcD` bug — which is why this now closes #2716.

Unrecognised types are still dropped silently by the `default:` case. `CleaveDissociationTypesParseRegardlessOfCase` pins that deliberately with a `UVPD` case. Turning the silence into a warning needs its own error channel — `GlobalVariables.ErrorsReadingMods` is not allocated until one call after `LoadCrosslinkers()` runs, so writing to it from crosslinker parsing would throw at startup — which is more than this PR should carry.

## Tests

Four new, plus one assertion pair corrected:

- `CidCleavableCrosslinkerIsAlsoCleavedByLowCid` — CID, HCD and LowCID cleave; ETD does not; LowCID-declared does not imply CID; uncleavable stays uncleavable.
- `ShippedCidCleavableCrosslinkersAreCleavedByLowCid` — the change reaches DSSO, DSUB and DisulfideBond as actually shipped in `Data/Crosslinkers.tsv`, and does not sweep in the uncleavable entries.
- `CleaveDissociationTypesParseRegardlessOfCase` — enum-cased input, GUI-cased input, a hand-edited field with padding and lower case, an unrecognised type, and a `ToString(true)` round trip.
- `TestTheoreticalFragmentsCleavableCrosslinkLowCid` — the output change above, asserted as a LowCID-equals-CID equivalence plus explicit stub masses, rather than as a golden number.
- `TestMs2Ms3` — both alpha assertions now read 2 and both pairs use `Is.EqualTo`, so a failure names the value rather than reporting `Expected: True`.

## Verified

`upstream/master` is merged in — the branch was 49 commits behind — and the whole solution builds on macOS/arm64 .NET 10 with `-c UbuntuMac -p:EnableWindowsTargeting=true`: 0 errors, 106 warnings, `Test.csproj` included, so the `XLTest.cs` changes do compile against current master.

The four crosslinker tests run green through a net10.0 harness against `EngineLayer`, and they are not vacuous. Dropping only the LowCID clause from `IsCleavedBy` fails three of the four, the first break being the `ProductType.M` equivalence at `XLTest.cs:1431` with the 11799.35996-vs-814.35996 values quoted above; reverting only the `ToUpperInvariant()` parse fails `CleaveDissociationTypesParseRegardlessOfCase`. So the two halves of the fix are independently covered. Restoring both returns all four to green.

Not verified here: `TestMs2Ms3` still cannot run on macOS, because `Test.csproj` needs `Microsoft.WindowsDesktop.App`. @trishorts ran `--filter FullyQualifiedName~XLTest` on Windows and got 32/32 with the twin assertion corrected, but that was on 13 Aug against his own local edit of the then-current branch. Nobody has run `XLTest` on Windows since the two later tests were added or since this base merge, so CI is the first check on that.

## Related

mzLib hard-codes the same equivalence at `PeptideWithSetModifications.cs:337`. That is the only copy of it there — the other four `LowCID` branches in that file are deliberately LowCID-only, for star and degree ions. A `DissociationType.IsCid()` helper next to the enum in `MassSpectrometry/Enums/DissociationType.cs` would stop the two repos drifting; filed as mzLib#1310, with a fix proposed in mzLib#1311.


MetaMorpheus/Test/XLTest.cs
```diff
@@ -1588,7 +1588,9 @@ public static void TestMs2Ms3()
 
             // test child scan (low-resolution CID, alpha peptide signature ion)
             Assert.That(csm.ChildMatchedFragmentIons.First().Key == 4);
-            Assert.That(csm.ChildMatchedFragmentIons.First().Value.Count == 1);
+            // 2 rather than 1 since #1797: DSSO is CID-cleavable, so its MS3 LowCID child scans are now
+            // searched with the cleaved masses, which matches one additional a* ion here
+            Assert.That(csm.ChildMatchedFragmentIons.First().Value.Count == 2);
```
trishorts: **Updating this one leaves its twin behind, and `TestMs2Ms3` fails on this branch.** The same quantity is asserted again at line 1608, after the TSV round trip, still expecting 1. I ran it — `Test` builds and runs fine on Windows:

```
Failed TestMs2Ms3 [1 s]
  Assert.That(psmFromTsv.ChildScanMatchedIons.Count == 4
           && psmFromTsv.ChildScanMatchedIons.First().Key == 4
           && psmFromTsv.ChildScanMatchedIons.First().Value.Count == 1, Is.True)
  Expected: True   But was: False
  at Test.XLTest.TestMs2Ms3() in XLTest.cs:line 1606
```

This assertion (in memory) passes; the one at 1608 (after `WritePsmCrossToTsv` → `ReadPsmTsv`) does not. The beta pair at 1597/1612 agrees at 5 in both places, which is the giveaway that these two are meant to track each other.

Changing line 1608 to `2` as well makes the whole class green — I ran `--filter FullyQualifiedName~XLTest` and got **32/32 passed**. So 2 is right in both places and nothing is being dropped in serialization; the second assertion was just missed.

Worth considering the constraint form (`Is.EqualTo(2)`) for both — as written, the compound `Assert.That(a && b && c)` at 1606 reports only `Expected: True / But was: False`, so you have to read the expression to find which conjunct broke.

MetaMorpheus/EngineLayer/CrosslinkSearch/CrosslinkedPeptides.cs
```diff
@@ -15,7 +15,7 @@ public static IEnumerable<Tuple<int, List<Product>>> XlGetTheoreticalFragments(D
             List<int> possibleCrosslinkerPositions, double otherPeptideMass, PeptideWithSetModifications peptide)
         {
             List<double> massesToLocalize = new List<double>();
-            if (crosslinker.Cleavable && crosslinker.CleaveDissociationTypes.Contains(dissociationType))
+            if (crosslinker.IsCleavedBy(dissociationType))
```
trishorts, line 18: **This gate is not on the path the description points at.** The primary cleaved-MS3 case never reaches `XlGetTheoreticalFragments` at all.

`ScoreChildScan` splits first (`CrosslinkSearchEngine.cs:724`):

```csharp
if (Crosslinker.Cleavable && childScan.TheScan.MsnOrder == 3 && (shortMassAlphaMs3 || longMassAlphaMs3))
{
    double massToLocalize = shortMassAlphaMs3 ? Crosslinker.CleaveMassShort : Crosslinker.CleaveMassLong;
    ...
    peptideWithMod.Fragment(CommonParameters.MS3ChildScanDissociationType, FragmentationTerminus.Both, childProducts);
}
```

That branch infers cleavage from the child precursor matching `peptide + CleaveMassShort/Long`, builds the fragments inline, and **never consults `CleaveDissociationTypes`**. So an MS3 scan that genuinely is a cleaved signature ion was already localized against the stub mass before this PR, LowCID or not — the "MS3 fragments get localized against the intact crosslinker plus the partner peptide" case in the description isn't this branch.

The four call sites this gate actually governs are the parent-scan alpha/beta localization (`:510`, `:564`, via `CommonParameters.DissociationType`), the MS2-child path (`:751`), and the MS3 `else` fallback (`:758`) whose own comment reads *"It tried to corver the situation that the MS3 scan have the same precursor mass as the parent MS2."*

I instrumented `TestMs2Ms3` to see which branch its child scan 4 takes, since that's the only coverage of this change:

```
DSSO  CleaveMassShort=54.01056  CleaveMassLong=85.982635  CleaveDissociationTypes=[CID,HCD]
parent scan 2  MsnOrder=2  PrecursorMass=3437.5264
  child scan 4  MsnOrder=3  PrecursorMass=1519.7202   (child - parent) = -1917.8062
```

Beta's count is unchanged at 5, alpha's went 1 → 2. Since branch one never calls the changed code, alpha's child scan 4 must be falling through to `:758` while beta takes branch one — i.e. scan 4 is beta's signature ion, now annotated against alpha's stub masses. Note the fallback's comment doesn't describe this case either: the precursor is nowhere near the parent's.

So two questions rather than an assertion, since which behaviour is correct here is your call:

1. Is the case #1797 concerns actually reaching `:758`, or is it the MS2-child path at `:751` and the parent scan at `:510`/`:564` when the search dissociation type is LowCID? That last is what the issue literally asks for, and the part lonelu was sceptical anyone runs.
2. For a scan that fell through branch one because it did **not** match this peptide's stub masses, is localizing it against `CleaveMassShort`/`CleaveMassLong` right? That extra matched ion in the test is exactly this branch, so the changed assertion is evidence about it.

Two smaller notes, neither blocking:

- `CustomCrosslinkerWindow.xaml.cs:49` does `txtDissociationType.Text.ToUpper()`, so a crosslinker entered through the GUI arrives as `"LOWCID"` and the new case-sensitive `case "LowCID":` never matches — which is why the existing cases are all uppercase. As written the new case is reachable only from a hand-edited tsv or a `ToString` round trip. Comparing case-insensitively would cover both and incidentally fix `EThcD` (written `"EThcD"`, parsed only as `"ETHCD"`), though that overlaps #2716.
- mzLib already hard-codes this exact equivalence at `PeptideWithSetModifications.cs:337` (`dissociationType == DissociationType.CID || dissociationType == DissociationType.LowCID`, skipping b1). Good evidence the premise is right — and an argument that a `DissociationType.IsCid()` helper belongs next to the enum upstream rather than as a second copy here, so the two can't drift.

trishorts (commented):
Ran the crosslink suite locally on Windows, since you flagged that you couldn't. Two things inline: `TestMs2Ms3` fails as the branch stands (the round-trip assertion at 1608 was missed), and the gate this changes doesn't govern the MS3 path the description describes. The premise itself looks right - mzLib already treats LowCID as CID elsewhere.

MetaMorpheus/Test/XLTest.cs
acesnik: 🤖 Fixed — both now read 2. You're right that the pair is meant to track each other; I updated the in-memory assertion and missed its twin after the round trip.

Also took the constraint-form suggestion for both pairs. The compound `Assert.That(a && b && c)` is now three separate `Is.EqualTo` assertions, so a failure names the value instead of reporting `Expected: True / But was: False`. That's what cost you the extra step of reading the expression to find which conjunct broke.

Thanks for running it — I said in the description that CI would be the first real check, and it should have been you telling me it failed rather than me shipping it that way.

MetaMorpheus/EngineLayer/CrosslinkSearch/CrosslinkedPeptides.cs
acesnik: 🤖 You're right, and the description is wrong. I verified it: the branch at `CrosslinkSearchEngine.cs:724` builds its fragments with `peptideWithMod.Fragment(...)` inline and never calls `XlGetTheoreticalFragments`, so it only ever consults `Crosslinker.Cleavable` — the `bool`, not the list. A genuinely cleaved MS3 signature ion was already localized against the stub mass before this PR. I'll rewrite that paragraph.

**On your first question** — the literal #1797 case is `:510`/`:564`, the parent-scan alpha/beta localization driven by `CommonParameters.DissociationType`, and `:751` for an MS2 child. Those are the paths where a CID-cleavable crosslinker under a LowCID setting previously fell to the intact mass. So the change does what the issue asks, and it is exactly the scenario lonelu was sceptical anyone runs. `:758` is not the case #1797 is about; it's collateral, and `TestMs2Ms3` only exercises it because scan 4 doesn't match alpha's stub masses.

**On your second** — I don't think that extra alpha ion is a good thing, and I'd rather say so than claim it as a win. Scan 4 is beta's signature ion. When alpha is scored, `shortMassAlphaMs3`/`longMassAlphaMs3` are computed against alpha and both miss, so it falls to `:758` and gets annotated against alpha anyway. Before this PR that annotation used intact-crosslinker-plus-beta; now it uses alpha's stub masses. Alpha isn't really present in that scan under either reading, so the additional match looks coincidental rather than correct — one more matched ion feeding alpha's score for a scan that isn't alpha's.

That's a pre-existing property of the fallback rather than something this PR introduces: `:758` annotates a child scan against a peptide whose stub masses it failed to match, and its own comment describes a different situation entirely ("the MS3 scan have the same precursor mass as the parent MS2"), which isn't true here either — the precursor is 1917 Da away. I'd rather not narrow it in this PR, since the fix is right for the three call sites #1797 is actually about, but I'll file the fallback question separately unless you'd prefer it handled here.

**Case sensitivity** — caught, and worse than you flagged: the case was dead from the GUI *and* the parse switch matches upper case while `ToString` writes `"LowCID"`, so it was unreachable from both directions. Now matching on `ToUpperInvariant()`, which makes it reachable either way and fixes the identical `EThcD` bug, so this closes #2716 as you predicted. Added `CleaveDissociationTypesParseRegardlessOfCase` covering enum-cased input, GUI-cased input, and a `DissociationTypes2String` round trip.

**On `IsCid()` in mzLib** — agreed, and `PeptideWithSetModifications.cs:337` is a better argument for it than anything in this PR. Two copies of the same equivalence in two repos will drift. I'd rather not add the mzLib dependency to this change, but I'll open an issue there unless you want to.

Also added `CidCleavableCrosslinkerIsAlsoCleavedByLowCid`, which you offered — it covers the negative directions too, since the equivalence shouldn't run backwards (LowCID-declared does not imply CID) or make an uncleavable crosslinker cleavable.

acesnik:
🤖 Both findings confirmed and fixed in `138de61`. Thanks for running the suite; the failure was real and I'd flagged the wrong branch in the description. Detail in the two inline replies — the second one concedes the mechanism and answers your two questions, though the answer to the first is essentially "you're right, it's the path lonelu was sceptical about."

MetaMorpheus/Test/XLTest.cs
trishorts: Verified at `92117f66`: both assertions read 2, the compound `Assert.That(a && b && c)` is split into named `Is.EqualTo` constraints in both places, and the in-memory one carries a comment explaining why it moved. Resolving.

MetaMorpheus/EngineLayer/CrosslinkSearch/CrosslinkedPeptides.cs
trishorts: Verified at `92117f66`. The parse switch is `x.Trim().ToUpperInvariant()` with `LOWCID` and `ETHCD`, so the GUI path and the `ToString` round trip both reach it; `CleaveDissociationTypesParseRegardlessOfCase` and `CidCleavableCrosslinkerIsAlsoCleavedByLowCid` cover it in both directions, and the description now maps the four call sites correctly.

Answering your two questions rather than leaving them open:

1. **The `:758` fallback** — separately, please, and I'd rather it were filed now than after this merges. Your reading matches mine: alpha isn't in scan 4 under either annotation, so the extra matched ion is coincidental, and the fallback's own comment describes a case that isn't the one it's handling. That is worth an issue on its own terms whoever fixes it, and it is pre-existing, so it should not hold this PR up.
2. **`IsCid()` in mzLib** — yes, open the issue there. Two copies of the same equivalence in two repos is exactly the drift I was worried about, and `PeptideWithSetModifications.cs:337` is the better home.

I don't consider either a condition of this PR. One practical note: it is currently BEHIND master, so it needs a base merge before it can go in.

Resolving.

MetaMorpheus/EngineLayer/CrosslinkSearch/Crosslinker.cs
```diff
@@ -47,12 +47,30 @@ public Crosslinker(string crosslinkerModSites, string crosslinkerModSites2, stri
...
+
         private List<DissociationType> GetCleaveDissociationTypes(string cleaveDissociationTypesInString)
         {
             List<DissociationType> cleaveDissociationTypes = new List<DissociationType>();
             foreach (var x in cleaveDissociationTypesInString.Split('|'))
             {
-                switch (x)
+                // ToString writes "EThcD" and "LowCID" but the GUI upper-cases its input, so the two
+                // sources disagree on case and a case-sensitive match silently drops entries (#2716)
+                switch (x.Trim().ToUpperInvariant())
```
trishorts, line 73: Right fix, and the comment explains the non-obvious part (two sources disagreeing on case) rather than restating the code.

Two small notes, neither blocking:

`case "":` still works after `Trim().ToUpperInvariant()`, but it is now also reachable from a field of pure whitespace, which it was not before. That is an improvement -- just noting the `break` is quietly doing more than its name suggests.

`default: break;` swallowing an unrecognised type is pre-existing, and your `unknownType` case pins it deliberately. Given that a hand-edited tsv is explicitly a supported input here, a warning rather than silence would be kinder to whoever typos `UVPD`. Out of scope for this PR, but it is the same class of silent-drop bug you are fixing one line up.

trishorts (commented):
Nice piece of archaeology, and I appreciate the correction of the earlier claim being left visible in the description rather than quietly edited away. The honest version -- "this changes results for searches configured with LowCID as the main or MS2-child dissociation type, which is the case lonelu was sceptical anyone runs" -- is far more useful to a reviewer than the issue title.

I checked the blast radius rather than trusting the table: `CleaveDissociationTypes.Contains(dissociationType)` appears exactly once in the whole repo outside tests, at `CrosslinkedPeptides.cs:18`, which is the line you replaced. There is no second copy of this gate that got missed. The remaining `Cleavable` checks (`CrosslinkedPeptides.cs:61`, `CrosslinkSearchEngine.cs:315`, `:657`, `:724`) do consult only the bool, exactly as you say.

The case-insensitivity fix is honestly the more valuable half of this PR. `ToString` writing `EThcD` while the custom-crosslinker GUI upper-cases its input meant a round trip silently dropped entries, and that bites people who are not running LowCID at all.

**Merge-order note:** this and #2736 both edit `Crosslinker.cs` and both insert a new test immediately after `XlTestCrosslinker()` in `XLTest.cs`. They are independent changes but will conflict textually in `XLTest.cs` whichever lands second. Worth choosing an order now rather than finding out at merge time.

acesnik:
🤖 Merged `upstream/master` — the branch was 49 commits behind, now 0, and it merged clean.

Two of the intervening commits do touch `CrosslinkSearchEngine.cs`: #2742 retyped the fragment index from `List<int>[]` to `Indexing.FragmentIndex`, and #2759 generalised the peptide index to `IBioPolymerWithSetMods` with a narrower `PeptideIndex` view over it. Both are signature plumbing, and neither goes near the cleavage gate or the branch structure in `ScoreChildScan`, which is why nothing conflicted. They did shift every line number quoted in this thread by +9, so I have updated the description: the four gated call sites are now `:519`, `:573`, `:760` and `:767`, the inline-`Fragment` MS3 branch is `:733`, and the bare `Cleavable` checks are `CrosslinkedPeptides.cs:61` plus `CrosslinkSearchEngine.cs:324`, `:666`, `:733`.

Your blast-radius check survives the merge, and is slightly stronger than when you ran it: `CleaveDissociationTypes.Contains` now appears on exactly two lines in the whole repository, both of them inside `IsCleavedBy` itself (`Crosslinker.cs:62-63`), and nowhere in the tests. No caller is left that could drift.

The description was also out of date on the tests — it listed two and there are four; `ShippedCidCleavableCrosslinkersAreCleavedByLowCid` and `TestTheoreticalFragmentsCleavableCrosslinkLowCid` arrived in `92117f66` after I last edited the body. The second is the one that actually pins the output change: under LowCID the M, b and y masses come out identical to CID and equal to the two cleaved stubs, while undeclared ETD still yields intact crosslinker plus partner peptide.

On merge order, I would land #2736 first and take the conflict here. It already has nbollis's approval, it is the smaller diff, and it is 39 behind master so it needs a base merge of its own regardless — resolving one insertion point in `XLTest.cs` during that merge is one operation rather than two. The conflict really is only `XLTest.cs`: both PRs add a test immediately after `XlTestCrosslinker()`, but the `Crosslinker.cs` hunks are in different methods (`IsCleavedBy` and `GetCleaveDissociationTypes` here, `ParseCrosslinkerFromString` there), so that file merges without help. Happy to do it the other way round — say which and I will update whichever lands second.

Follow-ups filed as you asked: #2826 for the `:767` fallback and mzLib#1310 for `DissociationType.IsCid()`, each with a fix proposed alongside it (#2827 and mzLib#1311). mzLib turned out cleaner than the single line suggested — `PeptideWithSetModifications.cs:337` is the only place the equivalence is spelled out there, and the other four `LowCID` branches in that file are deliberately LowCID-only, for star and degree ions.

MetaMorpheus/EngineLayer/CrosslinkSearch/Crosslinker.cs
acesnik: 🤖 Both fair.

The empty case turns out to be load-bearing rather than incidental, in two ways, so I gave it a one-line comment naming them. `DissociationTypes2String` appends `|` after every entry rather than joining, so `ToString` writes `CID|HCD|` and the split always yields a trailing empty field — every round trip goes through `case "":`. Separately, the shipped `Crosslinkers.tsv` leaves the column empty for the uncleavable entries, which splits to a single `""`. You are right that `Trim()` now widens it to whitespace-only fields as well, and that is welcome; it just is not the only thing the case is carrying.

On `default: break;` — you are right that it is the same class of bug, and I would rather answer it here than open a third issue. I looked at what a warning would cost. The precedent channel is `GlobalVariables.ErrorsReadingMods`, which the GUI already surfaces, but `LoadCrosslinkers()` is called at `GlobalVariables.cs:112` and that list is not allocated until `LoadModifications()` at `:426`, one call later — so routing a crosslinker parse warning into it would throw a `NullReferenceException` at startup, on exactly the malformed input it is meant to report. Doing it properly wants its own `ErrorsReadingCrosslinkers`, allocated before `LoadCrosslinkers()`, plus somewhere in the GUI to show it. That is a bigger change than the one-line fix above, and I would rather not fold it into this PR. In the meantime `CleaveDissociationTypesParseRegardlessOfCase` pins the current behaviour with a `UVPD` case, so the silence is at least deliberate and documented, and whoever does change it will see the test move.

PR #2718, opened 2026-08-13
Tell FlashLFQ which digestion agent produced each identification

### LINKED ISSUE #1034: Match-between-runs and multiprotease search are not compatible
MBR attempts to map identifications from one run onto peaks in another run. This is incompatible with the multiprotease search because it does not make sense to map IDs from one protease onto another (e.g. gluC peptide IDs onto chymotrypsin peaks). There should probably be one FlashLFQ engine created per protease to match-between-runs between those files, and the results should be combined somehow (??).

acesnik:
**Closes #1034**, together with smith-chem-wisc/mzLib#1148.

That issue has been open since 16 March 2018, when rmillikin filed it: match-between-runs maps identifications from one run onto peaks in another, which does not make sense across proteases, and the closing question — "the results should be combined somehow (??)" — is the part that kept it open. rmmiller22 worked out in the thread that normalization should happen within a protease but not between them. zrolfs noted in August 2019 that it was the last issue left in the 1.0 milestone. It was still a bug when rmillikin last confirmed it in March 2019, and it still is on master today.

This PR tells FlashLFQ which digestion agent produced each identification, so match-between-runs stays within an agent instead of transferring tryptic peptides into a Glu-C file — while normalization and protein quantification keep spanning every file. That is the resolution to the "combined somehow" question: nothing needs combining if the engine is never split by protease in the first place, and normalizing within a protease rather than across them falls out for free.

### When it actually matters

Match-between-runs only earns its keep in a multiprotease experiment that has replication *within* a protease, which is worth spelling out because it bounds how much this fix is worth:

| Design | Is MBR useful? | Affected? |
|---|---|---|
| Coverage / protein inference, one sample per protease (e.g. MassIVE MSV000083304) | No — no within-agent donor exists | Yes, purely harmful |
| Multiprotease with replicates and conditions | Yes, fully | Yes, and worst here |
| Pooled or sequential digest (trypsin+LysC mix) | Yes | No — single search agent |
| Cross-protease validation | N/A | Partitioning is what you want |

Only the second row earns its keep, and it is also where the damage is worst — the same row the normalization observation in the issue thread applies to. Everything else is either unaffected or wants match-between-runs turned off. Note also that `MatchBetweenRuns` defaults to `false`, so nobody hits this without asking for it.

The failure in that second row is worse than the spurious peaks the issue describes, and the mechanism lives on the mzLib side — see smith-chem-wisc/mzLib#1148 for the donor-selection detail. In short: one global donor is chosen per peptide sequence, so for a peptide two agents both produce, the donor can land in the wrong agent's file, and the legitimate same-agent transfer is then dropped rather than merely misdirected. Spurious peaks are visible in the output; that loss is silent.

### The change

```csharp
digestionAgentName: psm.DigestionParams.DigestionAgentName()
```

`DigestionAgentName()` reads `SpecificProtease` for proteolytic parameters, and `DigestionAgent` for everything else (RNA, and any implementation that names no agent). A fully non-specific search replaces the protease with `singleN`/`singleC`, so reading `DigestionAgent` directly would make every protease report the same name -- collapsing exactly the distinction this PR exists to draw, and doing it silently, since `FlashLfqEngine` treats agreement as "one agent" and switches the restriction off rather than reporting a conflict.

Plus deleting `proteaseSortedPsms` and `proteaseSortedFlashLFQResults`, which have been built and never read since 2018. The original intent survives in commit `91f24bd6` as a commented-out block labelled `//MultiProtease MBR capability code`; a later cleanup deleted the comment and left the dictionaries. The same dead pair is removed from `PostGlycoSearchAnalysisTask.cs`. That deletion takes the last reader of `PostGlycoSearchAnalysisParameters.ListOfDigestionParams` with it, so the property -- and the unguarded `(DigestionParams)` downcast in `GlycoSearchTask` that existed only to populate it -- are removed as well.

### Why it is this small

The obvious fix is one `FlashLfqEngine` per digestion agent with the results merged, and I wrote that first — 110 lines. It works, but the engine boundary cuts across two unrelated questions at once: it partitions match-between-runs, which is what we want, and it also partitions normalization and protein-level median polish, which is a separate decision entirely.

Restricting donor selection inside FlashLFQ instead keeps those independent. Match-between-runs is constrained by digestion; normalization and median polish continue to span every file exactly as they do today. That is also the safer default, given Woessmann et al. 2022 report peptides from different proteases disagreeing by up to 4-fold on the same protein — a reason to decide protein-level pooling deliberately rather than as a side effect of where an engine boundary falls.

### Effect on output

- **Single digestion agent — nothing changes.** Verified against master by running a two-file search with match-between-runs enabled on the bundled yeast data: `AllQuantifiedPeaks.tsv`, `AllQuantifiedPeptides.tsv` and `AllPeptides.psmtsv` are byte-identical. That check exercises the search task; the glyco path takes the same one-line change and is covered by the existing glyco tests.
- **Multiple digestion agents — MBR peaks in a file are no longer sourced from other agents.** In a two-agent run (trypsin plus `Lys-C|P` via file-specific toml) master produced 27 MBR peaks in the Lys-C file, 12 of them ending in R, which Lys-C cannot produce.

Both checks used the same tryptic spectra re-searched under a second agent, which exercises the code path but is not a real multiprotease digestion — so treat the second number as a demonstration, not a measurement.

### Dependency (resolved)

Requires `Identification.DigestionAgentName` from smith-chem-wisc/mzLib#1148, which merged on
2026-08-27 and shipped. The `PackageReference` here is mzLib 1.0.587, which carries the API, and
#2710 (.NET 10) merged ahead of it — so the blocker that held this in draft is cleared and there is
no ordering left to respect.

### Tests

`DigestionAgentNameTests` covers the extension method directly: the specific-search cases, the
non-specific search where the enzyme survives as `SpecificProtease`, the per-terminus clone that
flips `singleN` to `singleC`, the RNA fall-through, and null.

`PostSearchAnalysisTaskTests.QuantificationTellsFlashLfqWhichEnzymeDigestedEachIdentification`
covers the wiring, which those cannot: it drives `QuantificationAnalysis` and asserts the name on
the `Identification`s FlashLFQ actually ends up holding. Dropping the `digestionAgentName` argument
at the call site leaves every unit test above passing, so without this the argument could be removed
silently. Both mutants were verified to fail it -- dropping the argument reports `null`, reverting to
`DigestionAgent` reports `"singleN"`.

The read of `SpecificProtease` is unguarded, because a null is not a state `DigestionParams` can be
reached in: both constructors assign `Protease` from a `ProteaseDictionary` lookup that throws rather
than returning null, then `RecordSpecificProtease()` copies it across unconditionally, and the setter
is private. mzLib depends on the same invariant, dereferencing it bare in `ToString`, `Equals` and
`Clone`. `ProteolyticParametersAlwaysCarryASpecificProtease` pins it across the parameterless toml
constructor, full, semi and non-specific searches, and the per-terminus clone; a toml round-trip of a
non-specific Glu-C search was also checked by hand, since that is the path that actually runs and
`SpecificProtease` is private-set — it comes back `Glu-C`, not the constructor's default.
`ParametersThatNameNoAgentReportNullRatherThanThrowing` covers the other branch, where a null is
reachable and is what `FlashLfqEngine` already treats as "do not restrict match-between-runs".


MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -566,7 +553,10 @@ private void QuantificationAnalysis()
                                 psmToProteinGroups[psm],
                                 psmScore: psm.Score,
                                 qValue: psmsForQuantification.FilterType == FilterType.QValue ? psm.FdrInfo.QValue : psm.FdrInfo.PEP_QValue,
-                                decoy: psm.IsDecoy));
+                                decoy: psm.IsDecoy,
+                                // lets FlashLFQ keep match-between-runs within a digestion agent, while
+                                // normalization and protein quantification still span every file
+                                digestionAgentName: psm.DigestionParams?.DigestionAgent?.Name));
```
trishorts: **A non-specific search reports `singleN`/`singleC` here, not the real protease — which voids the guard in exactly the case this PR targets.**

`DigestionParams.RecordSpecificProtease()` swaps the protease out for a synthetic placeholder when the search mode is non-specific, keeping the real one in `SpecificProtease`:

```csharp
SpecificProtease = Protease;
if (SearchModeType == CleavageSpecificity.None) //nonspecific searches, which might have a specific protease
{
    Protease = FragmentationTerminus == FragmentationTerminus.N ?
       ProteaseDictionary.Dictionary["singleN"] : ProteaseDictionary.Dictionary["singleC"];
}
```

Since `DigestionAgent => Protease`, `psm.DigestionParams?.DigestionAgent?.Name` returns the placeholder. I ran every combination:

```
protease=trypsin  mode=Full  terminus=N  =>  DigestionAgent.Name='trypsin'  SpecificProtease.Name='trypsin'
protease=trypsin  mode=Semi  terminus=N  =>  DigestionAgent.Name='trypsin'  SpecificProtease.Name='trypsin'
protease=trypsin  mode=None  terminus=N  =>  DigestionAgent.Name='singleN'  SpecificProtease.Name='trypsin'
protease=trypsin  mode=None  terminus=C  =>  DigestionAgent.Name='singleC'  SpecificProtease.Name='trypsin'
protease=Glu-C    mode=None  terminus=N  =>  DigestionAgent.Name='singleN'  SpecificProtease.Name='Glu-C'
protease=Glu-C    mode=None  terminus=C  =>  DigestionAgent.Name='singleC'  SpecificProtease.Name='Glu-C'
```

Two consequences, both silent, and they pull in opposite directions:

- **Non-specific multiprotease run** — the trypsin file and the Glu-C file both report `singleN`, so `SharesAnalytePool` treats them as the same agent and permits the transfer. The fix does nothing in the configuration it exists for. mzLib's own comment on that branch says "nonspecific searches, *which might have a specific protease*", so this is a supported combination rather than a hypothetical.
- **Same protease, differing `FragmentationTerminus`** — one file reports `singleN`, the other `singleC`, so legitimate same-protease transfers are refused.

`SpecificProtease` is assigned unconditionally before the swap, so it carries the real protease in all three search modes — it is the right thing to read, and it keeps this a one-liner:

```csharp
digestionAgentName: psm.DigestionParams is DigestionParams dp
    ? dp.SpecificProtease?.Name
    : psm.DigestionParams?.DigestionAgent?.Name
```

`SpecificProtease` lives on the concrete `DigestionParams` rather than `IDigestionParams`, hence the pattern match; RNA falls through to the existing expression. Worth a note in the PR body too, since "one argument at the `Identification` call site" is no longer quite the whole story.

Two smaller things while you are here, neither blocking:

- The PR body says "The same dead pair still exists in `PostGlycoSearchAnalysisTask.cs` and is untouched here", but this diff does remove it from that file. Worth correcting so the description matches the diff — and the "byte-identical output" verification, as written, covers only the search task.
- After the deletion, `using Omics.Digestion;` looks unused in the glyco file, and `Parameters.ListOfDigestionParams` loses its last reader there (it is still read at `PostSearchAnalysisTask.cs:403` and `:616`).

For what it is worth, the deletion is better than dead-code removal: on master that block also carries a hard `IDigestionParams`→`DigestionParams` cast in the glyco `foreach` and an unguarded `proteaseSortedPsms[...]` indexer, both of which could throw. Verified the dictionaries are never read anywhere, and that they trace back to the commented-out `//MultiProtease MBR capability code` in `91f24bd6` (2018-07-24), exactly as you describe.

trishorts (commented):
Reviewed alongside smith-chem-wisc/mzLib#1148. The approach is right and the archaeology checks out, but there is one case where the agent name is not the protease - details inline. Noting this is a draft blocked on the mzLib release, so nothing here is urgent.

acesnik:
🤖 @trishorts caught a real defect in this PR while reviewing the mzLib half (smith-chem-wisc/mzLib#1148): for a fully non-specific search, the agent name this sends to FlashLFQ was the search strategy rather than the enzyme. Fixed in `9ce16ac8`.

### The defect

`DigestionParams.RecordSpecificProtease` moves the real protease into `SpecificProtease` and overwrites `Protease` with `singleN`/`singleC` when `SearchModeType == CleavageSpecificity.None`, and `DigestionAgent => Protease`. So `psm.DigestionParams?.DigestionAgent?.Name` reported `singleN` for every non-specifically searched file, whatever the enzyme.

Measured against the mzLib 1.0.584 this PR pins:

```
Full/trypsin     DigestionAgent.Name=trypsin    SpecificProtease.Name=trypsin
Semi/trypsin     DigestionAgent.Name=trypsin    SpecificProtease.Name=trypsin
None/trypsin     DigestionAgent.Name=singleN    SpecificProtease.Name=trypsin
None/Glu-C       DigestionAgent.Name=singleN    SpecificProtease.Name=Glu-C
after Clone(C):  DigestionAgent.Name=singleC    SpecificProtease.Name=trypsin
```

A trypsin file and a Glu-C file therefore reported the same agent, `SharesAnalytePool` returned true, and the restriction this PR exists to add silently did nothing — for precisely the datasets it is meant to protect. It failed open, so the behaviour was current master's rather than anything worse, but the pair did not deliver for non-specific searches.

Two things worth noting from the measurement. Full and Semi were never affected, because `RecordSpecificProtease` only substitutes for `None` — so the blast radius is fully non-specific searches only. And `Clone` deliberately re-seeds from `SpecificProtease.Name` in that case, which is what makes `SpecificProtease` the durable record of the wet-lab enzyme rather than an incidental leftover.

### The fix

A `DigestionAgentName()` extension on `IDigestionParams` in `EngineLayer/Util/MzlibExtensions.cs`, preferring `SpecificProtease` and falling back to `DigestionAgent` for oligos and anything else non-proteolytic. That file already extends this interface and already dispatches on the concrete type for `RnaDigestionParams`, so this follows the existing pattern rather than introducing one. Both call sites — search and glyco — become `psm.DigestionParams.DigestionAgentName()`.

Kept out of mzLib on purpose. A `WetLabDigestionAgent` property on `DigestionParams` would arguably be the better home, but it would put this PR behind a second mzLib release for a three-line dispatch.

`Test/DigestionAgentNameTests.cs` adds eight tests. Reverting the helper to the naive read fails exactly the two regression cases and leaves the other six green, so neither is padding.

### Still waiting on the mzLib release

This PR stays a draft until then, and the failing checks are expected rather than new: `Identification` in mzLib 1.0.584 has no `digestionAgentName` parameter, so `TaskLayer` cannot compile against the pinned package. Every check here has been failing on that one error since the PR opened.

Order of operations:

1. merge smith-chem-wisc/mzLib#1148
2. tag an mzLib release (`release.yml` runs on tag push, version from `git describe --tags`)
3. bump the `mzLib` `PackageReference` in the six csprojs that carry it — `CMD`, `EngineLayer`, `GUI`, `GuiFunctions`, `TaskLayer`, `Test`
4. CI here goes green for the first time, then un-draft

Because of step 3 I could only verify the parts that do not need the new mzLib API: the extension method and its eight tests compile and pass against `EngineLayer`. The two call-site edits are same-typed one-token substitutions, but the compiler will not confirm them until the package ships.

One known gap left, filed rather than fixed here: smith-chem-wisc/mzLib#1152, IsoTracker borrows identifications across files by a separate path that never reaches the digestion agent guard. `IsoTracker` defaults to false, so the default quantification path is unaffected.

trishorts (dismissed):
Re-reviewed after `9ce16ac8`. My substantive point is fixed: `DigestionAgentName()` now reads
`SpecificProtease` before `DigestionAgent`, so a fully non-specific search no longer collapses every
protease onto `singleN`. That was the case that mattered, because it silently disabled the
restriction for exactly the searches this PR exists to protect.

I pushed one commit (`92c9214d6`) rather than leaving the rest as comments, since two of them were
loose ends from the deletion:

**Completed the glyco cleanup.** Removing the protease-sorted dictionaries took the last reader of
`PostGlycoSearchAnalysisParameters.ListOfDigestionParams` with it, leaving a property that was only
ever written — and an unguarded `(DigestionParams)` downcast in `GlycoSearchTask` that existed
solely to populate it. Both removed, plus the now-orphaned `using Omics.Digestion`.

**Pinned the wiring.** `DigestionAgentNameTests` is good, but it only covers the extension method —
delete the `digestionAgentName:` argument at either call site and all eight still pass. That
failure is silent rather than loud: `FlashLfqEngine` reads a null name as "agent unknown" rather
than as a conflict, so the restriction switches itself off and quantification completes normally.
The new test drives `QuantificationAnalysis` and asserts the name on the `Identification`s FlashLFQ
actually holds. It uses a non-specific search so it also fails if the call site reverts to reading
`DigestionAgent`. Both mutants verified to fail it — `null` and `"singleN"` respectively.

**On the 12.50% patch coverage: that number is a broken upload, not a testing gap.** Codecov has
`sessions=1` for `ff2d7e3ed` and total coverage 28.3%; every other commit on this branch has
`sessions=2` and ~93.4%. One of the two Windows test jobs never uploaded, so the patch was measured
against half a report — which is also why `FilteredPsms.cs` appears in the impacted files at 0.00%
despite this PR not touching it. Measured locally, `DigestionAgentName` is at **100% line coverage**
from the existing tests alone (branch coverage is 58%, the uncovered arm being the `?? DigestionAgent`
fallback, which is unreachable: `SpecificProtease` is assigned in the `DigestionParams` constructor
and mzLib dereferences it unguarded in its own `ToString`/`Equals`).

Also corrected three claims in the PR body that had gone stale: the code snippet still showed
`psm.DigestionParams?.DigestionAgent?.Name`; the description said the dead pair was untouched in
`PostGlycoSearchAnalysisTask.cs` when the diff removes it; and the Blocker section still described
mzLib 1.0.584, though #1148 merged on 2026-08-27 and the pin is 1.0.587.

174/174 tests pass locally across the glyco, post-search and digestion-agent fixtures.

Approving — though note this approval now covers a commit of my own, so it is not an independent
second pair of eyes on that part.

trishorts:
Pushed `788ed21b5` to close the patch-coverage gap. One part of that is a change to production code
rather than a test, so it deserves an explanation rather than being buried in a diff.

**The two partials were not the same kind of problem.** Codecov reported them together as "2 lines
missing coverage" in `DigestionAgentName`, but measuring locally with coverlet separates them:

| Line | Branches hit | Reachable? |
|---|---|---|
| `proteolytic.SpecificProtease?.Name ?? proteolytic.DigestionAgent?.Name` | 2/6 | no |
| `digestionParams?.DigestionAgent?.Name` | 3/4 | yes |

**The four uncovered branches on the first line cannot be reached, so I removed them instead of
writing a test that pretends to.** `SpecificProtease` cannot be null on a `DigestionParams`:

- `RecordSpecificProtease()` assigns it unconditionally, as the first thing it does, from `Protease`;
- `Protease` comes from `ProteaseDictionary.Dictionary[protease]`, an indexer that throws
  `KeyNotFoundException` on a miss rather than yielding null;
- the setter is `private set`, and nothing in either repo subclasses `DigestionParams`.

mzLib settles it independently of any of that reasoning: `DigestionParams.ToString()` dereferences
`SpecificProtease.Name` bare, and `Equals` calls `SpecificProtease.Equals(...)` bare, both
unconditionally. A `DigestionParams` with a null `SpecificProtease` would fault in mzLib long before
it reached this extension method, so the fallback could never run.

So the line is now:

```csharp
return proteolytic.SpecificProtease.Name;
```

with a comment recording *why* it is unguarded. That comment is the point of the change as much as
the deletion is — an unguarded dereference invites the next reader to "fix" it, and the guard would
come straight back along with its dead branch.

This is the same thing I noted [in my earlier review](https://github.com/smith-chem-wisc/MetaMorpheus/pull/2718#pullrequestreview-5053613032)
as an unreachable fallback. I left it alone then because it was harmless; it stopped being harmless
once it was the reason the patch could not go green, since with no `status` block in `codecov.yml`
the patch target defaults to `auto`, i.e. the base's 93.41%. Covering only the reachable branch gets
to 8/9 ≈ 88.9%, which still misses.

**The second line's missing branch is real and now has a test.** It needs an `IDigestionParams`
implementation that is neither proteolytic nor RNA and names no agent. That is worth pinning on its
own merits: both call sites reach this once per identification while assembling the FlashLFQ input,
so a `NullReferenceException` here would abort quantification for the whole run rather than degrade
it — and null is precisely what `FlashLfqEngine` already handles, reading an unknown agent as "do not
restrict match-between-runs". Covered with a minimal double; `ParameterTest` has a similar one, but
private to its own fixture.

I also added `ProteolyticParametersAlwaysCarryASpecificProtease`, which pins the invariant the
unguarded read now depends on — across the parameterless toml constructor, full, semi and
non-specific searches, and the per-terminus clone. If `SpecificProtease` ever stops being populated,
that test fails with a clear name instead of the change surfacing as a `NullReferenceException`
inside quantification.

All three patch lines now report full branch coverage locally (2/2, no branches, 4/4). 69 tests green
across the digestion-agent, parameter and post-search fixtures.

Happy to revert the production half and leave the partial standing if you would rather keep the
defensive read, @acesnik — it is your call, and the tests stand either way.

nbollis (approved):

acesnik:
🤖 Re-reviewed your two commits. Both good; merging is fine by me. Checked the load-bearing claims rather than taking them, since `788ed21b5` removes a code path.

**The unguarded `SpecificProtease.Name` is safe.** All four premises hold against mzLib 1.0.587: both `DigestionParams` constructors chain to the one doing `Protease = ProteaseDictionary.Dictionary[protease]` then `RecordSpecificProtease()`; `Dictionary` is a plain `Dictionary<string, Protease>`, so the indexer throws rather than yielding null; the setter is private and nothing derives from the type; and mzLib dereferences it bare at `DigestionParams.cs:94`, `:124` and `:132`. `SetAllFileSpecificCommonParams` already does the same at `MetaMorpheusTask.cs:545`, so this is consistent with what MetaMorpheus was doing anyway.

**One thing worth adding to that.** `ProteolyticParametersAlwaysCarryASpecificProtease` pins the constructors, including the parameterless one, but not a toml round-trip — and that is the path that runs. `tomlConfig` does not `IgnoreProperty` `SpecificProtease`, while `SpecificProtease`, `SearchModeType` and `FragmentationTerminus` are all private-set, so if Nett could not write them back, a non-specific Glu-C search read from toml would report `trypsin` from the parameterless constructor's default. Wrong rather than null, so the unguarded read would not throw and the test would not catch it. Checked it:

```
in memory              Protease=singleN  SpecificProtease=Glu-C  SearchModeType=None  Terminus=N
toml                   SpecificProtease = "Glu-C" | Protease = "singleN" | SearchModeType = "None"
after toml round-trip  Protease=singleN  SpecificProtease=Glu-C  SearchModeType=None  Terminus=N
```

Nett does write private setters, so there is no gap. Noted in the PR body rather than adding another test, since the invariant test already fails loudly if the constructor side ever changes.

**Glyco deletion confirmed.** No remaining reader of the glyco `ListOfDigestionParams` at branch HEAD; every surviving reference is the search-task property of the same name, still read at `PostSearchAnalysisTask.cs:410` and `:625`.

**The wiring test is the right shape.** Asserting `DigestionAgent.Name == "singleN"` as a premise means it cannot quietly go vacuous if mzLib changes, and asserting `agentNames` non-empty before asserting content avoids passing over an empty peak set — which is how a couple of tests in this repo have passed for the wrong reason.

**Two stale lines your second commit left in the PR body, now fixed** — same class of thing you caught in mine. The description still said `DigestionAgentName()` "reads `SpecificProtease` before falling back to `DigestionAgent`", which stopped being true when you dropped the `??`; and the Tests section did not mention the two tests `788ed21b5` adds. Rewritten.

On your approval caveat: it resolves itself. The push dismissed your review, and @nbollis approved at 20:07, after `788ed21b5` — so the standing approval is independent of both your commits.

trishorts (approved):
Approving. Re-instating the approval that was dismissed on 2026-08-28 when `788ed21b5` was pushed —
the dismissal was a new commit, not a change of mind.

Recording what this approval does and does not cover, because three of the four commits on this
branch are mine and an approval that hides that is worth less than one that states it:

- **`92c9214d6` and `788ed21b5` are my own commits.** So on those, this is not an independent second
  pair of eyes, and it should not be counted as one. What *is* independent is @nbollis's approval,
  submitted against `788ed21b5` — the current head — so both of my commits have had a genuine
  outside reviewer.
- **The part worth a third reader is the deleted null-guard** in `DigestionAgentName`:
  `SpecificProtease?.Name ?? DigestionAgent?.Name` became `SpecificProtease.Name`. That is a
  behavioural change to @acesnik's production code, made by me, to close a codecov patch gate whose
  uncovered branches were unreachable. The reasoning and the offer to revert it are in the comment
  above; it is still open, and reverting costs one partial line of coverage and nothing else.

The substance of the PR — reading `SpecificProtease` before `DigestionAgent` so match-between-runs
stays within a digestion agent — is unchanged from what I reviewed originally and still looks right.

Alexander-Sol (approved):

PR #2721, opened 2026-08-14
Refuse to update a spectral library when none was given

### LINKED ISSUE #2291: Update Spectral Library Bug
If the update spectral library option is checked, but no library is present, the search will continue as usual, but throw an exception when trying to update the library. 

This exception does not cause the program to crash, but rather hang indefinitely during the 'Writing PSM results' component of the search task. 

The results.txt file contained the following error after forcing the software to close.
MetaMorpheus: version 1.0.2
Data files were processed on a computer running Microsoft Windows 10.0.22621 with a 64-bit Intel(R) Xeon(R) Silver 4214R CPU @ 2.40GHz processor with 24 threads and 96GB installed RAM.
.NET version: .NET 6.0.14
e: System.NullReferenceException: Object reference not set to an instance of an object.
   at TaskLayer.PostSearchAnalysisTask.UpdateSpectralLibrary() in C:\projects\metamorpheus\MetaMorpheus\TaskLayer\SearchTask\PostSearchAnalysisTask.cs:line 631
   at TaskLayer.PostSearchAnalysisTask.Run() in C:\projects\metamorpheus\MetaMorpheus\TaskLayer\SearchTask\PostSearchAnalysisTask.cs:line 92
   at TaskLayer.SearchTask.RunSpecific(String OutputFolder, List`1 dbFilenameList, List`1 currentRawFileList, String taskId, FileSpecificParameters[] fileSettingsList) in C:\projects\metamorpheus\MetaMorpheus\TaskLayer\SearchTask\SearchTask.cs:line 407
   at TaskLayer.MetaMorpheusTask.RunTask(String output_folder, List`1 currentProteinDbFilenameList, List`1 currentRawDataFilepathList, String displayName) in C:\projects\metamorpheus\MetaMorpheus\TaskLayer\MetaMorpheusTask.cs:line 493e.Message: Object reference not set to an instance of an object.e.InnerException: e.Source: TaskLayere.StackTrace:    at TaskLayer.PostSearchAnalysisTask.UpdateSpectralLibrary() in C:\projects\metamorpheus\MetaMorpheus\TaskLayer\SearchTask\PostSearchAnalysisTask.cs:line 631
   at TaskLayer.PostSearchAnalysisTask.Run() in C:\projects\metamorpheus\MetaMorpheus\TaskLayer\SearchTask\PostSearchAnalysisTask.cs:line 92
   at TaskLayer.SearchTask.RunSpecific(String OutputFolder, List`1 dbFilenameList, List`1 currentRawFileList, String taskId, FileSpecificParameters[] fileSettingsList) in C:\projects\metamorpheus\MetaMorpheus\TaskLayer\SearchTask\SearchTask.cs:line 407
   at TaskLayer.MetaMorpheusTask.RunTask(String output_folder, List`1 currentProteinDbFilenameList, List`1 currentRawDataFilepathList, String displayName) in C:\projects\metamorpheus\MetaMorpheus\TaskLayer\MetaMorpheusTask.cs:line 493e.TargetSite: Void UpdateSpectralLibrary()

acesnik:
🤖 Fixes #2291.

Asking MetaMorpheus to update a spectral library without giving it one has always ended badly, because `UpdateSpectralLibrary` dereferences the library without checking whether it exists. What makes this worth more than a one-line null check is where in the run it happens: the search completes in full, every result is written, and only then does the code reach for a library that was never there.

Thanks @nbollis for the report, and particularly for the detail in it. The stack trace, the version, and the observation that the task stalled specifically during "Writing PSM results" between them identified the failing line and the point in the run it was reached from, which is most of the diagnosis.

## The behaviour described in the report no longer reproduces

When #2291 was filed against version 1.0.2, the failure announced itself: the task appeared to hang, and the `NullReferenceException` surfaced in results.txt only after force-quitting. Somewhere since then `UpdateSpectralLibrary` acquired a `try`/`catch`, which means it no longer hangs, and also that it no longer visibly fails. The exception goes to `EngineCrashed`, which writes a crash file and lets the task return as though all were well, so a test asserting a throw against current master finds nothing to catch:

```
Expected: <EngineLayer.MetaMorpheusException>
But was:  null
```

The live defect is therefore a run that reports success while quietly producing no updated library, with the reason sitting in an `UpdateSpectralLibrary_crash.txt` that nobody has reason to open. In one respect that is worse than the original hang, which at least demanded attention.

## Refusing early, where the answer is already known

The information needed to prevent this has been available all along. `LoadSpectralLibraries` returns null when the database list contains no library, and the code immediately below the search loop already accounts for that possibility:

```csharp
if (spectralLibrary != null && SearchParameters.UpdateSpectralLibrary == false)
```

Nothing, however, consulted it before committing to a search. `SearchTask` now checks the combination as soon as the libraries are loaded and refuses it there, so the cost of the mistake is a database load rather than an entire search, and the message explains the two ways out: supply a library to update, or ask for a new one to be written instead.

`UpdateSpectralLibrary` retains a null guard that warns and returns, which matters for any caller assembling `PostSearchAnalysisParameters` on its own. That guard deliberately does not throw. By the time it runs the search has finished and its results are on disk, and discarding all of that over a request that cannot be honoured would be a poor trade.

## Test

A single test in `Test/SpectralLibraryUpdateTests.cs` asserts the refusal and checks that the message names what is missing rather than merely reporting that something went wrong. It finishes in roughly half a second, which is as much the substance of the fix as the assertion is: reaching the check no longer requires searching anything. It fails against unfixed code, where, as above, nothing is thrown at all.

`Test` targets `net10.0-windows` and cannot run on macOS, so I exercised it through a throwaway net10.0 harness over `TaskLayer`. CI remains the real check.

## One thing deliberately left alone

A tempting alternative is to fall back to generating a library from the search results, which `SpectralLibraryGeneration` does a few lines above and which is plausibly what someone ticking this box actually wants. I have not done that, because it turns a bug fix into a behaviour change: it would emit a library through a path the user did not select, quietly reinterpreting one request as another. Refusing clearly and letting the user choose seemed the more honest option, though I am happy to take the other road if you would prefer it.


trishorts (commented):
Nice diagnosis, and I agree with both the framing and the decision not to silently fall back to generating a library. Three things below, plus a note on coverage.

I verified each of these against the branch rather than reasoning from the diff — file:line references are to this PR's head.

**Patch coverage.** Codecov is red at 57.14% (7 patch lines, 4 hit / 2 miss / 1 partial). All three uncovered lines are the new guard in `PostSearchAnalysisTask.cs:823-826` — unreachable through `SearchTask` by construction, which is exactly the point of it. I've written a set of fast direct-construction tests that cover it (and the update logic generally, which today is only exercised by a 7-second full search) and will open them as a PR against this branch so you can take or leave them.

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
```diff
@@ -174,6 +174,15 @@ protected override MyTaskResults RunSpecific(string OutputFolder, List<DbForTask
             // load spectral libraries
             var spectralLibrary = LoadSpectralLibraries(taskId, dbFilenameList);
 
+            // Checked here rather than where the library is updated, which happens after the whole search:
+            // there is nothing to update, and reporting that once the search has finished wastes the run.
+            if (SearchParameters.UpdateSpectralLibrary && spectralLibrary == null)
```
trishorts: **The check lands nine lines too late to deliver what the PR describes.**

`var proteinLoadingTask = dbLoader.RunAsync();` already fired at line 171, and `RunAsync` is `Task.Run(Run)` (`EngineLayer/MetaMorpheusEngine.cs:318`) with no cancellation token — there is no `CancellationToken` overload anywhere in TaskLayer/EngineLayer, and `DatabaseLoadingEngine.RunSpecific` never checks `GlobalVariables.StopLoops`. Throwing here abandons that task: it runs the FASTA load, decoy generation and `ScrambleHomologousDecoys` to completion on a thread-pool thread after `RunTask` has unhooked `FinishedSingleEngineHandler` (`MetaMorpheusTask.cs:685`) and rethrown, and any exception it raises is never observed.

The check doesn't need any of that work. `LoadSpectralLibraries` is a pure filter over `dbFilenameList` — `dbFilenameList.Where(p => p.IsSpectralLibrary)`, returning `null` when the list is empty (`MetaMorpheusTask.cs:741-746`), no I/O. So the whole test can move above line 170 and the cost of the mistake becomes nothing at all, rather than "a database load".

In practice the leak is small (thread-pool work, and `WriteTargetDecoyFasta` defaults to false so nothing is written), and `SearchTask` already abandons this task on other throw paths — `:301` for an unknown isobaric mass tag. So this isn't a new class of problem. But it is the one line of the PR whose placement is the whole argument, and hoisting it is free.

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
```diff
@@ -174,6 +174,15 @@ protected override MyTaskResults RunSpecific(string OutputFolder, List<DbForTask
...
+            {
+                throw new MetaMorpheusException(
```
trishorts: **Worth knowing before you settle on `throw`: in the GUI this surfaces as a crash report prompt, not as a configuration message.**

I traced it end to end. `RunTask` catches, writes a full dump into `results.txt` — system info, stack trace, `TargetSite` — and rethrows (`MetaMorpheusTask.cs:683-700`). `EverythingRunnerEngine.Run` has no try/catch around `RunTask`. `MainWindow` wires the runner with `t.ContinueWith(EverythingRunnerExceptionHandler, TaskContinuationOptions.OnlyOnFaulted)` (`GUI/MainWindow.xaml.cs:1020`), and that handler (`:431-490`) does not special-case `MetaMorpheusException`:

```
MessageBox.Show(message + "\n\nWould you like to report this crash?", "Runtime Error", MessageBoxButton.YesNo);
```

Answering Yes composes a mailto to `mm_support@chem.wisc.edu` with the stack trace and the toml. So a user who ticked one checkbox too many is told MetaMorpheus crashed and invited to file a bug about it. (Since you exercised this through a net10 harness on macOS, you'd have had no way to see that.)

The house channel for refusing a bad database/task combination is `EverythingRunnerEngine.Run`, which already does three of these with `Warn(...)` + `FinishedAllTasks(...)` + `return` — including one that inspects `IsSpectralLibrary` (`EverythingRunnerEngine.cs:78-86`, "Cannot proceed. No protein database files selected."). That gate runs before `RunTask`, has access to both the task list and the database list, and already has a test asserting on its warning string (`Test/MyTaskTest.cs:839` `MissingDbInSpectralLibrarySearch`). Putting the refusal there would be earlier still, and would reach the user as a notification rather than a crash dialog.

Counter-argument, which is real: `CalibrationTask.GenerateIndexes` (`CalibrationTask.cs:349-361`) throws `MetaMorpheusException` for exactly this shape of impossible combination and is the newest precedent in the repo, and `SearchTask.cs:301` does the same. So the inline throw isn't out of keeping. I'd still rather this one didn't read as a crash — either move it to the runner, or keep it here and teach `EverythingRunnerExceptionHandler` to present `MetaMorpheusException` as a message rather than a crash report (that second one is its own PR).

Either way, not something to fix silently — your call, and worth saying out loud in the PR description if you keep the throw.

MetaMorpheus/Test/SpectralLibraryUpdateTests.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            var task = new SearchTask();
+            task.SearchParameters.UpdateSpectralLibrary = true;
+
+            string database = Path.Combine(TestContext.CurrentContext.TestDirectory,
+                "TestData", "hela_snip_for_unitTest.fasta");
+            string spectra = Path.Combine(TestContext.CurrentContext.TestDirectory,
+                "TestData", "TaGe_SA_A549_3_snip.mzML");
+
+            var thrown = Assert.Throws<MetaMorpheusException>(() => task.RunTask(
```
trishorts: **The test name and doc comment promise "before searching"; the assertions only prove "at all".**

The three assertions are `Assert.Throws<MetaMorpheusException>` plus two substring checks on the message. Nothing observes that the search loop was skipped. The check could migrate anywhere upstream of `PostSearchAnalysisTask` — after the per-file search loop, say — and this test stays green, which is the half of the fix you describe as "as much the substance of the fix as the assertion is".

(To be precise about the one place it *would* catch: moving the throw into `UpdateSpectralLibrary()` itself would be swallowed by the `try`/`catch` at `PostSearchAnalysisTask.cs:820/892-895` and `Assert.Throws` would fail on "But was: null" — the same failure you quote in the description. So the test does pin *some* earliness. It just doesn't pin "before the search".)

Cheapest fix that pins it, after the expected exception:

```csharp
Assert.That(Directory.GetFiles(outputFolder, "*.psmtsv", SearchOption.AllDirectories), Is.Empty,
    "the refusal has to come before the search, not after it");
Assert.That(Directory.Exists(Path.Combine(outputFolder, "Individual File Results")), Is.False);
```

Two smaller things while you're in here:

- The message assertions cover the diagnosis but not the remedy. `Does.Contain("no spectral library was given")` subsumes `Does.Contain("spectral library")`, and neither pins the half of the message that tells the user what to do about it — the string could be trimmed to the diagnosis and this stays green. One more substring on the "add one to the list of databases / write a new one instead" clause would cover it.
- `RunTask` writes its toml to `Directory.GetParent(output_folder)/"Task Settings"` (`MetaMorpheusTask.cs:627`), i.e. `SpectralLibraryUpdateTests/Task Settings/`, one level above the folder you delete — so it survives every run. And the delete is the last statement rather than a `finally`, so a failing assertion leaves `NoLibrary/` behind too. `CalibrationTests.cs:273` and `BinGenerationTest.cs:80` are the sibling pattern. (The test PR I'm sending you handles this with a `[OneTimeTearDown]` on the whole `SpectralLibraryUpdateTests` root.)

acesnik:
🤖 All three taken, and your test PR merged as-is. Four commits on top: `ab0b0330` (yours), `d4cf63d2`, `a46a8405`, `c551b165`.

## Coverage

Took acesnik#7 wholesale rather than writing a second set — it covers `PostSearchAnalysisTask.cs:823-826` exactly, and the fact that you checked it fails with the guard deleted is what made it worth taking unedited. Merged as `ab0b0330`, so #7 closes with it.

## The check landing nine lines too late — `d4cf63d2`

Right, and hoisted, with one change to what moves. Moving the `LoadSpectralLibraries` call itself would cost something on the happy path: `SpectralLibrary`'s constructor opens every library file and builds a byte-offset index (`IndexSpectralLibrary`, mzLib `SpectralLibrary.cs:132-144`), so putting it above `RunAsync` delays the background protein load for every *normal* library search to save work only in the error case.

So only the predicate moved. `AnySpectralLibrary(dbFilenameList)` sits next to `LoadSpectralLibraries` — your `LoadSpectralLibrariesReturnsNullOnlyWhenNoLibraryIsInTheDatabaseList` is exactly the test that keeps the two from drifting. Reaching the refusal now costs no I/O at all, so the doc comment's "costs a database load, not a search" is gone too.

## Reading as a crash — `c551b165`

Moved to the runner. You're right that this is the house channel, and it's now a fourth gate beside the three existing `Warn` + `FinishedAllTasks` + `return` ones.

Two things worth flagging about the placement. It goes **inside** the per-task loop, not before it: `CurrentXmlDbFilenameList` grows as tasks chain, so a search that writes a library followed by one that updates it has to stay legal, and an up-front check would break that. And I kept the task-level throw as a backstop for callers that use `RunTask` directly — it's now unreachable through the runner, which is the same relationship this PR already sets up between `SearchTask` and the `PostSearchAnalysisTask` guard. Happy to drop it if you'd rather it were strictly either/or.

I confirmed by grep that `EverythingRunnerExceptionHandler` doesn't special-case `MetaMorpheusException`, and that there are **32** `throw new MetaMorpheusException` sites (14 in TaskLayer) that would all present as crash reports. I did not run the GUI myself — that part of the trace is yours. Teaching the handler to present them as messages is the fix worth having, and it's its own PR as you say.

## The test promising more than it proves — `c551b165`

The point stands, but the assertion you suggested doesn't discriminate. I tried it: with the guard moved down to just before `PostSearchAnalysisTask`, `Directory.GetFiles(outputFolder, "*.psmtsv", AllDirectories)` is still empty and the test still passes. Every result file is written *by* that task, so absence of output can't tell "search skipped" from "search ran, post-processing skipped".

Engine start events do discriminate. With the guard late:

```
the refusal has to come before any work starts, but these ran: DatabaseLoadingEngine, ClassicSearchEngine
```

Also added the remedy-clause assertion and dropped `Does.Contain("spectral library")`, which the other substring subsumed. Cleanup is your `[OneTimeTearDown]` now.

There's a new `TheRunnerRefusesAnUpdateWithNoLibraryAndRunsNothing` for the runner path — warning text plus no engines started. It reads `runner.Warnings` rather than subscribing, because `EverythingRunnerEngine` has its own static `WarnHandler` separate from `MetaMorpheusTask`'s and your `CaptureWarnings` silently returns nothing for it. `MyTaskTest.MissingDbInSpectralLibrarySearch` was the pattern to copy.

## One thing your tests turned up — `a46a8405`

Five of your eight failed on macOS and pass on Windows. Not the tests: both spectral library writers concatenate a literal backslash onto the output folder.

```csharp
string spectrumFilePath = outputFolder + "\\updateSpectralLibrary" + "_" + ... + ".msp";
```

On macOS and Linux a backslash is an ordinary filename character, so the updated library is written into the **parent** directory under a single name containing a backslash, and the path handed to `NewDatabases` points at something the next task can't resolve. `Path.Combine` on `MetaMorpheusTask.cs:1123` and `:1137` fixes it; all 8 then pass on both. It's a no-op on Windows, which is also why CI could never have caught it — the test project runs on windows-latest only.

Pre-existing and outside what this PR is about, so it's in its own commit if you'd rather drop it. The same pattern is in ~10 more places in `PostGlycoSearchAnalysisTask`, left alone.

## Not verified

9 tests green locally in under a second, but via a throwaway net10.0 harness over `TaskLayer`/`EngineLayer` — `Test` can't run on macOS. `Test.csproj` compiles with `EnableWindowsTargeting`. CI is the real check, and it's the only place the Windows path behaviour gets exercised at all.

nbollis (dismissed):

zhuoxinshi (dismissed):

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
```diff
@@ -223,6 +223,28 @@ protected override MyTaskResults RunSpecific(string OutputFolder, List<DbForTask
...
+            // Checked before the protein load starts, rather than where the library is updated after the
+            // whole search: there is nothing to update, and reporting it once the search has finished wastes
+            // the run. It has to precede RunAsync as well -- that is Task.Run with no cancellation, so
+            // throwing after it leaves the load running unobserved. Reads the same precondition
+            // LoadSpectralLibraries does, without opening the library.
+            // Checked before the protein load starts, rather than where the library is updated after the
+            // whole search: there is nothing to update, and reporting it once the search has finished wastes
+            // the run. It has to precede RunAsync as well -- that is Task.Run with no cancellation, so
+            // throwing after it leaves the load running unobserved. Reads the same precondition
+            // LoadSpectralLibraries does, without opening the library.
```
trishorts, line 232: **This comment block is pasted three times.**

Lines 226-240 at this head are the same five-line comment repeated verbatim, three times, immediately above the guard:

```
226-230:  // Checked before the protein load starts, ... without opening the library.
231-235:  // Checked before the protein load starts, ... without opening the library.
236-240:  // Checked before the protein load starts, ... without opening the library.
```

Almost certainly an editor accident during the rework rather than anything intentional. Two of the three want deleting before this merges.

The comment itself is good and worth keeping once — the reasoning about `RunAsync` being `Task.Run` with no cancellation, so a throw placed after it leaves the load running unobserved, is the part of this fix that is least obvious from the code.

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
trishorts: Confirmed fixed. The guard now sits at `SearchTask.cs:241`, ahead of both the `DatabaseLoadingEngine` construction and `RunAsync` at :249-250, so it no longer lands after the work it was meant to prevent. Resolving.

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
trishorts: Confirmed addressed, and better than what I suggested. Rather than swapping the throw for a return, you added the refusal to `EverythingRunnerEngine` (:90-101) where it reads as a warning and a clean stop, and kept the throw in `SearchTask` as a backstop for a caller invoking `RunTask` directly. That is the right split — every user-facing entry point goes through the runner, so the GUI gets a notification rather than a crash report, and the direct-call path still cannot proceed silently.

I checked the "checked per task rather than up front" reasoning too, since it is the load-bearing claim for putting it inside the loop: `EverythingRunnerEngine.cs:116` reassigns `CurrentXmlDbFilenameList = myTaskResults.NewDatabases`, so an earlier task really can introduce a library and a write-then-update pair really does stay legal. Resolving.

MetaMorpheus/Test/SpectralLibraryUpdateTests.cs
trishorts: Confirmed fixed, and this is exactly the distinction I was asking for. `UpdatingASpectralLibraryWithoutOneIsRefusedBeforeSearching` now subscribes to `MetaMorpheusEngine.StartingSingleEngineHander` and asserts `enginesStarted` is empty, which actually discriminates "refused before searching" from "refused at all".

The comment explaining why the obvious alternative does not work — that absence of output files proves nothing, because every result file is written by `PostSearchAnalysisTask`, so any check ahead of that leaves the folder empty even if the whole search ran — is the part worth keeping. That was the trap in the original version.

Ran the file locally: 9/9 green. Resolving.

trishorts:
Re-reviewed at `d89c7b4`. All three of my earlier threads are properly addressed and resolved — the guard moved ahead of the background load, the refusal moved into `EverythingRunnerEngine` so it reads as a notification rather than a crash report, and the ordering test now proves "before searching" by observing engine-start events instead of an empty output folder. Built clean and `SpectralLibraryUpdateTests` is 9/9 green locally.

One blocker, and it is purely mechanical: the comment above the guard is **pasted three times** (`SearchTask.cs:226-240`). Thread opened.

One nit, non-blocking: `AnySpectralLibrary`'s doc says it is "kept beside `LoadSpectralLibraries` so the two cannot drift apart", but `LoadSpectralLibraries` does not call it — it still inlines `dbFilenameList.Where(p => p.IsSpectralLibrary)` at `:751`, and `EverythingRunnerEngine:94` inlines a third copy. Three encodings of "is there a spectral library", which is what the helper was meant to prevent. Either route the other two through it or soften the comment; both are fine, but the claim as written is not true today.

trishorts:
🤖 Re-checked at `d89c7b47`. Nothing has been pushed since 25 Aug, and the triplicated comment block is still there — `SearchTask.cs:226-240`, the same five lines three times above the guard. My other three threads are confirmed fixed and resolved.

That is the only thing holding an approval from me, and it is a fifteen-line deletion. Everything else about the change is right, including the `EverythingRunnerEngine` refusal, which is better than what I originally suggested.

acesnik:
🤖 Duplicated comment deleted — that was an editor accident during the rework, exactly as you guessed, and it was the only one of your four still open. One copy left. `SpectralLibraryUpdateTests` is 9/9 green here too, and the solution and `Test.csproj` both build clean.

Thank you for the three you closed yourself; the `RunAsync` placement one in particular I would not have found from the diff, since the reason it matters is that `Task.Run` has no cancellation rather than anything visible at the call site.

On coverage: the three unhit lines are still the `PostSearchAnalysisTask` guard, which is unreachable through `SearchTask` by construction now that the refusal happens in the runner — so the 57% is describing a backstop rather than a gap. You mentioned sending the direct-construction tests as a PR against this branch; nothing has landed here, so either it slipped or you thought better of it. Happy either way — say if you would still like to send them and I will hold this, or I will land it as is and the guard stays covered only by the argument for why it cannot fire.

One thing worth flagging: pushing the comment fix dismissed nbollis' and zhuoxinshi's approvals.

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
trishorts: Confirmed fixed in `f8331faa`. The comment now appears exactly once, at `SearchTask.cs:228-232`, still ahead of the `DatabaseLoadingEngine` construction and `RunAsync`. The commit is a pure deletion of the two duplicate copies -- +0/-10, comment lines only. Resolving.

trishorts (approved):
Approving. All four of my comments are addressed and I re-verified each against the current head.

The guard sits at `SearchTask.cs:233`, ahead of both the `DatabaseLoadingEngine` construction and `RunAsync`, so it precedes the work it exists to prevent. The refusal was added to `EverythingRunnerEngine` where it reads as a clean stop, with the throw kept in `SearchTask` as a backstop for a direct `RunTask` caller — a better split than the return I suggested. `UpdatingASpectralLibraryWithoutOneIsRefusedBeforeSearching` now asserts no engine started, which actually discriminates "before searching" from "at all". And `f8331faa` deletes the two duplicate copies of the guard comment (+0/-10, comment lines only).

CI is green.

trishorts (commented):
Refusing this up front instead of running a whole search and then throwing `NullReferenceException` out of `UpdateSpectralLibrary` is clearly right, and the reasoning about *where* the check has to sit -- before `RunAsync`, because that is `Task.Run` with no cancellation and throwing after it leaves the load running unobserved -- is the kind of thing worth having in a comment.

The test that asserts **no engine starts** rather than "no output files exist" is the correct discriminator, and the comment explaining why the obvious assertion would not work is exactly right: every result file comes from `PostSearchAnalysisTask`, so an empty folder proves nothing.

Two design points, inline. Both are about placement rather than correctness -- the behaviour you want is the behaviour you get, on the path users actually take.

The `Path.Combine` fixes to `WriteSpectrumLibrary` / `UpdateSpectralLibrary` are unrelated to the stated scope but are real bugs (`outputFolder + "\\..."` is broken off Windows), so I would keep them and just mention them in the description so they do not read as churn.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -746,6 +746,16 @@ public MyTaskResults RunTask(string output_folder, List<DbForTask> currentProtei
 
         #region Database Loading
 
+        /// <summary>
+        /// The precondition <see cref="LoadSpectralLibraries"/> reads before it opens anything, so a caller
+        /// can test for a library without paying for the byte-offset index the SpectralLibrary constructor
+        /// builds. Kept beside it so the two cannot drift apart.
+        /// </summary>
+        protected static bool AnySpectralLibrary(List<DbForTask> dbFilenameList)
```
trishorts, line 754: **This helper cannot serve the purpose its docstring claims, and there is already a copy it does not fix.**

The docstring says it is "kept beside it so the two cannot drift apart" -- but `protected static` on `MetaMorpheusTask` is reachable only from tasks. `EverythingRunnerEngine` is not a `MetaMorpheusTask`, which is why `EverythingRunnerEngine.cs:94` has to write `!CurrentXmlDbFilenameList.Any(p => p.IsSpectralLibrary)` inline. So the drift risk the helper exists to remove is present in this very PR.

It is also not the first copy. `GPTMDTask.cs:212` already contains `dbFilenameList.Any(p => p.IsSpectralLibrary)` -- character-identical to this method body -- and is untouched here. After this PR the same test is written three ways in three files.

The conventional home is an extension in `EngineLayer/Util/`, which already holds exactly this shape of helper: `DictionaryExtensions.cs` and `MzlibExtensions.cs`, both declaring the file-scoped `namespace EngineLayer;`. Something like

```csharp
public static bool AnySpectralLibrary(this List<DbForTask> dbs) => dbs.Any(p => p.IsSpectralLibrary);
```

is then visible to `MetaMorpheusTask`, `EverythingRunnerEngine` (which already has `using EngineLayer;`), `GPTMDTask` and `DatabaseLoadingEngine` alike -- and `GPTMDTask.cs:212` can collapse onto it. `DbForTask` itself has no static members today, so an extension is the better fit than putting it on the type.

PR #2723, opened 2026-08-14
Docs: Thermo .raw reading works on macOS, not just Windows and Linux

acesnik:
🤖 The README tells macOS users that Thermo `.raw` files are off limits to them:

```
* Thermo .raw (Windows and Linux only)
```

That is not true, and as far as I can tell it has not been true for some time. It matters more than a
typo would, because it is the first thing a prospective macOS user reads about input formats, and it
points them at a file conversion step they do not actually need.

This changes the line to `(Windows, macOS and Linux)`. The one-line diff is the whole change.

## Verified rather than reasoned

I did not want to correct a platform-support claim on the strength of an argument about managed
assemblies, so I ran the real thing: a search task over a Thermo `.raw` file, on macOS 26.6, once on
a **native arm64** runtime and once on an **x64** runtime under Rosetta.

```
dotnet CMD.dll -t SearchTask.toml -s slicedMouse.raw -d slicedMouseDatabase.fasta.gz -o out
```

| runtime | MS2 scans | target PSMs ≤ 1% FDR | peptides | protein groups |
|---|---|---|---|---|
| arm64 (native, .NET 8.0.423) | 5 | 7 | 7 | 7 |
| x64 (Rosetta, .NET 8.0.204) | 5 | 7 | 7 | 7 |

Both runs read the file, deconvoluted it, searched it, quantified it and wrote the full output set
including `.mzID`. The two architectures agree exactly. Input was `slicedMouse.raw` with its paired
database from `Test/TestData/SpectralLibrarySearch/`, copied into a scratch directory so nothing in
the repository was touched.

## Why it works, and why one might have expected otherwise

The ThermoFisher CommonCore readers are managed assemblies, so nothing on the `.raw` path is
platform-specific. Reading them needs no native code.

The reasonable objection is `ThermoFisher.CommonCore.MassPrecisionEstimator`. It is consumed as a
managed `<Reference>` in mzLib's `Readers.csproj` rather than as a P/Invoke target, and its PE machine
field is stamped `AMD64`, which CoreCLR on Apple Silicon refuses to load. If the loader ever had to
resolve it, it would throw `BadImageFormatException` and no amount of managed portability elsewhere
would help.

It never has to. The assembly is referenced by the project file and used by no code: there is not a
single `.cs` reference to it anywhere in mzLib or MetaMorpheus, and no Thermo assembly names it as a
string either, so it is not loaded reflectively. The CLR resolves assemblies lazily, on first use of a
type, so a reference nothing touches is never loaded and its bitness never comes up. That is why the
runs above succeed despite an x64-stamped binary sitting in the output directory.

Worth being explicit that this was fragile in one direction: the day something called into
`MassPrecisionEstimator`, `.raw` reading would break on every non-x64 platform at once, and the
failure would surface as an assembly load error a long way from the call that caused it. That is now
addressed on the mzLib side — the dead reference is dropped and a test checks that no assembly in the
build output references an architecture-stamped one, so this cannot be reintroduced quietly.

## One thing this turned up, separate from the docs

The first `.raw` search I attempted died like this:

```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at MetaMorpheusCommandLine.Program.Run(CommandLineSettings settings) in CMD/Program.cs:line 160
```

`CMD/Program.cs:160` reads the Thermo licence agreement from the console:

```csharp
string res = Console.ReadLine().ToLowerInvariant();
```

`Console.ReadLine()` returns `null` at end of input, so any non-interactive first run with a `.raw`
file — a CI job, a scheduled cluster batch job, `nohup`, or anything with stdin redirected — crashes
with an unhandled `NullReferenceException` instead of saying what it wanted. It only bites before
`settings.toml` records the agreement, which is exactly the situation on a fresh machine or in a
container. Recording it here rather than fixing it in this PR, which is meant to stay a one-line docs
change.


trishorts:
🤖 The claim checks out from the code side too, so this is safe to merge as the one-liner it is.

I traced the read path in mzLib at `smith/master @ 03081569`: `MsDataFileReader.GetDataFile` dispatches `.raw` to `ThermoRawFileReader`, which calls `RawFileReaderAdapter.FileFactory` and `RawFileReaderFactory.CreateThreadManager` directly. There is no platform check, no P/Invoke, and no licence gate anywhere in that path — `ThermoRawFileReaderLicence` is a bare `public static string` whose only consumer in mzLib is a test asserting it's longer than 100 characters. Nothing in it could have been Windows-and-Linux-only, which matches your two runs.

Your `MassPrecisionEstimator` note is the right thing to have chased, and the mzLib-side removal plus the architecture-stamp test is the durable fix.

### One thing the README change surfaces, worth a follow-up rather than a change here

The line now correctly says macOS can read `.raw`. What it can't say yet is that agreeing to the licence is meaningfully harder there.

`GlobalVariables.SetUpDataDirectory()` (`EngineLayer/GlobalVariables.cs:357-370`) only redirects `DataDir` to `%LOCALAPPDATA%\MetaMorpheus` when the executable sits under `ProgramFiles`. On Unix `GetFolderPath(SpecialFolder.ProgramFiles)` returns empty, the `IsNullOrWhiteSpace` guard trips, and `DataDir` becomes `AppDomain.CurrentDomain.BaseDirectory` — the extracted folder. So the agreement recorded in `settings.toml` is stored **per extracted copy** on macOS and Linux, versus per user on a Windows MSI install. Every fresh download, conda env, and container layer asks again, and a container has no tty to answer with, which is exactly the NRE you hit and documented here.

There's also no GUI on macOS to click Agree in, and no CLI flag or environment variable for it — the option list is `-t -d -s -o -g -v --test --mmsettings`.

I've written the full flow up on #2724, and I'm following it with a small PR adding an explicit accept-the-licence flag so non-Windows and scripted users have a supported route instead of hand-editing a toml. Nothing needed in this PR — the one-line correction is right on its own.

trishorts:
🤖 Opened it: #2725.

It adds `--acceptThermoLicence` and the operational half of the story this PR corrects: reading a `.raw` works on macOS, and now agreeing to the licence there does too, without a console and without hand-editing a toml.

The README edits are in different sections — yours is the format bullet, mine is a new subsection at the end of Spectra Requirements — so they should not overlap.

nbollis (approved):

trishorts:
Verified rather than taken on trust — this is correct.

mzLib issue [#1158](https://github.com/smith-chem-wisc/mzLib/issues/1158) is the tracking issue that measured cross-platform reading, and it puts Thermo `.raw` in the working set on all five targets (Windows, Linux x64/arm64, macOS x64/arm64), from a full search run on both Mac architectures — 7/7 target PSMs at 1% FDR, identical results. It names this PR as the fix for the README claim.

So the line is now accurate, and it is the last place the old "Windows and Linux only" wording survived.

One adjacent note, not a change request: the same table shows Bruker `.d` and `.baf` are Windows-only, and for `.d` that is not fixable by shipping a missing file — Bruker publish no macOS build of their SDK at all. The README does not currently qualify the Bruker formats by platform. Out of scope here, but if someone is doing a platform-accuracy pass on this section, that is the other half of it.

trishorts (approved):
🤖 Approving. One line, correct, and the right kind of evidence behind it — two runs on real hardware beat any argument about managed assemblies, and you named the objection (`MassPrecisionEstimator` stamped `AMD64`) instead of hoping nobody would raise it.

Two things I checked rather than took on trust:

**The mzLib-side claim in the body is now true, and was not when you opened this.** smith-chem-wisc/mzLib#1156 merged yesterday (2026-08-27, `3829034d`), dropping the dead `<Reference>` and adding the guard against an architecture-stamped assembly returning. Until MetaMorpheus picks that up the README line rests entirely on the lazy-loading argument — which is sound, and which your two runs demonstrate directly, so the claim is true either way. Worth noting the pin here is already `1.0.587`, so the bump that carries the guard is a normal release away.

**The rest of the README agrees with the new line.** Line 36 already says the MSI is Windows-only while the command-line version runs anywhere .NET does, and line 34 already names macOS on both Intel and Apple Silicon. The `.raw` line was the one place that contradicted them, so this removes an inconsistency rather than introducing one.

I have filed the `Console.ReadLine()` crash you recorded at the end of the body as #2770 — it is real on master and there are two sites, not one (`CMD/Program.cs:171` and `:307`), the second returning a bare `5` that is already taken. Correct call keeping it out of a one-line docs change.

trishorts (approved):
Approving. I re-verified the claim independently rather than taking the runs on trust: nothing on the Thermo `.raw` path is platform-specific, and the one assembly that would have made it so — `ThermoFisher.CommonCore.MassPrecisionEstimator`, whose PE machine field is stamped AMD64 — has no `.cs` reference anywhere in mzLib or MetaMorpheus and is not named as a string, so the loader never resolves it and its bitness never comes up.

The README line was the last place still telling macOS users to go and convert their files. One-line diff, correct claim.

trishorts (commented):
One-line docs change, so I read it directly rather than running the engine. No findings, and nothing inline to attach one to. What I did do was check the claim independently rather than take the run table on trust.

**The load-bearing claim is durable, not pending.** The body says the `MassPrecisionEstimator` fragility "is now addressed on the mzLib side". That is merged, not in flight -- on mzLib master the only remaining mentions are the guard itself, `Test/FileReadingTests/ReadersInfrastructure/TestNoArchitectureSpecificAssemblyReferences.cs`, and the x64-stamped fixture DLL it loads via `Test.csproj`. `Readers.csproj` no longer references it. So the one thing that could have made this README line wrong again is both removed and pinned.

**The fix is complete, not half-applied.** `grep -rn -i "Windows and Linux|Linux only|Windows only"` across every `.md`, `.cs` and `.xaml` in the repo returns exactly one hit: `README.md:42`, the line this PR changes. Nothing else in the tree repeats the stale claim, and the surrounding text is already consistent -- line 36 draws the installer-vs-CLI distinction correctly, and the "Agreeing to the Thermo licence" section right below already documents `--acceptThermoLicence`.

**The deferred bug is already tracked.** The `Console.ReadLine().ToLowerInvariant()` NRE is real -- now at `MetaMorpheus/CMD/Program.cs:171` rather than the 160 in the body, master has moved -- and the second prompt at line 307 has the same hole, `result.ToLowerInvariant()` on an unchecked return. Both are covered by #2770. Right call keeping it out of a one-line docs PR, and no issue needs filing.

Already approved; this is the verification behind it rather than a new ask. Only thing left is that the branch is `BEHIND`.

Alexander-Sol (approved):
.raw file reading works on my (Intel) mac

Issue #2256, opened 2023-02-27
Error Reading Experimental Design File when Re-Running a Search


conorjenkins:
**Describing the issue**
After a search has been successfully completed with a calibration, gtpmd and search task, I received an Error on reading the experimental design file when attempting to re-run the search with an added a modification to the gtpmd task.

**Reproducing the error**
Attempt to re run a search after completion. Error has occurred twice so far. 

- Protein database
Downloaded XML file from uniprot

- Mass spectrometry file types
Thermo Raw/ Calibrated MZML files

- Example datafiles
This is the experimental design file containing the updated calibrated files that metamorpheous generated. it looks fine to me. Had to conver from tsv to txt to upload here.
[ExperimentalDesign.txt](https://github.com/smith-chem-wisc/MetaMorpheus/files/10840901/ExperimentalDesign.txt)

**Expected behavior for GUI**
Expected to re run the samples in another search.

**Screenshots**
Error Recieved
![Errror with Experimental design file when attempting to re-run a search](https://user-images.githubusercontent.com/85502162/221597467-418e5a80-7b3e-4f87-9f4f-12f7f35c4754.PNG)

Experimental design shown in gui. Looks ok to me
![Errror with Experimental design file when attempting to re-run a search image 2](https://user-images.githubusercontent.com/85502162/221597621-a49368f3-9f9a-45c4-a450-bc919073217d.PNG)

**Desktop information**
[AutoGeneratedManuscriptProse.txt](https://github.com/smith-chem-wisc/MetaMorpheus/files/10840929/AutoGeneratedManuscriptProse.txt)

**Additional context**
Not fatal to the program and I can start from the beginning by setting up a new run with the same samples. I have not tried rerunning a search with the calibrated mzml files to see if that works. Let me know if you want me to upload anything to a google drive.

acesnik:
🤖 Reproduced, and it is our bug — thanks for attaching the design file, that is what pinned it down.

[...]

PR #2724, opened 2026-08-14
Explain the Thermo licence requirement instead of crashing when there is no console

acesnik:
🤖 The first time MetaMorpheus is asked to search a Thermo `.raw` file on a given machine, it prints the
RawFileReader licence and waits for a yes or no. If there is no console to answer from, it crashes:

```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at MetaMorpheusCommandLine.Program.Run(CommandLineSettings settings) in CMD/Program.cs:line 160
```

`Console.ReadLine()` returns `null` at end of input, and the result was lowercased without a check:

```csharp
string res = Console.ReadLine().ToLowerInvariant();
```

So a CI job, a scheduled cluster batch job, a container, `nohup`, or anything else with stdin redirected
or closed gets a null-reference stack trace instead of being told that a licence agreement is what stands
between it and the search. The message it needed was printed immediately above the crash, which makes
this easy to diagnose by eye and easy to miss entirely in a log.

It only happens before `settings.toml` records the agreement, so it is invisible on a machine that has
answered once and reliable on a fresh one — a new CI runner, a fresh container layer, a new user's
install. That is also why it has survived: interactive first runs work fine, and everyone's second run
works fine.

I found it while verifying #2723, by piping a search through `tail`.

## The change

Read the answer defensively, and treat "no answer available" as its own outcome rather than as a refusal:

```csharp
string res = Console.ReadLine()?.Trim().ToLowerInvariant();
```

A `null` answer now reports that nothing could be read, names the two ways to agree without a console —
pipe a response, or set the flag in `settings.toml`, with the resolved path printed — and returns a new
exit code **5**. That keeps it distinct from the existing **3**, which means the licence was actively
declined; a batch job that could not be asked is not the same as a user saying no, and a script should be
able to tell them apart.

Answers are also trimmed now, so a piped response carrying trailing whitespace or a stray `\r` is read as
intended rather than silently treated as a refusal.

## Verified against the real CLI

Built from this branch on macOS, `net10.0`, with the licence not yet agreed in the build output. Confirmed
first that the crash reproduces on unmodified `master` at `258d7484` — same `NullReferenceException`, same
line — then that each path behaves:

| invocation | exit | outcome |
|---|---|---|
| stdin closed (`< /dev/null`), unfixed master | — | `NullReferenceException`, unhandled |
| stdin closed (`< /dev/null`), this branch | **5** | explains the requirement and how to agree |
| `echo n \|` | 3 | declines, as before |
| `echo y \|` | 0 | full search completes: 5 MS2 scans, 7 target PSMs at 1% FDR |

`settings.toml` was checked after each: `UserHasAgreedToThermoRawFileReaderLicence` stays `false` after
the closed-stdin and declined runs, and flips to `true` only after an actual `y`. Nothing agrees to a
third-party licence on the user's behalf, which was the thing to be careful about here.

## On a regression test

I did not add one, and I would rather say so than let it pass unmentioned. `Program.Run` is private, and
reaching this branch needs three things arranged at once — a `.raw` in the spectra list, the agreement
flag unset in global state, and stdin at EOF. From the `Test` project that means either an
`InternalsVisibleTo` on `CMD` plus mutating `GlobalVariables.GlobalSettings` around the call, or spawning
the built CLI as a subprocess. Both are a fair amount of machinery for a null check, and a test that
found the agreement already recorded on the machine would pass without exercising anything.

The table above is what I trusted instead: the crash reproduced on master, the fix verified on all four
paths. Happy to add either form of test if reviewers would prefer one.


trishorts:
🤖 I went and mapped the whole licence flow before reviewing this, because I wanted to know whether the null check was the only thing wrong with it. It isn't, and two of the others are the same bug wearing different clothes. Everything below is against `upstream/master @ 258d74848` and mzLib `smith/master @ 03081569`.

The change itself is right. `?.Trim()` and a distinct exit code are both correct, and I agree with keeping 3 and 5 apart — a job that could not be asked is not a job that said no.

### The trim fixes a real, separately-reported annoyance

Worth putting in the PR body, because it is a second justification for a change you framed as defensive. `res` is compared to the literal `"y"` (`CMD/Program.cs:162`), untrimmed on master. So piping from PowerShell:

```powershell
"y" | & MetaMorpheusCommandLine.exe -t Task.toml -s data.raw ...
```

delivers `y\r`, which is not `"y"`, so it takes the else branch and exits **3 — licence declined**. That has burned me on this machine, and I had written it off as "the pipe is unreliable." It isn't unreliable, it's `\r`. Your `.Trim()` is the fix.

While you are there: this prompt accepts only the literal `y`, but the ExperimentalDesign prompt about a hundred lines down accepts `"y" || "yes"`. `yes` currently declines and exits 3, which is a surprising way to lose a run. Since you are already touching the parse, matching the sibling prompt seems free.

### Same bug, second instance: the GUI dialog

The GUI has the identical unchecked-nullable pattern, and it is reachable by clicking the wrong thing.

`GUI/MainWindow.xaml.cs:1759-1767`:

```csharp
var dialogResult = thermoLicenceWindow.ShowDialog();

var newGlobalSettings = new GlobalSettings
{
    UserHasAgreedToThermoRawFileReaderLicence = dialogResult.Value,
```

`ThermoLicenceAgreementWindow` sets `DialogResult` only in its two button handlers (`GUI/Views/ThermoLicenceAgreementWindow.xaml.cs:15-25`). Close it with the title-bar **X** and WPF returns `null`, so `dialogResult.Value` throws `InvalidOperationException`. Closing a dialog with X instead of the decline button is an ordinary thing to do.

Not asking you to fix it here — it's the GUI and this PR is CMD — but it is the same "`bool?` dereferenced without a check" mistake in the same feature, so it may be worth one issue covering both so the second one doesn't outlive the first.

### Same bug, third instance: read-only data directory

`GlobalVariables.cs:543-551` only assigns `GlobalSettings` if `settings.toml` exists, and skips creating it when the directory is read-only. `GlobalSettings` then stays null and `Program.cs:155` dereferences it — one line *above* the one you're fixing, and with the same stack shape.

### On the regression test

Your reasoning for not adding one is sound and I'd rather have the verified table than a test that passes because the machine already agreed. One cheap option if reviewers push: the branch needs three things arranged at once, and the third is already injectable — `GlobalVariables.GlobalSettings` is settable, so an `InternalsVisibleTo` on `CMD` plus a `TextReader` seam on the read would make it a unit test. Both are more machinery than the null check deserves today. I'm satisfied with the four-path table.

### Two things this flow does that surprised me, for the record

**There is no first-run licence dialog in the GUI.** I had assumed a fresh install prompts on launch. It doesn't — `MainWindow()` and `MyWindow_Loaded` have no licence check at all. The dialog appears the first time a `.raw` is *added* (`AddPreRunFile`, `case ".raw"`). And on decline it's a bare `return` with no message, so the file just silently doesn't appear in the list.

**mzLib enforces nothing.** `ThermoRawFileReaderLicence` is a single `public static string` with no method, flag, or check, and its only consumer in mzLib is a test asserting the string is longer than 100 characters. `MsDataFileReader.GetDataFile` takes no licence parameter. So the entire §3.3 "require your end users to agree" obligation is discharged in exactly two places, both in MetaMorpheus: `Program.cs:155` and `AddPreRunFile`. Which means **MetaDraw reads `.raw` with no licence check at all** — there is no Thermo or Licence reference in any MetaDraw file at this ref.

### The macOS/Linux side, which is where I'm taking this next

Related to #2723. The reason this crash is "reliable on a fresh machine" is not evenly distributed across platforms:

`GlobalVariables.SetUpDataDirectory()` (`GlobalVariables.cs:357-370`) sends `DataDir` to `%LOCALAPPDATA%\MetaMorpheus` only when the executable is under `ProgramFiles`. `Environment.GetFolderPath(SpecialFolder.ProgramFiles)` is empty on Unix, the `IsNullOrWhiteSpace` guard catches it, and `DataDir` falls to `AppDomain.CurrentDomain.BaseDirectory`.

| | Windows MSI | macOS / Linux zip, conda, docker |
|---|---|---|
| `DataDir` | `%LOCALAPPDATA%\MetaMorpheus` | the extracted app folder |
| agreement persists | per user, survives upgrades | **per extracted copy** |
| GUI dialog available | yes | no — WPF |

A Windows user agrees once and never thinks about it again. A macOS user agrees again for every new download, every conda env, and every container layer — and since a container has no tty, every one of those is the crash you're fixing. Your PR turns that into a clear message, which is the right first move. But the message it prints will tell them to hand-edit a toml, and there is no supported alternative: the full option list is `-t -d -s -o -g -v --test --mmsettings`, no flag and no environment variable touches the licence.

So I'm going to follow this with a small PR adding an explicit accept-the-licence flag — print the text, record the agreement, exit — so scripted and non-Windows users have a first-class route rather than a documented workaround. It stacks naturally on the message yours adds. Happy to fold it into this PR instead if you'd rather it land as one change.

trishorts:
🤖 Opened it: #2725.

`--acceptThermoLicence` prints the licence, records the agreement and does not prompt — standalone as a setup step, or alongside a run. It is the thing your new message can point people at, rather than telling them to hand-edit a toml.

It does not touch `Program.cs:156-177`, so it should not conflict with this PR in either direction. The cost of staying out of those lines is that the six that write `settings.toml` now appear twice, once in your prompt and once in a `RecordThermoLicenceAgreement()` helper. Happy to collapse both onto the helper as a follow-up after yours merges, or to rebase mine onto your branch instead if you would rather it land as one change — your call, since yours is the one in flight.

Verified with stdin closed throughout: your crash reproduces (exit 139), the flag records the agreement (exit 0), and then the *identical* command that crashed completes a search. Table is in the PR body.

trishorts:
🤖 Correcting myself on the one thing I got wrong above, before you act on it.

I wrote *"I agree with keeping 3 and 5 apart — a job that could not be asked is not a job that said no."* The principle is right and I still hold it. The number is not: **5 is already taken**, three times, in the same method, for a completely unrelated failure.

### Exit code 5 is not new

`Run` already returns `5` at three points, all in the experimental-design / normalization path. On this branch (`d703bc1`):

| line | condition |
|---|---|
| `CMD/Program.cs:277` | `ExperimentalDesign.tsv` missing while a search task has `Normalize = true` |
| `CMD/Program.cs:295` | errors reading `ExperimentalDesign.tsv` while `Normalize = true` |
| `CMD/Program.cs:313` | user answered anything but `y`/`yes` to "continue with an empty experimental design?" |

They are bare `return 5;` statements rather than assignments to `errorCode`, which is why they are easy to miss — I missed them, and so did the PR body. `Run`'s return value is `Main`'s (`Program.cs:36-39`), so all three are real process exit codes today.

The full set the process can return is therefore `{0, 1, 2, 3, 4, 5}` — `0` success/`--help`, `1` parse error (`:70`), `2` settings validation (`:108`), `3` licence declined (`:184` on this branch), `4` run threw (`:351`), `5` experimental design. There is no enum, no const, and nothing in the README or the wiki that lists them; the only in-repo documentation is the `>0 means failure` comment at `Program.cs:27-29`, and the three CI checks in `InstallRunAndArtifact.yml` only test `-ne 0`. So nothing in-repo breaks — but a user script that already distinguishes these does, and `5` is reachable in production right now.

The practical consequence is exactly the thing this PR set out to prevent: a wrapper that sees `5` and retries with `echo y |` will do that to a run whose real problem is a missing `ExperimentalDesign.tsv`, and a wrapper that sees `5` and prints "fix your experimental design" will say that to someone who just needed to agree to a licence. Two failures, opposite remedies, one code.

**Suggested change: use `6`.** It is the next free integer, it preserves the 3-vs-not-3 distinction you and I both want, and it is a one-character edit plus a line in the PR body. Renumbering the three experimental-design returns instead would be the cleaner world but it touches unrelated paths and is a different PR.

If you want to go further at some point, the codes are worth an `enum ExitCode` — six values with no single place naming them is how this happened in the first place. Not asking for that here.

### While I'm here: a fourth instance of the same bug

Adding to the "same bug wearing different clothes" list from my comment above, because this one is closer to home than the GUI dialog and I missed it the first time too.

`CMD/Program.cs:305`, about 140 lines below the line you're fixing:

```csharp
var result = Console.ReadLine();

if (result.ToLowerInvariant() == "y" || result.ToLowerInvariant() == "yes")
```

Identical `NullReferenceException`, identical trigger — closed or exhausted stdin — in the same file, and reachable in the same container and CI runs. These two are the only `Console.ReadLine()` call sites in the entire repository, and after this PR one of them is guarded and the other is not.

I'd rather not expand your scope; a null check on a prompt you didn't write isn't your job. But given there is no read-a-console-answer helper anywhere in the repo (I looked — no `ReadKey`, no `IsInputRedirected`, no `UserInteractive`, no confirm utility outside WPF), this PR is establishing the convention rather than following one. A small `TryReadConsoleAnswer(...)` in `CMD` used by both sites would make the convention explicit and close both crashes at once. Your call entirely — the alternative is that it goes in the issue I owe you for the GUI dialog and the read-only-data-dir null, and I'll write that up either way.

Everything else in my review above stands. `?.Trim()` is right, the three-way branch is right, the licence invariant is preserved, and the four-path verification table is more convincing than the test would have been.

MetaMorpheus/CMD/Program.cs
```diff
@@ -157,7 +157,16 @@ private static int Run(CommandLineSettings settings)
...
                 Console.WriteLine("\nIn order to search Thermo .raw files, you must agree to the above terms. Do you agree to the above terms? y/n\n");
-                string res = Console.ReadLine().ToLowerInvariant();
+                // null when there is nothing to read: redirected or closed stdin, as in a CI or batch run
+                string res = Console.ReadLine()?.Trim().ToLowerInvariant();
+                if (res == null)
+                {
+                    Console.WriteLine("No response could be read, because MetaMorpheus is not running interactively. Searching Thermo .raw files requires agreeing to the terms above.");
+                    Console.WriteLine("To agree without a console, pipe a response (echo y | metamorpheus ...) or set UserHasAgreedToThermoRawFileReaderLicence to true in " + Path.Combine(GlobalVariables.DataDir, @"settings.toml") + ".");
+                    Console.WriteLine("You can still search .mzML and .mgf files without agreeing to the Thermo licence.");
+                    errorCode = 5;
```
trishorts: **Exit code 5 is already taken — three times.**

The PR body says this "returns a new exit code **5**" that "keeps it distinct from the existing **3**". It is distinct from 3, but it collides with the experimental-design failures. At this head, `CMD/Program.cs`:

```
167:  errorCode = 5;   <- this PR
277:  return 5;        \
295:  return 5;         > experimental design missing / unreadable
313:  return 5;        /
```

The three pre-existing sites are bare `return 5;` rather than `errorCode = 5;`, which is exactly why they are easy to miss — a grep for `errorCode\s*=` finds five of the six meanings and silently skips 5 entirely. Worth grepping `return\s+\d+\s*;` too.

This matters because it defeats the PR's own argument. The case for a distinct code is that "a batch job that could not be asked is not the same as a user saying no, and a script should be able to tell them apart" — but a script switching on 5 now cannot tell "no console to answer the Thermo prompt" from "experimental design file was missing", which are unrelated failures with unrelated fixes.

The next genuinely free integer is **6**.

trishorts: **The same crash is still live 144 lines below this one.**

`Console.ReadLine()` has exactly two call sites in the whole repository, both in this file, and both were null-unsafe. This PR fixes the first:

```
161:  string res = Console.ReadLine()?.Trim().ToLowerInvariant();   <- fixed here
305:  var result = Console.ReadLine();
306:  if (result.ToLowerInvariant() == "y" || ...)                  <- still NREs on null stdin
```

Line 305 is the experimental-design prompt, and it fails the same way for the same reason: redirected or closed stdin returns null, and it is dereferenced unguarded. The CI job, container, or `nohup` run that this PR rescues from one NRE can still hit the other.

I am not asking you to widen the PR if you would rather not — the title is specifically about the Thermo prompt. But it is one `?.` plus a branch, and leaving the pair half-fixed means the next person has to rediscover it. If you would prefer to keep this PR narrow, say so and I will file it as an issue instead.

trishorts:
Reviewed at `d703bc1`. The diagnosis is right and the fix is the correct shape — `Console.ReadLine()` returning null at end of input is a real crash on any non-interactive first run, and turning "could not be asked" into its own outcome rather than a refusal is the right call.

Two things before it merges, one of them blocking.

**Blocking — exit code 5 is already in use, three times over.** The body presents 5 as new and distinct; it collides with the experimental-design/normalization failures at `Program.cs:277`, `:295` and `:313`. Those are bare `return 5;` rather than `errorCode = 5;`, which is why they do not turn up in the obvious grep. Next free integer is 6. Thread opened with the detail.

**Non-blocking — the sibling call site is still unguarded.** `Console.ReadLine()` has exactly two call sites in the repo, both in this file; `:305` is the experimental-design prompt and still dereferences a possible null. Thread opened; happy for it to become a separate issue if you would rather keep this PR narrow.

Worth knowing either way: nothing exercises this path. There are no `.raw` references anywhere under `.github/`, so CI never reaches the licence branch at all, and no test touches `UserHasAgreedToThermoRawFileReaderLicence`. That is not a reason to block a ten-line fix, but "CMD cannot be tested" would not be a good reason to leave it uncovered either — `Test.csproj` already references `CMD`, `Main` is public, and `InstallRunAndArtifact.yml` already runs the CLI as a subprocess asserting `$LASTEXITCODE` in three places. Three `.raw` fixtures exist in the repo. One extra CI invocation would cover both the licence prompt and whichever exit code you settle on.

trishorts (commented):
Explaining instead of crashing is right, and `Console.ReadLine()` returning null on redirected stdin is exactly the failure. Two things on the way it exits, and one that the same fix arguably belongs to.

**The identical crash is still live 140 lines down.** `Program.cs:305` does `var result = Console.ReadLine();` then `result.ToLowerInvariant()` at :307 — the experimental-design prompt, reached under the same non-interactive conditions and throwing the same `NullReferenceException`. It is outside this diff so I can't mark it inline, but it is the same bug in the same method, and a reader of this PR title would reasonably expect it covered. Worth folding in, or worth an issue so it isn't found the same way twice.

trishorts:
Housekeeping: I posted a review a few minutes ago that restated both points from my 25 Aug review -- the exit-code-5 collision and the second null-unsafe `Console.ReadLine()`. I have deleted the duplicate inline comment. The original threads stand and nothing new is being asked for; sorry for the noise.

MetaMorpheus/CMD/Program.cs
trishorts: `5` already means *experimental design failure* in this same file — `return 5;` at `:277` (design missing while normalizing), `:295` (design read with errors) and `:313` (user declined an empty design).

This makes `5` ambiguous for exactly the caller the PR is written for: a script that wants to tell "could not be asked" from "declined" now cannot tell "could not be asked" from "your experimental design is broken".

**Next free code is 6.** These three are bare `return 5;` statements, so they do not show up in a grep for `errorCode = ` — which is why this is an easy one to miss.

trishorts (commented):
Good catch, and the diagnosis is right — this genuinely is unreachable once `settings.toml` records the agreement, which is why it survived. Two things before it goes in, one of which I'd call a blocker.

## Exit code 5 is already taken, three times over

`5` is not a free code. On this branch it already means *experimental design failure* at `CMD/Program.cs:277`, `:295` and `:313`:

- `:277` — the design file is missing and a search task wants normalization
- `:295` — the design file was read but returned errors
- `:313` — the user declined to continue with an empty design

So after this change a batch job that exits 5 cannot tell "nobody could answer the Thermo licence prompt" from "your experimental design is broken" — which is precisely the distinction the PR description argues for when it separates 5 from 3. **The next free code is 6.**

Worth knowing why this is easy to miss: those three are bare `return 5;` statements, so they are invisible to a grep for `errorCode = `. That is how I nearly missed it too.

## The same crash is still live 140 lines below

`Console.ReadLine()` is dereferenced unguarded a second time in this file, at `CMD/Program.cs:305` on this branch:

```csharp
var result = Console.ReadLine();

if (result.ToLowerInvariant() == "y" || result.ToLowerInvariant() == "yes")
```

Same `NullReferenceException`, same cause, and reachable in the same scenario this PR exists to fix: a non-interactive run whose experimental design has errors, at `minimal` or `normal` verbosity. A CI job that gets past the Thermo prompt can still die here.

It is in scope by the PR's own framing — the title is *"instead of crashing when there is no console"*, and this is the other place there is no console. Leaving it means the next person to hit it gets an identical stack trace from a line the fix was supposed to have covered.

I have both changes ready in a branch if it is easier to take than to write — see the follow-up comment.

Everything else here I agree with: `?.Trim()` handling the stray `
` from a piped answer is a real improvement, the message naming the resolved `settings.toml` path is the right thing to print, and treating "could not be asked" as distinct from "declined" is the correct model.

trishorts:
@acesnik — the two fixes from my review are on a branch against yours, so you can merge them instead of retyping: **acesnik/MetaMorpheus#9**.

Exit code `5` → `6`, and the second unguarded `Console.ReadLine()` at `Program.cs:305`. `dotnet build CMD/CMD.csproj -c Release` is clean. Merge it, cherry-pick it, or ignore it and write your own — no attachment to the patch, only to the two findings.

trishorts:
@acesnik — **acesnik/MetaMorpheus#15** takes `codecov/patch` here from **0.00%** to covering every line this PR adds. Four tests, no production change.

They run the built CLI as a **subprocess** with stdin closed rather than calling `Program.Main` in-process. That is deliberate and I found out the hard way: the in-process version passed 4/4 alone and then failed **eleven** unrelated parsimony tests later in the same suite, because `Program.Run` subscribes handlers to `MetaMorpheusEngine`'s static events and never unsubscribes. I filed that separately as #2788 and assigned it to you — it is a master defect, not this PR's, and it is the same hazard your Avalonia view model already guards against by being `IDisposable`.

A subprocess sidesteps it and is the more honest test anyway: the process really has no stdin, and the assertion is on the exit code a script would see. Coverage is unaffected by the choice — coverlet instruments the assemblies on disk, so the child process records hits; I checked line by line that the `errorCode = 6` block and the guarded second `ReadLine` are covered.

Verified: mutation-checked (reverting the two production changes fails two of the four), and the full `Category!=ExternalService` core suite is **1689 passed, 0 failed**.

trishorts:
@acesnik — **acesnik/MetaMorpheus#17** closes the rest of the `codecov/patch` gap. After #15 merged it went **0.00% → 80.00%** (and `CMD/Program.cs` 14.87% → 33.2%, which confirms in CI that the subprocess tests do record coverage — coverlet instruments the assemblies on disk).

The remaining 20% was two lines:

1. **The y-branch of the experimental-design prompt was never taken.** Codecov had it as *partial*, not missed — the condition was only ever evaluated on the paths that decline, so `File.Delete` sat at 0 hits. That is the branch you least want unexercised, since it deletes a user's file. Three cases now take it (`y`, `yes`, padded `  YES  `), asserting the file really is deleted and that the run does not exit 5. `File.Delete` → 3 hits.

2. **A comment line reported as an uncovered diff line**, while the coverage report contains no entry for either comment line. Folded to one line. Calling that out plainly because it is the one change made for the metric rather than for the code — if you would rather keep the two-line comment and carry the 1-line miss, drop that hunk; the branch coverage is the part that matters.

Full `Category!=ExternalService` core suite: **1692 passed, 0 failed**.

Separately, and not urgent: your ask on #2733 from 16 Aug — eyeballing the statistics plots on Windows — is still open, and it is now the gap that matters most there. The 588 MetaDraw tests prove the OxyPlot 2.2.0 rewrite does not *throw*, but nothing checks that 16 plots still *look* right, and that PR re-renders all of them through a hand-written StackedColumnSeries.

trishorts:
@acesnik — this went `CONFLICTING` when #2775 merged. **acesnik/MetaMorpheus#18** resolves it, and the resolution is mostly deletion: **#2775 solved half of this PR, better than this PR did.**

It moved the experimental-design handling into `Program.ResolveExperimentalDesign` with an injectable `readLine`, which fixes the same `NullReferenceException` this branch fixed and goes further — it no longer deletes a design that will not parse (citing #2256, a user who lost theirs), and `ResolveExperimentalDesignTests` already covers `y` / `YES` / `n` / `""` / `null` in-process.

So master's version is taken wholesale, and **the two experimental-design tests I added earlier are deleted with it** — they asserted the file gets deleted on "y", which is exactly the behaviour #2775 removed. They would have been wrong against master.

What still stands: the Thermo licence prompt on master is still `Console.ReadLine().ToLowerInvariant()` at `CMD/Program.cs:171`, unguarded — the original crash — and 5 is still taken three times, so exit code 6 remains right. The three licence tests are unchanged.

18 tests pass across this fixture and `ResolveExperimentalDesignTests`; full core suite **1906 passed, 0 failed**.

PR #2727, opened 2026-08-15
Cite UniDec in the manuscript prose when IsoDec deconvolution was used

acesnik:
🤖 UniDec's licence requires that software redistributing IsoDec **clearly indicate to the end user that UniDec is in use**, and pass on the request to cite Marty et al. Anal. Chem. 2015:

> If UniDec is redistributed or incorporated into other software, it must be clearly indicated to the end user that UniDec is being used, and the request to cite Marty et al. Anal. Chem. 2015. DOI: 10.1021/acs.analchem.5b00140 must be passed on to the end user.

MetaMorpheus ships `isodeclib` through mzLib and offers IsoDec deconvolution, and said so nowhere a user would see. This writes the citation into `AutoGeneratedManuscriptProse.txt` when a task actually ran with IsoDec.

That file is the natural home: it already collects the citations a user is asked to make, and it is the thing they paste into a methods section. smith-chem-wisc/mzLib#1161 records the same dependency on the library side.

## The citation follows what ran, not what was configured

`FileSpecificParameters` holds each file's settings already merged over `CommonParameters`, so it is authoritative when populated. A file-specific TOML can switch IsoDec on for one file — and can equally switch it **off** for every file, in which case the task-level setting never applied and citing IsoDec would be a false claim in someone's methods section.

**My first version got this wrong**, and the existing `TestFileSpecificDeconvolutionParams` is what exposed it: one of its four runs sets IsoDec on the task and overrides every file back to Classic. The original check was `IsIsoDec(CommonParameters) || any file`, which would have cited UniDec for a run that used Classic throughout. It now prefers the merged per-file list, falling back to `CommonParameters` only when no per-file entries exist.

## Tests

The prose assertions went into that existing test, because it already runs all four combinations that matter:

| run | task setting | file TOML | expected |
|---|---|---|---|
| `IsoDecDirect` | IsoDec | none | cites UniDec |
| `IsoDecFromToml` | Classic | IsoDec | cites UniDec |
| `ClassicDirect` | Classic | none | no citation |
| `ClassicFromToml` | IsoDec | Classic | **no citation** |

That last row is the one that caught the bug, and it is the reason to assert on absence as well as presence.

`TestUniDecCitationDecision` adds an eleven-case matrix driven through the helper by reflection, following how `CalibrationTests` exercises private members. Each case is chosen to fail if one specific part of the rule is removed, rather than to raise a number:

- the `DoPrecursorDeconvolution` gate, in both directions
- the product half of the check, which has no equivalent flag
- per-file precedence when the list covers every file
- the fallback when the list is short because a TOML failed to parse
- descent into `MultipleDeconParameters`, and that it does not over-trigger
- zero spectra files, where nothing ran and nothing is owed

Against the pre-review logic exactly four of them fail, one per defect found in review; deleting the product clause fails three more. `TestUnparseableFileSpecificTomlLeavesFileSpecificParametersShort` separately proves the short-list state is reachable, using Classic so it does not need the IsoDec native library.

End to end, a real search over `SmallCalibratible_Yeast.mzML` with classic deconvolution writes prose containing no mention of UniDec or Marty, with the rest of the prose unchanged.

**The four-run assertions need Windows CI to execute.** IsoDec's native library has no macOS build — `isodeclib.dll` is Windows x86-64 only — so the positive cases cannot run on the machine I verified the rest on. That gap is what smith-chem-wisc/mzLib#1158 tracks, and I have a portable-CMake change for UniDec that builds `isodeclib.dylib` natively on arm64, which would eventually let these run everywhere.


MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -732,6 +740,30 @@ public MyTaskResults RunTask(string output_folder, List<DbForTask> currentProtei
...
 
+        /// <summary>
+        /// Whether this run deconvoluted with IsoDec, for either precursors or products, at the task level
+        /// or for any individual file. Drives the UniDec citation in the manuscript prose, which its licence
+        /// requires be passed on to end users rather than left implicit.
+        /// </summary>
+        private bool UsedIsoDecDeconvolution()
+        {
+            static bool IsIsoDec(CommonParameters parameters) =>
+                parameters?.PrecursorDeconvolutionParameters?.DeconvolutionType == DeconvolutionType.IsoDecDeconvolution
```
trishorts: **This can cite UniDec for a run that deconvoluted with Classic.**

The precursor clause decides on `PrecursorDeconvolutionParameters.DeconvolutionType` alone, but whether those parameters are ever applied is gated by `CommonParameters.DoPrecursorDeconvolution` — the guard is at `MetaMorpheusTask.cs:272`, and the contract comment you kept just above the merge says so outright (`:573-575`):

> DoPrecursorDeconvolution and DoProductDeconvolution flow from CommonParameters only; file-specific PrecursorDeconvolutionParameters / ProductDeconvolutionParameters are stored independently and **take effect when the corresponding Do\* flag is true**.

So with `DoPrecursorDeconvolution == false`, IsoDec precursor parameters are inert, yet this returns `true` whenever the product parameters are Classic. The prose then states "Isotopic envelope deconvolution was performed using IsoDec" for a run in which IsoDec never executed — the same false methods-section claim `ClassicFromToml` was added to prevent, arriving through the opposite door.

There is a second route that is harder to see: `SearchTask.cs:244-248` force-disables precursor deconvolution for `Mgf`/`Ms2Align` inputs —

```csharp
if (myMsDataFile is Mgf or Ms2Align)
{
    combinedParams.DoPrecursorDeconvolution = false;
    combinedParams.UseProvidedPrecursorInfo = true;
}
```

— and it does that on a *transient* `combinedParams`, which never reaches the `FileSpecificParameters` list this helper reads. So even reading the flag off the merged per-file entries would not catch the mgf/msalign case.

Gating the precursor clause on the merged `DoPrecursorDeconvolution` fixes the reachable-by-setting case. The mgf/msalign case is genuinely unobservable from configuration; if you want it airtight, the citation has to be recorded where deconvolution is dispatched rather than inferred from parameters. Happy with the narrower fix plus a comment naming the limitation.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -732,6 +740,30 @@ public MyTaskResults RunTask(string output_folder, List<DbForTask> currentProtei
...
         }
 
+        /// <summary>
+        /// Whether this run deconvoluted with IsoDec, for either precursors or products, at the task level
+        /// or for any individual file. Drives the UniDec citation in the manuscript prose, which its licence
+        /// requires be passed on to end users rather than left implicit.
+        /// </summary>
+        private bool UsedIsoDecDeconvolution()
+        {
+            static bool IsIsoDec(CommonParameters parameters) =>
```
trishorts, line 785: **`DeconvolutionType.Multiple` slips through this check.**

`MultipleDeconParameters` pins its own type rather than exposing the ones it wraps:

```csharp
public override DeconvolutionType DeconvolutionType { get; protected set; } = DeconvolutionType.Multiple;
public DeconvolutionParameters[] Parameters { get; protected set; }
```

(mzLib `MassSpectrometry/Deconvolution/Parameters/MultipleDeconParameters.cs:10,12`)

So an equality test against `IsoDecDeconvolution` returns `false` for a parameter set that contains IsoDec, and MetaMorpheus does offer exactly that combination — `MultipleDeconParamsViewModel.cs:19-23`:

```csharp
public ObservableCollection<DeconvolutionType> AvailableSubTypes { get; } = new()
{
    DeconvolutionType.ClassicDeconvolution,
    DeconvolutionType.IsoDecDeconvolution
};
```

This is the under-citing direction, so it is the one that misses the licence obligation rather than over-claiming.

**Reachability caveat, so you can weigh this properly:** the toml converter at `MetaMorpheusTask.cs:98-102` accepts only `"ClassicDeconvolution"` and `"IsoDecDeconvolution"` and throws on anything else, so `Multiple` cannot arrive through a task config or a file-specific toml — it is in-memory only (GUI or programmatic). I did not confirm end-to-end that a user can actually commit a `Multiple` selection through the GUI, so this may be latent today. If it is, a recursive descent through `MultipleDeconParameters.Parameters` is cheap insurance against it becoming live later; if you would rather not, it is worth a comment saying `Multiple` is knowingly unhandled.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -732,6 +740,30 @@ public MyTaskResults RunTask(string output_folder, List<DbForTask> currentProtei
...
+                parameters?.PrecursorDeconvolutionParameters?.DeconvolutionType == DeconvolutionType.IsoDecDeconvolution
+                || parameters?.ProductDeconvolutionParameters?.DeconvolutionType == DeconvolutionType.IsoDecDeconvolution;
+
+            // FileSpecificParameters holds each file's settings already merged over CommonParameters, so it
+            // is authoritative when populated: a file-specific toml can turn IsoDec on for one file, and it
+            // can equally turn it off for every file, in which case the task-level value was never used and
+            // citing IsoDec would be wrong. CommonParameters is the fallback only when no per-file entries
+            // were built.
+            if (FileSpecificParameters is { Count: > 0 })
```
trishorts: **The "authoritative when populated" invariant has a hole: a toml that fails to parse.**

The comment above states the list "holds each file's settings already merged over CommonParameters", and for the normal paths that is exactly right — `:652` for files with a toml, `:667` for files without. But both catch blocks warn and fall through *without* adding an entry:

```csharp
catch (MetaMorpheusException e)   // :654
{
    Warn("Problem parsing the file-specific toml " + ...);
}
catch (KeyNotFoundException e)    // :660
{
    Warn("Problem parsing the file-specific toml " + ...);
}
```

Meanwhile `fileSettingsList[i]` stays null, so that file genuinely runs on plain `CommonParameters`.

Two files, task set to IsoDec, one toml malformed and the other setting Classic: the list is non-empty but holds a single Classic entry, `Any(...)` is `false`, and the task-level IsoDec that actually governed the malformed file is never consulted. No citation, and the user sees only an unrelated toml-parsing warning. Under-citing again, so it fails the licence obligation rather than merely over-claiming.

(`GlobalVariables.StopLoops` at `:642` breaks the loop the same way, though there the run is aborting anyway.)

Cheapest fix inside this PR is to OR in `IsIsoDec(CommonParameters)` when the entry count is short of `currentRawDataFilepathList.Count`. Adding the `CommonParameters` entry in the catch blocks is arguably the more correct fix since those files really do run on it — but that changes behaviour for every other consumer of the list, so it probably belongs in its own PR.

MetaMorpheus/Test/TestToml.cs
```diff
@@ -407,6 +407,32 @@ public static void TestFileSpecificDeconvolutionParams()
...
+                //
+                // UniDec's licence requires software redistributing IsoDec to tell the end user it is in
+                // use and pass on the citation request, so this is a licence obligation rather than a
+                // nicety. The four runs above are exactly the cases that matter: the citation must follow
+                // what actually ran, not what the task was configured with, and two of these four have a
+                // file-specific toml that disagrees with the task.
+                static string Prose(string dir, string run) =>
+                    File.ReadAllText(Path.Combine(dir, run, "AutoGeneratedManuscriptProse.txt"));
+
+                Assert.Multiple(() =>
```
trishorts, line 421: **Two behaviours the PR names as requirements aren't covered by anything committed here.**

The description lists product-deconvolution IsoDec and a null `FileSpecificParameters` among the six cases checked "directly against the private helper", but all four committed runs set `precursorDeconParams` and `productDeconParams` to the *same* type together — `TestToml.cs:346-347`, `:360-361`, `:381-382`, `:394-395`. So nothing distinguishes the precursor clause from the product clause of `IsIsoDec`, and the null path is untested by construction, since `RunTask` always assigns a list at `:631` before the helper runs.

Concretely: deleting either half of the `||`, or the `?.` guards, leaves this whole suite green. Product-only IsoDec is a common top-down configuration, so that half going missing in a later refactor would be quiet.

Given the helper is private, the low-friction option is a fifth run whose file-specific toml sets IsoDec on products and Classic on precursors. Making it `internal` with `InternalsVisibleTo` would let you assert the six cases directly and cheaply, including the null ones — your call which is worth the surface area.

Not asking for the non-Windows gap to be addressed here; mzLib#1158 clearly owns that.

trishorts (commented):
Reviewed the detection logic against `master @ 258d74848`. The precedence rule is right, and the `ClassicFromToml` case is a genuinely good catch — I confirmed `FileSpecificParameters` really is populated for *every* spectra file (`MetaMorpheusTask.cs:667` adds `CommonParameters` when no toml exists), so preferring the per-file list is sound.

Four comments inline. One is a merge blocker; the rest are your call.

The blocker is that `IsIsoDec` reads the precursor parameters without consulting `DoPrecursorDeconvolution`, which is exactly the false-methods-claim failure the PR sets out to prevent, just in the over-citing direction.

Two smaller things not worth their own threads:

**Negative assertions only check the DOI.** `ClassicDirect` and `ClassicFromToml` assert `Does.Not.Contain("10.1021/acs.analchem.5b00140")`, and `Does.Contain("UniDec")` is applied to `IsoDecDirect` alone. The PR body says the classic prose contains "no mention of UniDec or Marty", but neither substring is asserted absent anywhere. A regression that emitted the sentence unconditionally while reformatting the DOI would keep both negative cases green. Asserting absence of `"UniDec"` in the two classic runs would close that, and it is the name the licence actually requires be indicated.

**The sentence is unconditional but the condition is `Any(...)`.** On a mixed run — one file's toml on IsoDec, the rest Classic — the prose claims "Isotopic envelope deconvolution was performed using IsoDec" for the whole analysis. The licence obligation is still discharged, so this is accuracy rather than compliance; worth either softening the wording or noting the deliberate choice in the comment.

For context, not a request: mzLib has no citation API to consume today — `DeconvolutionParameters` exposes no `Citation`/`Name` member — and smith-chem-wisc/mzLib#1161 is docs-only and explicitly delegates the runtime obligation here, so hard-coding the string is currently the only option. If the two copies of the DOI drift, a `Citation` member on `DeconvolutionParameters` would make mzLib the single source; #1161's Chronologer entry raises the same need, so that seam has two customers already. Separate PR, clearly.

Also note #2715 touches this same file about 80 lines away. No textual conflict, but the two are related enough to be worth landing in a known order.

acesnik:
🤖 All four inline comments addressed in `bac4aa04`, plus a follow-up in `48f69889`. Three were real defects; details in the threads. Two notes on the body:

**Negative assertions** — done. Both classic runs now also assert `Does.Not.Contain("UniDec")`. You're right that the DOI alone would survive a reformatting regression, and UniDec is the name the licence requires be indicated.

**Unconditional sentence vs. `Any(...)`** — reworded rather than commented around. It now reads *"IsoDec, part of UniDec (…), was used for isotopic envelope deconvolution in this analysis."* That stays true when only some files used IsoDec, and the deliberate choice is noted at the call site. I kept it file-agnostic rather than enumerating which files, since the licence obligation is to indicate UniDec is in use and pass on the citation, not to attribute per file.

**`Citation` member on `DeconvolutionParameters`** — agreed, and worth its own PR. With #1161's Chronologer entry that seam has two customers, and it would remove the second copy of the DOI. Not doing it here.

**#2715 ordering** — happy to land after it, since it is the smaller of the two and this one now has a larger test surface to rebase. Say if you'd rather the reverse.

I also **edited the PR description**: the Verification section claimed six detection cases were checked directly against the private helper, and no such tests were committed. It now describes what is actually there.

One correction to the framing, which the stale comment at `:573` caused: **there is no `DoProductDeconvolution`.** More in the blocker thread.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
acesnik: 🤖 Fixed, and you're right that it was the same false-methods-claim failure arriving through the opposite door. The precursor clause is now gated:

```csharp
&& ((parameters.DoPrecursorDeconvolution && ContainsIsoDec(parameters.PrecursorDeconvolutionParameters))
    || ContainsIsoDec(parameters.ProductDeconvolutionParameters));
```

**One correction: there is no `DoProductDeconvolution`.** `CommonParameters` declares only `DoPrecursorDeconvolution` (`CommonParameters.cs:157`), and `Ms2ScanWithSpecificMass.GetNeutralExperimentalFragments` (`:96`) deconvolutes products unconditionally with no flag. So the product clause stays ungated deliberately — gating it would suppress citations for deconvolution that does happen.

The contract comment you quoted is the culprit: it asserts a symmetric pair of flags that has never existed. Corrected in `48f69889` so the next reader doesn't make the same inference.

I took the narrower fix plus a comment on the mgf/msalign limitation, as offered. Recorded at the call site that a transient `combinedParams` never reaches these lists, so that route can still over-cite and closing it needs the citation recorded where deconvolution is dispatched.

Covered by `TestUniDecCitationDecision("TaskPrecursorIsoDecButDeconvolutionOff", …)`, which fails against the pre-review logic.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
acesnik: 🤖 Implemented the recursive descent — and it is **not latent**, so this is better than cheap insurance.

`DeconHostViewModel` builds its selectable lists by iterating `Enum.GetValues<DeconvolutionType>()` and skipping only `ExampleNewDeconvolutionTemplate` and `FromFile` (`DeconHostViewModel.cs:47-51`). `Multiple` therefore gets a view model, because `GetDefaultDeconParams` returns a real `MultipleDeconParameters` for it (`MzLibExtensions.cs:67-69`, defaulting to a Classic inner set). Combined with the `AvailableSubTypes` you found, a user can select `Multiple` and add IsoDec to it today.

```csharp
static bool ContainsIsoDec(DeconvolutionParameters parameters) => parameters switch
{
    null => false,
    MultipleDeconParameters multiple => multiple.Parameters.Any(ContainsIsoDec),
    _ => parameters.DeconvolutionType == DeconvolutionType.IsoDecDeconvolution
};
```

Your reachability caveat about the toml converter is right and I kept it in the comment, since it explains why this cannot be tested through a task config.

Covered both ways: `TaskMultipleContainingIsoDec` (true) and `TaskMultipleAllClassic` (false), so the descent cannot over-trigger either.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
acesnik: 🤖 Confirmed and fixed with the cheap version you suggested:

```csharp
return perFileParameters.Count < spectraFileCount && IsIsoDec(CommonParameters);
```

I had independently hit this from the other direction — it is exactly the branch codecov flagged as partial, and it was uncovered because the rule it encoded was wrong. A test written to reach the old branch would have locked in the under-citing bug.

Agreed that adding the `CommonParameters` entry in the catch blocks is the more correct fix and belongs in its own PR; it would change what every other consumer of the list sees. Your `StopLoops` note is in the comment too, since it produces the same short list.

Two tests, deliberately split:

- `TestUniDecCitationDecision("ShortListBecauseATomlFailedToParse", 2, true)` — the logic. Fails against the pre-review version.
- `TestUnparseableFileSpecificTomlLeavesFileSpecificParametersShort` — that the state is *reachable*: two files with one malformed toml, asserting the parse warning fires and `FileSpecificParameters` has one entry. Uses Classic throughout so it does not need the Windows-only `isodeclib`; my first attempt ran IsoDec end to end and could not execute off Windows.

MetaMorpheus/Test/TestToml.cs
acesnik: 🤖 You're right, and the PR description was wrong: it claimed six cases "checked directly against the private helper" when nothing of the kind was committed. I have corrected the description.

Went with reflection rather than `InternalsVisibleTo` — `CalibrationTests` already drives private members that way (`GenerateIndexes`), so it adds no production surface. `TestUniDecCitationDecision` is an eleven-case matrix, each case chosen to fail if one specific part of the rule is removed rather than to raise a number.

Measured, not asserted. Against the **pre-review** logic, exactly four fail:

```
TaskPrecursorIsoDecButDeconvolutionOff   the Do flag gate
ShortListBecauseATomlFailedToParse       the short-list fallback
TaskMultipleContainingIsoDec             the Multiple descent
NoSpectraFiles                           citing when nothing ran
```

That last one was not in your list: with zero spectra files and task-level IsoDec, the old code fell through and cited.

On your specific point that deleting either half of the `||` leaves the suite green — deleting the **product** clause now fails three cases (`TaskProductIsoDecOnly`, `TaskProductIsoDecOnlyAndDeconvolutionOff`, `TaskMultipleContainingIsoDec`); deleting the precursor clause fails `TaskPrecursorIsoDec`. Both halves are pinned.

The `?.` guards are the one thing I did **not** cover, and I do not think I honestly can: `CommonParameters` defaults both parameter sets to Classic when null (`CommonParameters.cs:109-111`), so a null never arrives through the public constructor. The `null =>` arm survives only to protect the recursion against a null element inside a `MultipleDeconParameters` array, which its constructor does not reject. Happy to add a case for that if you want it pinned, but it would be testing an arrangement no caller can currently produce.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
trishorts: Confirmed fixed. `IsIsoDec` now gates the precursor clause on `DoPrecursorDeconvolution` (`MetaMorpheusTask.cs:774-776`), and `TestUniDecCitationDecision`'s `TaskPrecursorIsoDecButDeconvolutionOff` case pins it.

I checked the asymmetry you assert in the new comment rather than taking it on trust: `DoProductDeconvolution` genuinely does not exist anywhere in the repo, and `Ms2ScanWithSpecificMass.cs:33` computes `GetNeutralExperimentalFragments` unconditionally, which deconvolutes with `ProductDeconvolutionParameters` at line 96. So leaving the product clause ungated is correct, and the comment correction in the third commit is right.

Resolving.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
trishorts: Confirmed fixed. `ContainsIsoDec` descends into `MultipleDeconParameters` via the recursive switch arm (`MetaMorpheusTask.cs:759-764`), and both directions are pinned — `TaskMultipleContainingIsoDec` and `TaskMultipleAllClassic`.

Agreed it was not latent; nesting IsoDec inside a `MultipleDeconParameters` is reachable in memory today. Resolving.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
trishorts: Confirmed fixed, and I traced the population site to be sure the invariant is the one you describe rather than assuming it.

`MetaMorpheusTask.cs:667` adds `(path, CommonParameters)` for a spectra file that has no toml at all, so the list is short **only** when a toml threw (`MetaMorpheusException` at :655 or `KeyNotFoundException` at :661). That is exactly the case the `perFileParameters.Count < spectraFileCount` fallback at `:787` now catches, and `TestUnparseableFileSpecificTomlLeavesFileSpecificParametersShort` proves the short list really happens end-to-end rather than only in the unit matrix.

Resolving.

MetaMorpheus/Test/TestToml.cs
trishorts: Confirmed covered. The eleven-case `TestUniDecCitationDecision` matrix plus `TestUnparseableFileSpecificTomlLeavesFileSpecificParametersShort` pin every branch of the rule, including the two the earlier version had nothing on — the product clause and the short-list fallback. Thanks for correcting the PR description rather than just the code.

One thing I deliberately did **not** flag: the `BindingFlags.NonPublic` reflection into `UsedIsoDecDeconvolution`. I checked it against house convention first, and 27 files under `Test/` already reach private members that way, so this is the established pattern here rather than a deviation. The `Assert.That(method, Is.Not.Null, "...was renamed; this test needs updating")` guard is the right mitigation for the string-name brittleness.

Resolving.

trishorts:
Re-reviewed at `48f6988`. All four of my earlier threads are addressed and I have resolved them — each fix verified against the checkout, not just read in the diff:

- the `DoPrecursorDeconvolution` gate, plus the product-side asymmetry checked independently (`DoProductDeconvolution` really does not exist; `Ms2ScanWithSpecificMass.cs:33/96` deconvolutes products unconditionally)
- the recursive descent into `MultipleDeconParameters`
- the short-list fallback, with the `MetaMorpheusTask.cs:667` population site confirming the list is short only when a toml threw
- test coverage for the two behaviours that previously had nothing pinning them

No new findings. One loose end that is not mine to fix: `codecov/patch` is red at 75% of the diff hit against a 93.32% target — three lines of a twelve-line payload. Worth a glance to confirm it is the unreachable-today `MultipleDeconParameters` arm rather than something you meant to exercise.

trishorts:
Pushed `211464ef` to this branch to close the `codecov/patch` gap — four test cases, no production change.

The uncovered lines were all defensive branches the eleven-case matrix structurally cannot reach, because every scenario there is built from fully populated `CommonParameters`:

| line | branch | before |
|---|---|---|
| `MetaMorpheusTask.cs:761` | `null => false` | 0 hits |
| `MetaMorpheusTask.cs:759` | `ContainsIsoDec` switch | 75% (3/4) |
| `MetaMorpheusTask.cs:774` | `parameters is not null` | 83.33% (5/6) |
| `MetaMorpheusTask.cs:781` | `FileSpecificParameters ?? []` | 50% (1/2) |

`TestUniDecCitationNullGuards` adds four cases, each of which fails if its guard is deleted rather than just painting the line: a null `FileSpecificParameters` list (the state of a task before `RunTask` builds it at :631, so the null-coalesce genuinely has to hold), a per-file entry carrying null parameters, and a `MultipleDeconParameters` holding a null element in **both** orders. That last pair is the one I think earns its place beyond coverage — mzLib's ctor validates only that the wrapped collection is non-null and non-empty, so a null element is constructible today, and `Multiple(null, IsoDec())` proves the descent skips it rather than abandoning the entries after it.

Verified with coverlet rather than assumed. Against your original eleven cases alone the four lines reproduce Codecov's report exactly (761 at 0 hits, 759 at 75%, 774 at 83.33%, 781 at 50%); with these four added all reach 100%, which should take the patch to 16/16. `TestToml` is 36/36 green locally on Windows.

Kept to the file's existing idiom — same `[TestCase]` scenario-switch shape and the same reflection helper — but as a separate `[Test]` rather than more rows on `TestUniDecCitationDecision`, so that method's "each case fails if one piece of the rule is removed" claim stays about the rule and not about null handling. Rip it out or fold it in as you prefer.

trishorts (approved):
Approving.

All four of my earlier threads are fixed and resolved, each verified against the checkout rather than read off the diff — the `DoPrecursorDeconvolution` gate (with the product-side asymmetry checked independently), the descent into `MultipleDeconParameters`, the short-list fallback for an unparseable toml, and test coverage for the two behaviours that previously had nothing pinning them.

`codecov/patch` is now green at 100% (20 hits, 0 misses, 0 partials) against the 93.32% target, and all thirteen checks pass.

Two notes for the record:

- The four test cases in `211464ef` are mine, so my approval covers the production change — which is entirely yours — and not those tests. If you would rather they were reviewed by someone who did not write them, or folded into `TestUniDecCitationDecision`, say so and I will not treat this approval as settling that.
- The branch is behind master. A base-only merge does not dismiss reviews here, so updating it will not cost you this approval.

PR #2729, opened 2026-08-15
Run the WPF-free half of the GUI tests on Linux and macOS

acesnik:
🤖

## Run the WPF-free half of the GUI tests on Linux and macOS

`GuiFunctions` targets `net10.0-windows` because parts of it use WPF. That means the GUI and MetaDraw
tests only run on Windows, even though most of what they cover has nothing to do with WPF. This splits
the UI-free half into its own project and runs its tests on all three platforms.

Offered separately from #2709 and independent of it. @trishorts recommended exactly this as the next
step there: *"the cross-platform `GuiFunctions` extraction, alone: it retains full value even if the
GUI is abandoned, and doing it before Milestone 1 stops the duplicate being hardened."* It stands on
its own merits whatever happens to the Avalonia front end.

### What moves

`GuiFunctions.Core` (`net10.0`) holds the 25 files that reference no UI type at all — the deconvolution
view models, `BaseViewModel`, `GuiGlobalParams`, `MzLibExtensions`, `SpectralMatchComparer`,
`HungarianAlgorithm`, `PlotModelStatParameters` and friends.

The set was found by **compiling candidates against `TaskLayer` alone and iterating to a fixed point**,
not by reading imports. That distinction matters: four files with no `System.Windows` reference anywhere
still fail to compile without WPF, so an import scan would have produced the wrong answer and a broken
build.

Namespaces are unchanged, so this is not a source-breaking change for callers. `GuiGlobalParams` keeps
its internal setters through `InternalsVisibleTo` rather than having them widened to public to cross
the new assembly boundary.

### The tests are linked, not copied

`GuiFunctions.Core.Tests` (`net10.0`) `<Compile Include>`s the existing test files out of `Test/`. The
same source compiles into `Test.csproj` on Windows and into this project everywhere, so there is **one
copy of each test** and no way for the two to drift. The `<Compile Remove>` list is the set that still
needs WPF, found the same fixed-point way.

**187 tests run and pass on macOS arm64** — all of `Test/GuiTests` bar three files, and eight of the
`Test/MetaDraw` files. A new CI job runs them on ubuntu-latest, macos-latest and windows-latest.

### Deliberately no `PlatformTarget`

`GuiFunctions.csproj` sets `x64`; this project sets nothing. An x64 stamp cannot be loaded by an arm64
runtime (#2503, #2504), and this project exists to be loadable anywhere. Verified rather than assumed:
`file` reports `PE32 ... Intel 80386 Mono/.Net assembly`, not `PE32+ ... x86-64`. Left unset rather
than set to `AnyCPU`, which `Microsoft.ML` rejects at build time.

### Two notes for review

- **The `[SetUpFixture]` is in namespace `Test` on purpose.** NUnit applies one to its own namespace
  and below, and the linked tests live in `Test.GuiTests` and `Test.MetaDraw`. I had it in a namespace
  of its own first, and it failed silently — 175 of 187 tests still passed, and only the twelve that
  actually read global state failed, which reads like a porting problem rather than a missing fixture.
  Ten of those twelve looked like genuine macOS incompatibilities (`FileShare.None` is a mandatory lock
  on Windows and advisory on Unix) and I nearly annotated them as Windows-only before finding the real
  cause.
- **`dotnet sln add` maps `UbuntuMac` to `Debug` and drops `Build.0`**, which makes the solution build
  "successfully" while skipping the project. The rows here are set by hand to match the other projects.

### Scope

Additive. `GuiFunctions`, `GUI` and `Test` are otherwise untouched, nothing about the Windows build or
the WiX installer changes, and no test is deleted or rewritten — they are the same tests, now compiled
somewhere they can run.

This is also a prerequisite for the MetaDraw port: the remaining WPF coupling in `GuiFunctions` is
mostly `System.Windows.Media` colour types, and separating the UI-free half first makes that boundary
visible.


MetaMorpheus/MetaMorpheus.sln
```diff
@@ -22,58 +22,200 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Test", "Test\Test.csproj",
...
 Project("{B7DD6F7E-DEF8-4E67-B5B7-07EF123DB6F0}") = "Bootstrapper", "Bootstrapper\Bootstrapper.wixproj", "{E0EA5AC4-24A9-43DC-8FBC-CCEB3B9935B6}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "GuiFunctions.Core", "GuiFunctions.Core\GuiFunctions.Core.csproj", "{7B63C758-BA12-4BDB-879F-DBD8AF82A959}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "GuiFunctions.Core.Tests", "GuiFunctions.Core.Tests\GuiFunctions.Core.Tests.csproj", "{9AB2EDBA-4D49-4AD2-B9B3-5B5410DE05AA}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
```
trishorts: This adds **six new solution platforms** — `x64` and `x86` across Debug, Release and UbuntuMac. Master has exactly three (`Debug|Any CPU`, `Release|Any CPU`, `UbuntuMac|Any CPU`); I checked against `upstream/master`.

The PR body mentions hand-fixing the `UbuntuMac` rows but not this, and it has a real consequence further down: the new `UbuntuMac|x64` and `UbuntuMac|x86` rows give `.Build.0` to GUI, MetaMorpheusSetup and Bootstrapper — the three projects that `UbuntuMac|Any CPU` deliberately withholds it from. Selecting `UbuntuMac|x64` in Visual Studio would try to build the WPF GUI and both WiX projects under a configuration that exists to exclude them.

CI is unaffected (no workflow passes `-p:Platform`, so everything resolves to `Any CPU`), which is exactly what makes it easy to miss. This is the same class of sln churn as smith-chem-wisc/mzLib#1127.

Suggest dropping the `x64`/`x86` solution platforms entirely and keeping only the two new project rows under the three existing `Any CPU` configs.

MetaMorpheus/GuiFunctions.Core/GuiFunctions.Core.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <Nullable>disable</Nullable>
```
trishorts, line 5: Every other project in the solution declares `<Configurations>Debug;Release</Configurations>` — `GuiFunctions.csproj:5`, `GUI.csproj:15`, `TaskLayer.csproj:5`. That declaration is *why* the sln maps their `UbuntuMac|Any CPU` to `Release|Any CPU`: there is no `UbuntuMac` configuration for them to build.

Neither new project declares it, and the sln correspondingly maps both to `UbuntuMac|Any CPU`. So under `dotnet build MetaMorpheus.sln -c UbuntuMac` — which `build.yml` runs — these two land in `bin/UbuntuMac/net10.0/` while every sibling lands in `bin/Release/`.

It builds today, so this is consistency rather than breakage. But `InstallRunAndArtifact.yml`'s `$foldersToCheck` and `BuildReleaseArtifact.yml` both reach into `bin/Release`, so a project that sometimes writes elsewhere is a trap worth closing now: add `<Configurations>Debug;Release</Configurations>` to both csproj files and point the two `UbuntuMac|*` sln rows at `Release|Any CPU` like their siblings.

.github/workflows/Test.yml
```diff
@@ -42,6 +42,37 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup .NET
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: ${{ env.DOTNET_VERSION }}
+
+      - name: Run GuiFunctions.Core tests with coverage
+        run: dotnet test ./MetaMorpheus/GuiFunctions.Core.Tests/GuiFunctions.Core.Tests.csproj --configuration Release --verbosity normal --collect:"XPlat Code Coverage"
```
trishorts: This omits `--filter "Category!=ExternalService"`, which the Windows job a few blocks up carries with a comment explaining why, and which the repo applies at every other run site.

Harmless today — I checked, and the only file under the linked directories carrying that category is `Test/GuiTests/GuiFunctionsTest.cs`, which is already in your `<Compile Remove>` list. So the filter would currently be a no-op.

The reason to add it anyway is that the link glob is recursive and opt-out. A future test added to `Test/GuiTests/` that is external-service-flavoured but *not* WPF-bound gets adopted automatically, and lands in a three-OS job that is blocking — turning a UniProt outage into red CI on someone else's PR. That is precisely the failure the non-blocking `external-service-tests` job exists to prevent.

One-line fix, and it keeps this job honest with the convention.

trishorts (commented):
Reviewed at `36e34ef1`, with the `/oracle` consulted on prior art. **Nothing here blocks** — all 14 checks are green and the extraction itself is sound.

On the installer: I verified the *Validate Installer Contents* failure independently before seeing your fix, and the fix is right — both components with fresh explicit GUIDs, sourced from `$(var.GUI_TargetDir)`, adjacent to `src_GUIFunctions.pdb`, pdb parity included. Worth noting the PR body still says "nothing about the Windows build or the WiX installer changes"; that sentence is now false and is the kind of claim a future reader would trust.

The oracle backs the design: MetaMorpheus master has no UI-free MVVM home (EngineLayer/TaskLayer/CMD host zero view models), no source-linking pattern anywhere in the repo, and no test project has ever run on Linux/macOS. `GuiFunctions.Core` duplicates nothing. I also checked `TypeForwards.cs` exhaustively — 33 public types in the new assembly, 33 forwards, exact set match.

Four inline notes below. Three more that don't anchor to the diff:

**`ColorMapping/README.md:26` is now stale** — its file tree still lists `ColorGradientType.cs` under `GuiFunctions`. More interesting than the stale line: the enum now sits alone in Core, across an assembly boundary from `ColorGradientFactory` and all five palette classes that switch on it. That split may be unavoidable, but it is worth a word in the README so the next person does not "fix" it by moving the enum back.

**The body's file counts do not match the csproj.** It says "all of `Test/GuiTests` bar three files" — the Remove list drops five, so 11 of 16 are linked. And "eight of the `Test/MetaDraw` files" — actually 3 of 24 (21 removed). The 187-test figure may well be right; it is the two ratios that overstate coverage.

**`DownloadUniProtDatabaseFunctions.GetUniprotFilename` hardcodes a `\` path separator**, and has now moved into an assembly whose entire premise is platform neutrality. Pre-existing and moved unchanged, so out of scope for this PR — but it is a good argument for filing it now, since the file's new address implies a guarantee it does not yet meet.

Nice piece of work — the fixed-point method for finding the file set, and linking the tests rather than copying them, are both the right calls.

MetaMorpheus/GuiFunctions.Core/AssemblyInfo.cs
```diff
@@ -0,0 +1,7 @@
+using System.Runtime.CompilerServices;
+
+// GuiGlobalParams and friends deliberately keep internal setters so only the settings layer writes
+// them. Splitting this assembly out of GuiFunctions put those writers on the other side of an
+// assembly boundary, so grant them access rather than widening the public API.
+[assembly: InternalsVisibleTo("GuiFunctions")]
+[assembly: InternalsVisibleTo("Test")]
```
trishorts, line 7: `GuiFunctions.Core.Tests` is not granted access here, so the two projects do not compile the linked sources under the same visibility — which narrows the "same source, one copy, no drift" property slightly.

Nothing is broken today: every internal setter is on `GuiGlobalParams`, and the only test exercising them (`GuiGlobalParamsTests.cs`) is legitimately WPF-excluded — it goes through `GuiGlobalParamsViewModel`, which stays in `GuiFunctions` and imports `System.Windows`. So the asymmetry is latent.

But it means a future linked test touching an internal compiles in `Test.csproj` and fails only in the cross-platform project, with an error that points at visibility rather than at the real cause. Adding `[assembly: InternalsVisibleTo("GuiFunctions.Core.Tests")]` costs a line and keeps the two compilation contexts genuinely identical.

(The `"GuiFunctions"` grant above is definitely needed — `GuiGlobalParamsViewModel.cs:45` writes those setters.)

nbollis (dismissed):

MetaMorpheus/MetaMorpheus.sln
trishorts: Still open, and I can now put numbers on it rather than leave it as a worry. This is the one thing I would hold the PR on.

`dotnet build MetaMorpheus.sln -c UbuntuMac` is exactly what `build.yml:28-30` runs on `ubuntu-latest` and `macos-latest`. On this branch (`8f4798a`) that build succeeds with 0 errors and puts the output here:

```
GuiFunctions/bin/Release/                <- every pre-existing project
GuiFunctions.Core/bin/UbuntuMac/         <- new
GuiFunctions.Core.Tests/bin/UbuntuMac/   <- new
```

The cause is in the rows this PR adds. Every sibling project maps `UbuntuMac|Any CPU.ActiveCfg = Release|Any CPU`, but `{7B63C758-...}` and `{9AB2EDBA-...}` map `UbuntuMac|Any CPU.ActiveCfg = UbuntuMac|Any CPU` (sln lines 195-196 and 213-214). `UbuntuMac` is not a configuration either project defines, so MSBuild falls through to its unknown-configuration defaults:

```
$ dotnet msbuild GuiFunctions.Core/GuiFunctions.Core.csproj \
    -getProperty:Optimize -getProperty:DefineConstants -getProperty:OutputPath \
    -p:Configuration=UbuntuMac

Optimize:        ""                   (Release: "true")
DefineConstants: "TRACE;UBUNTUMAC"    (Release: "TRACE;RELEASE")
OutputPath:      bin\UbuntuMac\net10.0\
```

So on both Unix legs the extracted assembly is compiled **unoptimized and without `RELEASE`**, into a directory nothing else looks in. CI stays green because nothing asserts on it — and note that `build.yml`'s own arch-neutrality guard reads `bin/Release/<project>.dll`, so if `GuiFunctions.Core` were ever added to that loop it would silently find no file rather than check one.

The fix is small: point the `UbuntuMac|*` rows for both new project GUIDs at `Release|Any CPU` the way every sibling does, and add `<Configurations>Debug;Release</Configurations>` to both csprojs so the mismatch cannot quietly come back. That second half is the other thread I opened — same root cause, so treat them as one change.

Separately, while you are in this file: the six new solution **platforms** (`Debug|x64`, `Debug|x86`, `Release|x64`, `Release|x86`, `UbuntuMac|x64`, `UbuntuMac|x86`) are `dotnet sln add` churn. Master carries only `Any CPU` for all three configurations, and none of the new rows is reachable from CI or from the installer build. Worth dropping them so the platform list stays the three it was.

MetaMorpheus/GuiFunctions.Core/GuiFunctions.Core.csproj
trishorts: Same root cause as the `MetaMorpheus.sln` thread, so I have put the evidence and the full fix there rather than repeating it here.

Short version: `<Configurations>Debug;Release</Configurations>` on both new csprojs is only half the fix — on its own it does not change the build, because MSBuild accepts an unlisted configuration and `<Configurations>` mainly drives the VS dropdown. It is the sln rows that actually route these two projects into a `UbuntuMac` configuration they do not define. Add both together and this class of mistake stops being silent.

Leaving this open until the sln rows go with it.

.github/workflows/Test.yml
trishorts: Downgrading this — I checked, and it does not fire today.

The only `[Category("ExternalService")]` test under the linked directories is in `Test/GuiTests/GuiFunctionsTest.cs`, and that file is in the `<Compile Remove>` list, so nothing external compiles into `GuiFunctions.Core.Tests` right now. My original phrasing implied a live problem, and that was wrong.

It is still worth the one-line guard, because the include is a wildcard (`..\Test\GuiTests\**\*.cs` and the `MetaDraw` equivalent): the next ExternalService test added under either directory gets picked up automatically and lands in a **blocking** three-platform job, which is precisely what the separate non-blocking `External-service tests` job exists to prevent. Appending `--filter "Category!=ExternalService"` to the `dotnet test` line costs nothing and matches what the Windows job already does.

Non-blocking — your call.

MetaMorpheus/GuiFunctions.Core/AssemblyInfo.cs
trishorts: Withdrawing this one; I was wrong.

`dotnet build MetaMorpheus.sln -c UbuntuMac` completes with 0 errors on this branch, and CI is green on all three platforms, so `GuiFunctions.Core.Tests` compiles fine without the grant. The reason is that the linked test which actually touches the internal setters, `Test/GuiTests/GuiGlobalParamsTests.cs`, is in the `<Compile Remove>` list — I had not checked that list against the internals before writing the comment.

Resolving.

trishorts:
Re-reviewed at `8f4798a`. Net: one blocker, one nit, one withdrawal.

**Blocker** — the solution rows route `GuiFunctions.Core` and `GuiFunctions.Core.Tests` into a `UbuntuMac` configuration neither project defines, so the two Unix legs of `build.yml` compile the extracted assembly unoptimized, without `RELEASE`, into `bin/UbuntuMac/` while every sibling lands in `bin/Release/`. Proven by building rather than inferred; details and the two-part fix are in the `MetaMorpheus.sln` thread. The `<Configurations>` thread is the same root cause and should be fixed with it.

**Nit** (non-blocking) — the missing `--filter "Category!=ExternalService"` on the new job. I confirmed it does not fire today; it is a latent consequence of the wildcard `<Compile Include>`.

**Withdrawn** — my `InternalsVisibleTo` comment was wrong and I have resolved it. The linked test that touches the internal setters is in the `<Compile Remove>` list, so the project compiles fine without the grant.

Everything else holds up well. The fixed-point method for choosing the extracted set, the linked-not-copied tests, the type forwards for the XAML markup compiler, and catching the missing `coverlet.collector` are all the right calls, and each is justified in the commit that made it.

trishorts:
Cross-reference, since it affects how the open `MetaMorpheus.sln` thread here gets resolved.

#2733 is cut from *this* branch rather than from master -- `git merge-base` on the two heads gives `36e34ef1`, a commit on this branch. It carries 34 files byte-identical to this PR, including the whole `GuiFunctions.Core` project, `AssemblyInfo.cs`, `Test.yml` and `Product.wxs`. But this branch has moved on since (`8f4798ae`), so the two are no longer ancestor-related and nothing propagates between them.

Concretely: **#2733 adds the same nine `SolutionConfigurationPlatforms` entries** that the thread above is still open on. I checked both heads and they are identical on that point. So fixing it here does not fix it there -- #2733 needs either the same fix applied again or, better, a rebase onto this branch's current tip, which would also clear its merge conflict with master.

I have flagged it on #2733 as well. Mentioning it here so this thread is not closed on the assumption that it covers both.

MetaMorpheus/MetaMorpheus.sln
trishorts: These six new solution platforms (`Debug|x64`, `Debug|x86`, `Release|x64`, `Release|x86`, `UbuntuMac|x64`, `UbuntuMac|x86`) take the matrix from 3 to 9 and generate rows for all ten projects — most of the +142 lines here.

Master has only the three `Any CPU` configurations, and nothing in this PR needs more than that: #2709 adds two projects to the same solution with a 16-line diff and no new platforms.

Recommend dropping all six and keeping only the three `Any CPU` rows per new project. That removes the two problems flagged below as a side effect.

trishorts (commented):
The extraction itself is the right shape and I'd take it. I checked the part that is easiest to get wrong — whether the 25 relocated files are genuinely untouched — and they are: rename-aware `git diff` reports **0 changed lines** across all 25, so the only reviewable payload is the eight build/CI files. I also confirmed the fixed-point claim holds where it is most suspicious: `GuiGlobalParamsTests.cs` has no `System.Windows` reference anywhere, and it is still correctly excluded, because `GuiGlobalParamsViewModel` itself uses `System.Windows` at `GuiFunctions/GuiGlobalParamsViewModel.cs:5`. Reading imports really would have given the wrong answer.

The WiX side is fine too — both new components land inside `ExecutablesAndLibrariesComponentGroup`, which `ProductFeature` already references, so nothing is orphaned. `OxyPlot.Core 2.0.0` matches the `OxyPlot.Wpf 2.0.0` that `GuiFunctions` already pins.

The solution file is where I'd push back, and it is worth taking seriously because it is the one part of this PR that changes behaviour for projects the PR is not otherwise touching.

## The solution gained six new platforms

Master has three configurations: `Debug|Any CPU`, `Release|Any CPU`, `UbuntuMac|Any CPU`. This adds `Debug|x64`, `Debug|x86`, `Release|x64`, `Release|x86`, `UbuntuMac|x64` and `UbuntuMac|x86` — a 3 to 9 matrix — and generates rows for all ten projects, which is most of the +142 lines. The description mentions setting rows by hand for the new projects; it does not mention expanding the matrix for the existing nine.

That expansion is what produces the next two problems.

## The new `UbuntuMac|x64` / `x86` rows re-enable builds master deliberately disables

Three projects carry an `ActiveCfg` with **no** `Build.0` under `UbuntuMac|Any CPU` on master. That omission is not an oversight, it is the mechanism that keeps them out of the Linux/macOS build:

- `{C654FC97...}` **GUI** — WPF
- `{BE72541A...}` **MetaMorpheusSetup** — WiX
- `{E0EA5AC4...}` **Bootstrapper** — WiX

The new rows give all three a `Build.0` under `UbuntuMac|x64` and `UbuntuMac|x86`. It is latent rather than live — `build.yml` passes `--configuration UbuntuMac` with no `-p:Platform`, so it still resolves `UbuntuMac|Any CPU` — but it arms a trap that fires the first time anyone selects `UbuntuMac|x64` in Visual Studio or a workflow adds a platform argument. This is the same footgun as mzLib #1127, where a new solution platform mirroring the others silently re-enabled a configuration a project was deliberately excluded from, and PR CI could not catch it.

## The two new projects map `UbuntuMac` differently from every other project

Every pre-existing project maps `UbuntuMac|Any CPU` to **`Release|Any CPU`**. The two new ones map it to **`UbuntuMac|Any CPU`** — a project configuration that does not exist: no `.csproj` in the repo declares an `UbuntuMac` configuration, which is exactly why everything else redirects to `Release`.

I measured the consequence rather than reasoning about it. Building `GuiFunctions.Core` under this branch:

```
dotnet build GuiFunctions.Core/GuiFunctions.Core.csproj -c UbuntuMac
-> 0 Error(s), output in GuiFunctions.Core/bin/UbuntuMac/
```

So on the Unix legs of `build.yml` these two projects land in `bin/UbuntuMac/` while all nine others land in `bin/Release/` — which is the directory the architecture-neutrality assert in the same workflow reads (`$project/bin/Release/$DOTNET_FRAMEWORK/$project.dll`). Nothing fails today because the assert only covers CMD, EngineLayer and TaskLayer, but the new projects are precisely the ones whose loadability on arm64 this PR cares about, and they are the two now sitting outside the directory that check looks in.

This also means the description's *"the rows here are set by hand to match the other projects"* is not quite what landed — they do not match; they are the only two that do not.

**#2709 gets this exactly right, for reference**: three rows per project, `UbuntuMac|Any CPU.ActiveCfg = Release|Any CPU` plus `Build.0`, no new platforms, 16 lines of sln diff for two projects. Doing the same here removes all of the above.

## Minor: the new CI job is missing the ExternalService filter

The sibling `unittests` job runs with `--filter "Category!=ExternalService"` so a UniProt outage cannot redden a required check. The new job has no filter. It is not a live problem — the only `ExternalService`-tagged file under the linked directories is `GuiTests/GuiFunctionsTest.cs`, and that is already in the `Compile Remove` list — but the new project globs `..\Test\GuiTests\**\*.cs`, so the first live-service test added there joins this job automatically and starts blocking on a third party's uptime. Cheap to close now.

Fixes for all of the above are in a branch — see the follow-up comment.

MetaMorpheus/MetaMorpheus.sln
```diff
@@ -22,58 +22,200 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Test", "Test\Test.csproj",
...
+		{C654FC97-FBD1-43D7-9F61-35FDD1A4E0AD}.Debug|x86.Build.0 = Debug|Any CPU
 		{C654FC97-FBD1-43D7-9F61-35FDD1A4E0AD}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{C654FC97-FBD1-43D7-9F61-35FDD1A4E0AD}.Release|Any CPU.Build.0 = Release|Any CPU
+		{C654FC97-FBD1-43D7-9F61-35FDD1A4E0AD}.Release|x64.ActiveCfg = Release|Any CPU
+		{C654FC97-FBD1-43D7-9F61-35FDD1A4E0AD}.Release|x64.Build.0 = Release|Any CPU
+		{C654FC97-FBD1-43D7-9F61-35FDD1A4E0AD}.Release|x86.ActiveCfg = Release|Any CPU
+		{C654FC97-FBD1-43D7-9F61-35FDD1A4E0AD}.Release|x86.Build.0 = Release|Any CPU
 		{C654FC97-FBD1-43D7-9F61-35FDD1A4E0AD}.UbuntuMac|Any CPU.ActiveCfg = Release|Any CPU
+		{C654FC97-FBD1-43D7-9F61-35FDD1A4E0AD}.UbuntuMac|x64.ActiveCfg = UbuntuMac|Any CPU
+		{C654FC97-FBD1-43D7-9F61-35FDD1A4E0AD}.UbuntuMac|x64.Build.0 = UbuntuMac|Any CPU
```
trishorts: This is the one that concerns me. `{C654FC97...}` is the **WPF GUI**, and on master it has an `ActiveCfg` with **no `Build.0`** under `UbuntuMac|Any CPU` — that omission is deliberate, and it is what keeps WPF out of the Linux/macOS build.

These new rows hand it a `Build.0` under `UbuntuMac|x64` and `|x86`. The same happens to `{BE72541A...}` MetaMorpheusSetup and `{E0EA5AC4...}` Bootstrapper, both WiX.

Latent today, because `build.yml` passes `--configuration UbuntuMac` with no `-p:Platform` and so still resolves the `Any CPU` row. It fires the moment anyone picks `UbuntuMac|x64` in VS or a workflow gains a platform argument — and PR CI cannot catch it, which is exactly how mzLib #1127 went in.

MetaMorpheus/MetaMorpheus.sln
```diff
@@ -22,58 +22,200 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Test", "Test\Test.csproj",
...
+		{7B63C758-BA12-4BDB-879F-DBD8AF82A959}.Debug|x64.Build.0 = Debug|Any CPU
+		{7B63C758-BA12-4BDB-879F-DBD8AF82A959}.Debug|x86.ActiveCfg = Debug|Any CPU
+		{7B63C758-BA12-4BDB-879F-DBD8AF82A959}.Debug|x86.Build.0 = Debug|Any CPU
+		{7B63C758-BA12-4BDB-879F-DBD8AF82A959}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{7B63C758-BA12-4BDB-879F-DBD8AF82A959}.Release|Any CPU.Build.0 = Release|Any CPU
+		{7B63C758-BA12-4BDB-879F-DBD8AF82A959}.Release|x64.ActiveCfg = Release|Any CPU
+		{7B63C758-BA12-4BDB-879F-DBD8AF82A959}.Release|x64.Build.0 = Release|Any CPU
+		{7B63C758-BA12-4BDB-879F-DBD8AF82A959}.Release|x86.ActiveCfg = Release|Any CPU
+		{7B63C758-BA12-4BDB-879F-DBD8AF82A959}.Release|x86.Build.0 = Release|Any CPU
+		{7B63C758-BA12-4BDB-879F-DBD8AF82A959}.UbuntuMac|Any CPU.ActiveCfg = UbuntuMac|Any CPU
```
trishorts: Every other project in this solution maps `UbuntuMac|Any CPU` to `Release|Any CPU`. These two map it to `UbuntuMac|Any CPU`, and no `.csproj` in the repo declares an `UbuntuMac` configuration — which is why everything else redirects to `Release`.

Measured on this branch:

```
dotnet build GuiFunctions.Core/GuiFunctions.Core.csproj -c UbuntuMac
-> 0 Error(s), output in GuiFunctions.Core/bin/UbuntuMac/
```

So on the Unix legs these two land in `bin/UbuntuMac/` while the other nine land in `bin/Release/` — the directory `build.yml`'s architecture-neutrality assert reads. Harmless right now only because that assert covers CMD/EngineLayer/TaskLayer, but these are the two projects whose arm64 loadability this PR is about.

Should be `= Release|Any CPU` on both rows, matching the rest.

.github/workflows/Test.yml
trishorts: Missing `--filter "Category!=ExternalService"`, which the sibling `unittests` job carries so a UniProt outage cannot redden a required check.

Not a live problem — the only `ExternalService` file under the linked directories is `GuiTests/GuiFunctionsTest.cs` and it is already in the `Compile Remove` list. But this project globs `..\Test\GuiTests\**\*.cs`, so the next live-service test added there joins this job automatically, and this job runs on three OSes.

trishorts:
@acesnik — fixes for the solution-file and CI findings are on a branch against yours: **acesnik/MetaMorpheus#10**. Merging it updates this PR in place.

Net **-131 lines**: the six added solution platforms and their rows come out, the two new projects redirect `UbuntuMac` to `Release|Any CPU` like the other nine, and the cross-platform job gains `Category!=ExternalService`. I verified the redirect through a solution build both ways (`bin/UbuntuMac/` → `bin/Release/`) and diffed the whole `ProjectConfigurationPlatforms` section to confirm every pre-existing project is byte-identical to master again.

The extraction itself I did not touch — that part is right, and the fixed-point method genuinely earned its keep on `GuiGlobalParamsTests`.

trishorts:
@acesnik — the red build here is not from the solution-file merge. **acesnik/MetaMorpheus#13** fixes it in one line.

`Test/MetaDraw/MetaDrawExportRegressionTests.cs` is not in this branch at all — it is on master, added by #2753 on 2026-08-25. CI compiles the PR *merge* commit, so master's file meets this branch's `..\Test\MetaDraw\**\*.cs` glob and lands in a project with no WPF. Your last build here was 2026-08-20, before that file existed; merging #10 just triggered the first build since master moved.

Verified with master merged in: builds clean, **196 tests pass**.

It is also the deny-list-against-a-glob hazard from the review showing up for real, just as a WPF test rather than an external-service one. Worth deciding at some point whether the exclusion list should become an allow-list — anyone adding a WPF-touching file under `Test/MetaDraw/` breaks three platforms and finds out from someone else's PR.

PR #2730, opened 2026-08-15
feat(gui): drag and drop to reorder tasks

### LINKED ISSUE #1087: drag and move to reorder tasks in GUI

acesnik:
Closes #1087.

Tasks can now be reordered by dragging them, in both the Tasks tab tree and the Run tab summary.

## Why this turned out to be tractable

The concern raised on the issue in 2018 was that the TreeView data structure made this hard. That no longer applies: `PreRunTask` is rendered by a plain `DataTemplate` (`MainWindow.xaml:333`), not a `HierarchicalDataTemplate`, so the task tree is a flat list bound to `ObservableCollection<PreRunTask>`. Reordering is a single `Move()` on the same collection that the `+`/`-` keys from #1122 already mutate, followed by the existing `UpdateGuiOnPreRunChange()` to renumber Task1/Task2/…

## Behavior

- Drop position follows which half of the target row you are over — insert above or below — with a line drawn on the target to show where the task will land. Hovering past the last task draws the line beneath it and drops at the end.
- Drag starts only past `SystemParameters.Minimum{Horizontal,Vertical}DragDistance`, so clicking and double-click-to-edit are unaffected.
- Reordering is blocked during a run, when both trees rebind to `InProgressTasks`.

## Two things worth a reviewer's attention

**File drops still work.** The window already has `Drop="Window_Drop"` for `.toml`/`.mzML`/`.fasta`. Setting `AllowDrop` on the tree makes the tree the drop target for those drags too, so the handlers only set `e.Handled` once the drag is confirmed to be a task reorder; anything else bubbles up to `Window_Drop` untouched. This was a regression I introduced and then fixed, so it is the thing most worth re-checking.

**The dragged task is captured on mouse-down**, not hit-tested when the drag threshold is passed. Hit-testing at threshold time means a fast flick picks up whichever task the cursor has already moved onto, and lets a press that began outside the tree drag a task it never grabbed.

## Verification status

Compiles clean, XAML markup compile included — which is what confirms the handler signatures resolve on both trees.

**The drag interaction itself is untested.** This was written on macOS, where WPF cannot run, hence the draft. Two things specifically want a Windows pass:

1. The insertion line assumes the default `TreeViewItem` template template-binds `BorderBrush`/`BorderThickness`. If it does not, the line simply will not render — reordering still works, it is only the visual cue.
2. There is no auto-scroll when dragging near the edge of a long task list. Task lists are typically short enough that this has not seemed worth the complexity, but say the word if you would like it.

No test: `Test.csproj` references `GuiFunctions`, not `GUI`, so this code-behind is not reachable from the test project — the same reason the existing `MoveSelectedTask_Click` is untested. Making the index arithmetic testable would mean lifting it into `GuiFunctions`; happy to do that if it is wanted.

### DIFF
--- a/MetaMorpheus/GUI/MainWindow.xaml.cs
+++ b/MetaMorpheus/GUI/MainWindow.xaml.cs
@@ -16,6 +16,7 @@ using System.Windows.Controls;
 using System.Windows.Controls.Primitives;
 using System.Windows.Documents;
 using System.Windows.Input;
+using System.Windows.Media;
 using System.Windows.Navigation;
 using Omics.Modifications;
 using TaskLayer;
@@ -37,6 +38,9 @@ namespace MetaMorpheusGUI
         private readonly ObservableCollection<RawDataForDataGrid> SelectedSpectraFiles = new ObservableCollection<RawDataForDataGrid>();
         private readonly ObservableCollection<ProteinDbForDataGrid> SelectedProteinDatabaseFiles = new ObservableCollection<ProteinDbForDataGrid>();
         private ObservableCollection<InRunTask> InProgressTasks;
+        private Point DragStartPoint;
+        private PreRunTask TaskToDrag;
+        private TreeViewItem DropIndicatorItem;
         public static string NewestKnownMetaMorpheusVersion { get; private set; }
 
         public MainWindow()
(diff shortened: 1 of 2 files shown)


acesnik:
🤖 @trishorts this is #1087, which you filed in April 2018. Two asks: do you still want it, and if so would you take it from here? Maintainer edits are on, so you can push straight to `feat/gui-drag-reorder-tasks-1087`.

I can't test it. This was written on macOS, where WPF won't run, so the drag has never been performed. The only thing verified is that it compiles, markup compile included — that confirms the handler signatures resolve on both trees and nothing beyond that.

The issue has no body, so I read the title as insert-at-position reordering: drop above or below the row you're over, with a line showing where it lands. Worth saying in case you meant something else.

Two things to check on Windows:

- Dropping `.toml`/`.mzML`/`.fasta` onto the task list should still add the file. `AllowDrop` on the tree makes it the drop target for those drags too, so the handlers only set `e.Handled` once the drag is confirmed to be a task reorder. Unverified, and the most likely thing to be wrong.
- The insertion line assumes the default `TreeViewItem` template template-binds `BorderBrush`/`BorderThickness`. If it doesn't, reordering still works and only the visual cue is missing.

trishorts (commented):
Reviewed at `63bbb9af`, with `/oracle` consulted on prior art. **No blockers** — all 11 checks are green, and the parts you flagged as unverified I was able to settle from the source. Taking your three open questions in turn.

### The index arithmetic is correct

I brute-forced `TaskTreeView_Drop`'s destination computation against the intended semantics — every list size 2 through 7, every drag source, every target row, both halves, plus drop-past-the-end. **332/332 cases agree.** That includes the `if (oldPosition < newPosition) newPosition--` correction, which is exactly where this normally goes wrong, and the drop-onto-itself no-op.

### The insertion line will render

You listed this as the main visual unknown. It is fine: there is no custom `ControlTemplate` for `TreeViewItem` anywhere in `MainWindow.xaml` — the only `ItemContainerStyle` (line 802) sets `FontSize` and `FontWeight` and nothing else — so the default template applies, and it template-binds both `BorderBrush` and `BorderThickness`.

One caveat that falls out of the same template: its bordered element sits in `Grid.Column="1"`, after the ~19px expander column, so the line will be **indented rather than spanning the full row width**. Cosmetic, and arguably fine, but not what the description implies.

### File drops do still work

I traced both handlers. Every early-return path in `TaskTreeView_DragOver` and `TaskTreeView_Drop` leaves `e.Handled` false and never touches `e.Effects`, so a non-task drag bubbles to `Window_Drop` (`MainWindow.xaml.cs:504`) exactly as before. The reverse direction is safe too: a *task* dropped outside both trees reaches `Window_Drop`, where `GetData(DataFormats.FileDrop)` returns null and the `files != null` guard catches it — a clean no-op rather than a crash. The mitigation is the right shape. It still wants the Windows pass, but not because the logic looks wrong.

### The run guard holds

`TasksCanBeReordered` keys off `treeView.DataContext`, and `MainWindow.xaml.cs:1155-1156` really does rebind both trees to `InProgressTasks` when a run starts, so reordering is genuinely blocked mid-run rather than blocked by assumption.

### Answering the two things you offered

**Yes to lifting the index arithmetic** — you asked, so: it is worth doing. Not because it looks wrong (see above) but because those four lines are the part most likely to be broken by a later well-meaning edit, and they are pure integer logic with no WPF in them. `Test.csproj` references CMD, EngineLayer, GuiFunctions and TaskLayer but not GUI (lines 34-37), which is exactly why nothing here is reachable today. A static `GetDropIndex(int oldPosition, int targetIndex, bool dropAboveTarget, int count)` in `GuiFunctions/Util/` would be, and there is a clean precedent for the shape and placement: `GuiFunctions/Util/CustomMIonLossManager.cs` tested from `Test/GuiTests/CustomMIonLossTests.cs`, and `HungarianAlgorithm` tested from `Test/MetaDraw/HungarianAlgorithmTests.cs`. Pass ints, not the collection, so `GuiFunctions` stays free of a GUI-layer type. Worth noting that `HungarianAlgorithm` is one of the files #2729 moves into `GuiFunctions.Core` — if that lands first, this helper could live there and be tested on all three platforms, which is a neat demonstration of what that split buys.

**No to auto-scroll** for now. Agreed that task lists are short enough; not worth the complexity until someone actually hits it.

### One thing you did not flag

There is already a functional twin of this drag pattern in the codebase: `MetaDraw.xaml.cs:988-1042` reorders `PsmStatPlotFiles` with the same `_dragStartPoint` / `SystemParameters.Minimum*DragDistance` / `DoDragDrop` shape, and `GetDataFromListBoxItemUnderMouse` at :1035 mirrors your `GetTreeViewItemContaining`. Yours is strictly the better implementation — typed `DataObject`, a `GetDataPresent` guard, `e.Handled`, a drop indicator, the half-row rule, drop-at-end, and the `LogicalTreeHelper` fallback that the MetaDraw copy would trip over on a `Run` inside a `TextBlock`.

I am not asking you to touch MetaDraw in this PR. Flagging it so the duplication is a known, deliberate state rather than an accident, and so whoever unifies them later knows to converge on this one.

Nice work — the mouse-down capture decision and the "only set `e.Handled` once the drag is confirmed" rule are both the right calls, and the reasoning in the description made this much faster to review.

MetaMorpheus/GUI/MainWindow.xaml.cs
```diff
@@ -874,20 +878,204 @@ private void MoveSelectedTask_Click(object sender, RoutedEventArgs e, bool moveT
...
+            }
+
+            e.Effects = DragDropEffects.Move;
+            e.Handled = true;
+
+            var targetItem = GetTreeViewItemContaining(e.OriginalSource);
+            bool lineAboveTarget = targetItem != null && IsInUpperHalf(targetItem, e);
+
+            // hovering past the last task draws the line below it, matching the drop-at-the-end behavior
+            targetItem ??= treeView.ItemContainerGenerator.ContainerFromIndex(PreRunTasks.Count - 1) as TreeViewItem;
```
trishorts, line 962: This fallback is unconditional, so it fires for **any** cursor position inside the tree that is not over a row — not just below the last one.

In practice that means the thin strip above the first row, and the left/right padding beside the rows, all draw the line beneath the *last* task and drop there. Dropping at the end when the user is visually at the top is the one case where the indicator is honest about a surprising outcome rather than preventing it.

The empty area below the last task is the case you designed for and it works. The fix is to make the fallback conditional on actually being past the last row — compare the cursor's Y against the bottom of the last container, and simply draw nothing when the drag is over the tree but not over a meaningful drop position. `TaskTreeView_Drop` would want the matching guard, since it computes `newPosition = PreRunTasks.Count` from the same "no target item" condition.

Low severity — a short task list has very little non-row area — but it is cheap to tighten while this code is fresh.

MetaMorpheus/GUI/MainWindow.xaml.cs
```diff
@@ -874,20 +878,204 @@ private void MoveSelectedTask_Click(object sender, RoutedEventArgs e, bool moveT
...
+            e.Handled = true;
+
+            int oldPosition = PreRunTasks.IndexOf(draggedTask);
+            var targetItem = GetTreeViewItemContaining(e.OriginalSource);
+
+            // dropping past the last task moves it to the end of the list
+            int newPosition = PreRunTasks.Count;
+            if (targetItem?.DataContext is PreRunTask targetTask)
+            {
+                newPosition = PreRunTasks.IndexOf(targetTask) + (IsInUpperHalf(targetItem, e) ? 0 : 1);
```
trishorts, line 1001: These four lines are the whole reorder semantics, and they are the strongest candidate for lifting into `GuiFunctions` as you offered in the description.

To be clear about why: **it is not because they are wrong.** I brute-forced this computation against the intended insert-before/insert-after semantics for every list size 2 through 7, every drag source, every target row, both halves, and the drop-past-the-end path — 332 of 332 cases agree, including the `oldPosition < newPosition` decrement and the drop-onto-itself no-op.

The reason to move it is that it is correct *and* currently unprovable. `Test.csproj` references CMD, EngineLayer, GuiFunctions and TaskLayer but not GUI (`Test.csproj:34-37`), so no test can reach this file, and the next person to adjust the half-row rule or the end-of-list case has nothing to catch a regression.

A pure static taking ints — `GetDropIndex(int oldPosition, int targetIndex, bool dropAboveTarget, int count)` — has no WPF in it and would move cleanly to `GuiFunctions/Util/`. Precedent for both the shape and where the tests go: `GuiFunctions/Util/CustomMIonLossManager.cs` with `Test/GuiTests/CustomMIonLossTests.cs`, and `HungarianAlgorithm` with `Test/MetaDraw/HungarianAlgorithmTests.cs`. Keep the collection itself on this side of the boundary and pass indices, so `GuiFunctions` gains no dependency on a GUI-layer type.

MetaMorpheus/GUI/MainWindow.xaml.cs
```diff
@@ -874,20 +878,204 @@ private void MoveSelectedTask_Click(object sender, RoutedEventArgs e, bool moveT
...
+
+            DropIndicatorItem.ClearValue(Control.BorderBrushProperty);
+            DropIndicatorItem.ClearValue(Control.BorderThicknessProperty);
+            DropIndicatorItem = null;
+        }
+
+        /// <summary>
+        /// Walks up from a hit-tested element to the tree view item that contains it, if any.
+        /// </summary>
+        private static TreeViewItem GetTreeViewItemContaining(object source)
```
trishorts, line 1066: Worth knowing that this helper already has a near-twin in the codebase: `GetDataFromListBoxItemUnderMouse` at `MetaDraw.xaml.cs:1035`, which does the same walk for `ListBoxItem` as part of the `PsmStatPlotFiles` reorder at `MetaDraw.xaml.cs:988-1042`.

Yours is the better of the two. The `LogicalTreeHelper` fallback here is the meaningful difference — the MetaDraw version calls `VisualTreeHelper.GetParent` unconditionally, which is exactly the case your comment describes (a `Run` inside a `TextBlock` is not a `Visual`). It also starts from `e.OriginalSource` rather than re-hit-testing a `Point`, which is more robust.

No change requested in this PR — MetaDraw is not yours to touch here. Recording it so the duplication is deliberate rather than accidental, and so that if these ever converge, it is clear this is the version to keep.

PR #2731, opened 2026-08-16
Warn when a database annotates a modification we cannot match

acesnik:
🤖 Addresses smith-chem-wisc/mzLib#417, which asked for a message rather than a silent skip.

mzLib already collects these: `ProteinXmlEntry.ParseAnnotatedMods` puts every annotation it cannot resolve into `unknownModifications`. `DatabaseLoadingEngine` received that dictionary at both call sites and discarded it, so the annotation was dropped with nothing said.

## What it reports

Aggregated per database, added to the existing `errors` list that `RunSpecific` already routes through `Warn()`:

```
Warning: 15 annotated modification(s) in uniprot-mouse-reviewed-1-24-2018.xml.gz could not be
matched to a known modification and were ignored: '5-glutamyl polyglutamate', '5-glutamyl
polyglycine', 'Aspartate 1-(chondroitin 4-sulfate)-ester', 'Blocked amino end (Gln)', 'Blocked
amino end (Ser)', and 10 more
```

Five named then a count, because mzLib keys the dictionary by modification id and gives no accessions — a database using an unknown id throughout would otherwise warn once per occurrence. Warning rather than error, per the discussion in the issue.

## Measured against a real proteome

uniprot-mouse-reviewed, 16,946 entries: **15 distinct ids covering 249 annotations, 0.51% of the 48,606 modification-bearing features.** The full set is GPI anchors (140 across seven residues), `Omega-N-methylated arginine` (70), `N6-methylated lysine` (17), `PolyADP-ribosyl glutamic acid` (16), `5-glutamyl polyglutamate`/`polyglycine` (13), `Aspartate 1-(chondroitin 4-sulfate)-ester` (3) and two `Blocked amino end` entries.

Every one is a modification with **no single defined mass** or an ambiguous parent term, which is why they are absent from the loaded modification list and why this has gone unnoticed. **Nothing about which modifications get applied changes — search output is identical.** This only makes existing dropping visible.

## Tests

Three in `ProteinLoaderTest.cs`: one asserts the warning names both the modification and the database; one asserts a resolvable modification produces no warning, without which the first could be satisfied by warning unconditionally; and one annotates seven unmatched ids and asserts the message names five, reports `and 2 more`, and counts all seven.

Confirmed by removing the fix: **2 failed without it, 3 pass with it** — with the engine change reverted no warning is produced at all, so both positive tests fail and the negative one passes either way. `Test.csproj` is `net10.0-windows` so I ran them through a throwaway `net10.0` harness locally; Windows CI executes them for real. No new compiler warnings — the three in the touched file pre-exist on master.


nbollis (approved):

trishorts (commented):
Automated code review - 6 finding(s) posted as inline comments.

MetaMorpheus/EngineLayer/DatabaseLoading/DatabaseLoadingEngine.cs
```diff
@@ -126,6 +128,32 @@ public static List<IBioPolymer> LoadBioPolymers(string taskId, List<DbForTask> d
...
+        // Aggregated per database rather than per entry: mzLib keys the dictionary by modification id, so the
+        // annotated accessions are not recoverable here, and a database using an unknown id throughout would
+        // otherwise warn once per occurrence.
+        const int maxToName = 5;
+        var ids = unknownModifications.Keys.OrderBy(p => p, StringComparer.Ordinal).ToList();
+        string named = string.Join(", ", ids.Take(maxToName).Select(p => "'" + p + "'"));
+        if (ids.Count > maxToName)
+            named += $", and {ids.Count - maxToName} more";
+
+        errors.Add($"Warning: {ids.Count} annotated modification(s) in {Path.GetFileName(db.FilePath)} could not be "
```
trishorts, line 153: **Minor** - Warning counts distinct ids but reads as a count of annotations

**Problem:**
`ids.Count` is the number of distinct dictionary keys (modification ids), but the message renders it as "{count} annotated modification(s) ... were ignored". mzLib de-duplicates by id in `ProteinXmlEntry.ParseAnnotatedMods` (`if (!unknownModifications.ContainsKey(annotatedId))`), so the number of *annotations* dropped is strictly greater in general — the PR's own measurement is 15 ids covering 249 annotations. A user reading "15 annotated modification(s) ... were ignored" will reasonably conclude 15 features were dropped, not 249.

**Impact:**
Understates by ~16x on a real UniProt proteome how much annotation was discarded, which is the one number this warning exists to communicate.

**Suggested fix:**
Word the count for what it is — distinct ids / modification types — e.g. "{n} distinct annotated modification type(s)", without adding per-id occurrence counts (which the PR explicitly settled against and which the dictionary cannot supply anyway).

**Affected test scenarios:**
- UnmatchedModificationWarningNamesFiveThenCountsTheRest - seven distinct ids each annotated once - message states seven distinct modification types
- UnmatchedModificationWarningCountsIdsNotOccurrences - one id annotated on three residues - message states one, not three

MetaMorpheus/Test/DatabaseTests/ProteinLoaderTest.cs
```diff
@@ -247,6 +247,112 @@ public static void CatchesError()
...
+        /// A modification annotated in a database but absent from the known modifications is skipped by mzLib, which
+        /// records it in its unknownModifications dictionary. That dictionary used to be discarded here, so the
+        /// annotation vanished with nothing said about it (mzLib #417). It must now reach the user as a warning.
+        /// </summary>
+        [Test]
+        public static void UnmatchedModificationInDatabaseIsWarnedAbout()
+        {
+            string dbPath = WriteDbWithModificationDescription("unknownModDb.xml", "Nonexistent modification on S");
+
+            DatabaseLoadingEngine.LoadBioPolymers("TestTask", new List<DbForTask> { new DbForTask(dbPath, false) },
```
trishorts, line 298: **Minor** - New tests silently depend on process-wide GlobalVariables.AnalyteType

**Problem:**
All three added tests call `DatabaseLoadingEngine.LoadBioPolymers(...)` directly, and that method branches on the mutable process-wide static `GlobalVariables.AnalyteType` (`DatabaseLoadingEngine.cs:93`): `Oligo` routes to `LoadOligoDb`/`RnaDbLoader.LoadRnaXML`, anything else to `LoadProteinDb`. None of the three tests sets or restores that static. `Test/Transcriptomics/RnaDatabaseLoadingTests.cs:33,57,103` sets `GlobalVariables.AnalyteType = AnalyteType.Oligo` and never restores it (the fixture has no `TearDown`), and the only place it is set back to `Peptide` is the one-time `MySetUpClass.GlobalSetup` in `SetUpTests.cs:25`, which runs once per assembly.

**Impact:**
If the RNA fixture executes before this one — NUnit fixture ordering is not contractual and changes under filtered runs, `--where` selections, or IDE single-fixture runs — the protein XML written by these tests is fed to the RNA loader, `unknownMods` comes back empty, and the two positive tests fail while the negative control passes for the wrong reason. The failure would look like a bug in the feature rather than in the test setup.

**Suggested fix:**
Have the new tests pin the analyte type explicitly (a fixture-level `[SetUp]` setting `GlobalVariables.AnalyteType = AnalyteType.Peptide`, or set-and-restore inside each test) so they exercise the protein branch regardless of what ran before them.

**Affected test scenarios:**
- UnmatchedModificationInDatabaseIsWarnedAbout - run after a fixture that leaves AnalyteType=Oligo - still takes the protein branch and warns
- ResolvableModificationProducesNoUnmatchedModificationWarning - same - passes because the mod resolved, not because the DB was never parsed

MetaMorpheus/EngineLayer/DatabaseLoading/DatabaseLoadingEngine.cs
```diff
@@ -94,11 +94,13 @@ public static List<IBioPolymer> LoadBioPolymers(string taskId, List<DbForTask> d
             {
                 dbBioPolymers = LoadOligoDb(db.FilePath, searchTarget, decoyType, localizeableModificationTypes, db.IsContaminant, out Dictionary<string, Modification> unknownModifications, out int emptyOligoEntriesForThisDb, commonParameters, db.DecoyIdentifier);
                 emptyEntries += emptyOligoEntriesForThisDb;
+                AddUnknownModificationWarning(db, unknownModifications, errors);
```
trishorts, line 97: **Minor** - Oligo call site and the null-dictionary guard are both untested

**Problem:**
`AddUnknownModificationWarning` is wired into two call sites, but all three added tests exercise only the protein branch at line 103 (they run with the default peptide analyte type against a `mzLibProteinDb` XML). The Oligo call site at line 97 — reached via `LoadOligoDb`/`RnaDbLoader.LoadRnaXML` — has no test, and the FASTA path, which is the only path that sets the dictionary to `null` (`LoadOligoDb` line 168, `LoadProteinDb` line 197) and therefore the only thing exercising the `unknownModifications == null` guard at line 141, has none either. The PR body lists the null guard as one of the feature's invariants; nothing verifies it.

**Impact:**
A future change to the guard or to the Oligo wiring (for example dropping the null check, which would throw `NullReferenceException` on every FASTA load) passes the whole suite. `Test/Transcriptomics/RnaDatabaseLoadingTests.cs` already has an RNA XML fixture (`20mer1.xml`), so the RNA case is cheap to cover.

**Suggested fix:**
Add an RNA-side test that loads an oligo XML annotating an unmatched modification and asserts the same warning, and a test that loads a FASTA (e.g. `TestData/gapdh.fasta`) and asserts no exception and no `could not be matched` warning.

**Affected test scenarios:**
- UnmatchedModificationInOligoDatabaseIsWarnedAbout - AnalyteType=Oligo, RNA XML with an unresolvable modified-residue description - warning names the id and the database
- FastaDatabaseProducesNoUnmatchedModificationWarning - .fasta input (null unknownModifications) - no throw, no "could not be matched" entry in errors

MetaMorpheus/Test/DatabaseTests/ProteinLoaderTest.cs
```diff
@@ -247,6 +247,112 @@ public static void CatchesError()
...
+        /// </summary>
+        [Test]
+        public static void ResolvableModificationProducesNoUnmatchedModificationWarning()
+        {
+            string dbPath = WriteDbWithModificationDescription("knownModDb.xml", "Phosphoserine on S");
+
+            DatabaseLoadingEngine.LoadBioPolymers("TestTask", new List<DbForTask> { new DbForTask(dbPath, false) },
+                true, DecoyType.None, new List<string> { "UniProt" }, new CommonParameters(), out var errors);
+
+            Assert.That(errors.Any(p => p.Contains("could not be matched")), Is.False,
```
trishorts, line 322: **Minor** - Negative control passes vacuously if the test database never parses

**Problem:**
`ResolvableModificationProducesNoUnmatchedModificationWarning` discards the `List<IBioPolymer>` returned by `LoadBioPolymers` and asserts only the absence of `"could not be matched"` in `errors`. The whole point of the test is that `"Phosphoserine on S"` was seen, looked up in `ProteinDbLoader.IdWithMotifToMod`, and matched. But every failure mode that stops the annotation from ever reaching `ParseAnnotatedMods` — the hand-written XML in `WriteDbWithModificationDescriptions` losing its `<feature>` element, the entry failing to yield a `Protein`, the `modTypesToExclude` computation excluding UniProt — also produces an empty `unknownModifications` dictionary and therefore no warning, so the assertion holds.

**Impact:**
The negative control can go green while asserting nothing about modification resolution, which is exactly the failure mode it was added to prevent (the PR describes it as the guard against "warning unconditionally").

**Suggested fix:**
Capture the returned biopolymer list and assert it is non-empty and that the single protein carries a localized modification at position 3 whose `IdWithMotif` is `Phosphoserine on S`, so the test can only pass when the annotation was parsed *and* matched.

**Affected test scenarios:**
- ResolvableModificationProducesNoUnmatchedModificationWarning - feature element malformed so no annotation parsed - test fails instead of passing
- ResolvableModificationProducesNoUnmatchedModificationWarning - Phosphoserine resolvable - one protein returned with the mod attached at position 3

MetaMorpheus/Test/DatabaseTests/ProteinLoaderTest.cs
```diff
@@ -247,6 +247,112 @@ public static void CatchesError()
...
+        /// in ordinal order are named and the remainder is reported as a count, so a database using many unknown
+        /// ids produces one bounded message rather than an unbounded list.
+        /// </summary>
+        [Test]
+        public static void UnmatchedModificationWarningNamesFiveThenCountsTheRest()
+        {
+            string[] descriptions = Enumerable.Range(1, 7)
+                .Select(i => $"Nonexistent modification {i} on S")
+                .ToArray();
+            string dbPath = WriteDbWithModificationDescriptions("manyUnknownModsDb.xml", descriptions);
```
trishorts, line 337: **Minor** - Per-database aggregation, the PR's stated design, has no test

**Problem:**
The PR's central design choice is that the warning is aggregated *once per database* rather than once per entry or per occurrence, and `LoadBioPolymers` calls `AddUnknownModificationWarning` inside the `foreach (var db in dbFilenameList...)` loop with `Path.GetFileName(db.FilePath)` in the text. Every added test passes a single `DbForTask` containing a single `<entry>` with distinct ids. Nothing asserts that two databases produce two separate warnings each naming its own file, and nothing asserts that the same unknown id annotated on several entries of one database yields one warning rather than several.

**Impact:**
A regression that emitted the warning outside the per-db loop, reused the wrong `db` for the file name, or accumulated ids across databases into a single message would not be caught by any of the three new tests.

**Suggested fix:**
Add a test that passes two `DbForTask` entries with different file names and different unknown ids and asserts two distinct warnings, each naming its own file name and only its own ids; and extend one writer helper to emit several entries so a repeated id can be shown to warn once.

**Affected test scenarios:**
- New test - two databases each with one unmatched id - two warnings, each naming its own file and id
- New test - one database, same unmatched id on three entries - exactly one warning reporting 1 modification

MetaMorpheus/Test/DatabaseTests/ProteinLoaderTest.cs
```diff
@@ -247,6 +247,112 @@ public static void CatchesError()
...
+        private static string WriteDbWithModificationDescriptions(string fileName, params string[] modificationDescriptions)
+        {
+            string features = string.Concat(modificationDescriptions.Select(description =>
+                $"    <feature type=\"modified residue\" description=\"{description}\">\n" +
+                "      <location>\n" +
+                "        <position position=\"3\" />\n" +
+                "      </location>\n" +
+                "    </feature>\n"));
+
+            string path = Path.Combine(TestContext.CurrentContext.TestDirectory, fileName);
```
trishorts, line 273: **Minor** - Written test databases are never cleaned up, unlike the rest of the fixture

**Problem:**
`WriteDbWithModificationDescriptions` writes to `Path.Combine(TestContext.CurrentContext.TestDirectory, fileName)` — the test output root — and none of the three callers deletes the file afterwards. Every other test in this fixture that creates artifacts (`WriteTargetDecoyFasta_WhenEnabled_CreatesFastaFile`, `..._WhenDisabled_DoesNotCreateFile`, `..._ContainsBothTargetsAndDecoys`) ends with `Directory.Delete(outputFolder, true)`, so this breaks the file's own convention.

**Impact:**
`unknownModDb.xml`, `knownModDb.xml` and `manyUnknownModsDb.xml` accumulate in the build output directory and persist between runs, so a subsequent run reads a database left by the previous one rather than the one it intended to write if the writer is ever changed or fails partway.

**Suggested fix:**
Write into a per-test temp subdirectory (or the existing `TestData` scratch convention) and delete it in a `[TearDown]`/`finally`, matching the cleanup pattern the surrounding tests already use.

**Affected test scenarios:**
- UnmatchedModificationInDatabaseIsWarnedAbout - run twice in a row - second run reads a freshly written database
- UnmatchedModificationWarningNamesFiveThenCountsTheRest - after run - no XML left in the test output root

trishorts:
Checked at `cd6df7c`. The head is unchanged since my last pass — the only commit since is the master merge — so all six threads from that review are still open and none has been addressed. Nothing to close and nothing new to add; I have not re-posted them.

They are all Minors, so none of this blocks on its own, but the two I would actually act on before merging are the vacuous negative control (`ProteinLoaderTest.cs:322` passes whether or not the database parses) and the missing test for per-database aggregation (`:337`), since that is the PR's stated design and nothing currently pins it.

PR #2732, opened 2026-08-16
Do not write empty ProteinDetectionHypothesis elements to mzID

acesnik:
🤖 mzIdentML 1.1 requires at least one `PeptideHypothesis` in every `ProteinDetectionHypothesis`, and at least one `ProteinDetectionHypothesis` in every `ProteinAmbiguityGroup` — both `minOccurs="1"` in the HUPO-PSI schema. `MzIdentMLWriter` emits both empty, which ProteomeXchange rejects. This is part of the `.mzid` half of smith-chem-wisc/mzLib#259.

## Cause

`ProteinGroup.ConstructSubsetProteinGroup` builds the per-file view by keeping **all** of a group's `Proteins` while narrowing `AllPeptides` to the peptides seen in the current file. In a multi-file search a protein whose evidence came from another fraction therefore matches nothing in the writer's `peptide.Protein == protein` loop, and its hypothesis is written with no children. The same shape is reachable when a group's only supporting PSMs are ambiguous, since `FullSequence == null` PSMs are excluded from `PeptideEvidence`.

The list is now built with `List<>` instead of fixed-size arrays so an empty hypothesis, and a group left with none, can be skipped. `ProteinDetectionList/ProteinAmbiguityGroup` is `minOccurs="0"`, so dropping a group is valid. Retained elements keep their original `PDH_n` / `PAG_n` ids, so the diff against previous output is a pure deletion.

## Output change

**This changes the mzID.** Full 5-task vignette (Calibrate → GPTMD → Search as the vignette defines it), two fractions, mouse reviewed proteome + cRAP, run twice on identical inputs:

| task | file | PDH before | after | dropped | PAG before | after | dropped |
|---|---|---|---|---|---|---|---|
| Task1 Search | Frac4 | 2925 | 2826 | 99 | 2860 | 2763 | 97 |
| Task1 Search | Frac5 | 2963 | 2880 | 83 | 2888 | 2807 | 81 |
| Task3 Search (calib) | Frac4 | 2966 | 2844 | 122 | 2898 | 2782 | 116 |
| Task3 Search (calib) | Frac5 | 2986 | 2919 | 67 | 2913 | 2847 | 66 |
| Task5 Search (calib+GPTMD) | Frac4 | 3055 | 2922 | 133 | 2987 | 2860 | 127 |
| Task5 Search (calib+GPTMD) | Frac5 | 3043 | 2931 | 112 | 2972 | 2862 | 110 |
| **total** | | 17938 | 17322 | **616** | 17518 | 16921 | **597** |

Empty hypotheses after the fix: **0 in all six files.** Every "after" total is exactly `before − empty`, so only the invalid elements are removed and nothing else moves. 597 of the 616 were the sole protein in their group, taking the group with them; the other 19 sat in groups that survive with fewer hypotheses.

**Nothing outside the mzID changes.** `AllQuantifiedProteinGroups.tsv` (Task1 and Task5), `AllPSMs.psmtsv` (Task5) and the GPTMD-augmented `uniprot-mouse-reviewed-...GPTMD.xml` are **byte-identical** between the two runs. No search results, no quantification, no produced database is affected. Protein groups dropped from an mzID are still fully reported in the TSVs.

The count rises along the chain on both fractions, so the invalid tail is largest in the post-GPTMD search.

## Also fixed: a NullReferenceException

The SILAC pre-filter reads `!p.FullSequence.Contains("|")` guarded only by `p.BaseSequence != null`. An ambiguous PSM has a null `FullSequence`. Only reachable when `SilacLabels` is non-null — `SearchParameters` never initialises it, so non-SILAC searches pass null and skip the filter, but a SILAC search with any ambiguous PSM crashes mzID writing. Found while building the test, not from a report.

## This does not by itself make output submittable

Every `.mzID` MetaMorpheus writes is currently in namespace `http://psidev.info/psi/pi/mzIdentML/1.1.0` instead of `.../1.1`, which comes from mzLib 1.0.584 and makes the file unvalidatable regardless of this fix. That is smith-chem-wisc/mzLib#1171, and it needs an mzLib release plus a `PackageReference` bump before it reaches here. Both vignette runs above still show the wrong namespace.

So mzLib#259 needs this PR **and** mzLib#1171 **and** a release before a MetaMorpheus mzID passes ProteomeXchange.

## Not addressed

`sequence coverage` uses `proteinGroup.SequenceCoverageFraction.First()` for every protein in a group, so in a multi-protein ambiguity group the second and later hypotheses all report the first protein's coverage. `SequenceCoverageFraction` is ordered by `ListOfProteinsOrderedByAccession` while the writer iterates the unordered `Proteins` set, so a correct fix means pairing against the ordered list. It is a wrong value rather than a validity error and would change reported numbers, so it is left out.

I also have no explanation for why these counts are roughly ten times the checked-in 2024 vignette output (12 and 8 for the comparable files). The writer's `ProteinDetectionList` branch is unchanged since before that run, so the difference is upstream of it. Not investigated.

## Verification

**4 tests** in `MzIdentMlSchemaTest.cs` — the multi-file subset shape, the ambiguous-PSM shape, the SILAC crash, and a control asserting a well-formed group is *not* dropped. All four were confirmed to fail against the unfixed writer before being trusted.

Every pre-existing `MzIdentMLWriter` test passes an empty protein-group list, so the `ProteinDetectionList` branch had no coverage at all.

`Test` cannot run on macOS (the host needs `Microsoft.WindowsDesktop.App`), so these ran via a throwaway `net10.0` harness referencing `TaskLayer`/`EngineLayer`; `Test.csproj` compiles with `EnableWindowsTargeting`. **CI on windows-latest is the real check.**

### DIFF
--- a/MetaMorpheus/TaskLayer/SearchTask/MzIdentMLWriter.cs
+++ b/MetaMorpheus/TaskLayer/SearchTask/MzIdentMLWriter.cs
@@ -37,9 +37,14 @@ namespace TaskLayer
                         }
                     }
                 }
-                // FullSequence is null when the sequence could not be resolved; such PSMs are
-                // dropped further down anyway, and "|" only appears on a resolved ambiguous one.
-                psms = psms.Where(p => p.BaseSequence != null && p.FullSequence != null && !p.FullSequence.Contains("|") && !labelsToSearch.Any(x => p.BaseSequence.Contains(x)));
+                // SpectralMatch.FullSequence is assigned from PsmTsvWriter.Resolve(...).ResolvedValue,
+                // which is null whenever the candidate sequences disagree -- the pipe-joined form is
+                // ResolvedString and is discarded. So FullSequence is either a clean sequence or null,
+                // never "|"-delimited, and the old !FullSequence.Contains("|") conjunct could only ever
+                // be true or throw a NullReferenceException on an ambiguous PSM. Dropping it fixes the
+                // crash without narrowing psms, which also feeds peptides/proteins/filenames below;
+                // ambiguous PSMs are excluded from the identification data by unambiguousPsms instead.
+                psms = psms.Where(p => p.BaseSequence != null && !labelsToSearch.Any(x => p.BaseSequence.Contains(x)));
             }
 
             List<PeptideWithSetModifications> peptides = psms.SelectMany(i => i.BestMatchingBioPolymersWithSetMods.Select(v => v.SpecificBioPolymer as PeptideWithSetModifications)).Distinct().ToList();
--- a/MetaMorpheus/TaskLayer/SearchTask/MzIdentMLWriter.cs
+++ b/MetaMorpheus/TaskLayer/SearchTask/MzIdentMLWriter.cs
@@ -37,7 +37,9 @@ namespace TaskLayer
                         }
                     }
                 }
-                psms = psms.Where(p => p.BaseSequence != null && !p.FullSequence.Contains("|") && !labelsToSearch.Any(x => p.BaseSequence.Contains(x)));
+                // FullSequence is null when the sequence could not be resolved; such PSMs are
+                // dropped further down anyway, and "|" only appears on a resolved ambiguous one.
+                psms = psms.Where(p => p.BaseSequence != null && p.FullSequence != null && !p.FullSequence.Contains("|") && !labelsToSearch.Any(x => p.BaseSequence.Contains(x)));
             }
 
             List<PeptideWithSetModifications> peptides = psms.SelectMany(i => i.BestMatchingBioPolymersWithSetMods.Select(v => v.SpecificBioPolymer as PeptideWithSetModifications)).Distinct().ToList();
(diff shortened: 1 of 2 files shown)


trishorts (approved):
Approving. The diagnosis is right and the fix is well-scoped — I checked the design against the schema and the surrounding code rather than just the diff, and it holds up on every axis:

- **The `List<>` switch is the correct idiom here, not a mixed one.** The generated mzIdentML properties are `T[]` with no `Add`, which is *why* the writer pre-sizes. `PepXMLWriter.cs` — the repo's only other XSD-generated writer — already uses `List<>` → `.ToArray()` for exactly this problem (L37→L111, L125→L151).
- **Keeping the original ids with gaps is safe.** Nothing in the writer or in mzIdentML 1.1 references a `PAG_` or `PDH_` id; `id` is `xsd:string`, not `xsd:ID`, and the `*_ref` attributes are `xsd:string`, not `IDREF`. Dropping a whole `ProteinAmbiguityGroup` is legal because it is `minOccurs="0"` at the `ProteinDetectionList` level.
- **The mzLib seam is clean.** mzLib has no mzID *writer* at all — `MzidIdentifications.cs` only deserializes, and it never touches `ProteinDetection*` / `PeptideHypothesis`. There is no upstream home for this fix and no read-side blast radius. I also confirmed smith-chem-wisc/mzLib#1171: `mzIdentML1_1_0.cs:25-26` really does declare `.../1.1.0` on the root against 65 sibling declarations of `.../1.1`, so your "this alone doesn't make output submittable" caveat is accurate.

I pushed two commits rather than leaving comments, since both were small and I wanted them verified rather than described — sorry for editing your branch directly, and shout if you'd rather I'd left them as review notes.

**`3298aaf` — the SILAC pre-filter.** I dropped `!p.FullSequence.Contains("|")` entirely instead of adding the `p.FullSequence != null` guard. `SpectralMatch.FullSequence` is assigned from `PsmTsvWriter.Resolve(...).ResolvedValue`, and `Resolve` returns `(pipe-joined, null)` when the candidates disagree — the pipe form comes back as `ResolvedString` and is thrown away. So a `SpectralMatch.FullSequence` is never `"|"`-delimited; that conjunct could only ever be `true` or throw, which is the NRE you found. Two consequences for the comment you added: `"|"` doesn't appear on a resolved ambiguous PSM either, and the null guard, while it does fix the crash, narrows `psms` itself — and `psms` also feeds `peptides` / `proteins` / `filenames` a few lines down, so a SILAC search whose PSMs were all ambiguous would have written a file with zero `SpectraData` rather than throwing. Deleting the dead conjunct fixes the crash and leaves the rest of the document alone.

**`f7c020a` — the tests.** The main one: `ProteinGroupSupportedOnlyByAnAmbiguousPsmStillWritesValidMzid` was passing vacuously after the fix. `AssertMinimumChildCounts` only fires an assertion when it *encounters* a PDH or PAG, and post-fix the document has none — so it would have passed just as happily on a file with no protein-detection content whatsoever. It now asserts the shape you actually intend: group gone, `ProteinDetectionList` and `SpectraData` still there.

I also added `DroppedGroupsLeaveGapsButDoNotRenumberSurvivingIds`, because the "pure deletion" compatibility claim is the load-bearing part of the change and nothing pinned it — a dropped group 0 must leave the survivor as `PAG_1`/`PDH_1`, not renumber it to 0. Plus three smaller ones: `Does.Contain("DBS_ACC1")` was already satisfied by the `DBSequence` declaration in `SequenceCollection` (now asserts `dBSequence_ref="DBS_ACC1"`), the SILAC test's lone `Throws.Nothing` couldn't distinguish "NRE fixed" from "PSM silently discarded", and each `Write` moved inside its `try` so a writer failure can't leak a fixed-name `.mzID` into the shared test directory.

All 5 tests in the fixture pass, as do the 16 in `MzIdentMl*` + `SilacTest`, locally on `net10.0-windows`.

Two things I left alone:

- Your note that the control test can't have failed against the unfixed writer is right — it passes on both sides by construction. Worth a one-line edit to the PR body, not to the test.
- The `SequenceCoverageFraction.First()` mis-association you flagged as out of scope: agreed, and it's worse than a wrong number in the multi-protein case, because the writer iterates the unordered `HashSet` while the fractions are filled in `ListOfProteinsOrderedByAccession` order. I filed it with the rest in #2734.

Separately — #2734 covers the pre-existing over-allocation this PR's pattern belongs to: six more arrays in the same method sized from one set and filled from a smaller one, and the hardcoded `SpectrumIdentificationItem[500]` whose fill index is unbounded. Not this PR's problem, but the same defect class.

nbollis (approved):

zhuoxinshi (approved):

PR #2662, opened 2026-05-20
Add MetaDraw mirror plot workflow

nbollis:
## Summary
- Add a new MetaDraw mirror plot tab with dual PSM selection, mirrored spectrum rendering, relative-intensity mode, and full-view export support.
- Extend MetaDraw plotting and settings to support mirrored isotopic envelopes, abbreviated ion labels, transcript-aware descriptions, and bulk descriptor selection.
- Add targeted coverage for mirror plot behavior, settings persistence, export logic, and spectral match comparison.

## Screenshots
- [ ] Mirror Plot tab overview
- [ ] Relative intensity mode enabled
- [ ] Mirror plot export output
- [ ] Settings dialog showing short ion names and descriptor bulk actions

PR #2733, opened 2026-08-16
Make MetaDraw's model layer UI-neutral

acesnik:
🤖 Phase 1 of porting MetaDraw off WPF. Stacked on #2729 (the `GuiFunctions.Core` extraction) — review that first; the diff here includes its commits until it merges.

This moves MetaDraw's settings and model layer into the existing non-WPF assembly and leaves the drawing layer untouched. Purely preparatory — no user-visible behaviour changes.

### Why this is smaller than expected

The earlier estimate for de-WPF-ing `GuiFunctions` was ~17 files and a public API change. For the model layer it came to two type changes:

- `MetaDrawSettings` had **2 WPF-typed members against 115 `OxyColor` references**, and one of them — `ModificationAnnotationColor` — was **dead**: its only appearance repo-wide was its own declaration. Deleted rather than converted.
- `BioPolymerCoverageColors` becomes `Dictionary<BioPolymerCoverageType, OxyColor>`, matching every other colour there. That removes `System.Windows.Media` and with it `OxyPlot.Wpf`, whose only use was a `.ToOxyColor()` hop on a `Brush` during serialisation — now unnecessary. Three call sites convert at the WPF boundary via `DrawnSequence.ParseColorBrushFromOxyColor`, which already existed for exactly this.
- `PlotModelStat` and `ChimeraGroupViewModel` had **zero** WPF usings already.

### What moved

`MetaDrawSettings`, `PlotModelStat` and `ChimeraGroupViewModel` into `GuiFunctions.Core`, plus three files that were already WPF-free and merely mis-located: `MetaDrawSettingsSnapshot`, `PlotModelStatParametersViewModel`, and `LegendDisplayProperty` — a plain enum declared inside `ChimeraLegendCanvas.cs`, which is why the settings layer depended on a `Canvas` subclass.

One real coupling broken: `ChimeraGroupViewModel` reached `MetaDrawSettingsViewModel.Instance` solely for precursor deconvolution parameters, whose own types were already in Core. Now a constructor argument — the WPF call site passes exactly what it read before, and the class becomes unit-testable.

`TypeForwardedTo` for all six moved types. Not optional: `PlotModelStatParametersViewModel` is named in `MetaDraw.xaml` via `x:Static`, and without the forward the markup compiler fails `MC3050`.

### Histogram binning extracted

`SpectrumMatchHistograms.GetHistogramData` returns the bins for all 16 statistics plots as plain data; `PlotModelStat` delegates. This is groundwork for replacing the OxyPlot rendering later — it turns "did the numbers move?" into a question about data rather than about a chart.

**Verified as a pure move rather than assumed:** diffing the relocated binning against the original region line by line comes back identical. Only the signature and visibility changed.

### Tests

**245 cross-platform tests pass on macOS arm64, up from 187.** 17 are new characterisation tests asserting invariants rather than golden numbers — every plot type bins without throwing on real ambiguity-containing input, each bin dictionary accounts for exactly the values collected per source file, bin sizes are pinned where a silent re-binning would be invisible, and pipe-delimited notches split across bins rather than being parsed as one value.

The cross-platform exclusion list goes **26 → 25 files**. That is one file, carrying 41 tests. Fewer than I expected, and worth recording why: the remaining MetaDraw tests are written against the WPF drawing classes — `SpectrumMatchPlot`, `DrawnSequence`, `MetaDrawLogic`, `MetaDrawSettingsViewModel` — so relocating the model layer does not reach them. They come free when the plot tree is reparented and the drawn views are rewritten, not here.

### Not verified locally

`MetaDrawSettingsAndViewsTest` has 204 `MetaDrawSettings` references and is the canary for the colour-type change, but it lives in the Windows-only `Test.csproj`. **Windows CI is the only gate on this PR.** Everything else was verified here: `GuiFunctions.Core` builds on `net10.0` with no WPF, and the solution builds in Release apart from WiX, which cannot run off Windows.

### What comes next

Reparenting the plot tree off `mzPlot.Plot` (which references `PresentationFramework`, and is why the otherwise WPF-free `SpectrumMatchPlot` hierarchy is Windows-bound), then the OxyPlot 2.2.0 move, then rewriting the four hand-drawn views in Avalonia.


acesnik:
🤖 Phase 3 of the MetaDraw cross-platform work is green and ready for human eyes before it merges. @trishorts and others — could you help eyeball the statistics plots on Windows? That's the one gap automation can't close here.

**Why a human is needed.** OxyPlot removed `ColumnSeries` at 2.1.0 and nothing in 2.2 replaces it: `BarSeries` stacks but renders only horizontally, `LinearBarSeries` renders vertically but has no `IsStacked`. The 14 histograms on the statistics tab are now drawn by a `RectangleBarSeries` subclass that does the stacking itself. The values and the orientation are asserted in tests; **a purely cosmetic difference is not**, and I'm not claiming pixel-identical output.

**What to look at** — MetaDraw → statistics tab, all 16 plots:

- bar width and the gaps between bars
- stacking order and colour per source file, including multi-file
- the log-scale y-axis option
- grouped mode (the nested group axis below the category axis)
- hover tooltips (bin / value / total)
- PNG and PDF export from the tab — background should still be white

**What was measured, and what it caught.** Three things `ColumnSeries` did implicitly had to be rebuilt, all found by rendering 2.0.0 output and comparing geometry rather than by reading the API:

- bar width is `BarWidth / (1 + CategoryAxis.GapWidth)` — the `ColumnWidth = 200` in the old code did nothing at all
- a `CategoryAxis` took its `-0.5 … N-0.5` range from the categorized series drawn against it. `RectangleBarSeries` is a plain `XYAxisSeries`, so without reporting whole slots from `UpdateMaxMin`, **every bar rendered 4% off where `ColumnSeries` put it**
- series colour came from `PlotModel.DefaultColors`

Also fixed a tooltip that was reporting the wrong number: the tracker strings used `{2}`, which was `BarSeriesBase`'s third positional argument. Under `RectangleBarSeries` it resolves to something else — printed `3` where the bar's value was `1`. Bound by item property name now.

**Not verified:** legend placement. `PlotModel.LegendTitle` became a `Legends` collection in 2.2 and I haven't compared legend rendering against 2.0 output.

**CI:** 1,771/1,772 on Windows; 264 `GuiFunctions.Core` tests passing on macOS/ubuntu/Windows, up from 252. The new tests were checked against a deliberately broken renderer — rotating the bars fails 7 of them, dropping the stacking fails 1, and the width and axis-range regressions each fail their own.

`Validate Installer Contents` earned its keep here: OxyPlot.Wpf 2.2.0 splits out `OxyPlot.Wpf.Shared.dll`, which the MSI didn't ship, so an installed GUI wouldn't have opened MetaDraw at all.

**What's next.** Phases 1–3 are additive and stop cleanly at this point. Phase 4 is the four hand-drawn views (`DrawnSequence`, `ChimeraDrawnSequence`, `ChimeraLegendCanvas`, `BioPolymerCoverageMapViewModel.Redraw` — ~1,300 lines), which is where the shipping WPF GUI's own drawing code changes. Worth agreeing the approach before I start rather than after.

trishorts (commented):
Reviewed the model-layer move itself as sound: `GuiFunctions.Core` is the right home, `SpectrumMatchHistograms.GetHistogramData` is a promotion of the existing private `PlotModelStat.GetHistogramData` rather than a second binner, and taking the precursor decon parameters by constructor follows `DeconHostViewModel`'s precedent. The `ExportToPng` handling of OxyPlot 2.2 dropping `PngExporter.Background` is correct, and the `Save()` switch from `color.ColorBrush` to `ParseOxyColorFromName(color.SelectedColor)` is behaviour-preserving (the `SelectedColor` setter already derives `ColorBrush` through the same lookup and the same `FallbackColor`).

Six things below. The first is a user-visible regression; the rest are a scope mismatch and four test/coverage gaps that undercut claims the description makes.

One cosmetic note not worth its own thread: the BOM is stripped from `Product.wxs` and `ChimeraLegendCanvas.cs` (present at master, absent here), `ChimeraGroupViewModel.cs` has no trailing newline against `.editorconfig insert_final_newline = true`, and the six new `TypeForwardedTo` lines are appended after `GuiFunctions.XmlReaderWriter` rather than sorted into the existing alphabetical list.

Also worth a heads-up to the authors of #2195 and #2662 before this merges: both have in-flight edits to `GuiFunctions/MetaDraw/MetaDrawSettings.cs`, and #2195 also edits `MetaDrawSettingsSnapshot.cs` and `PlotModelStat.cs` — all relocated here, which conflicts unrecoverably rather than textually.

MetaMorpheus/GuiFunctions.Core/MetaDraw/PlotModelStat.cs
```diff
@@ -137,7 +165,7 @@ private void createPlot(string plotType)
 
         private void histogramPlot(int plotType)
         {
-            privateModel.LegendTitle = "Source file(s)";
+            privateModel.Legends.Add(new OxyPlot.Legends.Legend { LegendTitle = "Source file(s)" });
```
trishorts, line 168: **The two scatter plots lose their legend after the OxyPlot bump.**

At master this was `privateModel.LegendTitle = "Source file(s)"` (`PlotModelStat.cs:140`). OxyPlot 2.0 rendered legends from `PlotModel` itself, with `IsLegendVisible` defaulting to true — so *any* plot with titled series got a key, and `LegendTitle` only added the header. 2.1 moved legends into `PlotModel.Legends`, and a model with an empty `Legends` collection renders nothing at all.

This line is the only `Legends.Add` in the repository, and it is inside `histogramPlot`. `linePlot` — which serves "Precursor PPM Error vs. RT" and "Predicted RT vs. Observed RT" — registers no legend, but still adds two data-less `ScatterSeries` at lines 780 and 791 (`Title = "non-variant PSMs"`, `Title = "variant PSMs"`) whose *only* purpose is the legend key.

So both scatter plots silently lose the blue/dark-red variant key, and those two series become dead objects. Nothing catches it: `StackedColumnSeriesTests` filters on names starting with `"Histogram of"`, which excludes both line plots from every new assertion.

Adding a `Legend` on the `linePlot` path too — or once in the constructor for every plot type — restores the 2.0.0 default.

MetaMorpheus/GuiFunctions.Core/MetaDraw/SpectrumMatchHistograms.cs
```diff
@@ -0,0 +1,259 @@
...
+        public static int RoundToBin(double number, double binSize)
+        {
+            int sign = number < 0 ? -1 : 1;
+            double d = number * sign;
+            double remainder = d % binSize;
+            int i = remainder < 0.5 * binSize ? (int)(d / binSize + 0.001) : (int)(d / binSize + 1.001);
+            return i * sign;
+        }
+
+        public static IEnumerable<double> ParseAmbiguousNotchValues(string notch)
```
trishorts, line 237: **These two helpers were duplicated rather than moved, and only this copy is tested.**

`ParseAmbiguousNotchValues` (here) and `SplitAmbiguousNotch` (line 248) are byte-identical to the private copies still sitting in `PlotModelStat.cs:909` and `:898` — only visibility and member order differ. The description says the helpers were moved; the diff leaves both.

The two copies are on different code paths: the ungrouped notch histogram bins through this public copy, while the *grouped* notch histogram goes through `PlotModelStat.GetNumbersFromPsms` (`PlotModelStat.cs:882`) into the private one.

That matters because `AmbiguousNotchesParseToExpectedValues` pins only this copy. The exact regression commit `6b17615` fixed — a dropped `DefaultIfEmpty("0")` making `"|"` yield no values — can reappear in the private copy with every test still green, silently dropping PSMs from the grouped notch histogram only.

Worth noting `NormalizeAmbiguityLevel` and `RoundToBin` *were* de-duplicated in the same commit (`PlotModelStat.cs:390` and `:513` now call the `SpectrumMatchHistograms` versions), so the notch pair looks like an oversight rather than a decision.

MetaMorpheus/Test/MetaDraw/SpectrumMatchHistogramsTests.cs
```diff
@@ -0,0 +1,126 @@
...
+
+    /// <summary>
+    /// The bin dictionaries must account for exactly the values collected, per source file. This is the
+    /// invariant a renderer change could break silently.
+    /// </summary>
+    [TestCase(1)]
+    [TestCase(3)]
+    [TestCase(6)]
+    [TestCase(7)]
+    [TestCase(12)]
```
trishorts, line 49: **`TestCase(12)` asserts nothing.**

`case 12` in `SpectrumMatchHistograms.cs:173-184` populates only `dictsBySourceFile`; it never writes to `numbersBySourceFile`. So for `plotType == 12` the `foreach (var (file, numbers) in data.NumbersBySourceFile)` at line 55 iterates an empty dictionary and the body never executes — the case passes unconditionally.

`case 13` has the same shape, which is presumably why it isn't in the list. Either drop `TestCase(12)`, or give plot 12 a bin-total assertion that works off `DictsBySourceFile` alone (e.g. the summed bin counts equal the PSM count for that file), which is the invariant you actually want pinned for the retention-time histogram.

MetaMorpheus/GuiFunctions.Core/GuiFunctions.Core.csproj
```diff
@@ -0,0 +1,23 @@
...
+         cannot be loaded by an arm64 runtime (#2503, #2504), and this project exists to be loadable
+         anywhere. Leave it unset rather than setting AnyCPU, which Microsoft.ML rejects at build. -->
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\TaskLayer\TaskLayer.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="OxyPlot.Core" Version="2.2.0" />
```
trishorts, line 20: **The description files this bump — and two other substantial changes — under "What comes next", but they are in this diff.**

The body says the PR is "purely preparatory — no user-visible behaviour changes", and lists reparenting the plot tree off `mzPlot.Plot` and "the OxyPlot 2.2.0 move" as future work. Both are here, along with:

- `OxyPlot.Core` / `OxyPlot.Wpf` **2.0.0 → 2.2.0** across four csprojs
- the new 104-line `StackedColumnSeries.cs`, a hand-written `RectangleBarSeries` subclass reimplementing stacked vertical bars because 2.1 removed `ColumnSeries` — manually reproducing `BarWidth / (1 + GapWidth)` sizing and the `CategoryAxis` −0.5/N−0.5 range
- a new `src_OxyPlot.Wpf.Shared.dll` installer component
- commits `0ed17b0` (reparent off `mzPlot.Plot`) and `5ac943c` (OxyPlot 2.2.0)

That is a rendering rewrite of all 16 statistics plots riding along with a model-layer move, and the legend regression I flagged on `PlotModelStat.cs:168` is a direct consequence of it.

No objection to the work — the `StackedColumnSeries` comments explain the constraint well, and the `PngExporter.Background` handling is right. But it wants either its own PR or a description that says what is actually in this one, because "no user-visible behaviour changes" is what tells a reviewer how hard to look at the plots. Windows CI being the only gate here makes that more load-bearing, not less.

MetaMorpheus/GuiFunctions.Core.Tests/GuiFunctions.Core.Tests.csproj
```diff
@@ -0,0 +1,70 @@
...
+    <Compile Remove="..\Test\GuiTests\RelayAndDelegateCommandTests.cs" />
+    <Compile Remove="..\Test\MetaDraw\BioPolymerCoverageColorMapperFactoryTests.cs" />
+    <Compile Remove="..\Test\MetaDraw\BioPolymerCoverageColorMapperTests.cs" />
+    <Compile Remove="..\Test\MetaDraw\BioPolymerCoverageColorScaleTests.cs" />
+    <Compile Remove="..\Test\MetaDraw\BioPolymerCoverageMapViewModelTests.cs" />
+    <Compile Remove="..\Test\MetaDraw\BioPolymerGroupViewModelTests.cs" />
+    <Compile Remove="..\Test\MetaDraw\BioPolymerTabViewModelTests.cs" />
+    <Compile Remove="..\Test\MetaDraw\ChimeraAnalysisTabViewModelTests.cs" />
+    <Compile Remove="..\Test\MetaDraw\ChimeraDrawnSequenceTests.cs" />
+    <Compile Remove="..\Test\MetaDraw\ChimeraGroupViewModelTests.cs" />
```
trishorts, line 49: **The stated payoff for the constructor change isn't collected here.**

Making `ChimeraGroupViewModel` take the precursor decon parameters by constructor is the right call, and it matches how `DeconHostViewModel` and `ClassicDeconParamsViewModel` already work. But `ChimeraGroupViewModelTests.cs` is still on this `Compile Remove` list, so the class gains no cross-platform coverage in this PR — and the new fixture in that file is written as though it had:

> `// ... but without touching the WPF view model, so this fixture runs off Windows.`

It doesn't run off Windows; this line is why.

Either drop it from the Remove list and find out what still blocks it, or soften that comment so it doesn't claim coverage the build doesn't produce.

Small related note: the repo's ctor-injection style elsewhere is optional-with-default — `DeconHostViewModel.cs:31`, `ChimeraAnalysisTabViewModel.cs:83`, and `PlotModelStat.cs:60`'s `?? PlotModelStatParametersViewModel.Instance.GetParameters()`. Making this parameter required is why every call site had to change; an optional parameter would have got the same testability without the churn.

MetaMorpheus/Test/MetaDraw/StackedColumnSeriesTests.cs
```diff
@@ -0,0 +1,322 @@
...
+
+            // The axes hold the transforms the render pass just used, so this is the geometry
+            // OxyPlot actually drew, not a restatement of the item values.
+            var xAxis = plot.Model.Axes.First(a => a.Position == AxisPosition.Bottom);
+            var yAxis = plot.Model.Axes.First(a => a.Position == AxisPosition.Left);
+
+            // Bottom-most series only: with stacking, upper segments legitimately float.
+            var items = SeriesOf(plot).First().Items.OfType<StackedColumnItem>()
+                .Where(i => i.Y1 > i.Y0).ToList();
+            if (items.Count < 2)
```
trishorts, line 114: **This test can pass having asserted nothing.**

The skip itself is fair and well-explained — a single-bar histogram genuinely can't distinguish vertical from rotated. The gap is that nothing records whether *any* plot cleared it. If every name in `HistogramPlotNames` yielded fewer than two positive items, the loop would `continue` through all of them and the test would report green without evaluating a single assertion.

That is the failure mode this test exists to catch, so it is worth a guard: count the plots that get past line 114 and assert the count is greater than zero at the end.

Same shape one method up, for what it's worth: `AmbiguousNotchesAreSplitAcrossBins` asserts `binned >= totalPsms`. With splitting removed entirely each PSM contributes exactly one value, `binned == totalPsms`, and the assertion still passes — so it catches values being *dropped* (the `6b17615` regression) but not splitting being *lost*, which is what the doc comment claims. When some PSM has a pipe-delimited notch, `Is.GreaterThan` is the assertion that distinguishes the two.

MetaMorpheus/MetaMorpheus.sln
```diff
@@ -22,58 +22,200 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Test", "Test\Test.csproj",
...
 Project("{B7DD6F7E-DEF8-4E67-B5B7-07EF123DB6F0}") = "Bootstrapper", "Bootstrapper\Bootstrapper.wixproj", "{E0EA5AC4-24A9-43DC-8FBC-CCEB3B9935B6}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "GuiFunctions.Core", "GuiFunctions.Core\GuiFunctions.Core.csproj", "{7B63C758-BA12-4BDB-879F-DBD8AF82A959}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "GuiFunctions.Core.Tests", "GuiFunctions.Core.Tests\GuiFunctions.Core.Tests.csproj", "{9AB2EDBA-4D49-4AD2-B9B3-5B5410DE05AA}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
```
trishorts: **This carries the same six extra solution platforms that are still open on #2729, and fixing them there will not fix them here.**

Both branches take `SolutionConfigurationPlatforms` from master's three entries to nine, adding `x64` and `x86` across Debug, Release and UbuntuMac. I checked both heads directly and they are identical on this point.

Because the two branches are not actually stacked in git (merge-base `36e34ef1` is behind #2729's current tip), whatever you do about it on #2729 has to be done again here -- or this branch has to be rebased onto #2729 so it inherits the fix. The second is much the better option, and is the same rebase that would clear the merge conflict with master.

Flagging it here mainly so the thread on #2729 is not treated as covering both. If the fix lands there and this merges afterwards, the platforms come straight back.

trishorts (commented):
Revisiting this alongside #2729, because the relationship between the two branches turns out to matter and is not stated anywhere.

**These are stacked, but not in a way git will maintain.** `git merge-base` on the two heads gives `36e34ef1` -- which is a commit on #2729's branch, not on master. So this branch was cut from #2729 partway along, and carries its `GuiFunctions.Core` extraction wholesale: 34 files are byte-identical between the two PRs, including the new project, `AssemblyInfo.cs`, `GuiFunctions.Core.Tests.csproj`, `Test.yml` and `Product.wxs`. But #2729 has moved on since (`8f4798ae`), so `git merge-base --is-ancestor` is false and nothing propagates between them.

Three consequences:

1. **A reviewer reads the extraction twice**, and cannot tell from either diff which PR owns it. Worth saying in this description that the `GuiFunctions.Core` half belongs to #2729 and the reviewable payload here is the MetaDraw model layer on top of it.
2. **This branch is `CONFLICTING` with master** while #2729 is mergeable. Rebasing this onto #2729's current tip (rather than onto master) would resolve both that and the duplication.
3. **Fixes made to #2729 in response to review will not arrive here.** The concrete instance is inline below.

The model-layer move itself I still think is sound, for the reasons in my earlier review -- `SpectrumMatchHistograms.GetHistogramData` being a promotion of the existing private binner rather than a second one is the part that matters, and it holds up.

MetaMorpheus/GUI/GUI.csproj
```diff
@@ -65,8 +65,8 @@
     <PackageReference Include="mzLib" Version="1.0.584" />
     <PackageReference Include="Nett" Version="0.15.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
-    <PackageReference Include="OxyPlot.Core" Version="2.0.0" />
-    <PackageReference Include="OxyPlot.Wpf" Version="2.0.0" />
+    <PackageReference Include="OxyPlot.Core" Version="2.2.0" />
```
trishorts, line 68: This is the OxyPlot 2.2.0 upgrade, which the description lists under **What comes next** rather than as part of this PR.

It is load-bearing for the rest of the diff: 2.1 removed `ColumnSeries`, which is why `StackedColumnSeries.cs` had to be written and why all 16 statistics plots now render through hand-computed geometry. It also pulls in the new `OxyPlot.Wpf.Shared.dll` installer component below.

Not objecting to the upgrade — objecting to it arriving inside a PR described as "purely preparatory, no user-visible behaviour changes", when the gate on it is Windows CI plus a canary test the description says was not run locally.

MetaMorpheus/GuiFunctions.Core/GuiFunctions.Core.csproj
trishorts, line 20: Same 2.2.0 bump here. Flagging one consequence that is easy to miss because nothing in this repo shows it:

mzLib's `mzPlot` is built against OxyPlot.Wpf **2.0.0** and instantiates `ColumnSeries` at `mzPlot/Plots/Plot.cs:250`. NuGet unifies on the highest version, so the `mzPlot.dll` that ships with MetaMorpheus is now bound against a 2.2.0 without that type.

This PR happens to remove the last MetaMorpheus caller of `mzPlot`, so nothing loads it and nothing breaks — but `Product.wxs:437` still installs it. Worth either dropping that component or filing an mzLib issue to move `mzPlot` forward.

MetaMorpheus/GuiFunctions.Core/MetaDraw/StackedColumnSeries.cs
```diff
@@ -0,0 +1,104 @@
+using System;
+using System.Collections.Generic;
+using OxyPlot;
+using OxyPlot.Series;
+
+namespace GuiFunctions
+{
+    /// <summary>
```
trishorts, line 8: 104 lines re-deriving what `ColumnSeries` did before OxyPlot 2.1 removed it — stacking via a shared accumulator, bar width as `BarWidth / (1 + GapWidth)`, and category-axis range widened by hand in `UpdateMaxMin`.

The reasoning is sound and the comments are unusually good. My concern is placement rather than correctness: this is the piece that makes the PR a rendering change to all 16 statistics plots, and the description frames the PR as purely preparatory with the OxyPlot move still to come.

One real coupling worth calling out in the class docs: `GapWidth` must be set before `AddItem` **and** must match the `CategoryAxis` the series is drawn against, or every bar and gridline shifts. Nothing at the call site enforces that, and a wrong value produces a plot that looks plausible rather than one that throws.

MetaMorpheus/MetaMorpheusSetup/Product.wxs
```diff
@@ -1,4 +1,4 @@
-﻿<Wix xmlns="http://wixtoolset.org/schemas/v4/wxs" xmlns:ui="http://wixtoolset.org/schemas/v4/wxs/ui">
+<Wix xmlns="http://wixtoolset.org/schemas/v4/wxs" xmlns:ui="http://wixtoolset.org/schemas/v4/wxs/ui">
```
trishorts: The UTF-8 BOM was stripped from this file in this diff (`-\ufeff<Wix ...` to `<Wix ...`). Looks incidental rather than intended — worth restoring so an installer file does not carry an unrelated encoding change.

trishorts (commented):
The model-layer extraction is good work and the mechanical parts hold up: `SpectrumMatchHistograms.GetHistogramData` really is a pure move (I diffed the relocated binning against the original region and it comes back identical), deleting the dead `ModificationAnnotationColor` is right, and breaking `ChimeraGroupViewModel`'s reach into `MetaDrawSettingsViewModel.Instance` in favour of a constructor argument is a genuine improvement that makes the class testable.

My concern is not with the code, it is that **the description and the diff describe different PRs**, and the gap is in the direction that matters for review.

## This PR performs the OxyPlot 2.2.0 upgrade

The description puts *"then the OxyPlot 2.2.0 move"* under **What comes next**, and opens with *"Purely preparatory — no user-visible behaviour changes."* But the diff bumps all three projects from 2.0.0 to 2.2.0:

- `GUI/GUI.csproj` — `OxyPlot.Core` and `OxyPlot.Wpf`
- `GuiFunctions/GuiFunctions.csproj` — `OxyPlot.Wpf`
- `GuiFunctions.Core/GuiFunctions.Core.csproj` — `OxyPlot.Core`

and adds a new installer component for `OxyPlot.Wpf.Shared.dll`, which only exists because 2.2.0 split it out.

The upgrade is not incidental to the move — it forces the rest of it. `ColumnSeries` was removed in OxyPlot 2.1, which is why `StackedColumnSeries.cs` exists: 104 new lines re-deriving stacking, bar width (`BarWidth / (1 + GapWidth)`) and category-axis range by hand, on top of `RectangleBarSeries`. Every one of the 16 MetaDraw statistics plots now renders through that instead of through OxyPlot's own series.

That is a rendering change to every statistics plot, and it is a reasonable one — but it is not "purely preparatory", and it should not be reviewed as though it were. Two consequences:

- The bar geometry is now our arithmetic rather than the library's. `GapWidth` has to be set before `AddItem` and has to agree with the `CategoryAxis` it is drawn against, or every bar and gridline shifts — a coupling that is invisible at the call site.
- The description says `MetaDrawSettingsAndViewsTest` (204 `MetaDrawSettings` references, the canary for the colour-type change) was **not verified locally** and that Windows CI is the only gate. With a plot-rendering rewrite in the same diff, that is a thin gate for the amount that changed.

I would split this: the model-layer move on its own (which is what the description promises and is easy to approve), then the OxyPlot upgrade plus `StackedColumnSeries` as its own PR with the visual before/after. If it stays as one PR, the description needs to say what it actually does.

## Cross-repo seam: this leaves a live-but-unloaded `mzPlot.dll` in the install

Worth recording because it is the kind of thing that surfaces a year later. mzLib's `mzPlot` is compiled against **OxyPlot.Wpf 2.0.0** and instantiates `ColumnSeries` at `mzPlot/Plots/Plot.cs:250`. NuGet resolves to the highest requested version, so after this PR the shipped `mzPlot.dll` is bound against a 2.2.0 that no longer has that type — its bar-series path is a `TypeLoadException` waiting to happen.

It is latent only because this PR removes MetaMorpheus's last real use of `mzPlot` (every remaining reference is a comment). But `Product.wxs:437` still installs `mzPlot.dll`. Either drop that component here, or open an mzLib issue to move `mzPlot` to 2.2.0 — otherwise the next person to call into it gets a runtime failure a long way from the cause.

Related, and not a blocker: the new `PlotBase` is a near-duplicate of mzLib's `mzPlot.Plot` (`public abstract class Plot : INotifyPropertyChanged`, `RefreshChart()` at `Plot.cs:330`). Reimplementing locally is defensible — the mzLib one takes an `OxyPlot.Wpf.PlotView` in its constructor and cannot be inherited without WPF — but the durable fix is splitting `mzPlot` upstream into a UI-neutral core plus a WPF layer, which would serve the Avalonia work in #2709 too. Follow-up, not this PR.

## Trivial

`Product.wxs` lost its UTF-8 BOM in this diff. Unrelated to the change and worth restoring so the installer file does not pick up an incidental encoding change.

Fixes for the BOM and the mzPlot component are in a branch — see the follow-up comment. The scope question is yours to call.

trishorts:
@acesnik — **acesnik/MetaMorpheus#11** restores the `Product.wxs` BOM against your branch. That is the only part of my review a patch can settle.

The other two are yours to call: whether the OxyPlot 2.2.0 upgrade stays in this PR (and if so, that the description says so), and whether to drop the `mzPlot.dll` installer component now that nothing loads it. I deliberately did not touch the second one — I would rather you confirm nothing else reaches for it than have me guess.

trishorts:
@acesnik — this PR is `CONFLICTING`, so **no checks are running on it at all**. **acesnik/MetaMorpheus#14** unblocks it.

Three things in it:

1. **Conflict resolved** — master's `Newtonsoft.Json 13.0.4` and `mzLib 1.0.588`, this branch's `OxyPlot 2.2.0`.
2. **Caught up to #2729.** This branch forked before `acesnik/MetaMorpheus#10`, so it still carried the nine-platform solution matrix and the missing `Category!=ExternalService` filter. Merging your current #2729 head brings both fixes across.
3. The same `MetaDrawExportRegressionTests` glob exclusion as #13.

Verified locally, since CI cannot speak on this PR yet: GUI builds Release on OxyPlot 2.2.0, 273 cross-platform tests pass, and **the MetaDraw suite passes 588 tests** — including `MetaDrawSettingsAndViewsTest`, the canary the description says was never verified locally. It is green with the OxyPlot rewrite in place.

The two judgement calls from the review are untouched and still yours: whether the OxyPlot 2.2.0 upgrade belongs in this PR, and whether to drop the `mzPlot.dll` installer component.

Issue #2454, opened 2024-12-17
Crash associated with issue #2400


Dmorgen:
Hi,
I've used the patched version you've sent me. It crashed during the quantification stage. 
please see the attached log.

Thanks!
David.

[results_171224.txt](https://github.com/user-attachments/files/18165126/results_171224.txt)

trishorts:
were you doing tmt?

Dmorgen:
No. did I mark that by any chance?

trishorts:
first line of the error message reference line 209. 
![image](https://github.com/user-attachments/assets/e101e6b7-ec54-408c-966f-1d1a6b2a4694)
can you check the toml?

Dmorgen:
No. But I will make sure the next attempt won’t have it. Thanks!

D.

trishorts:
can you post the search toml for me

trishorts:
i ran all your data through to quant on my side w/ no crash.

Dmorgen:
Thanks! Will try again.

D.

acesnik:
🤖 Thanks for posting this, David, and sorry it took a while to circle back!

Closing this one — we could never reproduce it (your data ran through to quant cleanly on our side, and we never got the search TOML to compare settings), and the crash you hit can no longer stop a run: #2574 wrapped quantification in crash handling, so a failure there now writes a `Quantification_crash.txt` and lets the rest of the search finish.

We did find the mechanism, and there's a fix in flight. `QuantificationAnalysis` dereferenced a nullable monoisotopic mass without checking it, and the filter ahead of it only screens on sequence — a resolved sequence doesn't imply a resolved mass. Two ways that happens: `Residue.ResidueMonoisotopicMass` is `NaN` for B, X and Z, and every NaN comparison is false, so even a completely unambiguous match came out null; or two matches whose modifications share a name but not a mass, which stringify to the same `FullSequence`. Either one used to cost quantification for the whole run rather than just the affected PSM. The fix drops those PSMs with a warning naming the count.

Your build's `PostSearchAnalysisTask.cs:209` was a red herring, incidentally — it was a custom patch, and line 209 in the tree at the time is a plain `bool` check, which is why the TMT question didn't lead anywhere.

Happy to reopen if you see anything like this on a current release. Thanks again for the detailed logs!

PR #2735, opened 2026-08-18
Skip PSMs whose monoisotopic mass did not resolve instead of throwing out of quantification

acesnik:
## Problem

`QuantificationAnalysis` dereferences `psm.BioPolymerWithSetModsMonoisotopicMass.Value` without checking `HasValue` (`PostSearchAnalysisTask.cs:563`). The filter ahead of it screens on `BaseSequence`/`FullSequence` emptiness (`FilteredPsms.cs:117-118`), and **a resolved sequence does not imply a resolved mass** — so a null-mass PSM reaches the deref and throws `InvalidOperationException: Nullable object must have a value`.

Since #2574 that throw is caught by `EngineCrashed`, so the run survives. But it writes `Quantification_crash.txt` and leaves `FlashLfqResults` null — **one PSM discards quantification for every file in the run**, while PSM, peptide and protein output are still written, so it reads as a successful search that simply produced no quant.

## How the mass ends up null

`BioPolymerWithSetModsMonoisotopicMass` comes from `PsmTsvWriter.Resolve`, which returns null when `max - min >= 1e-6`. Two ways that happens with the sequence still resolved:

- **`NaN` residue mass.** `Residue.ResidueMonoisotopicMass` is `double.NaN` for `B`, `X` and `Z`. Every NaN comparison is false, so `max - min < tolerance` fails and **even a single, entirely unambiguous match resolves to null**.
- **Two best matches whose modifications share `ModificationType` and `IdWithMotif` but not a mass.** `DetermineFullSequence` builds the string from those two fields alone, so both peptides stringify identically while their masses differ.

On the second route: `AddMods` already rejects this from a mods file (`"Modification id and type are equal, but some fields are not!"`), so it cannot arise there. It *can* arise from protein XML — that branch does `RemoveAll` + `Add` with no error raised, and `LoadProteinDb` loads mods and builds proteins per file, so a database loaded earlier keeps referencing the mod object a later database replaced. Two XML databases in one run that disagree about a mod's mass is therefore enough; neither file need be malformed. Verified at the mod-loading level, not end-to-end through a two-database search.

Note ~7 other consumers of this property already guard with `.HasValue` (`NonSpecificEnzymeSearchEngine`, `BinTreeStructure`, `GlycoSpectralMatch`, `CrosslinkSpectralMatch`); this call site and `DataPointAquisitionResults.cs:48` do not.

## Change

Drop unresolved-mass PSMs from the quantification set and warn with the count, via `FilteredPsms.RemovePsmsWithoutResolvedMass()` — mutation kept in the class that owns the list, alongside the existing `SetSilacFilteredPsms`. The `try/catch` is left in place as the backstop for genuinely unforeseen quantification failures.

Dropping rather than picking one of the tied candidates is deliberate: for the `NaN` route there is nothing to pick, and for the collision route a silently chosen mass would hide a database conflict the user should hear about. Warning at the point the mods actually collide is the better fix for that route and is left for separate work.

## Output impact

For a run that hits this, quantification goes from **entirely absent** to **present minus the affected PSMs**, plus a warning naming how many were dropped. Runs that never hit it are unchanged.

## Tests

Three tests in `PostSearchAnalysisTaskTests.cs` — one per null-mass route, plus a regression that invokes `QuantificationAnalysis` and asserts no `Quantification_crash.txt`, a warning, and non-null `FlashLfqResults`.

The regression test was confirmed to fail against the unfixed call site (`Quantification_crash.txt` present, `FlashLfqResults` null) and pass with the fix. Because `Test.csproj` is `net10.0-windows` and cannot run on macOS, these were executed via a throwaway `net10.0` harness compiling the committed methods verbatim; `Test.csproj` itself compiles clean with `EnableWindowsTargeting`. **The full Windows suite has not been run locally — CI is the real check.**

Relates to #2454.


MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
```diff
@@ -333,6 +333,15 @@ private void QuantificationAnalysis()
                     includeAmbiguousMods: false,
                     includeHighQValuePsms: false);
 
+                // FlashLFQ needs a mass. The filter above screens on sequence, which does not imply a
+                // resolved mass: a NaN residue mass, or two best matches whose mods share an IdWithMotif
+                // but not a mass, both leave BioPolymerWithSetModsMonoisotopicMass null.
+                int psmsWithoutMass = psmsForQuantification.RemovePsmsWithoutResolvedMass();
```
trishorts, line 336: **Minor** - New mass guard runs before SILAC replaces the whole PSM list, so it misses that path

**Problem:**
`psmsForQuantification.RemovePsmsWithoutResolvedMass()` is called at line 339, immediately after `FilteredPsms.Filter(...)`. But at line 529 the SILAC block calls `psmsForQuantification.SetSilacFilteredPsms(silacPsms)`, which replaces `FilteredPsmsList` wholesale with PSMs that did not exist when the guard ran — `SilacConversions.GetLabeledPsm` builds a brand-new `PeptideWithSetModifications` from a synthesised labeled base sequence, and `GetSilacPsm` rebuilds every hypothesis via `CreateSilacPwsm`. The very next line, `silacPsms.ForEach(x => x.ResolveAllAmbiguities()); //update the monoisotopic mass`, recomputes `BioPolymerWithSetModsMonoisotopicMass` for those new objects, so the null-mass property is re-derived after the guard has already run. The deref this PR exists to protect (`psm.BioPolymerWithSetModsMonoisotopicMass.Value`, line 572) then reads the un-guarded list.

**Impact:**
For a SILAC + label-free run, a labeled or light-converted PSM whose re-resolved mass comes back null still throws out of `QuantificationAnalysis`, is caught by `EngineCrashed`, and reproduces exactly the whole-run quantification loss the PR set out to eliminate — while the PR body states the fix applies to `QuantificationAnalysis` generally.

**Suggested fix:**
Either move the `RemovePsmsWithoutResolvedMass()` call (and its warning) to after the SILAC block finishes replacing the list — i.e. just before `psmsGroupedByFile` is built — or repeat it after `SetSilacFilteredPsms`, summing the counts into a single warning. Placing it once at the point where the list is final covers both paths and keeps the guarantee adjacent to the deref it protects.

**Affected test scenarios:**
- QuantificationAnalysis - SILAC labels configured, one silac-derived PSM whose re-resolved mass is null - no `Quantification_crash.txt`, warning emitted, `FlashLfqResults` non-null
- QuantificationAnalysis - SILAC labels configured, all masses resolve - identical identification count and output to the pre-PR run
- RemovePsmsWithoutResolvedMass - called after `SetSilacFilteredPsms` replaced the list - operates on the silac list, not the original

MetaMorpheus/Test/PostSearchAnalysisTaskTests.cs
```diff
@@ -636,5 +643,202 @@ public static void WriteDigestionCountFiles_IncludesDecoys_WhenIntended()
...
+                    CommonParameters = commonParameters,
+                    FileSpecificParameters = new List<(string, CommonParameters)> { (mzmlPath, commonParameters) },
+                };
+
+                typeof(PostSearchAnalysisTask)
+                    .GetMethod("QuantificationAnalysis", BindingFlags.NonPublic | BindingFlags.Instance)!
+                    .Invoke(task, null);
+
+                Assert.That(File.Exists(Path.Combine(outputFolder, "Quantification_crash.txt")), Is.False);
+                Assert.That(warnings.Any(w => w.Contains("monoisotopic mass could not be determined")), Is.True);
```
trishorts: **Minor** - Dropped-PSM count is never asserted anywhere in the new tests

**Problem:**
`RemovePsmsWithoutResolvedMass()` returns the number of PSMs it dropped, and `QuantificationAnalysis` interpolates that number into the warning (`PostSearchAnalysisTask.cs:339-343`). The only assertion on the warning is `warnings.Any(w => w.Contains("monoisotopic mass could not be determined"))`, which matches the tail of the message and is insensitive to the number in front of it. No other test observes the return value at all, so the count half of the new method's contract is untested.

**Impact:**
A return value of `FilteredPsmsList.Count` (survivors) instead of the `RemoveAll` result, or any later refactor that miscounts, would put a wrong number in front of every user reading `1 PSM(s) were excluded…` and the suite would stay green.

**Suggested fix:**
Assert the full expected warning string for the one-dropped fixture (e.g. that a warning starting with `1 PSM(s)` fired), and/or assert the `int` returned by `RemovePsmsWithoutResolvedMass()` directly in a unit test over a hand-built list.

**Affected test scenarios:**
- QuantificationSkipsPsmsWithUnresolvedMass - one massless PSM, one quantifiable PSM - warning text names exactly 1 excluded PSM
- RemovePsmsWithoutResolvedMass - list with 2 null-mass and 3 mass-bearing PSMs - returns 2

MetaMorpheus/Test/PostSearchAnalysisTaskTests.cs
```diff
@@ -636,5 +643,202 @@ public static void WriteDigestionCountFiles_IncludesDecoys_WhenIntended()
...
+
+            Assert.That(psm.BaseSequence, Is.Not.Null.And.Not.Empty);
+            Assert.That(psm.FullSequence, Is.Not.Null.And.Not.Empty);
+            Assert.That(psm.BioPolymerWithSetModsMonoisotopicMass, Is.Null);
+
+            var survivors = FilteredPsms.Filter(new List<SpectralMatch> { psm }, commonParameters,
+                includeDecoys: false, includeContaminants: true, includeAmbiguous: false,
+                includeAmbiguousMods: false, includeHighQValuePsms: false);
+
+            Assert.That(survivors.Count(), Is.EqualTo(1));
```
trishorts, line 676: **Minor** - New public method has no direct test; route tests stop before invoking it

**Problem:**
`ModMassCollisionLeavesMassUnresolvedButPassesTheQuantFilter` builds exactly the fixture the new method exists to handle, then ends at `Assert.That(survivors.Count(), Is.EqualTo(1))` — it never calls `survivors.RemovePsmsWithoutResolvedMass()`. `AmbiguousResidueLeavesMassUnresolved` never constructs a `FilteredPsms` at all. Nothing asserts the complementary half of the predicate either: that PSMs which *do* have a mass are retained. The new public API is therefore only reachable through the reflection-driven `QuantificationSkipsPsmsWithUnresolvedMass` integration test.

**Impact:**
Delete the body of `RemovePsmsWithoutResolvedMass` (return 0, remove nothing) and both route tests still pass; only the heavyweight integration test — the one whose FlashLFQ dependency makes it most likely to be quarantined or rewritten — catches the regression. A predicate that over-removed (e.g. also dropping decoys, which `MatchBetweenRuns` searches need per `PostSearchAnalysisTask.cs:330`) would also go unnoticed, since the fixture has no decoy and no second mass-bearing PSM.

**Suggested fix:**
Extend the two route tests with a call to `RemovePsmsWithoutResolvedMass()` on the `FilteredPsms` they already build, asserting the return value and the emptied/retained list. Add one small unit test over a mixed list (null-mass, mass-bearing, decoy) asserting exactly which entries survive.

**Affected test scenarios:**
- ModMassCollisionLeavesMassUnresolvedButPassesTheQuantFilter - collided-mod PSM after Filter - RemovePsmsWithoutResolvedMass returns 1 and leaves the list empty
- RemovePsmsWithoutResolvedMass - mixed list of null-mass, mass-bearing and decoy PSMs - only the null-mass entries are removed

trishorts (commented):

MetaMorpheus/Test/PostSearchAnalysisTaskTests.cs
```diff
@@ -636,5 +643,202 @@ public static void WriteDigestionCountFiles_IncludesDecoys_WhenIntended()
...
+            Assert.That(psm.BioPolymerWithSetModsMonoisotopicMass, Is.Null);
+        }
+
+        /// <summary>
+        /// One PSM without a resolvable mass used to throw out of QuantificationAnalysis, which
+        /// EngineCrashed turns into a Quantification_crash.txt and a null FlashLfqResults -- losing
+        /// quantification for the whole run. It should cost only the offending PSM, plus a warning.
+        /// </summary>
+        [Test]
+        public static void QuantificationSkipsPsmsWithUnresolvedMass()
```
trishorts, line 796: **Minor** - No test covers the unaffected path (nothing dropped, no warning emitted)

**Problem:**
Every new test feeds at least one null-mass PSM. Nothing asserts the PR's other stated behaviour — that a run with no null-mass PSM drops nothing and emits no warning (`if (psmsWithoutMass > 0)` at `PostSearchAnalysisTask.cs:340`). The existing end-to-end fixtures in this file assert line offsets in `allResults.txt`, which is not where `Warn` output lands, so they would not notice a spurious warning either.

**Impact:**
An inverted or `>=`-style guard, or a method returning a nonzero count when nothing was removed, would print `0 PSM(s) were excluded from quantification…` on every single search MetaMorpheus runs, and no test in the suite would fail. That is a user-visible regression on the runs the PR promises are unchanged.

**Suggested fix:**
Add a case (or a second invocation of the existing regression test with only the quantifiable PSM) asserting that no warning containing "monoisotopic mass" fires and that the filtered PSM count is unchanged.

**Affected test scenarios:**
- QuantificationAnalysis - all PSMs have resolved masses - no "monoisotopic mass" warning is raised and every PSM reaches FlashLFQ

MetaMorpheus/Test/PostSearchAnalysisTaskTests.cs
```diff
@@ -636,5 +643,202 @@ public static void WriteDigestionCountFiles_IncludesDecoys_WhenIntended()
...
+            masslessPsm.ResolveAllAmbiguities();
+            masslessPsm.SetFdrValues(1, 0, 0.0, 1, 0, 0.0, 0, 0.0);
+
+            Assert.That(quantifiablePsm.BioPolymerWithSetModsMonoisotopicMass, Is.Not.Null);
+            Assert.That(masslessPsm.BioPolymerWithSetModsMonoisotopicMass, Is.Null);
+
+            var dataFile = new GenericMsDataFile(
+                new[]
+                {
+                    new MsDataScan(new MzSpectrum(new[] { 500.0 }, new[] { 1e6 }, false), 1, 1, true,
```
trishorts: **Minor** - Synthetic mzML has no peak for the surviving PSM, so quant is never shown to work

**Problem:**
The fixture mzML holds a single MS1 peak at m/z 500.0, while the surviving identification is `PEPTIDER` (monoisotopic ≈ 955.46, i.e. ≈ 478.74 at the charge 2 taken from `NullMassTestScan`). No peak can match at the 5 ppm `QuantifyPpmTol`, so FlashLFQ finds nothing and every intensity in `FlashLfqResults` is zero. `PeptideModifiedSequences` is populated from the identification list regardless, so `Count == 1` (line 791) asserts only that one `Identification` was constructed — not that quantification produced anything.

**Impact:**
The PR's stated output change is "quantification goes from entirely absent to present minus the affected PSMs". The test verifies the first half (no crash, non-null results object) but cannot distinguish working quantification from quantification that silently yields zeros for every surviving PSM.

**Suggested fix:**
Give the MS1 scan a peak (ideally a small isotopic envelope) at the surviving peptide's theoretical m/z at the charge the PSM reports, and assert a non-zero intensity for that peptide in `FlashLfqResults`.

**Affected test scenarios:**
- QuantificationSkipsPsmsWithUnresolvedMass - MS1 contains the surviving peptide's precursor envelope - that peptide's quantified intensity is greater than zero

nbollis (approved):

nbollis (dismissed):

MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
trishorts: Fixed in `1c88650`. Screened again immediately after `SetSilacFilteredPsms`, with its own warning, rather than moving the original guard — the first call still has to run for the non-SILAC path.

Worth noting `silacPsms.ForEach(x => x.ResolveAllAmbiguities())` on the line above does not close this: resolving can still leave the mass null, which is the whole premise of the PR.

MetaMorpheus/Test/PostSearchAnalysisTaskTests.cs
trishorts: Fixed in `1c88650`, and this one turned out to be the substantive finding.

Adding a peak at the right m/z was not enough — the intensity stayed at zero, because FlashLFQ accepts a feature only when it can build an isotopic envelope, and the fixture had a single monoisotopic peak. It now carries the peptide’s real distribution:

```csharp
ChemicalFormula quantifiableFormula = new Proteomics.AminoAcidPolymer.Peptide("PEPTIDER").GetChemicalFormula();
IsotopicDistribution quantifiableDistribution = IsotopicDistribution.GetDistribution(quantifiableFormula, 0.125, 1e-8);
double[] quantifiableMzs = quantifiableDistribution.Masses.Select(m => m.ToMz(2)).ToArray();
```

across three scans spanning the identification’s retention time, so there is a shape to integrate rather than a point.

The test now asserts a nonzero intensity for the survivor. **That assertion fails against the old fixture**, which is the point — previously every intensity was zero and `Count == 1` passed anyway.

MetaMorpheus/Test/PostSearchAnalysisTaskTests.cs
trishorts: Fixed in `1c88650`. Now asserts `"1 PSM(s) were excluded from quantification"` rather than the tail of the sentence, so a method returning the wrong count no longer passes.

MetaMorpheus/Test/PostSearchAnalysisTaskTests.cs
trishorts: Fixed in `1c88650`. `RemovePsmsWithoutResolvedMass_DropsOnlyTheMasslessOnes` calls it directly and asserts both halves — the massless PSM is dropped, the resolved one is kept (`Is.SameAs`), and the return value is the number dropped. It also asserts both survive sequence filtering first, so the test cannot pass by the fixture being wrong.

MetaMorpheus/Test/PostSearchAnalysisTaskTests.cs
trishorts: Fixed in `1c88650`. `RemovePsmsWithoutResolvedMass_KeepsEverythingWhenAllMassesResolve` feeds only resolvable PSMs and asserts nothing is dropped and the count is zero — the case an inverted guard would otherwise pass.

trishorts (dismissed):
Approving. All five review points are addressed in `1c88650`, replies in each thread.

The one worth calling out is the fixture. The end-to-end test held a single MS1 peak at m/z 500.0 while the surviving identification is `PEPTIDER` — roughly 478.7 at charge 2 — so nothing could match at the 5 ppm tolerance, every intensity was zero, and `PeptideModifiedSequences.Count == 1` passed on the strength of an `Identification` having been constructed rather than anything having been quantified. For a PR whose stated purpose is that one bad PSM must not cost quantification for the whole run, the surviving path was the half that was never exercised.

Putting a peak at the right m/z was not sufficient either: FlashLFQ accepts a feature only when it can build an isotopic envelope, so the fixture now carries the peptide’s real distribution across three scans spanning its retention time, and the test asserts a nonzero intensity. That assertion fails against the old fixture.

The production change is small and in the same spirit: `SetSilacFilteredPsms` replaces the PSM list wholesale after the guard has run, so SILAC-derived PSMs were reaching FlashLFQ without ever being screened — the exact path the PR exists to protect.

For the record: `1c88650` is mine, pushed while working through these findings, so this approval covers a change I wrote as well as yours.

trishorts (approved):
Re-approving after `fb4e4d05`, which dismissed both standing approvals. @nbollis — nothing in the production change moved; would you re-approve when you get a chance?

The push is test-only. `codecov/patch` was failing at 71.43% on two lines in the SILAC arm of the guard: `PostSearchAnalysisTask.cs:534` was partial (only the false branch of `if (silacPsmsWithoutMass > 0)` was ever taken) and `:536` was never executed. Every existing SILAC test reaches that guard with nothing to drop, so the half that matters was untested — the same shape of gap as the fixture problem in `1c88650`.

`SilacQuantificationSkipsLabeledPsmsWithUnresolvedMass` drives `QuantificationAnalysis` with a SILAC label whose heavy residue has no defined mass. That is the same NaN route as B, X and Z, arriving through the labeled clone rather than the original peptide, and it is the state this task genuinely sees when the label residues were never added to `Residue`'s process-wide table — `SearchTask` adds them, not `PostSearchAnalysisTask`. The label residue is `X`, which is in `GlobalVariables.InvalidAminoAcids`, so `UpdateAminoAcidLabel` never assigns it and nothing in the suite registers a mass for it; the test asserts that precondition up front rather than depending on run order.

Both input PSMs resolve, so the guard ahead of the SILAC block has nothing to do and only the second guard can account for what the test observes. It asserts the SILAC message *and* the absence of the non-SILAC one, so neither guard alone satisfies it, and it asserts the survivors were actually quantified — nonzero intensity against a real isotopic envelope — rather than merely present in the results.

Confirmed against a mutant with the SILAC guard deleted: `Quantification_crash.txt` is written and `FlashLfqResults` comes back null, which is the whole-run loss this PR exists to prevent.

The MS1 fixture the earlier regression test built inline is now a helper taking any number of peptides, since this test needs envelopes for two.

Patch coverage is 100% (7/7, no misses, no partials) and project coverage moved from 93.40% to 93.41%; misses and partials are unchanged from base, so the new production lines added no uncovered code. All 13 checks pass.

trishorts:
@nbollis your approval here was dismissed automatically when new commits landed, not withdrawn — this needs a re-look when you have a moment.

You approved `72cf9720` (the master merge). Two commits have landed since:

- `1c88650` addresses my own review. The substantive change is a second mass guard after `SetSilacFilteredPsms`; the rest is the test fixture, which now builds a real isotopic envelope. FlashLFQ returns zero intensity for a feature with only a monoisotopic peak, so the original fixture was passing for the wrong reason.
- `fb4e4d05` covers the SILAC arm of that guard, which had no test of its own — codecov saw the branch as partial and the `Warn` line as never executed. Verified against a mutant with the guard deleted: `Quantification_crash.txt` is written and `FlashLfqResults` comes back null, which is the whole-run loss this PR exists to prevent.

No production behaviour changed between your approval and now beyond that second guard. All checks are green at `fb4e4d05`.

nbollis (approved):

RayMSMS (approved):

PR #2736, opened 2026-08-18
Read both spellings of the Cleavable column in a crosslinker tsv

acesnik:
🤖 A custom crosslinker saved with the Cleavable box unticked came back **cleavable** on the next launch.

`Data/Crosslinkers.tsv` spells that column `T`/`F`. The custom-crosslinker GUI saves through `Crosslinker.ToString(true)` (`CustomCrosslinkerWindow.xaml.cs:66`), which interpolates the bool and so spells it `True`/`False`. `ParseCrosslinkerFromString` only recognised `"F"`, so `"False"` fell through to the `cleavable = true` default.

## Output change

Measured on the exact line the GUI writes for an uncleavable DSS-like entry:

```
file line    : MyNonCleavable<TAB>K<TAB>K<TAB>False<TAB><TAB>138.06808<TAB>0<TAB>0<TAB>156.0786<TAB>155.0946<TAB>259.142
as entered   : Cleavable=False  CleaveDissociationTypes=[]
after reload : Cleavable=True   CleaveDissociationTypes=[]
```

The type list stays empty, so `IsCleavedBy` stays false and the localized masses do not move. But `Cleavable` gates signature-ion generation on its own at `CrosslinkedPeptides.cs:61`, so the reloaded crosslinker generated **one extra theoretical M ion per candidate**, at the intact crosslinked precursor mass — for `PEPTIDEK` crosslinked to a 1000 Da partner, 15 theoretical fragments instead of 14, the extra one at 2065.523 = 927.455 + 138.068 + 1000.

Inferred, not measured: that mass is the precursor, which is often present unfragmented in an MS2 scan, so it can match and add to a score; `Cleavable` also gates `CrosslinkSearchEngine.cs:315` and `:657`, which then populate `ParentIonMaxIntensityRanks` from those matched ions. **I have not run a search on real data, so I cannot say whether any PSM or FDR value moves.** Only searches using a *custom* uncleavable crosslinker are affected at all — the shipped file's `T`/`F` always parsed correctly.

Two output fields also reported the wrong value and are fixed by the same change: the pepXML `Cross-linker cleavable` parameter (`WriteXlFile.cs:123`) and the prose line at `XLSearchTask.cs:70`.

The session in which the crosslinker is created was always fine — `AddCrosslinkers` keeps the in-memory object. The bug only appeared after a restart.

## Fix

Reader-side, so existing `Data/CustomCrosslinkers.tsv` files are repaired rather than just newly-written ones. Making `ToString(true)` emit `T`/`F` instead would align the writer with the shipped format but would leave files already on disk misparsing; it is also the only production caller of that overload, so it can be changed later without touching anything else.

## Test

`UncleavableCrosslinkerStaysUncleavableThroughAFileRoundTrip` builds the file the GUI writes, reads it back through `LoadCrosslinkers`, and asserts the bool survives, that no signature ion is generated, and that the shipped `T`/`F` spelling still reads in both directions. It fails against the unfixed reader (`Expected: False / But was: True`).

Verified: `Test.csproj` compiles with `EnableWindowsTargeting`, and the test was run through a net10.0 harness against `EngineLayer`, both against the fix and against the unfixed reader. The full `XLTest` fixture cannot run on macOS.

Found while working on #2717 — same save-and-reload path, where the dissociation-type column had a matching case-sensitivity bug (#2716).

### DIFF
--- a/MetaMorpheus/EngineLayer/CrosslinkSearch/Crosslinker.cs
+++ b/MetaMorpheus/EngineLayer/CrosslinkSearch/Crosslinker.cs
@@ -98,11 +98,10 @@ namespace EngineLayer
         public static Crosslinker ParseCrosslinkerFromString(string line)
         {
             var split = line.Split('\t');
-            bool cleavable = true;
-            if (split[3] == "F")
-            {
-                cleavable = false;
-            }
+            // Crosslinkers.tsv writes T/F but ToString(true) writes the bool's own True/False, so a
+            // custom crosslinker saved as uncleavable reloaded as cleavable
+            string cleavableField = split[3].Trim().ToUpperInvariant();
+            bool cleavable = cleavableField != "F" && cleavableField != "FALSE";
 
             Crosslinker crosslinker = new Crosslinker(
                 crosslinkerName: split[0],
--- a/MetaMorpheus/Test/XLTest.cs
+++ b/MetaMorpheus/Test/XLTest.cs
@@ -53,6 +53,48 @@ namespace Test
             Assert.That(crosslinkerString == "testCrosslinker\tK\tK\tTrue\tCID|HCD|\t100\t25\t60\t0\t0\t50");
         }
 
+        /// <summary>
+        /// A custom crosslinker has to survive the save-and-reload the GUI puts it through:
+        /// CustomCrosslinkerWindow writes ToString(true), which spells the bool "True"/"False", while the
+        /// shipped Crosslinkers.tsv spells it "T"/"F". The reader has to read both, or an uncleavable
+        /// custom crosslinker comes back cleavable on the next launch.
+        /// </summary>
+        [Test]
+        public static void UncleavableCrosslinkerStaysUncleavableThroughAFileRoundTrip()
+        {
+            var asEntered = new Crosslinker("K", "K", "MyNonCleavable", false, "", 138.06808, 0, 0,
+                138.06808, 156.0786, 155.0946, 259.142);
+
+            string customCrosslinkerFile = Path.Combine(TestContext.CurrentContext.TestDirectory,
+                "XlTestData", "RoundTripCustomCrosslinkers.tsv");
+            Directory.CreateDirectory(Path.GetDirectoryName(customCrosslinkerFile));
+            File.WriteAllLines(customCrosslinkerFile, new[]
+            {
+                "Name\tCrosslinkAminoAcid\tCrosslinkerAminoAcid2\tCleavable\tDissociationType\tCrosslinkerTotalMass\tCrosslinkerShortMass\tCrosslinkerLongMass\tQuenchMassH2O\tQuenchMassNH2\tQuenchMassTris",
+                asEntered.ToString(true)
+            });
+
+            var reloaded = Crosslinker.LoadCrosslinkers(customCrosslinkerFile).Single();
+            Assert.That(reloaded.Cleavable, Is.False);
+            Assert.That(reloaded.CleaveDissociationTypes, Is.Empty);
+
+            // the bool alone gates signature-ion generation, so getting it wrong adds a theoretical M ion
+            // at the intact crosslinked precursor mass
+            var alphaPeptide = new Protein("PEPTIDEK", "").Digest(new DigestionParams(),
+                new List<Modification>(), new List<Modification>()).First();
+            var fragments = CrosslinkedPeptide.XlGetTheoreticalFragments(DissociationType.HCD, reloaded,
+                new List<int> { 7 }, 1000, alphaPeptide).Single().Item2;
+            Assert.That(fragments.Where(v => v.ProductType == ProductType.M), Is.Empty);
+
+            // the shipped T/F spelling still reads, in both directions
+            Assert.That(Crosslinker.ParseCrosslinkerFromString(
+                "shipped\tK\tK\tF\t\t138.06808\t0\t0\t156.0786\t155.0946\t259.142").Cleavable, Is.False);
+            Assert.That(Crosslinker.ParseCrosslinkerFromString(
+                "shipped\tK\tK\tT\tCID|HCD\t158.0038\t54.01056\t85.982635\t176.0143\t175.0303\t279.0777").Cleavable, Is.True);
+
+            File.Delete(customCrosslinkerFile);
+        }
+
         [Test]
         public static void XlTestXlPosCal()
         {


trishorts (commented):

MetaMorpheus/Test/XLTest.cs
```diff
@@ -53,6 +53,48 @@ public static void XlTestCrosslinker()
...
+            Assert.That(reloaded.Cleavable, Is.False);
+            Assert.That(reloaded.CleaveDissociationTypes, Is.Empty);
+
+            // the bool alone gates signature-ion generation, so getting it wrong adds a theoretical M ion
+            // at the intact crosslinked precursor mass
+            var alphaPeptide = new Protein("PEPTIDEK", "").Digest(new DigestionParams(),
+                new List<Modification>(), new List<Modification>()).First();
+            var fragments = CrosslinkedPeptide.XlGetTheoreticalFragments(DissociationType.HCD, reloaded,
+                new List<int> { 7 }, 1000, alphaPeptide).Single().Item2;
+            Assert.That(fragments.Where(v => v.ProductType == ProductType.M), Is.Empty);
```
trishorts, line 87: **Minor** - M-ion assertion is negative-only and passes vacuously on an empty fragment list

**Problem:**
`Assert.That(fragments.Where(v => v.ProductType == ProductType.M), Is.Empty)` is the only assertion made about the fragment list returned by `CrosslinkedPeptide.XlGetTheoreticalFragments`. Nothing asserts that `fragments` contains anything at all, so if `XlGetTheoreticalFragments` ever returned an empty `Item2` for this input — a changed `Fragment` contract, a different default protease producing a different `alphaPeptide`, a changed `possibleCrosslinkerPositions` meaning — the filter is empty and the assertion still passes. The PR body states the concrete expectation (14 theoretical fragments for uncleavable vs. 15 for the mis-parsed cleavable case), but the test asserts neither the count nor any positive ion. The sibling tests in this same file assert positive content: `TestTheoreticalFragmentsNonCleavableCrosslink` (`XLTest.cs:1264-1268`) checks the exact b- and y-ion mass sequences.

**Impact:**
The half of the test that guards the actual downstream consequence of the bug (an extra signature ion at the intact crosslinked precursor mass) can degrade into a no-op assertion without anyone noticing; only the `reloaded.Cleavable` assert would still be load-bearing.

**Suggested fix:**
Before the M-ion check, assert the fragment list is non-empty — better, assert the expected fragment count (the 14 the PR body cites) or the b/y ion masses in the style of `TestTheoreticalFragmentsNonCleavableCrosslink`, so the absence of the M ion is asserted against a known-good, populated list.

**Affected test scenarios:**
- UncleavableCrosslinkerStaysUncleavableThroughAFileRoundTrip - reloaded uncleavable crosslinker, HCD, PEPTIDEK + 1000 Da partner - fragment list is non-empty and contains exactly the expected b/y ions, and no `ProductType.M`
- UncleavableCrosslinkerStaysUncleavableThroughAFileRoundTrip - same input parsed by the unfixed reader - exactly one extra `ProductType.M` fragment at 2065.523 appears, so the count assertion fails as well as the `Cleavable` assertion

MetaMorpheus/Test/XLTest.cs
```diff
@@ -53,6 +53,48 @@ public static void XlTestCrosslinker()
...
+
+            // the shipped T/F spelling still reads, in both directions
+            Assert.That(Crosslinker.ParseCrosslinkerFromString(
```
trishorts, line 90: **Minor** - The added Trim() on the Cleavable column has no covering test input

**Problem:**
The production change at `Crosslinker.cs:103` is `split[3].Trim().ToUpperInvariant()`, and the PR states that whitespace tolerance on this column is part of the intended contract. The `ToUpperInvariant()` half is exercised — `"False"` from `ToString(true)` only matches `"FALSE"` after upper-casing, so removing it fails the test. The `Trim()` half is exercised by nothing: every Cleavable field the test feeds in (`"False"`, `"F"`, `"T"`) is already whitespace-free, and `LoadCrosslinkers` uses `ReadLine`, which strips the line terminator, so no incidental whitespace reaches the field either.

**Impact:**
`Trim()` could be deleted from the parser and the whole suite would still pass, so a hand-edited `CustomCrosslinkers.tsv` with a padded field (`"K\tK\t False \t"`) would silently regress to `cleavable = true` — the exact failure mode this PR exists to close.

**Suggested fix:**
Add one `ParseCrosslinkerFromString` assertion whose Cleavable column carries surrounding whitespace (e.g. `" False "` or `"\tF "`) and expects `Cleavable == false`.

**Affected test scenarios:**
- ParseCrosslinkerFromString - Cleavable column is `" False "` - `Cleavable` is false
- ParseCrosslinkerFromString - Cleavable column is `" F "` - `Cleavable` is false

MetaMorpheus/Test/XLTest.cs
```diff
@@ -53,6 +53,48 @@ public static void XlTestCrosslinker()
...
+                "shipped\tK\tK\tF\t\t138.06808\t0\t0\t156.0786\t155.0946\t259.142").Cleavable, Is.False);
+            Assert.That(Crosslinker.ParseCrosslinkerFromString(
+                "shipped\tK\tK\tT\tCID|HCD\t158.0038\t54.01056\t85.982635\t176.0143\t175.0303\t279.0777").Cleavable, Is.True);
```
trishorts, line 93: **Minor** - Cleavable half of the GUI round trip ("True") is never asserted

**Problem:**
The test verifies the uncleavable direction of the round trip (`ToString(true)` writes `"False"` → reads back false) and the shipped `T`/`F` spelling, but never asserts that the GUI's *cleavable* spelling `"True"` reads back as `Cleavable == true`. That case currently works only because it falls through to the permissive default (`cleavableField != "F" && != "FALSE"`), not because anything recognises `"TRUE"`. `CustomCrosslinkerWindow.SaveButton_Click` writes exactly this spelling for a ticked Cleavable box (`Crosslinker.ToString(true)`, line 66 of that file), so it is half of the round trip the PR claims to fix.



Worth pairing with the shape of the new predicate: `cleavableField != "F" && cleavableField != "FALSE"` means every value that is not one of those two spellings - `"True"`, but equally `"0"`, `"No"`, a typo, or an empty column - resolves to cleavable. That is the same permissive-default mechanism that produced the bug being fixed here, just narrowed by two cases. Recognising the four intended spellings explicitly and warning on anything else would close the class rather than this instance, and would give the missing `"True"` assertion something real to assert against.

**Impact:**
A later tightening of the predicate to a positive test — e.g. `cleavable = cleavableField == "T"`, the natural shape if someone aligns the writer to `T`/`F` as the PR says may happen later — would silently turn every GUI-saved *cleavable* custom crosslinker into an uncleavable one, and this test would still pass.

**Suggested fix:**
Add an assertion that a line whose Cleavable column is `"True"` parses to `Cleavable == true`, pinning both directions of the writer's spelling, not just the failing one.

**Affected test scenarios:**
- ParseCrosslinkerFromString - Cleavable column is `"True"` (the `ToString(true)` spelling) - `Cleavable` is true
- UncleavableCrosslinkerStaysUncleavableThroughAFileRoundTrip - cleavable crosslinker written via `ToString(true)` and reloaded - `Cleavable` stays true

MetaMorpheus/Test/XLTest.cs
```diff
@@ -53,6 +53,48 @@ public static void XlTestCrosslinker()
...
+
+            File.Delete(customCrosslinkerFile);
```
trishorts, line 95: **Minor** - Written tsv is leaked into shared XlTestData when an assertion fails

**Problem:**
`File.Delete(customCrosslinkerFile)` is the last statement of the test body with no `try`/`finally` or `[TearDown]`, and every assertion above it can throw. The file is written directly into the shared `XlTestData` directory (the same folder that holds the fixture's committed inputs such as `YeastPol2.fasta` and `BSA.fasta`), rather than into a test-owned folder. The other artifact-producing tests in this fixture — `XLSearchTastWriteFileTest`, `TestWriteToPercolator`, `XLSearchWithGeneratedIndices` — create a dedicated directory and `Directory.Delete(..., true)` it.

**Impact:**
A failing run leaves `RoundTripCustomCrosslinkers.tsv` behind in the test output's data directory indefinitely. It is harmless to the current suite (nothing enumerates files in `XlTestData`), but it dirties the shared fixture directory and makes a failed run's state indistinguishable from a mid-run one.

**Suggested fix:**
Guarantee cleanup on failure — wrap the assertions in `try`/`finally` with the delete in the `finally`, or write the file into a test-owned subdirectory that is deleted the way the sibling tests delete theirs.

**Affected test scenarios:**
- UncleavableCrosslinkerStaysUncleavableThroughAFileRoundTrip - `reloaded.Cleavable` assertion fails - written tsv is still removed from XlTestData
- UncleavableCrosslinkerStaysUncleavableThroughAFileRoundTrip - repeat run after a failed run - no stale file influences the result

MetaMorpheus/Test/XLTest.cs
```diff
@@ -53,6 +53,48 @@ public static void XlTestCrosslinker()
...
+            var reloaded = Crosslinker.LoadCrosslinkers(customCrosslinkerFile).Single();
+            Assert.That(reloaded.Cleavable, Is.False);
+            Assert.That(reloaded.CleaveDissociationTypes, Is.Empty);
+
+            // the bool alone gates signature-ion generation, so getting it wrong adds a theoretical M ion
+            // at the intact crosslinked precursor mass
+            var alphaPeptide = new Protein("PEPTIDEK", "").Digest(new DigestionParams(),
+                new List<Modification>(), new List<Modification>()).First();
+            var fragments = CrosslinkedPeptide.XlGetTheoreticalFragments(DissociationType.HCD, reloaded,
+                new List<int> { 7 }, 1000, alphaPeptide).Single().Item2;
```
trishorts, line 86: **Minor** - Crosslink position 7 on PEPTIDEK is E, not the K the crosslinker targets

**Problem:**

Verified against the indexing convention rather than inferred: `CrosslinkSpectralMatch.GetPossibleCrosslinkerModSites` (`CrosslinkSpectralMatch.cs:202-213`) walks a 0-based `r` over `BaseSequence` and emits `r + 1`, so the list this method takes is **1-based residue numbers**. `CrosslinkedPeptides.cs:38` then stores the localized mass at key `crosslinkerPosition + 1`, which is the mod-dictionary key for residue `crosslinkerPosition`. On `PEPTIDEK` residue 7 is `E`; the `K` is residue 8. So this test localizes the crosslink on a glutamate for a crosslinker declared `K`/`K` - a site `GetPossibleCrosslinkerModSites` would never return for this peptide. The M-ion assertion still exercises the signature-ion path, so the test is not wrong about what it catches, but the setup does not represent a reachable crosslink and the `1000` partner mass in the PR body's arithmetic is attributed to the wrong residue.
`asEntered` is built with `CrosslinkerModSites` / `CrosslinkerModSites2` both `"K"`, but the fragment call passes `new List<int> { 7 }`. In `PEPTIDEK` (P1 E2 P3 T4 I5 D6 E7 K8) position 7 is a glutamate; `XlGetTheoreticalFragments` places the localized mass at `crosslinkerPosition + 1` = index 8, i.e. residue 7. The only K in the peptide is position 8. `GetPossibleCrosslinkerModSites` would never return 7 for this crosslinker, so the fixture localizes the crosslink at a residue the crosslinker cannot modify.

**Impact:**
None for the assertion — `ProductType.M` generation at `CrosslinkedPeptides.cs:61` is gated on `crosslinker.Cleavable` alone and is independent of the position — but the fixture no longer represents the search-engine scenario it is documented as standing in for, and a reader chasing the 2065.523 M-ion number from the PR body will not reproduce it from this input.

**Suggested fix:**
Use the position of the actual K (8), or pick a peptide/mod-site pair where the supplied position is a legal crosslink site, so the fixture matches what the engine would hand this method.

**Affected test scenarios:**
- UncleavableCrosslinkerStaysUncleavableThroughAFileRoundTrip - crosslink localized at the K in PEPTIDEK - no `ProductType.M` fragment generated
- UncleavableCrosslinkerStaysUncleavableThroughAFileRoundTrip - same input parsed by the unfixed reader - exactly one `ProductType.M` fragment at the intact crosslinked precursor mass

nbollis (approved):

MetaMorpheus/EngineLayer/CrosslinkSearch/Crosslinker.cs
```diff
@@ -98,11 +98,10 @@ public static IEnumerable<Crosslinker> LoadCrosslinkers(string CrosslinkerLocati
...
             var split = line.Split('\t');
-            bool cleavable = true;
-            if (split[3] == "F")
-            {
-                cleavable = false;
-            }
+            // Crosslinkers.tsv writes T/F but ToString(true) writes the bool's own True/False, so a
+            // custom crosslinker saved as uncleavable reloaded as cleavable
+            string cleavableField = split[3].Trim().ToUpperInvariant();
+            bool cleavable = cleavableField != "F" && cleavableField != "FALSE";
```
trishorts, line 104: Fix is correct. One note on what it preserves.

`cleavable = cleavableField != "F" && cleavableField != "FALSE"` keeps the original default-to-true semantics: anything not recognisably false is cleavable. So a typo (`Fasle`, `N`, `0`) still silently yields a *cleavable* crosslinker -- which is the more dangerous direction, because it adds signature ions and stub-mass localization the user never asked for, and nothing reports it.

That behaviour is pre-existing and I am not asking you to change it as part of this fix. But you are already on this line, and the PR's own premise is that a hand-written or GUI-written file is a supported input, so the whitelist form costs nothing and removes the failure mode:

```csharp
string cleavableField = split[3].Trim().ToUpperInvariant();
bool cleavable = cleavableField == "T" || cleavableField == "TRUE";
```

That flips the unrecognised-value default to non-cleavable, which is the safe direction. The evidence that it is safe is already in your test -- it covers all four spellings that the two writers actually produce.

trishorts (commented):
Real bug, minimal fix, and the test is the right shape -- asserting through `LoadCrosslinkers` on a file written by `ToString(true)` is what actually reproduces the GUI round trip, rather than testing the parser against a string a human typed.

I particularly like that it goes on to assert the *consequence* (`fragments.Where(v => v.ProductType == ProductType.M), Is.Empty`) rather than stopping at `reloaded.Cleavable, Is.False`. That is the difference between pinning the bug and pinning the symptom the user actually saw, and it is what stops a future refactor satisfying the assertion without fixing anything.

**Merge-order note:** this and #2717 both edit `Crosslinker.cs` and both add a test immediately after `XlTestCrosslinker()` in `XLTest.cs`. Independent changes, but they will conflict in `XLTest.cs` whichever merges second -- worth picking an order deliberately.

PR #2737, opened 2026-08-18
Sequence-tag candidate filtering for database-seeded truncation search

trishorts:

# Sequence-tag candidate filtering for database-seeded truncation search

Split out of #2666 at @acesnik's request, so it can be reviewed on its own.

> **Stacked on #2666.** Head branch is `truncation-search-tags`, which sits on top of `truncation-search-mm`. GitHub can only open cross-fork PRs against `master`, so **the diff below currently includes #2666's commits**; it reduces to this feature's ~585 lines across 9 files once #2666 merges. Please review #2666 first. Draft until then.

## Why

`SeedParentsFromDatabase` (in #2666) removes the requirement that a truncation's parent be observed intact, by digesting the database rather than taking only upstream-identified proteoforms. On a real database that is a parent explosion, which is what makes the mode impractical. This narrows the digest to the proteins the spectra themselves support.

## How

- **`SequenceTagExtractor`** reads short de-novo residue tags straight off a scan's peak list. A tag is emitted only for an **unambiguous** residue gap, so an I/L-style tie produces no tag rather than a wrong one. Only the most intense fragment peaks feed it — low-intensity noise peaks create spurious residue-sized gaps that otherwise let tags match nearly every protein.
- **`ProteinTagIndex`** is a k-mer index over the database mapping each tag to the proteins containing it. A protein becomes a candidate once it carries at least `MinTagHits` distinct extracted tags.
- Tag extraction (over scans) and the candidate digest (over proteins) both run in parallel.

Two modes:

| | candidate set |
|---|---|
| default (`UsePerScanTagRestriction = false`) | **union over all scans** — the indexed engine is reused unchanged; controls the parent count but leaves the candidate set global |
| `UsePerScanTagRestriction = true` | each scan is scored only against parents from its **own** tag-supported proteins, so a scan cannot be won by a protein its own tags never supported |

The per-scan restriction strips the `DECOY_` prefix before testing, so a target and its reverse decoy are allowed together and FDR balance is preserved.

Tags restrict candidates only — Pass 2/3 scoring and the FDR analysis still police precision.

## Parameters

`UseSequenceTagFilter`, `UsePerScanTagRestriction`, `TagLength` (4), `MinTagHits` (2). All off/default; the feature is inert unless `SeedParentsFromDatabase` is also on.

## Tests

`SequenceTagTests` covers tag extraction (unambiguous-gap rule, tolerance, tag length), the k-mer index and the `MinTagHits` threshold; `TruncationSearchEngineTests.PerScanRestriction_ExcludesParentFromDisallowedProtein` covers the engine-side restriction in both directions; `TruncationTask_TagFilteredSeeding_WritesWellFormedOutputs` runs the whole path end to end.


MetaMorpheus/EngineLayer/Truncation/SequenceTagExtractor.cs
```diff
@@ -0,0 +1,112 @@
...
+{
+    /// <summary>
+    /// Extracts short de-novo sequence tags from a deconvoluted MS2 scan: amino-acid strings spelled by runs
+    /// of fragment-mass gaps that each match a residue mass. Tags are I/L-normalized (the two residues are
+    /// isobaric, both -> 'L'). They are used only as a recall-oriented candidate-protein FILTER for the
+    /// database-seeded truncation search (docs/Truncation-Search.md, Sequence-tag filtering / pTop §3.4.4); Pass 2/3 scoring + FDR police precision.
+    ///
+    /// Top-down caveat: at large fragment masses the per-peak absolute tolerance is wide relative to
+    /// residue-mass differences, so a gap may match several residues (e.g. K vs Q). Such gaps emit one edge
+    /// per matching residue, deliberately over-generating tags so the true protein is not missed.
```
acesnik: This describes the pre-`1fbdaca7` behaviour. The summary says ambiguous gaps "emit one edge per matching residue, deliberately over-generating tags", and line 10 calls the filter "recall-oriented"; the code at `:79` requires `matchCount == 1` and the comment at `:65-67` argues the opposite, "high precision over recall". Whichever is right, the file should not say both.

MetaMorpheus/EngineLayer/Truncation/SequenceTagExtractor.cs
```diff
@@ -0,0 +1,112 @@
...
+                    int matchCount = 0;
+                    foreach (var (aa, mass) in Residues)
+                    {
+                        if (Math.Abs(gap - mass) <= window)
+                        {
+                            matched = Normalize(aa);
+                            if (++matchCount > 1) break;
+                        }
+                    }
+                    if (matchCount == 1)
```
acesnik, line 92: Worth stating the recall cost next to `MinTagHits`. K and Q differ by 0.03638 Da, so the combined two-peak window covers that gap above ~909 Da at 20 ppm, ~1,819 Da at 10 ppm, ~3,638 Da at 5 ppm — above which no K/Q gap ever emits an edge. In top-down that is most of the ladder, so tags crossing a K or Q stop being produced exactly where the feature is meant to work.

MetaMorpheus/TaskLayer/TruncationSearchTask/TruncationSearchTask.cs
```diff
@@ -0,0 +1,588 @@
...
+                proteins.AddRange(DatabaseLoadingEngine.LoadProteinDb(
+                        db.FilePath, generateTargets: true, decoyType: DecoyType.None,
+                        localizableModificationTypes, db.IsContaminant, out _, out _, CommonParameters)
+                    .Where(p => !p.IsDecoy));
+            }
+
+            int tagLength = TruncationSearchParameters.TagLength;
+            int minTagHits = TruncationSearchParameters.MinTagHits;
+            int threads = Math.Max(1, CommonParameters.MaxThreadsToUsePerFile);
+            var index = new ProteinTagIndex(proteins, tagLength, threads);
```
acesnik, line 482: `TagLength` is user-settable in the TOML and reaches the `ProteinTagIndex` constructor unvalidated, which throws `ArgumentOutOfRangeException` below 1. No GUI wiring exists for this task yet, so today that is a CMD stack trace rather than the crash-report dialog — but it becomes the dialog the moment it is wired, and the house channel for an impossible configuration is a `Warn` gate. Cheap to add here.

MetaMorpheus/TaskLayer/TruncationSearchTask/TruncationSearchTask.cs
```diff
@@ -0,0 +1,588 @@
...
+
+            var candidateProteinIds = new HashSet<int>();
+            foreach (HashSet<int> set in perThreadCandidates) candidateProteinIds.UnionWith(set);
+
+            Warn($"Sequence-tag filter selected {candidateProteinIds.Count} of {proteins.Count} database proteins " +
+                 $"(tag length {tagLength}, min {minTagHits} tag hits).");
+
+            // Digest only the candidate proteins (in parallel) into theoretical full-length parents.
+            var perThreadParents = new ConcurrentBag<List<TruncationParent>>();
+            Parallel.ForEach(candidateProteinIds, parallelOptions, () => new List<TruncationParent>(), (id, _, local) =>
```
acesnik: This is the first parent-building path that is parallel; #2666's `BuildParentsFromDatabase` is a serial nested `foreach`, so its parent order is deterministic and this one's is not. `TruncationSearchEngine.cs:139` sorts by mass, but `OrderBy` is stable, so equal-mass parents keep `ConcurrentBag` arrival order and Pass 2's `nScore > bestScore` gives the first one the win. Iterating `candidateProteinIds.OrderBy(id => id)`, or a `.ThenBy` on accession at the engine's sort, closes it.

acesnik (commented):
🤖 Reviewed as the ~585-line delta over #2666 (`git diff refs/pr/2666...refs/pr/2737`), which matches the figure in your header — thank you for stating the review unit, it saved re-reading a feature I already approved.

The problem is the one that makes database seeding unusable, and the shape of the answer is right: if you drop the requirement that a parent be observed intact, you have to replace it with *some* evidence that the spectra support the protein, and the scan's own de-novo tags are the cheapest such evidence there is. Building a k-mer index once and intersecting per scan is the correct way round — it costs one pass over the database rather than a digest per scan. I also like that the reduction is reported rather than assumed: the `Warn` at `TruncationSearchTask.cs:500` prints selected-of-total, so anyone can see whether the filter did anything on their data.

Two things I checked because they would have been serious, and both are clean. The `DECOY_` strip in `IsScanAllowed` really does recover the target accession — `TruncationParentBuilder.cs:98` builds decoys as `"DECOY_" + target.ProteinAccession`, so a target and its decoy stand or fall together and the per-scan restriction cannot skew target/decoy balance. And a scan missing from the dictionary lands on the static `NoAllowedProteins` empty set rather than on `null`, so it gets no candidates instead of silently escaping the restriction; the only scans that can take that path are ones with no deconvoluted fragments, which would score nothing anyway.

The one I would fix before this leaves draft is that the class summary on `SequenceTagExtractor` now describes the opposite of what the code does. It says ambiguous gaps "emit one edge per matching residue, deliberately over-generating tags so the true protein is not missed", and calls the filter "recall-oriented"; the code at `:79` emits an edge only when exactly one residue matches, and the inline comment right above it argues for "high precision over recall". Commit `1fbdaca7` is where the behaviour flipped, and the summary did not follow. A reader takes the summary as the contract.

That matters more than a stale comment normally would, because the precision rule has a real cost in the regime this feature is for. K and Q differ by 0.03638 Da, so the two-peak window swallows that difference above roughly 909 Da of fragment mass at 20 ppm, 1,819 Da at 10 ppm and 3,638 Da at 5 ppm. Above those, every K/Q gap matches two residues and emits nothing — so in top-down, where fragment masses sit squarely in that range, tags spanning a K or a Q quietly stop being produced. That may well be the right trade, but it is a recall characteristic worth stating next to `MinTagHits` rather than leaving to be discovered.

Smaller, and both cheap on a task this new:

`TagLength` is a TOML-serialized parameter read at `:460` and handed straight to `new ProteinTagIndex(...)`, whose constructor throws `ArgumentOutOfRangeException` for anything below 1. Today `TruncationSearchTask` has no GUI wiring — I grepped `GUI/` and `GuiFunctions/` and found none — so a CMD user gets a stack trace rather than the crash-report dialog. Worth closing before it is wired up, since at that point a config typo becomes "MetaMorpheus crashed, would you like to report it?". A large `TagLength` is gentler: every protein is shorter than *k*, the index is empty, and the `Warn` honestly says 0 of N.

The tag-filtered seeding is the first parent-building path that is parallel — `BuildParentsFromDatabase` in #2666 is a plain nested `foreach` in database order — so parent order now varies run to run, coming out of a `ConcurrentBag` over a `HashSet<int>`. `TruncationSearchEngine.cs:139` sorts by `MonoisotopicMass`, which normalizes most of that, but `OrderBy` is stable, so parents of *equal* mass keep their arrival order, and Pass 2 breaks ties with `if (nScore > bestScore)` — first one wins. Equal-mass parents are not exotic (duplicate database entries, isoforms sharing a chain). I have not measured how often it flips a call, so treat that as reasoned rather than observed, but a `.ThenBy` on accession would remove the question for the cost of one line, and it is much cheaper to do now than to debug later as irreproducibility.

One for the description rather than the code: the body offers I/L as the example of an ambiguous gap producing no tag, but I and L are a single entry in the residue table, so a 113.084 gap matches exactly one residue and *does* emit a tag, normalized to `L` — which is the behaviour you want, and it matches the index's own normalization. K/Q is the case that actually produces nothing.

acesnik (commented):
_(duplicate of the review above — posted twice by mistake; the inline notes on this one have been removed)_

MetaMorpheus/EngineLayer/Truncation/SequenceTagExtractor.cs
trishorts: Fixed — the code is right and the summary was wrong, so the summary changed. It now says an ambiguous gap emits **no** edge, drops the "recall-oriented" label, and says why the drop is the right call (emitting one edge per matching residue over-generates until the filter selects most of the database and stops being a filter). It also points at `ExtractTags` for what that costs, which is the next thread.

MetaMorpheus/EngineLayer/Truncation/SequenceTagExtractor.cs
trishorts: Stated, with your numbers checked rather than copied. Recomputed from the residue table: K − Q = 0.03638 Da, and since `window` is the sum of the two peaks' tolerances the crossover is 910 Da at 20 ppm, 1,819 Da at 10 ppm, 3,638 Da at 5 ppm — your figures, confirmed.

Added one thing while checking: no other residue pair is close enough to matter here. The next narrowest is K vs E at 0.948 Da, whose crossover is around 23.7 kDa at 20 ppm, so K/Q really is the whole effect.

The comment now records that above its ppm's crossover no K/Q gap can ever emit an edge — so over most of a top-down ladder, tags spanning a K or a Q stop being produced — and that this is what `MinTagHits` is trading against: raising it demands more tags from a scan whose tags are already thinned this way.

MetaMorpheus/TaskLayer/TruncationSearchTask/TruncationSearchTask.cs
trishorts: Added, as a Warn gate. `TagLength < 1` now warns and falls back to the default rather than reaching `ProteinTagIndex` and throwing `ArgumentOutOfRangeException`.

I did `MinTagHits` at the same time — it is the same TOML, the same unchecked path, and a value below 1 asks no protein to carry any tag, which selects the whole database and silently defeats the filter rather than crashing. Both defaults are now `public const` on `TruncationSearchParameters` (`DefaultTagLength`, `DefaultMinTagHits`) and the property initialisers use them, so the warning text and the actual default cannot drift apart.

MetaMorpheus/TaskLayer/TruncationSearchTask/TruncationSearchTask.cs
trishorts: Fixed at the source rather than only at the sort, and then at the sort as well.

`BuildParentsFromDatabaseTagFiltered` no longer accumulates into a `ConcurrentBag`. Candidate ids are ordered, digestion writes into one slot per candidate protein via `Parallel.For`, and the slots are concatenated in protein-id order — so the parent list is a function of the database, not of thread scheduling, while staying parallel.

I also took your second suggestion, because it makes the guarantee independent of how any caller built its parents: the engine's sort is now `.OrderBy(mass).ThenBy(ProteinAccession).ThenBy(Proteoform.FullSequence)`, both ordinal. Belt and braces, but the failure mode you describe — stable `OrderBy` plus `nScore > bestScore` handing the win to whoever arrived first — is invisible in a passing test run, so pinning it in the one place every path goes through is worth the two lines.

66 truncation tests pass unchanged, so no currently-pinned winner was resting on the old arbitrary order.

PR #2738, opened 2026-08-18
Direct internal-fragment search with its own FDR

trishorts:

# Direct internal-fragment search with its own FDR

Split out of #2666 at @acesnik's request, so it can be reviewed on its own.

> **Stacked on #2737, which is stacked on #2666.** Head branch is `truncation-search-internal`. GitHub can only open cross-fork PRs against `master`, so **the diff below currently includes both of those PRs' commits**; it reduces to this feature's ~437 lines across 10 files once they merge. Please review #2666, then #2737, then this. Draft until then.
>
> The stacking order is incidental — this feature and the sequence-tag filter are independent of each other.

## Why the terminal search cannot find these

A real internal fragment has lost **both** termini, so it shares no ion ladder with its parent. The terminal search deduces which terminus a scan retains from the asymmetry of matched ions; an internal fragment retains neither, so there is nothing for it to deduce. It needs a direct search scored on its own b/y ladder.

It also cannot be judged in the same FDR universe. The pooled universe in #2666 contains the inherited intact matches, which clear an easy threshold; internal candidates judged against that distribution would inherit a q-value the evidence does not support. So they get their own per-class FDR pass and their own output file.

## How

- **`InternalTruncationSearch`** enumerates internal spans of the **same parent set** — no database-wide subsequence search. Mass-matching spans are found by prefix-sum arithmetic and only the few matches are built, so enumeration is near-linear in parent length rather than quadratic.
- A **bilateral gate**: a candidate must show at least `InternalMinIonsPerTerminus` matched ions on **each** newly created terminus before it is scored. One-sided evidence is exactly what a terminal truncation looks like, so this is what separates the two.
- Scoring is the standard core weighted by **`FragmentationPropensity`**.
- Results are collapsed, run through their own per-class FDR, and written to `AllInternalTruncations.psmtsv`.

## Off by default

`SearchInternalTruncations = false`. Its FDR behaviour in dense spectra is the open question, and nothing in the terminal search depends on it — the terminal/intact path never reads any of this.

## Parameters

`SearchInternalTruncations`, `InternalMinIonsPerTerminus` (2).

## Tests

`FragmentationPropensityTests` covers the propensity model; `ChopInternal_RecoversInternalSpan_LabeledInternal` covers the chopper primitive (bilateral chop, total residues chopped, notch, label); `InternalSearch_RecoversInternalFragment_FromParent` is its engine-level companion — the direct search must recover `P1[6..40]` from its parent, clear the bilateral gate, and report exactly one internal PSM with the right span; `TruncationTask_DatabaseSeeded_WritesWellFormedOutputs` exercises the branch and its output file end to end.


MetaMorpheus/EngineLayer/Truncation/FragmentationPropensity.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// the standard matched-ion score. Pluggable later for ETD-specific or data-learned propensities.
+    /// </summary>
+    public static class FragmentationPropensity
+    {
+        /// <summary>Relative cleavage weight for the backbone bond between N-side residue <paramref name="before"/>
+        /// and C-side residue <paramref name="after"/> (1.0 = average bond).</summary>
+        public static double CleavageWeight(char before, char after)
+        {
+            double weight = 1.0;
+            if (before == 'D') weight *= 3.0;        // strong enhanced cleavage C-terminal to aspartate
```
acesnik, line 36: These stack multiplicatively, so a `D-P` bond scores 9.0 and one such matched ion outweighs nine at average bonds. LeDuc 2014 is cited for the concept, but not for these three numbers or for multiplying them — and since this score is what the separate internal FDR is computed over, the multipliers are setting q-values. Worth recording in the remarks whether they are from the paper or chosen as a first pass.

MetaMorpheus/EngineLayer/Truncation/FragmentationPropensity.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// contribution, multiplied by the cleavage propensity. A uniform propensity reproduces the standard score.
+        /// </summary>
+        public static double Score(IReadOnlyList<MatchedFragmentIon> matchedIons, string baseSequence, double totalIntensity)
+        {
+            double score = 0;
+            int length = baseSequence.Length;
+            foreach (MatchedFragmentIon ion in matchedIons)
+            {
+                Product product = ion.NeutralTheoreticalProduct;
+                if (!TryGetCleavageBond(product.Terminus, product.FragmentNumber, length, out int before, out int after))
```
acesnik: This is why the "reduces exactly to the standard matched-ion score" claim in the summary is not exact. `CalculatePeptideScore` adds `1 + Intensity / TotalIonCurrent` for every matched ion except `ProductType.D`; here a whole-length b/y ion maps to no bond, returns false, and contributes nothing. Small on a long ladder, but the claim is what licenses comparing these scores with terminal ones.

MetaMorpheus/EngineLayer/Truncation/FragmentationPropensity.cs
```diff
@@ -0,0 +1,75 @@
...
+            }
+            return beforeIndex >= 0 && afterIndex < length;
+        }
+
+        /// <summary>
+        /// Propensity-weighted match score: sum over matched fragment ions of the cleavage weight at the bond
+        /// the ion reports, each scaled by (1 + fractional intensity) — the standard MetaMorpheus per-ion
+        /// contribution, multiplied by the cleavage propensity. A uniform propensity reproduces the standard score.
+        /// </summary>
+        public static double Score(IReadOnlyList<MatchedFragmentIon> matchedIons, string baseSequence, double totalIntensity)
```
acesnik: The model is static HCD/CID, but `InternalTruncationSearch.Run` resolves a per-scan dissociation type at `:89` and nothing gates on it — so an ETD run has D/E/P propensities applied to N-Cα cleavages that do not exhibit them, and the result feeds the internal FDR. Falling back to uniform weights outside HCD/CID would match the documented intent, and per the note above uniform is very nearly the standard score.

MetaMorpheus/EngineLayer/Truncation/InternalTruncationSearch.cs
```diff
@@ -0,0 +1,163 @@
...
+            Parallel.For(0, scans.Length, parallelOptions, scanIndex =>
+            {
+                Ms2ScanWithSpecificMass scan = scans[scanIndex];
+                double precursorMass = scan.PrecursorMass;
+                double totalIntensity = scan.TheScan.TotalIonCurrent;
+                DissociationType dissociationType = TruncationSearchEngine.ResolveDissociationType(commonParameters, scan);
+                var products = new List<Product>();
+                TruncationPsm best = null;
+
+                for (int pi = 0; pi < parents.Count; pi++)
```
acesnik, line 117: Scan × parent × start with a binary search inside and no fragment index, so cost tracks the parent count directly. The code is independent of #2737 as the description says, but affordability is not: this is only tractable once something has cut the parent set, which is what #2737 does. A timing on a realistic database-seeded parent set would say whether the default should stay off for cost as well as for FDR.

acesnik (commented):
🤖 Reviewed as the ~437-line delta over #2737 (`git diff refs/pr/2737...refs/pr/2738`), which matches your header.

The argument in "Why the terminal search cannot find these" is the clearest statement of the problem anywhere in this stack, and it is right on both halves. A fragment that has lost both termini shares no ladder with its parent, so a search that infers the retained terminus from ion asymmetry has nothing to infer from — and giving it its own FDR follows from the same fact, because a score regime that includes inherited intact matches sets a threshold the internal class cannot honestly clear. The bilateral gate is the part I would have gotten wrong: requiring evidence on *each* new terminus is exactly what distinguishes an internal fragment from a terminal one, and one-sided evidence being indistinguishable from a terminal truncation is the reason a symmetric ion count would not do.

The enumeration is genuinely near-linear as claimed — prefix sums plus `LowerBound`, with the `break` at `:106` cutting off starts once even the longest remaining span is too light. The span bounds are strictly interior (`1 <= start`, `end <= length-2`), so a "internal" fragment can never be a disguised terminal one. And the `NotchSearchWindow` comment at `:29-34` is the kind of note I wish more constants carried: it says plainly that the constant is not derived from the acceptor and names what breaks if someone pairs it with a wider one.

I also chased the `products` list reused across candidates at `:90`, since the history shows `3aafb51b` adding a clear and `a9b35ac3` reverting it. The revert was right — `PeptideWithSetModifications.Fragment` opens with `products.Clear()`, so reuse is safe. Noting it so the next reader does not take the same detour.

Three things on the scoring, which is where I would spend the remaining effort, because it feeds a q-value directly.

`FragmentationPropensity.Score` does not reduce to the standard score under a uniform model, though the summary says it does "exactly". `CalculatePeptideScore` adds `1 + Intensity / TotalIonCurrent` for every matched ion bar `ProductType.D`; this one skips any ion for which `TryGetCleavageBond` fails, which is the whole-length b/y ions — they report no internal bond, so they contribute nothing here and 1-plus-intensity there. It is a small difference on a long ladder, but "exactly" is doing work in that sentence, and the claim is what tells a reader they can compare these scores to terminal ones.

The weights are the bigger question. `D → ×3.0`, `E → ×2.0`, `P → ×3.0` stack multiplicatively, so a matched ion at a `D-P` bond scores 9, and one such ion outweighs nine ions at average bonds. The aspartic-acid and proline effects are real and the paper is the right one to point at, but LeDuc 2014 is cited for the *concept* while these particular numbers and their multiplicativity are not attributed. Since this score is what the separate internal FDR is computed over, the multipliers set the q-values, and a 9× single bond is a lot of leverage for a first-pass constant. Worth saying in the remarks where the numbers come from, even if the answer is "chosen, to be replaced by a learned model".

And the model is static HCD/CID while `Run` resolves the dissociation type per scan at `:89` and fragments accordingly. An ETD run therefore gets D/E/P propensities applied to N–Cα cleavages that do not have them, silently, and the resulting scores feed their own FDR. The summary anticipates this ("Pluggable later for ETD-specific") but nothing gates it. Falling back to uniform weights outside HCD/CID would make the behaviour match the documentation, and by the point above, uniform is very nearly the standard score — which is the honest default for a regime the model does not describe.

Last, a scaling note rather than a defect. The search is scan × parent × start with a binary search inside and no fragment index, so its cost scales with the parent set. That is fine on a confident parent set and not fine on a database-seeded one — which means that although this feature and #2737 are independent as code, in practice the internal search is only affordable when something has already cut the parent count, and #2737 is that something. Given `SearchInternalTruncations` is off by default I would not hold anything for it, but a timing on a realistic parent set would tell you whether the default should stay off for cost as well as for FDR.

MetaMorpheus/EngineLayer/Truncation/FragmentationPropensity.cs
trishorts: Recorded, and your framing is the right one — the multipliers set q-values, so their provenance belongs in the file.

The remarks now say explicitly: LeDuc 2014 is cited for the **concept** (weight a matched ion by the cleavage propensity of the bond it reports) and for the three effects being real in collisional dissociation. The magnitudes 3.0 / 2.0 / 3.0, and the decision to stack them multiplicatively so a D-P bond scores 9.0, are a first pass chosen here, not values read out of that paper. Your arithmetic is called out as the reason it matters: one matched ion at a D-P bond outweighs nine at average bonds, which is a strong claim for unfitted numbers. Flagged as the thing to calibrate against a labelled set before this feature is ever defaulted on.

MetaMorpheus/EngineLayer/Truncation/FragmentationPropensity.cs
trishorts: Made exact rather than softened, since as you say the claim is what licenses comparing these scores with terminal ones — weakening it would weaken the comparison.

Two changes: a whole-length b/y ion, which reports no internal bond, now takes the average weight of 1.0 instead of being dropped; and `ProductType.D` is skipped, exactly as `CalculatePeptideScore` does. With uniform weights the two functions are now term-for-term identical.

Pinned by a test rather than by assertion: `Score_OutsideCollisionalDissociation_IsExactlyTheStandardScore` builds b1..bL, y1..yL and a diagnostic ion and asserts `Score(..., DissociationType.ETD)` equals `MetaMorpheusEngine.CalculatePeptideScore` to 1e-12. `Score_UnderCollisionalDissociation_WeightsTheEnhancedBonds` asserts the HCD score on the same ions is strictly greater, so the first test cannot pass for the trivial reason that nothing is ever weighted.

MetaMorpheus/EngineLayer/Truncation/FragmentationPropensity.cs
trishorts: Gated. `FragmentationPropensity.ModelApplies(DissociationType)` now names the types the static model describes — HCD, CID, LowCID, ISCID — and `Score` takes the scan's resolved dissociation type and applies uniform weights everywhere else. `InternalTruncationSearch.Run` already resolves it at `:89`, so it just passes it through; an ETD scan is now scored on the standard matched-ion score, with no D/E/P weights reaching the internal FDR.

EThcD is deliberately excluded rather than lumped in with the collisional types: it yields b/y and c/z together, so a single scalar weight per bond cannot be right for both halves of its ladder. Uniform is the honest answer until the model can be applied per ion series. That is recorded in the doc comment and pinned by `ModelApplies_OnlyToCollisionalDissociation`.

This composes with the thread above: uniform is now *exactly* the standard score, so the fallback is a well-defined scoring function rather than an approximation of one.

MetaMorpheus/EngineLayer/Truncation/InternalTruncationSearch.cs
trishorts: Measured, and the answer is not the one either of us expected — so it is now in a `COST` remarks block on `Run` rather than left as a judgement call.

Microbenchmark on this branch (Release, one scan so the outer `Parallel.For` over scans is not hiding anything, synthetic repeat-alphabet parents, exact-notch acceptor):

| parents | ms | ms/parent |
|---|---|---|
| 25 | 0.6 | 0.025 |
| 100 | 2.2 | 0.022 |
| 400 | 9.1 | 0.023 |
| 800 | 18.0 | 0.023 |

| parent length | ms/parent |
|---|---|
| 100 | 0.111 |
| 200 | 0.192 |
| 400 | 0.386 |

So it is linear in parent count — your reading — and also linear in parent length, not quadratic: about **1 µs per residue, per parent, per scan** on one thread.

Extrapolated to 5,000 MS2 scans against parents averaging 350 residues, on 60 threads: **~350 ms per scan and about half a minute total** against a tag-filtered ~1,000-parent set, and **~7 s per scan and about ten minutes total** against an unfiltered ~20,000-parent human set.

Which changes the conclusion: #2737 buys roughly a factor of twenty, but this search is **not gated on it** — the unfiltered case is minutes, not hours, on a many-core box. The default stays off for the FDR reason on the parameter, not for cost. Caveat recorded in the remarks: synthetic parents make these a floor, since real proteoforms carry modifications and those multiply the forms per parent.

PR #2739, opened 2026-08-18
Place non-specific terminal modifications by peptide-relative key

### LINKED ISSUE #2400: crash running a non-specific search.
I ran an analysis that includes spectra averaging, GPTMD and search.
The crash occurred during the search stage, giving a "key=10" error. 
see the attached file.

Thanks!
David.
[results.txt](https://github.com/user-attachments/files/16676304/results.txt)

acesnik:
## Problem

🤖 `NonSpecificEnzymeSearchEngine.Accepts` used **absolute protein coordinates** as keys into `AllModsOneIsNterminus`, which is **peptide-relative** (`1` = N-terminus, `2` = first residue, `length + 2` = C-terminus). Four sites were affected, each correct at exactly one start position:

| Site | Was | Should be | Correct only when |
|---|---|---|---|
| N-truncation, C-terminal mod | `endResidue` | `FragmentNumber + 2` | `OneBasedStartResidue == 3` |
| C-truncation, N-terminal mod | `startResidue - 1` | `1` | `OneBasedEndResidue == F + 1` |
| full-length fallback (N) | `OneBasedEndResidue + 1` | `Length + 2` | `OneBasedStartResidue == 2` |
| full-length fallback (C) | `OneBasedStartResidue - 1` | `1` | `OneBasedStartResidue == 2` |

Measured on a `singleN` search with a C-terminal variable modification, sweeping where the matched peptide starts in the protein:

| peptide starts at | before | after |
|---|---|---|
| residue 1 | `AAAAA[mod]K` — on residue 5 | `AAAAAK-[mod]` |
| residue 2 | `AAAAAK[mod]` — on the K as a *residue* mod | `AAAAAK-[mod]` |
| residue 3 | `AAAAAK-[mod]` (already correct) | unchanged |
| residue 5 | `AAAAAK` — **mod silently dropped**, though the mass still includes it | `AAAAAK-[mod]` |
| `singleC` | `K[mod]AAAAAAA` — residue 1, not the N-terminus | `[mod]KAAAAAAA` |

The out-of-range case is the worst: `MonoisotopicMass` includes a modification that `FullSequence` does not show.

When another modification already occupied the wanted key, `Dictionary.Add` threw `An item with the same key has already been added`.

## Why the #2453 guard is removed rather than mirrored

#2453 fixed the C-terminal twin by guarding with `!updatedMods.Keys.Contains(startResidue - 1)`. That guard **skips adding the terminal modification**, but the notch that admitted the match was computed *including* that modification's mass. So affected non-specific searches have been reporting peptides whose sequence omits a modification their precursor match depended on — a crash traded for silent data loss.

With the corrected keys the guard is unnecessary and actively harmful: the N-side copy filter admits only keys **below** `FragmentNumber + 2`, and the C-side shift produces only keys **≥ 2**, so the target slot is provably free. A guard on a correct key would drop the modification instead of placing it.

## Scope

The copy-filter bound is algebraically unchanged — `endResidue - OneBasedStartResidue + 3` is identically `FragmentNumber + 2` — so **matches without a terminal-restricted modification are unaffected**. Confirmed empirically: `TestNonSpecificEnzymeSearchEngineSingleNModifications` and `...SingleCModifications` both still pass (they use only `Anywhere.` mods).

`IndexingEngine.AddInteriorTerminalModsToPrecursorIndex` is the only other terminal-modification code path; it computes masses and bins and never builds a modification dictionary, so it needs no change.

## ⚠️ This changes output

For non-specific searches that use terminal-restricted modifications, reported sequences **and scores** change: `Accepts` re-fragments and re-scores the peptide it returns, so relocating a terminal modification changes the theoretical ion series, the score, FDR ranking, and which FDR category wins a scan. Runs without terminal-restricted modifications are unaffected.

## Tests

This branch of `Accepts` had **no placement coverage** before this PR — no existing test used a `C-terminal.`/`N-terminal.` `LocationRestriction`.

Worth knowing: the existing #2453 regression test cannot detect this class of bug. `TestNonSpecificEnzymeSearchEngine` yields **39 PSMs and 0 with a terminal-placed modification, both before and after this change** — it guards the exception, not the placement. It does still pass, which is the safety check that matters for the code being touched.

Added to `SearchEngineTests.cs`: a 4-case start-position sweep (`singleN`), the `singleC` mirror, and the collision case where a competing `Anywhere` modification occupies the wanted key. All run the real engine through `IndexingEngine` + `NonSpecificEnzymeSearchEngine`, in the style of the existing non-specific tests. **5 of the 6 fail against unfixed code; all 6 pass with the fix** — and the one that passes unfixed is `start=3`, exactly the single position the old arithmetic was right for.

Verified on macOS via a throwaway `net10.0` harness compiling the committed test methods verbatim, since `Test.csproj` targets `net10.0-windows`; `Test.csproj` and `CMD.csproj` build clean with `EnableWindowsTargeting`. **The full Windows suite has not been run locally — CI is the real gate.**

Closes #2400 — the reported crash was in the C-terminal branch and was fixed by #2453; this closes the remaining N-terminal path flagged on that issue, so the exception is no longer reachable from `Accepts`.

### DIFF
--- a/MetaMorpheus/EngineLayer/NonSpecificEnzymeSearch/NonSpecificEnzymeSearchEngine.cs
+++ b/MetaMorpheus/EngineLayer/NonSpecificEnzymeSearch/NonSpecificEnzymeSearchEngine.cs
@@ -350,17 +350,20 @@ namespace EngineLayer.NonSpecificEnzymeSearch
                     if (fragmentationTerminus == FragmentationTerminus.N)
                     {
                         int endResidue = peptide.OneBasedStartResidue + fragment.FragmentNumber - 1; //-1 for one based index
+                        // Mod keys are peptide-relative: 1 is the N-terminus, 2 the first residue, length+2 the C-terminus.
+                        int cTerminusModKey = fragment.FragmentNumber + 2;
                         Dictionary<int, Modification> updatedMods = new Dictionary<int, Modification>();
                         foreach (KeyValuePair<int, Modification> mod in peptide.AllModsOneIsNterminus)
                         {
-                            if (mod.Key < endResidue - peptide.OneBasedStartResidue + 3) //check if we cleaved it off, +1 for N-terminus being mod 1 and first residue being mod 2, +1 again for the -1 on end residue for one based index, +1 (again) for the one-based start residue
+                            if (mod.Key < cTerminusModKey) //check if we cleaved it off
                             {
                                 updatedMods.Add(mod.Key, mod.Value);
                             }
                         }
                         if (terminalMod != null)
                         {
-                            updatedMods.Add(endResidue, terminalMod);
+                            // The filter above admits only keys below this one, so it is always free.
+                            updatedMods.Add(cTerminusModKey, terminalMod);
                         }
                         updatedPwsm = new PeptideWithSetModifications(peptide.Protein, peptide.DigestionParams, peptide.OneBasedStartResidue, endResidue, CleavageSpecificity.Unknown, "", 0, updatedMods, 0);
                     }
@@ -377,9 +380,10 @@ namespace EngineLayer.NonSpecificEnzymeSearch
                                 updatedMods.Add(key, mod.Value);
                             }
                         }
-                        if (terminalMod != null && !updatedMods.Keys.Contains(startResidue - 1))
+                        if (terminalMod != null)
                         {
-                            updatedMods.Add(startResidue - 1, terminalMod);
+                            // Key 1 is the peptide N-terminus; the shift above only produces keys of 2 or more.
+                            updatedMods.Add(1, terminalMod);
                         }
                         updatedPwsm = new PeptideWithSetModifications(peptide.Protein, peptide.DigestionParams, startResidue, peptide.OneBasedEndResidue, CleavageSpecificity.Unknown, "", 0, updatedMods, 0);
                     }
@@ -421,11 +425,11 @@ namespace EngineLayer.NonSpecificEnzymeSearch
                                 //add the terminal mod
                                 if (fragmentationTerminus == FragmentationTerminus.N)
                                 {
-                                    updatedMods[peptide.OneBasedEndResidue + 1] = terminalMod;
+                                    updatedMods[peptide.Length + 2] = terminalMod;
                                 }
                                 else
                                 {
-                                    updatedMods[peptide.OneBasedStartResidue - 1] = terminalMod;
+                                    updatedMods[1] = terminalMod;
                                 }
 
                                 PeptideWithSetModifications updatedPwsm = new PeptideWithSetModifications(peptide.Protein, peptide.DigestionParams, peptide.OneBasedStartResidue, peptide.OneBasedEndResidue, CleavageSpecificity.Unknown, "", 0, updatedMods, peptide.NumFixedMods);
--- a/MetaMorpheus/Test/SearchEngineTests.cs
+++ b/MetaMorpheus/Test/SearchEngineTests.cs
@@ -2034,5 +2034,191 @@ namespace Test
             File.Delete(xmlName);
             File.Delete(mzmlName);
         }
+
+        /// <summary>
+        /// A C-terminal variable modification found by Accepts must land on the truncated peptide's
+        /// C-terminus regardless of where in the protein that peptide starts. The key used to be an
+        /// absolute protein coordinate, so it was only correct for peptides starting at residue 3.
+        /// The parameter is where the matched peptide begins in the protein.
+        /// </summary>
+        [Test]
+        [TestCase(1)]
+        [TestCase(2)]
+        [TestCase(3)]
+        [TestCase(5)]
+        public static void TestNonSpecificEnzymeSearchTerminalModPlacementSingleN(int startResidue)
+        {
+            SearchParameters SearchParameters = new SearchParameters
+            {
+                SearchTarget = true,
+                MassDiffAcceptorType = MassDiffAcceptorType.Exact,
+                LocalFdrCategories = new List<FdrCategory> { FdrCategory.NonSpecific }
+            };
+            DigestionParams dp = new DigestionParams("singleN", minPeptideLength: 5,
+                fragmentationTerminus: FragmentationTerminus.N, searchModeType: CleavageSpecificity.None);
+            CommonParameters CommonParameters = new CommonParameters(
+                dissociationType: DissociationType.HCD, digestionParams: dp, scoreCutoff: 5,
+                precursorMassTolerance: new PpmTolerance(5), addCompIons: true);
+
+            ModificationMotif.TryGetMotif("K", out ModificationMotif kMotif);
+            Modification cTerminalMod = new Modification(_originalId: "TerminalProbe", _modificationType: "TestTerminal",
+                _target: kMotif, _locationRestriction: "C-terminal.", _monoisotopicMass: 42.010565);
+            var variableModifications = new List<Modification> { cTerminalMod };
+            var fixedModifications = new List<Modification>();
+
+            // The searched peptide is always AAAAAK; a glycine prefix moves where it starts in the
+            // protein, which is the coordinate the old key depended on. Glycine differs in mass from
+            // alanine, so a window starting at residue 1 cannot match the same precursor.
+            PeptideWithSetModifications guiltyPwsm = new PeptideWithSetModifications(
+                "AAAAAK-[TestTerminal:TerminalProbe on K]",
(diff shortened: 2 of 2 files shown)


trishorts (commented):
Automated code review - 3 finding(s) posted as inline comments.

MetaMorpheus/EngineLayer/NonSpecificEnzymeSearch/NonSpecificEnzymeSearchEngine.cs
```diff
@@ -421,11 +425,11 @@ private Tuple<int, PeptideWithSetModifications> Accepts(List<Product> fragments,
                                 //add the terminal mod
                                 if (fragmentationTerminus == FragmentationTerminus.N)
                                 {
-                                    updatedMods[peptide.OneBasedEndResidue + 1] = terminalMod;
+                                    updatedMods[peptide.Length + 2] = terminalMod;
```
trishorts, line 428: **Major** - Full-length fallback key changes ship with no test that can fail

**Problem:**
The PR changes four key expressions, but the three tests added to `SearchEngineTests.cs` only reach the two truncation branches (lines 366 and 386). The fallback at lines 428/432 is entered only when `databaseAnnotatedMods.TryGetValue(peptide.Length, ...)` succeeds, i.e. when the terminal-mod motif sits on the last residue of the *indexed* peptide. In all three new tests the indexed peptide is longer than the matched window (`AAAAAK` + 10×`G`, `GGK`+7×`A`, `AAAAAAAAMK`+10×`G`), so the lookup key is 6/8/10 while `peptide.Length` is 16/10/20 — the fallback is never taken. The only existing test that does reach it, `TestTerminalModificationsForSnes`, uses protein `MAGIAAKLAKDREAAEGLGSHA` with `OneBasedStartResidue == 2`, where old and new arithmetic are numerically identical (`OneBasedEndResidue + 1 == 22 + 1 == 23 == Length + 2`, and `OneBasedStartResidue - 1 == 1`).

**Impact:**
Half of the sites the PR claims to fix are merged unverified: a sign error or off-by-one in `peptide.Length + 2` / `1` would produce a wrong `FullSequence` and a wrong re-scored ion series on the full-length path, and the whole suite would still be green.

**Suggested fix:**
Add a fallback-branch analogue of the start-position sweep: index a protein whose singleN/singleC peptide ends (respectively begins) exactly on the modification motif so `TryGetValue(peptide.Length)` hits, and sweep `OneBasedStartResidue` across at least one value other than 2 so the old and new expressions diverge. Assert on `FullSequence` and on the `AllModsOneIsNterminus` key, as `TestTerminalModificationsForSnes` already does.

**Affected test scenarios:**
- TestNonSpecificEnzymeSearchTerminalModFallbackSingleN - full-length peptide with a C-terminal-restricted mod on its last residue, peptide starting at protein residue 5 - mod key is `Length + 2`, `FullSequence` ends in `-[...]`
- TestNonSpecificEnzymeSearchTerminalModFallbackSingleC - full-length peptide with an N-terminal-restricted mod on its first residue, peptide starting at protein residue 5 - mod key is 1, `FullSequence` begins with `[...]`
- TestTerminalModificationsForSnes - unchanged fixture (start residue 2) - still passes, confirming the fallback is preserved where old and new agree

MetaMorpheus/EngineLayer/NonSpecificEnzymeSearch/NonSpecificEnzymeSearchEngine.cs
```diff
@@ -377,9 +380,10 @@ private Tuple<int, PeptideWithSetModifications> Accepts(List<Product> fragments,
                                 updatedMods.Add(key, mod.Value);
                             }
                         }
-                        if (terminalMod != null && !updatedMods.Keys.Contains(startResidue - 1))
+                        if (terminalMod != null)
                         {
-                            updatedMods.Add(startResidue - 1, terminalMod);
+                            // Key 1 is the peptide N-terminus; the shift above only produces keys of 2 or more.
+                            updatedMods.Add(1, terminalMod);
```
trishorts, line 386: **Minor** - Guard removal is on the C branch; the collision test is singleN only

**Problem:**
The `!updatedMods.Keys.Contains(startResidue - 1)` guard added by #2453 was deleted from the C-truncation branch, and `updatedMods.Add(1, terminalMod)` now runs unguarded. The only collision test added, `TestNonSpecificEnzymeSearchTerminalModDoesNotCollide`, is configured with `new DigestionParams("singleN", ..., fragmentationTerminus: FragmentationTerminus.N, ...)`, so it exercises the *other* branch's `updatedMods.Add(cTerminusModKey, terminalMod)` at line 366. Nothing in the suite drives a C-truncation match that carries competing modifications through the shift filter at line 380 into the unguarded `Add(1, ...)`.

**Impact:**
The specific line whose crash protection this PR removes is untested. If the "shifted keys are always ≥ 2" reasoning were ever broken by a change to the copy filter, the regression would resurface as the same `An item with the same key has already been added` exception #2453 was raised for, with no test to catch it.

**Suggested fix:**
Mirror the collision test on the singleC side: a protein whose matched window carries an interior variable modification that survives the `mod.Key > indexShift + 1` filter, plus an N-terminal-restricted mod, and assert both modifications appear in the resulting `FullSequence`.

**Affected test scenarios:**
- TestNonSpecificEnzymeSearchTerminalModDoesNotCollideSingleC - singleC match with an `Anywhere.` mod on an interior residue and an N-terminal-restricted mod on the first residue - no exception, both modifications present in `FullSequence`

MetaMorpheus/EngineLayer/NonSpecificEnzymeSearch/NonSpecificEnzymeSearchEngine.cs
trishorts, line 428: **Minor** - Fallback keys now target the slot an existing terminal mod occupies

**Problem:**
In the truncation branches the PR establishes that the target key is provably free. That argument does not extend to the fallback: `peptide.AllModsOneIsNterminus` is copied wholesale at lines 421-424 with no filtering, and the new keys `peptide.Length + 2` (N) and `1` (C) are exactly the slots a digestion-applied C-terminal / N-terminal modification already occupies. Because these two sites use indexer assignment rather than `Add`, the pre-existing modification is silently replaced. The notch that admitted the match was computed at line 415 as `totalMass + terminalMod.MonoisotopicMass.Value` where `totalMass` already includes the resident modification, so the reported peptide carries two modification masses but only one modification in its sequence. The old expressions collided this way only when `OneBasedStartResidue == 2`; the corrected keys collide at every start position.

**Impact:**
On the fallback path, `MonoisotopicMass` can exceed what `FullSequence` accounts for — the exact failure mode the PR describes as "the worst case" for the truncation branch — for any peptide that was already digested with a terminal-restricted modification at the same terminus.

**Suggested fix:**
Decide explicitly what should happen when the fallback's target key is already taken: either skip the candidate modification (and do not accept the notch computed with its mass), or keep both by rejecting the match, rather than letting the indexer overwrite. Whichever is chosen, cover it with a test in which the indexed peptide already carries a terminal-restricted modification at that terminus.

**Affected test scenarios:**
- TestNonSpecificEnzymeSearchFallbackTerminalModAlreadyPresent - full-length peptide already carrying a C-terminal mod at `Length + 2`, precursor matching mass + a second terminal mod - reported `MonoisotopicMass` and `FullSequence` agree, or the match is rejected

trishorts (commented):
Good catch, and the diagnosis holds up. I checked the arithmetic against master rather than taking the comments on trust:

- **N-terminal branch.** The old filter `mod.Key < endResidue - peptide.OneBasedStartResidue + 3` expands to `fragment.FragmentNumber + 2`, which is exactly the new `cTerminusModKey`. So that rewrite is a pure clarification with no behavioural change -- worth saying out loud, because it reads like it might be one.
- **The old `updatedMods.Add(endResidue, terminalMod)` was correct only when `OneBasedStartResidue == 3`**, since `endResidue == OneBasedStartResidue + FragmentNumber - 1` equals `FragmentNumber + 2` only there. That matches the claim in the description and is a real bug.
- **C-terminal branch.** Dropping the `!updatedMods.Keys.Contains(startResidue - 1)` guard is safe: the loop admits only `mod.Key > indexShift + 1`, so the smallest key it can produce is `2`. Key 1 is genuinely never occupied.

The `[TestCase]`-per-start-residue shape is the right test for this, and using a glycine prefix so a window at residue 1 cannot accidentally match the same precursor is a nice touch -- it makes the test fail for the reason it claims.

One concern on the third site, inline.

Issue #2087, opened 2021-07-30
Another crosslinking and database partitioning issue?


emirzakh:
Hi, I've noticed that when you adjust the database partition size, the number of identified inter and intra-protein crosslinks, above 1% FDR, increase 5-10 fold. My protein databases are fairly small, between 100-250 proteins, depending on the search. When using a database partition size of 1, the results are lackluster. Increasing the partition size to 2 brings the results closer in line with pLink2 and XlinkX. Increasing the partition size further doesn't have a significant impact on results when compared to a partition size of 2, until you get up to higher partitions like 25-50, which then results in a drop in crosslinks identified above 1% FDR.

For example, with an unenriched crosslink file, with all other parameters held constant (parameters aren't very stringent - 20 ppm precursor and product tolerance, 2 miss cleavages allowed, 2 variable mods, etc) I see the following:

Database partition  -  Inter  -    Intra     -   single -    loop -   deadend
         1                         0             2         10150         72        822
         2                       418         48         10067         77        814
         3                       295         57         10094         79         819
         5                       229         59         10112         81         836
        25                        70         62         10208         107       845

With an enriched crosslink sample, the results are also very dramatic:

Database partition  -  Inter  -    Intra     -   single -    loop -   deadend
           1                       8            846           101         781      2612
           2                     1092      1300           91          803       2622 

I appreciate any advise as to whether the results achieved with the increased partition size can be trusted based on q-value/score or if there is a bug that is artificially causing high scoring crosslinked peptide ids.

lonelu:
Thanks for providing the details of the issue. We are aware of the issue (#2039). We have a new update here which will solve this problem (#2084) in theory. 
Still, your information is very valuable and I need to run some analysis to confirm it. I didn't expect the difference of database partition to cause such a big change of ids. 
Please wait for more information and the update.

trishorts:
to clarify:
1. you get a different result now compared to earlier
2. now, when you change partitions, you get the same result each time. (I think this is the desired result, correct?)

emirzakh:
Hi, yes, the results now are different than prior to the update. 

Regarding question #2 - yes, the result stays the same when the partition is changed. However, this raises another question - were the prior results correct or are the current results correct or were they both correct? With the massive difference of 3,889 vs 38 inter-peptide crosslinks for just 1 experiment between the two updates, this seems like an important answer to know.

lonelu:
[...]

PR #2742, opened 2026-08-18
Speed up modern-search index creation, and fix four bugs that made results depend on the partition count

### LINKED ISSUE #412: Speed up index creation in modern search
get it below 30s

### LINKED ISSUE #1542: Modern Search: Binary search is flawed
With an array of 12 items, attempt to hit the 3rd item. The result is 0 instead of 2.

Example: 
peptideIdsInThisBin = {0,1,2,3,4,5,6,7,8,9,10,11}
peptideMassToLookFor=2
peptideIndex = {0,1,2,3,4,5,6,7,8,9,10,11} (monoisotopic masses)

{l,m,r}
1. original {0,0,11}
2. start loop {0,5,11}
3. else {0,5,4}
4. loop {0,2,4}
5. else {0,2,1}
6. loop {0,0,1}
break, return 0.

`
        protected static int BinarySearchBinForPrecursorIndex(List<int> peptideIdsInThisBin, double peptideMassToLookFor, List<PeptideWithSetModifications> peptideIndex)
        {
            int m = 0;
            int l = 0;
            int r = peptideIdsInThisBin.Count - 1;

            while (l <= r)
            {
                m = l + ((r - l) / 2);

                if (r - l < 2)
                    break;
                if (peptideIndex[peptideIdsInThisBin[m]].MonoisotopicMass < peptideMassToLookFor)
                    l = m + 1;
                else
                    r = m - 1;
            }
            if (m > 0)
                m--;
            return m;
        }
`

acesnik:
🤖 Closes #412, fixes #1542.

## What this started as, and what it turned into

Issue #412 asks for faster index creation in modern search. Indexing the mouse
proteome now takes **46 s instead of 172 s**, and a full search **154 s instead of
337 s**.

Getting there meant partitioning the database more aggressively when memory is
tight, and that exposed something worse than a speed problem: **the number of
partitions changed the search results**. Partitioning is a memory decision, so
reported identifications moved with how much RAM happened to be free. Chasing that
turned up four separate accuracy bugs. All four are fixed, and the mouse search is
now byte-identical at 6 and 7 partitions across every output file. Crosslink search
becomes partition-independent too, which is the symptom reported in #2087.

Fixing them changes search output. At 1 % FDR the PSM count is unchanged and the
peptide count drops by 12 of 5,488 on the mouse proteome; crosslink identifications
at 1 % FDR do not move. Details and reasoning in
[what changes in your results](#what-changes-in-your-results) — please read that
section before merging.

## The speed work

**The fragment index is now two flat arrays instead of a list per bin.**

<details>
<summary>What compressed sparse row is, and why it suits a fragment index (background)</summary>

The fragment index answers one question: *given a fragment mass bin, which peptides
have a theoretical fragment in it?* It was stored the obvious way, as an array with
one list per bin:

```
bin 0 ──> null
bin 1 ──> List<int>{ 41, 909, 12874 }      each arrow is a separate heap object
bin 2 ──> null
bin 3 ──> List<int>{ 77 }
...      30,000,001 bins at the default 30,000 Da maximum
```

Compressed sparse row (CSR) is a standard sparse-matrix layout — rows here are bins,
entries are peptide ids. It stores the same information as two flat arrays: all the
ids end to end, plus an offset array saying where each bin's run begins.

```
peptideIds  [ 41, 909, 12874, 77, ... ]
binStart    [ 0,  0,   3,     3,  4, ... ]
                 └ bin 1 = peptideIds[0..3)      bin 2 = peptideIds[3..3) i.e. empty
```

Bin `b` is `peptideIds[binStart[b] .. binStart[b+1])`. An empty bin costs one `int`
and no object at all.

Four reasons this is a large win for *this* structure specifically:

- **Allocation dominated the build.** A mammalian database populates several million
  bins, so the old layout allocated several million small objects, each with an
  object header, each growing by repeated doubling — every doubling copies the
  contents and abandons the old array. CSR allocates exactly two arrays, once, at
  exactly the right size. That is why the win is far bigger than the arithmetic
  suggests: the work was never the binning, it was the memory traffic.
- **The garbage collector has to walk what you allocate.** Several million live
  objects is millions of references to trace on every collection, throughout a search
  that is already memory-hungry. Two `int[]` are two objects, and arrays of a
  primitive contain no references for the collector to follow at all.
- **The read pattern is a contiguous scan.** Coarse scoring walks every id in a bin,
  then binary-searches within it by peptide mass. Both are linear memory access over
  one array, so they stay in cache; a list per bin scatters those ids across the heap.
  This is also why the bins can be handed out as `ReadOnlySpan<int>` — a view into the
  shared array, with no copy and no allocation per lookup.
- **It is built once and then never modified.** CSR's real weakness is that inserting
  into a row means shifting everything after it, so it suits write-once-read-many
  structures. The fragment index is exactly that: built during indexing, read-only for
  the rest of the search. (The *precursor* index is not, which is why it keeps the old
  layout — `AddInteriorTerminalModsToPrecursorIndex` appends to it after construction.)

The one cost is that CSR needs to know each bin's size before filling, so the build
makes two passes over the peptides — count, then fill — and fragments each peptide
twice. Measured on the mouse proteome that trade is strongly favourable: the second
fragmentation pass is cheaper than the allocation it removes.

</details>

The old layout was a `List<int>[]` over the whole bin space — 30,000,001 bins at the
default maximum fragment mass, each occupied one holding its own list object with its
own header and a backing array doubled as ids arrived. Populating it was the serial
half of index creation and was dominated by allocation, not by work.

It is now built by counting into the offset array, prefix-summing it in place, then
filling with those offsets consumed as write cursors and shifted back afterwards — so
the only bookkeeping array is the offset array itself, reused rather than copied.
Reading a cached index back is now a pair of bulk array fills rather than rebuilding
millions of objects, and writing it is two `memcpy`-shaped writes.

**Server GC** is enabled for the CLI and GUI, and **a memory guard** raises
`TotalPartitions` when one partition of the index would not fit. The guard sizes
from memory that is actually free, clamped by any Slurm allocation, which is what
the review asked for. That is only safe because partition count no longer changes
results.

Measured on the mouse proteome (17,947 proteins), seven partitions both sides:

| | indexing | wall |
|---|---|---|
| before | 133–172 s | 297–337 s |
| after | 46–61 s | 154–189 s |

## The four accuracy bugs

**1. Fine scoring stopped early, per partition.** Candidates were taken in
descending coarse-score order and the loop broke once no remaining candidate could
beat the best fine score. The bound is sound, but it was accumulated per partition,
so the number of candidates scored — and with it the runner-up and the Delta Score
derived from it — depended on how the database had been split. Every candidate is
now scored. Measured cost: none outside run-to-run variance.

**2. Peptides with an undefined mass corrupted the coarse-scoring binary search.**
The search needs each bin's masses to ascend. Every comparison against `NaN` is
false, so such an entry reads as "too heavy" wherever it sits, and sorting the
peptide index by mass puts them at the front of every bin they occupy. The
predicate then reads false, then true, then false along the bin, and the search
could settle in the leading false run and clip real candidates off the window.
Which bins held one depended on which partition its protein landed in.

On scan 4934 of the mouse search, peptide `TAGAAAK` sits in the same five visited
bins at 6 and at 7 partitions, with the same mass, window and accepted notch — but
the computed window covered it in four of them against five. Coarse score 4 against
5, the cutoff is 5, so at six partitions it was never a candidate and an unrelated
11-mer won the scan.

The fix is in the searches, not the index: both now read an undefined mass as
negative infinity, which is where `double.CompareTo` already sorts it, so both
predicates become monotone again. Such peptides stay indexed, and that matters. An
earlier revision of this branch dropped them instead, on the reasoning that a mass
comparison can never accept one — true of a tolerance acceptor, but `OpenSearchMode`
accepts anything, and an open search leaves both window bounds infinite so neither
search runs and the whole bin is scored. Dropping them silently stopped open and
glyco searches reporting peptides with an unknown residue. Classic search reaches
the same contract by a different route, and a test now asserts the two agree.

**3. The start of the precursor window used an upper-bound search.**

<details>
<summary>Which bound is which, and why asking for the wrong one loses candidates (background)</summary>

Coarse scoring needs the slice of a fragment bin whose peptides fall inside the
precursor window `[lowest, highest]`. A bin is sorted by peptide mass, so that slice
is a contiguous range `[start, end]`, and each end is found by binary search.

The two ends are *different questions*:

```
bin:      [ 900.1   901.4   901.4   901.4   903.7   905.2 ]
                      ^                       ^
window:            lowest = 901.4          highest = 903.7

end   = LAST index whose mass <= highest      -> 4    (upper bound)
start = FIRST index whose mass >= lowest      -> 1    (lower bound)
```

Both ends were taken from the same upper-bound search. Asked for the start, it
answers the question it knows: *last index at or below 901.4* — which is index **3**,
not 1. Entries 1 and 2 fall out of the window and are never scored, even though their
mass is exactly the value the window opens at.

A run of equal masses on the lower edge is not an unlucky coincidence, it is the
normal case:

- the same peptide sequence occurs in many proteins, each an entry with an identical
  mass;
- a reversed decoy has its target's residue composition, so it carries its target's
  exact mass;
- notch intervals are computed *from* a peptide mass, so the window's lower edge
  routinely lands exactly on one.

Whether a given copy survived depended on which other peptides shared its bin, and
that depends on how the database was partitioned — which is how a memory decision
reached the score. Measured on scan 27831 of the mouse search: five copies of
`QQAQNIEKMSK` share a mass and a bin set, and four of them scored zero of the five
bins they sit in. The survivor scored 5 of 5 bins at seven partitions and 2 of 5 at
six. At six partitions nothing cleared the score cutoff, so no runner-up was ever
recorded and the reported Delta Score was Score minus the score cutoff — a sentinel
value, not a measurement.

The rest of the codebase already got this right. `CrosslinkSearchEngine.BinarySearchGetIndex`
and `GlycoPeptides.BinarySearchGetIndex` both take the bitwise complement for the
insertion point and then walk back over a run of equal masses, which is a correct
lower bound. Modern search was the outlier, and a test now runs the two
implementations against each other over ~900 targets so they cannot drift apart again.

A second, smaller defect in the same function: the upper-bound search returned `0`
both for "index 0 is the answer" and for "nothing in this bin is light enough", so a
bin whose peptides were all too heavy scored its first entry anyway. That fired on a
third of all window lookups. On its own it was inert — such a peptide fails the
mass-difference acceptor and never becomes a candidate — but it has to be right for
the corrected start to be right. It now returns `-1`.

</details>

The end of the window wants an upper bound; the start wants a lower bound. Both came
from the same upper-bound search, which for the start answers the wrong question: it
returns the *last* entry at or below the bound, so a run of equal masses sitting
exactly on the window's lower edge collapsed to its final member and the rest scored
nothing. Equal masses are the normal case, not an edge case — see above.

**4. Three psmtsv columns printed in insertion order.** (`Mass Diff (Da)`, `Mass Diff (ppm)`
and the most-abundant-mode error) read the raw candidate list while every sibling
column reads the sorted view, so they printed in arrival order. `CompareTo`
tie-breaks on `PrecursorMassErrorPpm.First()`, so that ordering reached the psm
sort and the cumulative target/decoy counts.

## What changes in your results

**Confident identifications are flat at PSM level and slightly down at peptide
level.** This is the number to look at, and it is not a gain — an earlier draft of
this description said it was, which was wrong.

Full mouse proteome, at 1 % FDR, targets only:

| | before | after |
|---|---|---|
| PSMs at q &le; 0.01 | 14,292 | **14,292** |
| peptides at q &le; 0.01 | 5,488 | **5,476** |
| protein groups | 2,752 | 2,752 |

Unfiltered, 13 scans gain a match and none loses one — 16,696 matched scans become
16,709 and 24,111 PSM rows become 24,130 — but **every one of those 13 sits above
1 % FDR**. They are candidates that were being silently skipped, and they are weak:
10 target, 3 decoy. At peptide level enough of the newly-scored candidates are
decoys to tighten the q-value threshold, and 12 target peptides fall the wrong side
of it.

The crosslink fixture shows the same shape. Unfiltered, 13 interlinks are gained and
8 of them are decoys. At 1 % FDR the crosslinks themselves do not move at all —
Inter 55, Intra 83, Loop 8, all unchanged — and the attrition is in the
non-crosslinked classes: Single 229 &rarr; 221, DeadEndH2O 62 &rarr; 61.

So the honest summary is that this costs about 0.2 % of confident peptides on a
mammalian search and buys correctness plus reproducibility. The candidates being
admitted are genuinely inside the precursor window; scoring them is what the code
was always trying to do, and that some are decoys at low scores is FDR working as
intended rather than a defect. But it is a trade, and it should be made knowingly.

**Crosslink search stops depending on the partition count too.** This is the symptom
reported in #2087, where changing the partition count moved inter- and intra-link
counts several fold. On the crosslink fixture here, forcing 1, 2 and 4 partitions
over the same database:

| | before | after |
|---|---|---|
| single peptides at 1 vs 4 partitions | 241 vs 242, files differ | 237 vs 237, byte-identical |
| interlinks, intralinks | files differ | identical |

The only field that still varies is `AlphaIndexingRank`, which is a per-partition
ordinal and cannot be compared across partition counts by construction. This fixture
produces four interlinks against the hundreds in #2087, so it demonstrates the
mechanism rather than settling that report — someone with that data should re-run it.

**Results do not depend on the thread count either.** The peptide index is
concatenated in thread order, so its ids genuinely differ between a 1-thread and an
8-thread run, but a modern search over the 3,000-protein database is byte-identical
across all five output files.

**Partition count no longer matters.** 6 against 7 partitions on the full mouse:

| | before | after |
|---|---|---|
| AllPSMs differing rows | 24,087 | 0 |
| AllPeptides | differed | identical |
| AllQuantifiedPeaks | differed | identical |
| AllQuantifiedProteinGroups | differed | identical |
| percolator input | differed | identical |

PEP and q-values move broadly, as they do for any change — the PEP model is trained
on the run's own PSMs, so a single changed score retrains it and moves nearly every
row. That is amplification, not independent movement.

Other engines: crosslink search moves one `AlphaIndexingRank`, glyco one `Rank`,
neither changing an identification, score or q-value. Non-specific and semi-specific
search are byte-identical and partition-invariant. The bundled Calibrate → GPTMD →
Search chain is byte-identical.

## Verification

Every engine that touches the index was run against a pre-change build on the same
inputs: modern (full mouse and a 3,000-protein subset), non-specific at one and
three partitions, semi-specific over both termini, crosslink with and without a
second fragment index, glyco, and the bundled chain. Cold and warm runs were
compared for all three index files — fragment, precursor and second fragment — so
the readers are exercised and not just the writers.

Coverage of what changed, measured with coverlet: `FragmentIndex` 100 %,
`IndexPartitioning` 96.6 %, the compressed-sparse-row build
(`IndexingEngine.RunSpecific`) 92.3 %, and the eleven `ModernSearchEngine` methods
touched 79.5 %, up from about 35 %. Every regression test was checked by
reverting its fix and confirming it fails — which is how one of them was found to be
decorative: with a few undefined masses leading a large bin the binary search never
probes them, so seven tests passed against a reverted fix. The case that
discriminates is a bin that is mostly undefined masses, and that test now exists.

One test-fixture fault was fixed along the way, and it is worth knowing about because
it is the reason the suite looked far worse than it was. `XlTest_MoreComprehensive`
switches on a process-wide static, `FdrAnalysisEngine.QvalueThresholdOverride`, and
switches it off again on its last line. When this branch moved that test's expected
counts, its first assertion failed, the reset never ran, and every later test computed
q-values under the override — 21 MetaDraw tests failed as a result, none of which had
anything wrong with them. A `[TearDown]` now clears it, so a wrong expected value in
that test can no longer change unrelated results.

Six expected values in that test do change, all measured rather than guessed: the raw
counts, the counts at 1 % FDR, and one PEP feature. Three of the four counting blocks
were reproduced locally against the commit before the window fix; the PEP-derived
block does not reproduce off CI, so those two numbers come from CI rather than from a
guess.

Each of the four bugs is also tested on every engine that shares its code, not just
where it was found. The codebase was swept for further instances of all four
mistakes; the span-versus-null trap had five (all fixed), and the other three had
none beyond the ones described here.

## Known and not fixed

A low-resolution CID search does not find the peptide its own fragments came from.
The indexing side bins with `Math.Round` and the scoring side with a cast to `int`,
so they disagree on exactly half of all bins — 15,000 of 30,000 at the default
maximum fragment mass. This predates the branch and is not touched here; it needs a
real LowCID dataset to fix responsibly.

The 30-second target in #412 is not met on a mammalian database. 46 s of indexing
is the honest number on this hardware.


MetaMorpheus/EngineLayer/Indexing/IndexPartitioning.cs
```diff
@@ -0,0 +1,189 @@
...
+
+            double scale = totalResidues / (double)sampledResidues;
+            return ((long)(sampledPeptides * scale), (long)(sampledFragments * scale));
+        }
+
+        /// <summary>
+        /// Memory the process can still use: the physical/container limit less what the heap already
+        /// holds (the spectra of the file being searched are loaded by this point).
+        /// </summary>
+        private static long AvailableBytes()
```
trishorts, line 243: **Major** - AvailableBytes measures total physical RAM, not memory actually available

**Problem:**
`AvailableBytes()` returns `GC.GetGCMemoryInfo().TotalAvailableMemoryBytes - GC.GetTotalMemory(false)`. `TotalAvailableMemoryBytes` is the machine's *physical* memory (or the container limit) — not free memory — and `GC.GetTotalMemory(false)` accounts only for this process's own managed heap. Nothing in the expression sees memory held by other processes, the OS file cache, or this process's unmanaged allocations. The XML summary ("the physical/container limit less what the heap already holds") describes the arithmetic accurately, but the value is then consumed as if it were headroom: `PartitionsForBudget` multiplies it by `MemoryBudgetFraction` and treats the product as memory an index build may occupy. The `Math.Max(total / 8, free)` floor on line 172 compounds this in the worst case — once the heap exceeds 7/8 of physical memory the function reports *more* headroom than `free`, not less.

**Impact:**
On the machine the PR targets — a 16 GB laptop — a user with a browser and MetaDraw open has perhaps 8 GB genuinely free, but the guard budgets 0.55 × ~15 GB ≈ 8.3 GB per index build and picks half the partitions needed. The run pages and appears to hang, which is the exact symptom the PR reports as unfixable-looking today. The 0.55 fraction absorbs some of this, but it is a fixed constant that was calibrated on a machine with nothing else running.

**Measured:**
Probed the two APIs directly on a 512 GB Windows box:

```
TotalAvailableMemoryBytes : 511.9 GB
MemoryLoadBytes (in use)  :  35.8 GB
GC.GetTotalMemory(false)  :   0.19 GB
AvailableBytes() reports  : 511.7 GB   <- what this PR uses
actually free            : 476.1 GB
```

So the reported figure tracks installed RAM, not free RAM, exactly as described above.

One note on the fix: `GC.GetGCMemoryInfo().MemoryLoadBytes` is the system-wide in-use figure and gives the right answer (`total - MemoryLoadBytes` = 476.1 GB above), so it is a cheaper option than P/Invoking the OS counters - but it reads **0 until the first GC has run in the process**. I hit exactly that: the same probe before forcing a collection reported `MemoryLoadBytes = 0.0 GB`, which would make the budget look like the whole machine. `TotalAvailableMemoryBytes` was already populated at that point, so whichever source you pick, it needs a guard for the not-yet-collected case.

**Suggested fix:**
Derive the budget from memory that is actually free rather than from installed RAM — on Windows the available-physical-memory counter, on Linux `MemAvailable` — and fall back to the current expression only when no such figure can be obtained. Alternatively, subtract the process working set rather than the managed heap and document that the fraction has to carry the entire other-process margin. Either way, state in the XML doc which of the two the number is, since `PartitionIncreaseWarning` prints it to the user as "GB available".

**Affected test scenarios:**
- AvailableBytes - other processes holding half of physical memory - reported availability must track free memory, not installed memory
- PartitionsForBudget - budget derived from a machine with 2 GB genuinely free of 16 GB installed - must raise partitions rather than fit the build into a nominal 8 GB

MetaMorpheus/EngineLayer/Indexing/IndexPartitioning.cs
```diff
@@ -0,0 +1,189 @@
...
+                }
+
+                sampledResidues += proteins[i].BaseSequence.Length;
+                foreach (var peptide in proteins[i].Digest(digestionParams, fixedModifications, variableModifications, silacLabels, turnoverLabels))
+                {
+                    sampledPeptides++;
+
+                    // count the fragments this peptide would actually contribute to the index, applying the
+                    // same mass window the indexing loop applies
+                    peptide.Fragment(commonParameters.DissociationType, digestionParams.FragmentationTerminus,
```
trishorts, line 200: **Major** - Footprint estimate counts zero fragments for Custom dissociation searches

**Problem:**
`EstimateIndexSize` calls `peptide.Fragment(commonParameters.DissociationType, ...)` directly, outside any engine. For `DissociationType.Custom`, mzLib resolves the product types from the process-global `DissociationTypeCollection.ProductsFromDissociationType[Custom]`, and the `CommonParameters` constructor (CommonParameters.cs:139-142) *empties* that global entry every time a `CommonParameters` is built. Every other fragmentation site in MetaMorpheus is reached through `MetaMorpheusEngine.Run()`, which restores it via `SetCustomProductTypes()` (MetaMorpheusEngine.cs:310); this new code path does not. In the two call sites that construct a `CommonParameters` immediately beforehand — `SetAllFileSpecificCommonParams` when file-specific settings exist (SearchTask.cs:383) and `CloneWithNewTerminus` in the semi/non-specific branch (SearchTask.cs:461) — the global list is deterministically empty at that moment, so `sampledFragments` stays 0.

**Impact:**
For a Custom-dissociation search the fragment term of `estimatedBytes` collapses to zero, halving or worse the estimated footprint, so the guard picks too few partitions and the run pages exactly as it did before this PR — the failure mode this feature exists to prevent, now silently.

**Suggested fix:**
Apply the custom product types before sampling (e.g. call `commonParameters.SetCustomProductTypes()` at the top of `EstimateIndexSize`, restoring the prior global value afterwards if the estimator must not leak state), or resolve the product-type list once from `commonParameters.CustomIons` and pass it in rather than relying on ambient global state.

**Affected test scenarios:**
- EstimateIndexSize - `DissociationType.Custom` with a non-empty `CustomIons` and the global Custom entry cleared - sampled fragment count is greater than zero
- SuggestTotalPartitions - identical database at `DissociationType.HCD` vs `Custom` with equivalent ion types - estimates within a small factor of each other
- SuggestTotalPartitions - called after `CloneWithNewTerminus` (semi search path) - fragment term still non-zero

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
```diff
@@ -376,17 +377,21 @@ protected override MyTaskResults RunSpecific(string OutputFolder, List<DbForTask
                 // modern search
                 if (SearchParameters.SearchType == SearchType.Modern)
                 {
-                    // Assume modern search is for proteins. 
+                    // Assume modern search is for proteins.
                     var proteinList = bioPolymerList.Cast<Protein>().ToList();
-                    for (int currentPartition = 0; currentPartition < combinedParams.TotalPartitions; currentPartition++)
+                    // scoped to indexing/searching only, so the settings the task reports stay as configured
+                    CommonParameters indexParams = RaisePartitionsToFitMemory(proteinList, combinedParams, fixedModifications,
```
trishorts, line 385: **Major** - Partition count is re-derived per spectra file, so it can change mid-run

**Problem:**
`RaisePartitionsToFitMemory` is called inside the `for (spectraFileIndex ...)` loop (SearchTask.cs:383 and 461, GlycoSearchTask.cs:119, XLSearchTask.cs:122), and its inputs are not stable across iterations: `AvailableBytes()` subtracts `GC.GetTotalMemory(false)`, which grows as `allPsms` accumulates and as each file's spectra are loaded. Nothing pins the first file's decision, so file 1 can be indexed at `TotalPartitions = 1` and file 5 at 4. The `ref bool alreadyWarned` protocol makes the change invisible: only the first raise is reported.

**Impact:**
Two concrete consequences. The index cache key includes the partition count (`IndexingEngine.ToString()` writes `Partitions: k/N`), so a changed `N` invalidates every cached partition and silently re-indexes the whole database for that file — minutes to hours on a mammalian database. And the PR itself documents that partition count moves `Delta Score`, `PEP` and `QValue`, so files within one run would be searched under different partitionings and their PSM-level statistics would no longer be comparable.

**Suggested fix:**
Compute the partition decision once per task, before the spectra-file loop, and reuse that `CommonParameters` for every file. If a per-file decision is genuinely wanted, warn on every change rather than only the first, and state the new count.

**Affected test scenarios:**
- RaisePartitionsToFitMemory - same database probed twice with materially different heap occupancy - partition count is stable across the run
- SearchTask - multi-file modern search - all files indexed with the same `TotalPartitions`
- RaisePartitionsToFitMemory - partition count raised again on a later file - a second warning is emitted naming the new count

MetaMorpheus/Test/IndexPartitioningTest.cs
```diff
@@ -0,0 +1,674 @@
...
+            bool warned = false;
+
+            var warnings = new List<string>();
+            void Collect(object sender, EngineLayer.StringEventArgs e) => warnings.Add(e.S);
+            MetaMorpheusTask.WarnHandler += Collect;
+            try
+            {
+                CommonParameters result = task.Probe(absurd, parameters, ref warned);
+
+                Assert.That(result.TotalPartitions, Is.EqualTo(256), "capped at MaxPartitions");
```
trishorts: **Major** - BeyondPartitionCap test fails on any machine with ~60 GB+ free RAM

**Problem:**
`RaisePartitionsToFitMemory_BeyondPartitionCap_WarnsThatItStillWillNotFit` builds a fixed-size synthetic database (4,000,000 references to a 4,901-residue protein) and then asserts `result.TotalPartitions == 256` and `warnings.Count == 2`. Both assertions hold only when `IndexPartitioning.PartitionsForBudget` sets `cappedByLimit`, i.e. when `estimatedBytes / (0.55 * AvailableBytes()) > 256`. With `MaxSampledPeptides = 50_000` the sample breaks after ~24 proteins (≈50,400 peptides, ≈1.31 M fragments) giving a scale factor of ~166,667, so the estimate is ≈8.7 × 10¹² bytes. That exceeds 256 × budget only while `AvailableBytes()` stays below roughly 60 GB. Nothing in the test pins the memory side of that inequality — it reads the real machine through `GC.GetGCMemoryInfo()`.

**Impact:**
On a 64 GB or 128 GB workstation (ordinary hardware for this lab) the guard returns ~125–251 partitions and emits one warning, so both `Is.EqualTo(256)` and `Is.EqualTo(2)` fail and the suite goes red for a machine-dependent reason unrelated to any code change. If `BytesPerPeptide`/`BytesPerFragmentEntry` are ever recalibrated the threshold moves again silently.

**Suggested fix:**
Drive the cap-and-second-warning assertion through the injectable decision path (`IndexPartitioning.PartitionsForBudget` with an explicit `availableBytes`, as `PartitionsForBudget_ReportsCapAndRespectsProteinCount` already does), or make `RaisePartitionsToFitMemory` accept an injectable budget for tests. If the end-to-end helper must be exercised, assert only that a second "may page heavily" warning appears *when* `cappedByLimit` is true, rather than asserting an exact partition count and warning count derived from the runner's RAM.

**Affected test scenarios:**
- RaisePartitionsToFitMemory_BeyondPartitionCap_WarnsThatItStillWillNotFit - runner with ≥ ~60 GB available memory - currently fails; must pass
- RaisePartitionsToFitMemory_BeyondPartitionCap_WarnsThatItStillWillNotFit - budget supplied explicitly so estimate/budget > 256 - 256 partitions and two warnings
- PartitionsForBudget - estimate exactly 256× budget vs 257× budget - cappedByLimit false then true

MetaMorpheus/Test/IndexPartitioningTest.cs
```diff
@@ -0,0 +1,674 @@
...
+            CommonParameters clone = original.CloneWithNewTotalPartitions(9);
+
+            Assert.That(clone.TotalPartitions, Is.EqualTo(9));
+            Assert.That(original.TotalPartitions, Is.EqualTo(3), "the original must not be mutated");
+
+            var mismatches = new List<string>();
+            foreach (PropertyInfo property in typeof(CommonParameters).GetProperties(BindingFlags.Public | BindingFlags.Instance))
+            {
+                if (property.Name == nameof(CommonParameters.TotalPartitions) || property.GetIndexParameters().Length > 0)
+                {
```
trishorts, line 715: **Major** - Clone test cannot detect eight dropped constructor arguments

**Problem:**
`CloneWithNewTotalPartitions_PreservesEveryOtherSetting` claims that "a parameter added to the constructor later, and forgotten here, fails this test rather than silently resetting to its default." That only holds for parameters the fixture sets to a non-default value. The `original` at line 324 leaves `productMassTolerance_LowRes`, `deconvolutionMassTolerance`, `listOfModsVariable`, `listOfModsFixed`, `precursorDeconParams`, `productDeconParams`, `diaParameters` and `fragmentationParams` at their constructor defaults, so dropping any of those from `CloneWithNewTotalPartitions` reproduces the identical default and the sweep sees no mismatch. Compounding this, the comparison falls back to `Equals(a?.ToString(), b?.ToString())`, and neither `ClassicDeconvolutionParameters` nor `Omics.Fragmentation.FragmentationParams` overrides `ToString` — both render as their type name, so even a *distinctively configured* instance replaced by a freshly defaulted one of the same type compares equal.

**Impact:**
A future edit that drops `fragmentationParams` or `precursorDeconParams` from the clone silently hands the indexing/search engines default fragmentation and deconvolution settings mid-search while this test still passes — the exact silent-reset failure the test was written to prevent.

**Suggested fix:**
Give every constructor parameter a value distinguishable from its default in the fixture (a `ProductMassTolerance_LowRes` other than `AbsoluteTolerance(0.35)`, a non-null `DIAparameters`, a non-default `FragmentationParams`, explicitly-constructed decon parameter objects with a distinctive charge/ratio, non-default mod lists), and compare reference-typed properties by a value-bearing projection (member-wise or JSON) rather than `ToString()` for types that do not override it. Assert the fixture's own premise — that each property differs from a default-constructed `CommonParameters` — so the test cannot become vacuous.

**Affected test scenarios:**
- CloneWithNewTotalPartitions_PreservesEveryOtherSetting - clone stops passing `fragmentationParams` - test fails
- CloneWithNewTotalPartitions_PreservesEveryOtherSetting - clone stops passing `productMassTolerance_LowRes` - test fails
- CloneWithNewTotalPartitions_PreservesEveryOtherSetting - every fixture property differs from a default-constructed instance - premise assertion passes

trishorts (commented):

acesnik:
🤖 Updated the title and rewrote the PR body. Flagging what changed, since the PR is a fair bit larger than when it was reviewed.

**Title** is now "Speed up modern-search index creation, and fix three bugs that made results depend on the partition count". The original title described only the perf work, which is now about half of it.

**Why it grew.** Partitioning harder to fit memory made an existing problem measurable: the partition count changed the search results. Since partitioning is a memory decision, reported identifications moved with how much RAM happened to be free. Rather than paper over that with a heuristic budget, I tracked it to source. It was three separate bugs:

1. Fine scoring stopped early, and the bound was accumulated per partition.
2. Peptides with an undefined mass sit at the front of every bin they occupy and read as "too heavy" to the coarse-scoring binary search, so it could clip real candidates off the window. Which bins held one depended on the partitioning.
3. The start of the precursor window used an upper-bound search where it needs a lower bound, so a run of equal masses on the window's lower edge collapsed to its last member. Equal masses are the normal case: the same sequence occurs in many proteins, and a reversed decoy carries its target's exact mass.

**This changes results, and finds more.** On the full mouse proteome, 13 scans gain a match and none loses one — 16,696 matched scans become 16,709, 24,111 PSM rows become 24,130, 12,256 peptides become 12,268. Protein groups are unchanged at 2,752. PEP and q-values move broadly, because the PEP model is trained on the run's own PSMs and one changed score retrains it; that is amplification of the above, not separate movement.

6 against 7 partitions on the full mouse now produces byte-identical AllPSMs, AllPeptides, AllQuantifiedPeaks, AllQuantifiedProteinGroups and percolator input. Before, all five differed.

**The Windows test failure was mine**, and it was the same class of problem flagged in the review: a test coupled to how much RAM the runner has. `SuggestTotalPartitions_SmallSlurmAllocation_RaisesPartitions` hard-coded a 512 MB allocation, which stopped forcing a raise once the fragment-entry cost was corrected from 20 B to 4 B. It now walks a range of allocations and asserts only what holds regardless: a tighter allocation never asks for fewer partitions, and somewhere in the range it asks for more.

**Coverage**, measured with coverlet on the changed code: `FragmentIndex` 100 %, `IndexPartitioning` 96.6 %, the CSR index build 97.7 %, and the ten `ModernSearchEngine` methods this branch touches 79.4 %, up from about 35 %. Each new regression test was checked by reverting the fix and confirming it fails.

**One thing found and deliberately not fixed:** a low-resolution CID search does not find the peptide its own fragments came from. The indexing side bins with `Math.Round` and the scoring side with a cast to `int`, so they disagree on exactly half of all bins — 15,000 of 30,000 at the default maximum fragment mass. That predates this branch. It needs a real LowCID dataset to fix and verify, which I do not have, so it is described in the body rather than guessed at here.

Replies to the five inline comments below.

MetaMorpheus/EngineLayer/Indexing/IndexPartitioning.cs
acesnik: 🤖 Fixed, and the suggested source was the right one — including the trap you flagged with it.

`AvailableBytes()` now derives free memory from `GC.GetGCMemoryInfo().MemoryLoadBytes` (`TotalAvailableMemoryBytes - MemoryLoadBytes`) rather than from installed RAM less the managed heap. The `Math.Max(total / 8, free)` floor is gone.

The "reads 0 until the first GC" case is handled explicitly: if `MemoryLoadBytes` comes back at or below zero, a gen-0 collection is forced and it is read again; if it is still unusable the budget falls back to `total / 4`, which is deliberately pessimistic rather than reporting the whole machine.

Two things came with it. A Slurm allocation, if present, clamps the figure — `SLURM_MEM_PER_NODE`, or `SLURM_MEM_PER_CPU` times `SLURM_CPUS_ON_NODE`, both in MB, with 0 meaning unlimited — less this process's own working set, since that already counts against the allocation. And `PartitionIncreaseWarning` now says "55 % of 5.6 GB free" rather than "available", so the printed number states which of the two it is.

Sizing against free memory is only defensible because the partition count no longer changes results; while it did, a budget that moved with machine load made two runs of the same search disagree. That is what the rest of this PR grew into.

On your test scenarios: the Slurm clamp is covered by `SchedulerMemoryLimitBytes_ReadsSlurmAllocation` and `SuggestTotalPartitions_SmallSlurmAllocation_RaisesPartitions`, and the budget arithmetic by the `PartitionsForBudget_*` tests, which take an explicit `availableBytes` so they do not read the runner. The first scenario — other processes holding half of physical memory — is not directly tested, because it would need `GC.GetGCMemoryInfo()` behind an injectable seam and I did not want to add one for a test. It was verified by hand instead: on this machine the guard reported 5.6 GB free against 64 GB installed and chose 7 partitions accordingly.

MetaMorpheus/EngineLayer/Indexing/IndexPartitioning.cs
acesnik: 🤖 Fixed as suggested, with the restore.

`EstimateIndexSize` now saves `ProductsFromDissociationType()[DissociationType.Custom]`, calls `commonParameters.SetCustomProductTypes()` before sampling, and restores the previous value in a `finally`, so the estimator does not leak the global state it needs.

Covered by `SuggestTotalPartitions_CustomDissociation_StillCountsFragments`, which clears the global Custom entry first — the state the two call sites you identified leave behind — and asserts the fragment term is non-zero.

The related scenario, that HCD and Custom with equivalent ion types estimate within a small factor of each other, is not asserted: the two resolve to different product-type sets, so any bound I picked would be a number I made up rather than one I measured.

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
acesnik: 🤖 Fixed: decided once per task.

`RaisePartitionsToFitMemory` takes a `ref int?` that carries the decision, initialised before the spectra-file loop and reused by every later file. All four call sites pass it — `SearchTask` in both the modern and the semi/non-specific branch, `XLSearchTask` and `GlycoSearchTask` — so a run cannot index file 1 at one partition count and file 5 at another, and the cache key cannot change mid-run.

`RaisePartitionsToFitMemory_DecisionAlreadyMade_IsReusedWithoutReDeriving` covers it.

The second half of your point has since been addressed at the source rather than by warning about it. The PR did document that partition count moves Delta Score, PEP and q-values; it no longer does. 6 against 7 partitions on the full mouse proteome now produces byte-identical AllPSMs, AllPeptides, AllQuantifiedPeaks, AllQuantifiedProteinGroups and percolator input, where every one of them differed before. Three separate bugs were behind that, described in the PR body.

MetaMorpheus/Test/IndexPartitioningTest.cs
acesnik: 🤖 Removed. The cap-and-second-warning assertion is now driven through `PartitionsForBudget` with an explicit `availableBytes`, as you suggested, so nothing in it reads the runner's memory. `PartitionsForBudget_ReportsCapAndRespectsProteinCount` covers the cap, the protein-count ceiling and the `cappedByLimit` flag; `PartitionWarnings_ReportTheRaiseAndWhetherItStillWillNotFit` covers the second warning.

Worth saying plainly that the general point caught a second instance later: `SuggestTotalPartitions_SmallSlurmAllocation_RaisesPartitions` hard-coded a 512 MB allocation and went red on the Windows runner once the fragment-entry cost was corrected from 20 B to 4 B, for exactly the reason you describe here — an assertion whose truth depended on both a calibration constant and the runner's RAM. It now walks a range of allocations and asserts the two properties that hold regardless: a tighter allocation never asks for fewer partitions, and somewhere in the range it asks for more.

MetaMorpheus/Test/IndexPartitioningTest.cs
acesnik: 🤖 Fixed, including the premise assertion.

Every constructor parameter now gets a value distinguishable from its default, the eight you listed included: `productMassTolerance_LowRes` is `AbsoluteTolerance(0.44)`, `diaParameters` is a distinct instance, `fragmentationParams` and both decon parameter objects are explicitly constructed, and the mod lists are non-empty. `CustomIons` is set by reflection rather than through the constructor, since the constructor reads it from a global dictionary rather than taking it as an argument — dropping it from the clone was in fact undetectable before, and is what this found.

The `ToString` fallback is gone. Reference-typed properties are compared by `ReferenceEquals`, so a distinctively configured instance replaced by a freshly defaulted one of the same type now fails; value types and strings compare by equality.

The test asserts its own premise, as you asked: any property whose fixture value is indistinguishable from a default-constructed `CommonParameters` is collected and fails the test with its own message, so the sweep cannot quietly become vacuous as parameters are added.

acesnik:
🤖 Correcting a claim I made in the PR body and in my earlier comment, and adding the reasoning behind the precursor-window fix.

**The correction.** I described this as finding more identifications. It does not. I counted matched scans without checking them at the FDR threshold, which was careless — the added matches are all weak ones. Measured on the full mouse proteome at 1 % FDR, targets only:

| | before | after |
|---|---|---|
| PSMs at q ≤ 0.01 | 14,292 | **14,292** |
| peptides at q ≤ 0.01 | 5,488 | **5,476** |
| protein groups | 2,752 | 2,752 |

PSM level is flat; peptide level loses 12 of 5,488, about 0.2 %. The 13 scans that gain a match are all above 1 % FDR (10 target, 3 decoy). At peptide level enough of the newly-scored candidates are decoys to tighten the q-value threshold, and 12 target peptides fall the wrong side of it.

Isolated per commit, only the precursor-window fix moves it. Confident PSMs are 14,292 at every step; confident peptides are 5,488 through CSR and the sentinel fix, then 5,476 once the window's lower bound is corrected.

Crosslink search shows the same shape, and there the outcome is cleaner: at 1 % FDR **the crosslinks do not move at all** — Inter 55, Intra 83, Loop 8 all unchanged — and the attrition is in the non-crosslinked classes, Single 229 → 221 and DeadEndH2O 62 → 61. Unfiltered, 13 interlinks are gained and 8 of them are decoys, which is what an unfiltered count of weak matches looks like against a half-decoy database.

**Why keep it.** The window's two ends are different questions — the end wants an upper bound, the start a lower bound — and both were taken from the same upper-bound search. Asked for the start, it returns the *last* entry at or below the bound, so a run of equal masses sitting exactly on the window's lower edge collapsed to its final member and the rest were never scored. Equal masses are the normal case: the same sequence occurs in many proteins, a reversed decoy carries its target's exact mass, and notch intervals are derived from a peptide mass so the lower edge lands on such a run routinely. The rest of the codebase already did this correctly — `CrosslinkSearchEngine.BinarySearchGetIndex` and `GlycoPeptides.BinarySearchGetIndex` both take the bitwise complement and walk back over the run — so this aligns modern search with the existing convention rather than inventing one, and a test now runs the two implementations against each other so they cannot drift apart.

The candidates it admits are genuinely inside the precursor window. Scoring them is what the code was always trying to do, and that some are decoys at low scores is FDR working rather than a defect. But it is a trade, and I would rather it were made knowingly than buried — so the PR body now leads with these numbers and the reasoning is written out in full there.

**Also corrected in the body:** the undefined-mass fix. An earlier revision dropped those peptides from the index on the reasoning that a mass comparison can never accept one. That is true of a tolerance acceptor but not of `OpenSearchMode`, which accepts anything, and an open search leaves both window bounds infinite so the whole bin is scored. Dropping them stopped open and glyco searches reporting peptides with an unknown residue. They are indexed again, and both searches now read an undefined mass as negative infinity, which is where `double.CompareTo` already sorts it.

MetaMorpheus/EngineLayer/Indexing/IndexPartitioning.cs
trishorts: Verified at head. `AvailableBytes()` is `TotalAvailableMemoryBytes - MemoryLoadBytes`, the `Math.Max(total / 8, free)` floor is gone, the read-0-before-first-GC case forces a gen-0 collect and falls back to `total / 4` rather than reporting the whole machine, and the warning says "free" rather than "available". The Slurm clamp is a bonus I had not asked for. Not testing the other-processes case rather than adding an injectable seam for it is the right call. Resolving.

MetaMorpheus/EngineLayer/Indexing/IndexPartitioning.cs
trishorts: Verified at head: `EstimateIndexSize` saves `ProductsFromDissociationType()[Custom]`, calls `SetCustomProductTypes()`, and restores in a `finally`; `SuggestTotalPartitions_CustomDissociation_StillCountsFragments` clears the global first, which is the state the two call sites leave behind. Agreed on dropping the HCD-vs-Custom bound — a made-up factor would have been worse than no assertion. Resolving.

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
trishorts: Verified at head: `decidedPartitions` is declared at `SearchTask.cs:271`, before the spectra-file loop at :280, and both the modern and semi/non-specific call sites take it by `ref`; `XLSearchTask` and `GlycoSearchTask` do the same. A run cannot index file 1 and file 5 under different partition counts, so the cache key is stable. Making the results partition-invariant at the source is a better answer than the warning I suggested. Resolving.

MetaMorpheus/Test/IndexPartitioningTest.cs
trishorts: Verified at head: `RaisePartitionsToFitMemory_BeyondPartitionCap_WarnsThatItStillWillNotFit` is gone, and the cap/second-warning properties are pinned through `PartitionsForBudget_ReportsCapAndRespectsProteinCount` and `PartitionWarnings_ReportTheRaiseAndWhetherItStillWillNotFit` with an explicit `availableBytes`. The `SmallSlurmAllocation` case you found afterwards is the same defect and the range-walk is the right shape for it. Resolving.

MetaMorpheus/Test/IndexPartitioningTest.cs
trishorts: Verified at head: every constructor parameter now takes a distinguishable value (`productMassTolerance_LowRes` at `AbsoluteTolerance(0.44)`, explicit decon and fragmentation params, non-empty mod lists), the `ToString` fallback is replaced by `ReferenceEquals` for reference types, and the premise assertion collects any property matching a default-constructed `CommonParameters` and fails naming them. `CustomIons` being set by reflection because the constructor reads it from a global — and having been undetectable before — is a better find than the one I reported. Resolving.

trishorts:
All five of my earlier findings are addressed at head, and CI is green. One coordination problem before this goes in, which is my fault rather than yours.

**This PR and #2758 independently implement the same data structure.** I opened #2758 today without noticing `FragmentIndex.cs` had been sitting here since the 18th. They are not stacked — `git merge-base` puts them both on master at `99711097`, so neither contains the other.

The two classes are the same design down to the member names:

| | #2742 `FragmentIndex` | #2758 `MassBinIndex` |
|---|---|---|
| layout | CSR: `BinStart[]` + `PeptideIds[]` | CSR: `_binStarts[]` + `_entries[]` |
| bin count | `Length => _binStart.Length - 1` | `Length => _binStarts.Length - 1` |
| accessor | `ReadOnlySpan<int> this[int bin]` | `ReadOnlySpan<int> this[int bin]` |
| empty bins | empty, not null | empty, not null |
| overflow | `MetaMorpheusException` past `int.MaxValue` | `MetaMorpheusException` past `int.MaxValue` |

They also touch the same fourteen files, including `IndexingEngine`, `IndexingResults`, `ModernSearchEngine`, `NonSpecificEnzymeSearchEngine`, `CrosslinkSearchEngine` and `GlycoSearchEngine`. Whichever merges second is not a rebase — it is rewriting its own half against a codebase that already has the other one.

Where they actually differ is the build, and it is worth knowing before choosing:

- **This PR** counts into the offset array, prefix-sums in place, then fills — two passes, serial. `peptides[peptideId].Fragment(...)` is called at `IndexingEngine.cs:179` and again at `:236`, so every peptide is fragmented twice.
- **#2758** fragments once into per-thread emission runs, `Parallel.For` over blocks, then a stable counting sort into the flat array.

So the layout win is common to both, and the build strategies are alternatives rather than complements.

This PR also carries the four partition-dependence bug fixes, which #2758 does not and which are the more valuable half of it. My suggestion is that **this one goes first on its own terms**, and I rebuild #2758 as a build-strategy change on top of `FragmentIndex` — dropping my class entirely rather than asking you to drop yours. I'll close or convert #2758 accordingly; nothing is needed from you here.

Flagging it on this PR only so the collision is on the record where the reviewer of either will see it.

trishorts:
Following up on my comment above, and superseding the last paragraph of it.

I said there that I'd take this one first and rebuild #2758 on top of `FragmentIndex`. That was me deciding it unilaterally, and it isn't mine to decide alone. Withdrawing that.

**These two need to become one thing before either merges. Pick one and close the other, or fold them into a single PR.** What is not acceptable is both staying open — whichever merges second is rewriting its half against a codebase that already has the other one, in 14 shared files, and reviewers of either are reviewing a design decision that was already made somewhere else.

The facts to decide on, all verified:

- Not stacked. `git merge-base` puts both on master `99711097`.
- Same data structure, down to the member names: CSR layout, `Length => binStart.Length - 1`, `EntryCount`, `ReadOnlySpan<int> this[int bin]`, empty-not-null bins, `MetaMorpheusException` past `int.MaxValue`.
- The layout win is common to both. The only real difference is the build: this PR counts then fills, serial, fragmenting every peptide twice (`IndexingEngine.cs:179` and `:236`); #2758 fragments once into per-thread emission runs, `Parallel.For` over blocks, stable counting sort.
- This PR additionally carries the four partition-dependence bug fixes. #2758 does not, and they are the more valuable half of the two.

My read is that one PR carrying this PR's bug fixes plus #2758's build is strictly better than either alone, but the call is yours. Same comment on #2758.

trishorts:
Resolved from my side — you don't need to pick anything.

I've closed #2758. Its CSR layout is dropped; `FragmentIndex` here is the one that survives, and this PR is unaffected by anything I did.

The two parts of #2758 that weren't duplicated went elsewhere:

- **[acesnik/MetaMorpheus#8](https://github.com/acesnik/MetaMorpheus/pull/8)**, opened against `perf/index-creation-412` — the parallel index build, applied to your `FragmentIndex`, with `FragmentIndex.cs` untouched. Entirely optional, and it can equally land after this merges.
- **#2761**, against master — the per-scan scoring table and coarse-score sort, which never touch the index.

One number from #8 that bears on this PR whether or not you take the PR itself: count-then-fill calls `Fragment(...)` at `IndexingEngine.cs:179` and again at `:236`, so the whole database is fragmented twice. On the full reviewed human proteome that is 47.1 s and 51.2 GB allocated, against 23.4 s and 34.1 GB fragmenting once. Same 6,907,559 peptides and 289,901,771 entries either way.

Also worth flagging for whoever merges second: #2761 edits the same four engine files this PR does. The conflict is narrow — the `scoringTable` parameter type on three signatures — but it exists.

nbollis (dismissed):

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -1311,25 +1356,158 @@ private static DigestionParams GetDigestionParamsFromFile(string path)
...
+        // 2: the fragment index went from bin counts + concatenated ids to the compressed sparse row pair
+        private const int FragmentIndexFormatVersion = 2;
+
+        /// <summary>
+        /// The two arrays behind a <see cref="FragmentIndex"/>, written as raw little-endian int32 in bulk.
+        /// NetSerializer walked every list and every element individually; a fragment index has millions of
+        /// bins and hundreds of millions of entries, so the per-object cost dominated. Now that the in-memory
+        /// form is already flat, reading it back is a pair of array fills rather than a rebuild.
+        /// </summary>
+        private static void WriteFragmentIndex(FragmentIndex fragmentIndex, string fragmentIndexFileName)
```
trishorts, line 1374: **The serialized fragment-index format changes, and nothing marks it as changed.**

`WriteFragmentIndex`/`ReadFragmentIndex` go from `List<int>[]` to `FragmentIndex`, and the NetSerializer `messageTypes` set changes with them. But `IndexEngineParamsFileName` is untouched, and `SameSettings` compares the *parameters* text, not the payload format. So a `DatabaseIndex` folder written by any previous MetaMorpheus version still satisfies `CheckFiles` (`:1368-1372`) -- the three files exist and the settings match -- and is then handed to a deserializer that cannot read it.

In `GenerateIndexes` this self-heals: the read throws, the `catch` at `:1424` swallows it, and a fresh index is built and written to a new timestamped folder. So users are not broken, which is presumably why this has not shown up. But the cost is a silent full re-index on the first run after upgrading, with no explanation, plus a stale folder left behind forever -- `GenerateOutputFolderForIndices` always mints a new one rather than reusing the unreadable one.

Two things worth considering:

- **Put a format version in `indexEngine.params`** so `SameSettings` rejects stale folders cleanly, rather than the rejection happening by exception several layers down. That turns "mystery re-index" into a normal cache miss, and it would keep working for the next format change.
- **At minimum, warn.** #2748 is adding exactly this warning (`"Could not read the existing peptide index, so it is being rebuilt. Reason: ..."`) to `GenerateIndexes_PeptideOnly`. The same line in `GenerateIndexes` would cover this case, and this PR is the reason it becomes common.

Worth confirming this was considered and is intended, rather than being a consequence nobody priced -- given the PR is otherwise so careful about what moves and what does not, I suspect it was, and just is not written down.

trishorts (commented):
Following up on my earlier threads -- all five of those are verified fixed at head, and I have said so on each. This adds one thing none of them covered, about the on-disk index rather than the in-memory one.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
acesnik: 🤖 Old index folders are already rejected — but you were right that something here is unpriced, just not that.

**The rename does the version check for released versions.** The index file names changed in the same commit as the format (`MetaMorpheusTask.cs:222-225`):

```
fragmentIndex.ind        -> fragmentIndex.bin
secondFragmentIndex.ind  -> secondFragmentIndex.bin
precursorIndex.ind       -> precursorIndex.bin
```

A `DatabaseIndex` folder from any released version fails `CheckFiles` at `File.Exists(FragmentIndexFileName)` and is never handed to a deserializer — a normal cache miss at the existence check, not an exception several layers down. `peptideIndex.ind` keeps its name because its format is unchanged, and a downgrade works the same way in reverse. The payload is self-describing as well: a 4-byte magic (`MMFI`/`MMPI`) plus a format version, checked in both readers at `:1395` and `:1467`, with `FragmentIndex_ReadRejectsForeignFile` (`IndexPartitioningTest.cs:1936`) covering foreign magic, a stale version, and a precursor index in the fragment index slot.

I think `:1368-1372` and `:1424` are master's line numbers — master's `CheckFiles` is at 1366 and its silent catch at 1424, where at head they are 1544 and 1612. That would explain why the rename and the header were not visible in what you read.

**What is real.** Two things, and the second one is a crash.

1. A folder that passes `CheckFiles` but fails to read is re-picked, not retired. `GetExistingFolderWithIndices` returns the *first* folder that passes, `GenerateOutputFolderForIndices` always mints a new timestamped one, and nothing deletes or marks the bad folder. Depending on directory enumeration order that is a silent re-index on every run rather than once. Pre-existing — reachable on master through any unreadable `peptideIndex.ind` — but a version-checked payload makes it a normal thing to hit rather than a rare one.

2. `GenerateSecondIndexes` (`:1695`) reads the second fragment index with no try/catch, so it has no self-healing path at all. Same shape as master. The rename protects released versions but not this branch's own history: `7a070247` wrote `FragmentIndexFormatVersion = 1` to `secondFragmentIndex.bin` and `edbaf5aa` bumped it to 2 under the same name, so anyone who ran the branch before Aug 19 has a v1 `.bin` cached. `GenerateIndexes` throws on that and recovers; `GenerateSecondIndexes` throws and does not — a crosslink search over data whose child scans use a different dissociation type crashes outright. (`XLSearchTask.cs:157` is its only caller; I said non-specific and semi-specific in the first version of this comment, which was wrong.) Narrow blast radius, since no released version can produce that file, but the same shape recurs on any future bump that keeps a name.

**Proposal** — your first bullet, with the version living in the payload rather than in `indexEngine.params`:

- **Validate the header in `CheckFiles`**, not just file existence: two small reads per candidate folder. That makes a stale folder a genuine cache miss, which fixes both problems at once — the stale folder stops being re-selected, and `GenerateSecondIndexes` is handed a fresh folder instead of an unreadable file. `SameSettings` is string equality over `indexEngine.ToString()`, so I would rather the format version sit next to the bytes it describes, but you are right that the check belongs at folder-selection time.
- **Take #2748's warning** into both `catch` blocks, so the read failures the header check cannot anticipate say why they are rebuilding.
- **Test it**: a folder whose `fragmentIndex.bin` carries a v1 header is not selected, and a second-index search over that folder completes rather than throwing.

Happy to do those here, or split them out if you would rather this merge on what it already does — the released-user path is correct either way, and the crash is only reachable from this branch's intermediate builds.

acesnik: 🤖 Done in `12ac96f6`.

- `CheckFiles` now reads the magic and format version out of each binary index before selecting the folder — `fragmentIndex.bin` always, `precursorIndex.bin` when the engine asks for one, and `secondFragmentIndex.bin` when it is present, since that one is written on demand. Anything that stops it reaching the header, including a truncated or locked file, is a miss.
- The silent `catch` in `GenerateIndexes` now warns with #2748's wording. I left `GenerateIndexes_PeptideOnly` alone, contrary to what I said above: #2748 rewrites that method and adds the warning there, so a second copy here would only conflict.
- `Test/IndexCacheHeaderTest.cs`, seven cases — the header check accepting a current index and rejecting a previous version, a foreign magic and a header too short to read; the precursor index validated only when required; the second index absent versus stale; and the crosslink sequence end to end over a cache aged to the previous version.

Two notes on the tests, since neither was obvious.

The end-to-end case passed against the unfixed code on the first attempt. It was leaning on `Directory.GetDirectories` order: with both indexes aged, `GenerateIndexes` rebuilds into a second folder, and whether the stale one gets picked again is filesystem-dependent — on APFS it was not. It now ages only the second index, so the folder still reads clean in `GenerateIndexes`, stays the only folder there is, and the stale read is forced. Against the unfixed code it fails with the actual crash:

```
MetaMorpheusException: .../secondFragmentIndex.bin is not a fragment index this version can read.
   at TaskLayer.MetaMorpheusTask.ReadFragmentIndex(String fragmentIndexFileName)
   at TaskLayer.MetaMorpheusTask.GenerateSecondIndexes(...)
```

Six of the seven fail with the header check mutated away; the seventh is the positive control. `IndexEngineTest` and `IndexPartitioningTest` are unaffected — 72 green in total, though that is a net10.0 harness rather than `Test.csproj`, which needs Windows, so CI is the real check.

One thing I did not change: `GenerateSecondIndexes` still reads without a try/catch. With the folder retired at selection time that is no longer reachable from a format change, and adding recovery there is wider than this thread needs — happy to make it belt and braces if you would rather.

MetaMorpheus/TaskLayer/GlycoSearchTask/GlycoSearchTask.cs
```diff
@@ -112,29 +113,35 @@ protected override MyTaskResults RunSpecific(string OutputFolder, List<DbForTask
                 
                 myFileManager.DoneWithFile(origDataFile);
 
-                for (int currentPartition = 0; currentPartition < CommonParameters.TotalPartitions; currentPartition++)
+                // scoped to indexing/searching only, so the settings the task reports stay as configured.
+                // every TotalPartitions read below comes from indexParams, including the loop bound and the
+                // protein-range slicing, which previously read from two different objects.
+                CommonParameters indexParams = RaisePartitionsToFitMemory(proteinList, combinedParams, fixedModifications,
```
trishorts, line 119: **Major** - The memory guard can raise glyco partitions, but glyco results still depend on partition count

The PR says the guard "is only safe because partition count no longer changes results", but that doesn't hold for glyco search. `GlycoSearchEngine.RunSpecific` applies its `GlycoSearchTopNum` cut per partition (the `OrderByDescending(p => scoringTable[p])` loop), so N partitions send up to N×TopN coarse candidates to glycan matching instead of TopN.

I checked by running it: GlycoSnip.toml with `GlycoPepMix_snip.mzML`, the 7-protein fasta plus human reviewed UniProt, `GlycoSearchTopNum = 5`, at `TotalPartitions` 1 and 4:
- AllPSMs rows: 13 vs 10
- scan 115: `LPPNVVEESAR` at 1 partition, `TTPPT[O-linked glycosylation:H1N1A1 on T]TATPIR` at 4

(The 7-protein fixture alone is identical at 1 vs 7, even with TopN = 1. It never has more than TopN candidates per scan, which is probably why the PR's glyco comparison moved only one `Rank`.)

Since the guard picks the count from free memory, the same glyco search can now report different glycopeptides on different machines. Suggest keeping the guard warn-only for glyco until the TopN cut is partition-independent, and adding a 1-vs-N partition test on a database large enough to exceed TopN.

MetaMorpheus/EngineLayer/Indexing/FragmentIndexBuilder.cs
```diff
@@ -0,0 +1,167 @@
...
+                _peptideCount++;
+                _peptideOpen = false;
+            }
+        }
+
+        private static void Append(ref int[] array, int index, int value)
+        {
+            if (index == array.Length)
+            {
+                Array.Resize(ref array, array.Length * 2);
```
trishorts: **Major** - Emission buffer doubling overflows past 2^30, and the guard never models the int.MaxValue entry cap

Two related gaps in the new flat index limits:

1. `Append` grows with `Array.Resize(ref array, array.Length * 2)`. Once a block's buffer reaches 2^30, `Length * 2` wraps to `int.MinValue`. I checked: `Array.Resize` with that size throws `ArgumentOutOfRangeException: Non-negative number required. (Parameter 'newSize')`, not the `MetaMorpheusException` that `Build` throws for the same situation. With `MaxThreadsToUsePerFile = 1`, one block holds every emission, so this fires at about 1.07 B fragment entries, before the `totalEntries > int.MaxValue` check in `Build` is reached.
2. `IndexPartitioning.PartitionsForBudget` sizes only by bytes against free memory. On a large-memory node it keeps one partition even when the estimated fragment count exceeds `int.MaxValue`, so the guard doesn't prevent the "Too many fragments to index; try ... more partitions" failure. The old `List<int>[]` index had no such cap.

Suggest overflow-safe growth capped at `Array.MaxLength` that throws the same `MetaMorpheusException`, and a per-partition entry ceiling (with margin) in the partition suggestion.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -1386,19 +1564,52 @@ private static string GetExistingFolderWithIndices(IndexingEngine indexEngine, L
...
                 return folder.FullName;
             }
             return null;
         }
 
+        /// <summary>
+        /// The magic and format version the readers check, read at folder-selection time. Any failure to get
+        /// at them is a miss, since the reader would fail on the same file.
+        /// </summary>
+        private static bool HasReadableIndexHeader(string indexFileName, int expectedMagic)
```
trishorts, line 1631: **Major** - Header check still accepts a truncated index file

`HasReadableIndexHeader` reads only magic + version (8 bytes). `WriteFragmentIndex` and `WritePrecursorIndex` write the header first and then the payload straight to the final file name, so an interrupted write (killed process, full disk) leaves a file that passes `CheckFiles`. The "CheckFiles accepts a current fragment index" test case already accepts a file that `ReadFragmentIndex` can't load.

What follows:
- `GenerateIndexes` catches the read failure and rebuilds into a new timestamped folder. The broken folder sorts first by name, though, so every later run selects it and rebuilds again.
- `GenerateSecondIndexes` has no catch (`ReadFragmentIndex` at line 1800), so a truncated `secondFragmentIndex.bin` still crashes crosslink search. That is the case the new doc comment says this check exists for.

This isn't a regression, since the base check was existence-only, but the new check doesn't close the gap it describes. Either compare the header's lengths against the file size (`16 + 4*(header[2] + header[3])` for the fragment index), or write to a temp name and `File.Move` it into place after flushing.

trishorts (commented):
Automated code review - 3 finding(s) at severity >= Major, posted as inline comments.

MetaMorpheus/EngineLayer/Indexing/FragmentIndex.cs
```diff
@@ -0,0 +1,51 @@
...
+        public int Length => _binStart.Length - 1;
+
+        /// <summary>Total peptide ids across all bins.</summary>
+        public int EntryCount => _peptideIds.Length;
+
+        /// <summary>The peptide ids in a bin, ascending. Empty when nothing fragments into that bin.</summary>
+        public ReadOnlySpan<int> this[int bin] => _peptideIds.AsSpan(_binStart[bin], _binStart[bin + 1] - _binStart[bin]);
+
+        /// <summary>Bin offsets; exposed so the index can be written and read as raw arrays.</summary>
+        public int[] BinStart => _binStart;
```
nbollis, line 46: [review-oop] Immutable-by-convention index leaks its mutable backing arrays

FragmentIndex stores _binStart/_peptideIds as readonly fields but exposes them via public getters returning the live arrays. Nothing enforces the documented write-once-read-many contract: a caller can write into the backing arrays and corrupt every ReadOnlySpan returned by the indexer, including spans held concurrently by parallel modern-search workers. A shared, concurrently-read core structure turns an internal representation choice into a data-integrity hazard at the public API surface. (Also reported by review-generic.)

Suggestion: Make BinStart/PeptideIds internal (serialization is in the same assembly) or return a read-only projection; keep the public surface to the indexer/Length/EntryCount.

MetaMorpheus/EngineLayer/ModernSearch/ModernSearchEngine.cs
```diff
@@ -231,68 +231,132 @@ protected void IndexScoreScan(Ms2ScanWithSpecificMass scan, byte[] scoringTable,
...
+            while (low <= high)
+            {
+                int mid = low + ((high - low) / 2);
+
+                if (MassForBinSearch(peptideIndex[bin[mid]].MonoisotopicMass) <= peptideMassToLookFor)
                 {
-                    l = m + 1;
+                    result = mid;
+                    low = mid + 1;
                 }
```
nbollis, line 343: [review-backend] FineScorePeptides now fine-scores every coarse candidate unconditionally; search-speed regression risk

Removing the early-break bound makes results partition-independent (the intended fix) but every peptide reaching the fine-scoring list is now fully re-scored for every scan. The 'costs nothing' claim is specific to tryptic mouse runs; open/tolerant mass-diff acceptors leave both mass bounds infinite so far more peptides are fine-scored per scan. Consensus across review-backend and review-generic.

Suggestion: Keep the partition-independent loop but retain a safe, partition-independent cap, or at minimum document worst-case candidates-per-scan growth for open/tolerant searches.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -1334,25 +1379,158 @@ private static DigestionParams GetDigestionParamsFromFile(string path)
...
+
+            WriteInt32Bulk(file, fragmentIndex.BinStart);
+            WriteInt32Bulk(file, fragmentIndex.PeptideIds);
+        }
+
+        private static FragmentIndex ReadFragmentIndex(string fragmentIndexFileName)
         {
-            var messageTypes = GetSubclassesAndItself(typeof(List<int>[]));
-            var ser = new NetSerializer.Serializer(messageTypes);
+            using var file = new FileStream(fragmentIndexFileName, FileMode.Open, FileAccess.Read, FileShare.Read, 1 << 20);
```
nbollis, line 1441: [review-backend] Index readers allocate arrays from unchecked header counts before validating against file size

ReadFragmentIndex/ReadPrecursorIndex trust header[2]/header[3] directly to size arrays; CheckFiles only validates magic and format version. A cached index with intact magic but corrupted count fields allocates arrays up to int.MaxValue (multi-GB) before any read failure is caught.

Suggestion: Validate header counts against stream length and sanity-cap before allocating; throw MetaMorpheusException so the cache is treated as a miss.

MetaMorpheus/EngineLayer/Indexing/IndexPartitioning.cs
```diff
@@ -0,0 +1,333 @@
...
+
+            // No usable in-use figure: fall back to a deliberately pessimistic slice rather than reporting the
+            // whole machine, which would under-partition exactly when the guard matters most.
+            long free = inUse > 0 && inUse < total ? total - inUse : total / 4;
+
+            long schedulerLimit = SchedulerMemoryLimitBytes();
+            if (schedulerLimit > 0)
+            {
+                // our own footprint already counts against the allocation
+                free = Math.Min(free, Math.Max(schedulerLimit / 4, schedulerLimit - Environment.WorkingSet));
```
nbollis, line 316: [review-backend] Scheduler-limit clamp can over-report free memory when the process working set is at/over its allocation

free = Math.Min(free, Math.Max(schedulerLimit / 4, schedulerLimit - WorkingSet)). When the working set already equals/exceeds the allocation (the exact memory-pressure case the guard exists for), the floor schedulerLimit/4 wins, reporting up to 25% of the allocation as free. That under-partitions precisely when the guard matters most.

Suggestion: When schedulerLimit - WorkingSet <= 0, report no partition fits (clamp to smallest viable count / set cappedByLimit) rather than falling back to a quarter of the allocation as free.

MetaMorpheus/EngineLayer/Indexing/IndexPartitioning.cs
```diff
@@ -0,0 +1,333 @@
...
+                totalResidues += protein.BaseSequence.Length;
+            }
+
+            // DissociationType.Custom resolves its product types from a process-global list that the
+            // CommonParameters constructor clears, and only MetaMorpheusEngine.Run() restores it. This
+            // samples outside any engine, so without restoring it here a Custom search would fragment into
+            // an empty product list, count zero fragments, and under-partition — silently, and precisely in
+            // the case the guard exists for. Put the global back afterwards so the estimate leaks no state.
+            bool isCustom = commonParameters.DissociationType == DissociationType.Custom;
+            List<ProductType> previousCustomProducts = null;
```
nbollis, line 210: [review-oop] EstimateIndexSize reaches into process-global state and does a save/restore dance

New static sampling code reaches into the process-global DissociationType.Custom product list and reassigns it in a finally so a Custom search isn't under-partitioned. Correct today but fragile: depends on SetCustomProductTypes replacing rather than mutating the dictionary value, and on the estimate winning any race with other config code touching the same global.

Suggestion: Pass the custom product list in as a parameter instead of mutating/restoring the global dictionary entry; at minimum guard that the restore re-attaches the exact prior list.

MetaMorpheus/EngineLayer/CommonParameters.cs
```diff
@@ -296,6 +296,65 @@ public CommonParameters CloneWithNewTerminus(FragmentationTerminus? terminus = n
                                 RTPredictorName);
         }
 
+        /// <summary>
+        /// Copy with a different TotalPartitions. Returns a new instance rather than mutating, because
+        /// SetAllFileSpecificCommonParams hands back the task's own CommonParameters when a file has no
+        /// file-specific settings — mutating that would rewrite the settings the task reports to the user.
+        /// DigestionParams is shared rather than cloned so the digestion identity peptides carry is unchanged.
+        /// </summary>
+        public CommonParameters CloneWithNewTotalPartitions(int totalPartitions)
```
nbollis, line 305: [review-generic] CloneWithNewTotalPartitions is a hand-maintained field-by-field constructor fork that silently drops future fields

CloneWithNewTotalPartitions re-lists ~40 constructor args; any future field stored as a settable property rather than a ctor param (like CustomIons, already a special case) is silently dropped and diverges the clone from the task's reported settings. The new partition guard distributes these clones into every search path, so a dropped field propagates without a compile error.

Suggestion: Use a private copy constructor or reflective/serial bulk copy shared with CloneWithNewDissociationType, and add a test comparing all public properties to catch future divergence.

MetaMorpheus/GUI/GUI.csproj
```diff
@@ -17,6 +17,8 @@
     <DebugSymbols>true</DebugSymbols>
     <ApplicationIcon>Icons\MMnice.ico</ApplicationIcon>
 	<EnableWindowsTargeting>true</EnableWindowsTargeting>
+    <!-- Indexing allocates tens of millions of short-lived objects; workstation GC cannot keep up. -->
+    <ServerGarbageCollection>true</ServerGarbageCollection>
```
nbollis, line 21: [review-infra] Server GC forced on an interactive WPF client

ServerGarbageCollection=true on the GUI allocates a dedicated managed heap per logical processor, raising resident memory and is generally discouraged for interactive desktop apps. It can increase end-user RAM and add UI-thread latency while only coincidentally speeding up indexing, and partially works against the PR's own memory-bounding guard (IndexPartitioning estimates a doubled 1500 B/peptide under Server GC).

Suggestion: Verify GUI peak memory and UI responsiveness at typical thread counts; consider limiting Server GC to the CMD project or documenting the desktop memory tradeoff.

MetaMorpheus/EngineLayer/Indexing/IndexPartitioning.cs
```diff
@@ -0,0 +1,333 @@
...
+        /// </summary>
+        private const double MemoryBudgetFraction = 0.55;
+
+        /// <summary>
+        /// Heap bytes per peptide in the peptide index. Measured on mouse at 9.21 M peptides: 6,926 MB of
+        /// live heap after digestion, i.e. ~750 B/peptide. Server GC keeps far more garbage before
+        /// collecting, so the same work occupies about twice the footprint and the estimate has to know
+        /// which mode it is running in.
+        /// </summary>
+        private static long BytesPerPeptide => System.Runtime.GCSettings.IsServerGC ? 1500 : 750;
```
nbollis, line 36: [review-infra] Memory-guard calibration keyed off host GC mode diverges between tests and shipped executables

BytesPerPeptide doubles (750->1500) based on GCSettings.IsServerGC, which only becomes true in CMD/GUI after the csproj change; Test.csproj does not set Server GC, so the 2150-line IndexPartitioningTest suite validates only the workstation-GC branch while production ships the Server-GC path. Because EngineLayer is a library, the effective budget now depends on the hosting executable's GC mode. (Also reported by review-oop.)

Suggestion: Have IndexPartitioningTest enable Server GC (test host runtimeconfig/gcServer) so the shipped branch is exercised, or decouple the budget from host GC mode.

nbollis (commented):
Automated multi-specialist review of head 887f1ece by review-backend, review-oop, review-generic, review-infra.

## TL;DR
Strong, well-tested correctness + performance PR. The CSR fragment index (FragmentIndex/FragmentIndexBuilder) and the parallel block-emission build are internally consistent and heavily tested; the four partition-dependence fixes (fine-scoring early stop, NaN-massed peptides, precursor-window lower bound, psmtsv insertion order) are real and each has a regression test verified by reverting the fix. No reviewer found a correctness defect in the index build or search math. The earlier inline flag that AvailableBytes() measured installed RAM is genuinely addressed at head — it now uses GC.GetGCMemoryInfo().MemoryLoadBytes with a forced-first-collection guard and is clamped by the Slurm allocation minus process working set.

## Findings posted inline (8)

- FragmentIndex.cs:46 (major) — Immutable-by-convention index leaks its mutable backing arrays
- ModernSearchEngine.cs:457 (minor, consensus of 2 reviewers) — Fine-score-everything removes the early-exit bound; open/tolerant searches may fine-score far more per scan
- MetaMorpheusTask.cs:1423 (minor) — on-disk index readers allocate arrays from unchecked header counts before validating against file size
- IndexPartitioning.cs:268 (minor) — scheduler-limit clamp over-reports free memory when the process working set is at/over its allocation
- IndexPartitioning.cs:162 (minor) — EstimateIndexSize reaches into process-global DissociationType.Custom state with a fragile save/restore
- CommonParameters.cs:305 (minor) — CloneWithNewTotalPartitions is a hand-maintained ~40-arg constructor fork that silently drops future non-constructor fields
- GUI.csproj:21 (minor) — Server GC forced on an interactive WPF client; per-CPU heaps raise memory/latency and partially work against the PR's own memory guard
- IndexPartitioning.cs:36 (suggestion) — memory-guard per-peptide calibration keyed off host GC mode; Test.csproj does not set Server GC so the 2150-line suite exercises only the workstation-GC branch

trishorts (approved):

nbollis (approved):

RayMSMS (approved):

PR #2743, opened 2026-08-18
Split the Windows test job across two runners (21m -> ~13m)

acesnik:
🤖 `Test on Windows` takes 21 minutes. This splits it across two runners and takes the wall clock to ~13.

## Measured, not estimated

Three jobs on one commit on a throwaway branch — the current single job as a control, plus the two halves:

| Job | Tests | Test step | Job total |
|---|---|---|---|
| Current single job | 1762 | 1062s | **21m01s** |
| `Test on Windows (glyco)` | 50 | 503s | **11m37s** |
| `Test on Windows (core)` | 1712 | 536s | **12m13s** |

503s + 536s = 1039s against the single job's 1062s, so test time is conserved rather than duplicated.

Where the 21 minutes went: the test step is 1062s of it, and one test is 324s of that — `OGlycoTest_Run5_WriteDecoys` copies a 1.8 MB mzML 19 times and searches all 20 to manufacture enough PSMs for one decoy to survive q≤0.01. That single test is 32% of the suite. Fixed per-job overhead is ~188s (checkout 15s, setup-dotnet 35s, restore 31s, build 101s).

## Why this partition

`--filter` on `FullyQualifiedName~Test.TestOGlyco` and its complement. Exhaustive by construction — a test either contains that substring or it doesn't, so nothing can fall between the halves. The counts in the table above corroborate that on the commit they were measured on: 50 + 1712 = 1762, matching the single job.

Those counts drift as tests are added upstream, and have: the run on this PR reports 50 + 1717, the extra five coming from `MzIdentMlSchemaTest.cs`, `TestCmd.cs` and `MassDiffAcceptorTypeTests.cs` landing on master since the experiment. So the count is corroboration at a point in time, not a standing check — which is the weakness, because a partition that silently drops tests is invisible in a green run.

That is what the runsettings guard is for; see below. It turns the drop scenario into a red build on every run, rather than relying on someone re-deriving two numbers from two job logs.

`TestOGlyco` is the half pulled out for two reasons. It holds the 324s test. And it is **not** one of the fixtures using `EverythingRunnerEngineTestCase`, so neither runner rebuilds a cached search the other already built — splitting `PostSearchAnalysisTaskTests` or the MetaDraw fixtures instead would have paid for `BottomUpQValue` twice, since four files share it.

## This is not NUnit parallelism

Separate jobs, separate VMs, separate filesystems. Each runner still executes its own tests **sequentially in one process**, so tests that share an output folder never run concurrently — `TESTGlycoData` is used by five test files, 24 times in `TestOGlyco.cs` alone, each ending in `Directory.Delete(outputFolder, true)`. In-process parallelism is not safe in this suite and this does not introduce any; there are already four `[NonParallelizable]` attributes from previous encounters with it.

Across both experiment runs, **no test failed in either half**, so no cross-fixture state dependency surfaced.

## Two details that would otherwise break

- **Both halves upload to Codecov.** It merges uploads for the same commit. If only one uploaded, that half would be reported as the whole suite and the project/patch percentages would collapse.
- **A `Test on Windows` aggregator job** depends on both halves and fails if either did. It uses `!cancelled()` rather than `always()`, so a cancelled run reports as cancelled instead of claiming a half failed. An existing required status check of that name keeps working with no branch-protection change. If you would rather require the two halves directly, delete that job and update the settings instead.

## Guarding the partition

A `--filter` that matches nothing exits 0. Renaming `TestOGlyco` or mistyping either filter string would silently stop running that half with every job still green, which is exactly the failure mode the partition section calls out. Of the three candidate mechanisms, only a runsettings file works — measured on the .NET 10 SDK:

| mechanism | zero-match filter | matching filter |
|---|---|---|
| plain `dotnet test --filter` | exit 0 | exit 0 |
| `-p:TreatNoTestsAsError=true` | exit 0 (no effect) | exit 0 |
| `--settings` with `TreatNoTestsAsError` | **exit 1** | exit 0 |

`.github/no-empty-test-run.runsettings` is passed to both halves. Confirmed against the production compound filters rather than simplified ones: the glyco half with its fixture renamed exits 1, both real halves still pass.

`codecov.yml` sets `after_n_builds: 2` so no status is computed off whichever upload lands first — `Test.yml` is the only workflow that uploads, and it has exactly two upload steps — and both uploads set `fail_ci_if_error: true` so a lost upload reddens its own half instead of silently halving reported coverage.

## Known interaction, worth stating plainly

There is a pre-existing intermittent failure where coverlet's instrumentation access-violates before any test runs:

```
Fatal error.  0xC0000005
   at Mono.Cecil.Pdb.ISymUnmanagedWriter2.Close()
   at Coverlet.Core.Instrumentation.Instrumenter.InstrumentModule()
   at Coverlet.Collector.DataCollection.CoverletCoverageCollector.OnSessionStart(...)
```

It hit 1 of the last 40 `Test.yml` runs on master (`truncation-search-mm`, 1824 passed / 1825 total, zero test failures, job red), and it hit one of the two experiment runs. **Two jobs means two chances per run to hit it, so this change roughly doubles exposure to it** — ~2.5% to ~5% at the observed rate.

That is an argument for fixing the flake, not for keeping one job, and the attempted fix is **included here**: every csproj set `<DebugType>full</DebugType>`, which emits native Windows PDBs, and Mono.Cecil routes those through the `ISymUnmanagedWriter2` path that crashes. All six now set `portable`, which avoids that path. This was #2746, folded in and closed.

**It is not proven to fix the crash.** The failure is intermittent — roughly 1 run in 40 — so a green run demonstrates nothing, and I have not run it enough times to claim otherwise. It is here rather than separate because splitting into two jobs doubles the exposure per run, so shipping the split without it makes the flake worse. Weigh it on that basis rather than on the wall-clock measurements, which say nothing about it.

`coverlet.collector` also moves into `Test.csproj`, replacing the `dotnet add package` step that mutated the project mid-workflow and forced the test step to re-compile (~15s per job).

Not addressed here: the 19-copy glyco test, which is the floor on any further split — ~8m30s however many ways the suite is divided.

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -35,11 +35,8 @@ jobs:
       - name: Build
         run: dotnet build --no-restore ./MetaMorpheus/MetaMorpheus.sln --configuration Release
 
-      - name: Install Coverlet for code coverage
-        run: dotnet add ./MetaMorpheus/Test/Test.csproj package coverlet.collector -v 6.0.2
-
       - name: Run unit tests with coverage
-        run: dotnet test .\MetaMorpheus\Test\Test.csproj --configuration Release --verbosity normal --filter "Category!=ExternalService&FullyQualifiedName~Test.TestOGlyco" --collect:"XPlat Code Coverage"
+        run: dotnet test .\MetaMorpheus\Test\Test.csproj --configuration Release --no-build --verbosity normal --filter "Category!=ExternalService&FullyQualifiedName~Test.TestOGlyco" --collect:"XPlat Code Coverage"
 
       # Both halves upload. Codecov merges uploads for the same commit, so a single upload would
       # report the suite as covering only its own half and sink the project/patch percentages.
@@ -65,11 +62,8 @@ jobs:
       - name: Build
         run: dotnet build --no-restore ./MetaMorpheus/MetaMorpheus.sln --configuration Release
 
-      - name: Install Coverlet for code coverage
-        run: dotnet add ./MetaMorpheus/Test/Test.csproj package coverlet.collector -v 6.0.2
-
       - name: Run unit tests with coverage
-        run: dotnet test .\MetaMorpheus\Test\Test.csproj --configuration Release --verbosity normal --filter "Category!=ExternalService&FullyQualifiedName!~Test.TestOGlyco" --collect:"XPlat Code Coverage"
+        run: dotnet test .\MetaMorpheus\Test\Test.csproj --configuration Release --no-build --verbosity normal --filter "Category!=ExternalService&FullyQualifiedName!~Test.TestOGlyco" --collect:"XPlat Code Coverage"
 
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v4
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -11,10 +11,18 @@ env:
   DOTNET_FRAMEWORK: net10.0
 
 jobs:
-  test:
-    name: Test on Windows
+  # The suite is split across two runners purely for wall clock. The partition is on fixture name and
+  # is exhaustive by construction: a test either matches ~Test.TestOGlyco or matches !~Test.TestOGlyco.
+  # TestOGlyco is the half chosen because it holds OGlycoTest_Run5_WriteDecoys, which alone is ~32% of
+  # the suite's runtime, and because it is NOT one of the fixtures sharing EverythingRunnerEngineTestCase
+  # -- so neither runner rebuilds a cached search the other already built.
+  #
+  # These are separate jobs, not NUnit parallelism. Each runner still executes its own tests
+  # sequentially in one process, so tests that share an output folder (TESTGlycoData is used by five
+  # test files) never run concurrently. In-process parallelism is not safe in this suite.
+  test-glyco:
+    name: Test on Windows (glyco)
     runs-on: windows-latest
-
     steps:
       - uses: actions/checkout@v4
       - name: Setup .NET
@@ -23,17 +31,45 @@ jobs:
           dotnet-version: ${{ env.DOTNET_VERSION }}
       - name: Restore dependencies
         run: dotnet restore ./MetaMorpheus/MetaMorpheus.sln
-        
+
       - name: Build
         run: dotnet build --no-restore ./MetaMorpheus/MetaMorpheus.sln --configuration Release
-        
+
       - name: Install Coverlet for code coverage
         run: dotnet add ./MetaMorpheus/Test/Test.csproj package coverlet.collector -v 6.0.2
-        
+
       - name: Run unit tests with coverage
-        # Exclude live external-service tests (UniProt, etc.) from the required run; they run in the
-        # separate, non-blocking external-service-tests job below.
-        run: dotnet test .\MetaMorpheus\Test\Test.csproj --configuration Release --verbosity normal --filter "Category!=ExternalService" --collect:"XPlat Code Coverage"
+        run: dotnet test .\MetaMorpheus\Test\Test.csproj --configuration Release --verbosity normal --filter "Category!=ExternalService&FullyQualifiedName~Test.TestOGlyco" --collect:"XPlat Code Coverage"
+
+      # Both halves upload. Codecov merges uploads for the same commit, so a single upload would
+      # report the suite as covering only its own half and sink the project/patch percentages.
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v4
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          verbose: true
+          flags: unittests
+
+  test-core:
+    name: Test on Windows (core)
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup .NET
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: ${{ env.DOTNET_VERSION }}
+      - name: Restore dependencies
+        run: dotnet restore ./MetaMorpheus/MetaMorpheus.sln
+
+      - name: Build
+        run: dotnet build --no-restore ./MetaMorpheus/MetaMorpheus.sln --configuration Release
+
+      - name: Install Coverlet for code coverage
+        run: dotnet add ./MetaMorpheus/Test/Test.csproj package coverlet.collector -v 6.0.2
+
+      - name: Run unit tests with coverage
+        run: dotnet test .\MetaMorpheus\Test\Test.csproj --configuration Release --verbosity normal --filter "Category!=ExternalService&FullyQualifiedName!~Test.TestOGlyco" --collect:"XPlat Code Coverage"
 
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v4
@@ -41,6 +77,23 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           verbose: true
           flags: unittests
+
+  # Aggregator so the required status check keeps its existing name. Branch protection can go on
+  # keeping "Test on Windows" required without a settings change; this fails if either half failed
+  # or was skipped.
+  test:
+    name: Test on Windows
+    runs-on: ubuntu-latest
+    needs: [ test-glyco, test-core ]
+    if: always()
+    steps:
+      - name: Check both halves passed
+        run: |
+          echo "glyco=${{ needs.test-glyco.result }} core=${{ needs.test-core.result }}"
+          if [ "${{ needs.test-glyco.result }}" != "success" ] || [ "${{ needs.test-core.result }}" != "success" ]; then
+            echo "::error::One or both halves of the Windows test suite did not pass."
+            exit 1
+          fi
 
   # Runs tests that depend on live external web services (UniProt, ...). Kept separate and
   # intentionally NOT a required status check, so a third-party outage never blocks a PR. The tests
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -36,7 +36,10 @@ jobs:
         run: dotnet build --no-restore ./MetaMorpheus/MetaMorpheus.sln --configuration Release
 
       - name: Run unit tests with coverage
-        run: dotnet test .\MetaMorpheus\Test\Test.csproj --configuration Release --no-build --verbosity normal --filter "Category!=ExternalService&FullyQualifiedName~Test.TestOGlyco" --collect:"XPlat Code Coverage"
+        # Exclude live external-service tests (UniProt, etc.) from the required run; they run in the
+        # separate, non-blocking external-service-tests job below. --settings makes a filter that
+        # matches nothing fail rather than silently pass; see the runsettings file for why.
+        run: dotnet test .\MetaMorpheus\Test\Test.csproj --configuration Release --no-build --verbosity normal --filter "Category!=ExternalService&FullyQualifiedName~Test.TestOGlyco" --settings .github\no-empty-test-run.runsettings --collect:"XPlat Code Coverage"
 
       # Both halves upload. Codecov merges uploads for the same commit, so a single upload would
       # report the suite as covering only its own half and sink the project/patch percentages.
@@ -46,6 +49,7 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           verbose: true
           flags: unittests
+          fail_ci_if_error: true
 
   test-core:
     name: Test on Windows (core)
@@ -63,7 +67,10 @@ jobs:
         run: dotnet build --no-restore ./MetaMorpheus/MetaMorpheus.sln --configuration Release
 
       - name: Run unit tests with coverage
-        run: dotnet test .\MetaMorpheus\Test\Test.csproj --configuration Release --no-build --verbosity normal --filter "Category!=ExternalService&FullyQualifiedName!~Test.TestOGlyco" --collect:"XPlat Code Coverage"
+        # Exclude live external-service tests (UniProt, etc.) from the required run; they run in the
+        # separate, non-blocking external-service-tests job below. --settings makes a filter that
+        # matches nothing fail rather than silently pass; see the runsettings file for why.
+        run: dotnet test .\MetaMorpheus\Test\Test.csproj --configuration Release --no-build --verbosity normal --filter "Category!=ExternalService&FullyQualifiedName!~Test.TestOGlyco" --settings .github\no-empty-test-run.runsettings --collect:"XPlat Code Coverage"
 
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v4
@@ -71,6 +78,7 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           verbose: true
           flags: unittests
+          fail_ci_if_error: true
 
   # Aggregator so the required status check keeps its existing name. Branch protection can go on
   # keeping "Test on Windows" required without a settings change; this fails if either half failed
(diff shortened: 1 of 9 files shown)


.github/workflows/Test.yml
```diff
@@ -11,10 +11,45 @@ env:
...
+        with:
+          dotnet-version: ${{ env.DOTNET_VERSION }}
+      - name: Restore dependencies
+        run: dotnet restore ./MetaMorpheus/MetaMorpheus.sln
+
+      - name: Build
+        run: dotnet build --no-restore ./MetaMorpheus/MetaMorpheus.sln --configuration Release
+
+      - name: Run unit tests with coverage
+        run: dotnet test .\MetaMorpheus\Test\Test.csproj --configuration Release --no-build --verbosity normal --filter "Category!=ExternalService&FullyQualifiedName~Test.TestOGlyco" --collect:"XPlat Code Coverage"
```
trishorts: **Major** - Filter partition has no CI guard against a half running zero tests

**Problem:**
The PR's central invariant — "a partition that silently drops tests is the failure mode worth guarding against, since it is invisible in a green run" — is verified once by hand (50 + 1712 = 1762) and then left unenforced. Nothing in the workflow re-checks it on subsequent runs. `Test.TestOGlyco` is a hand-typed substring of a single fixture (verified: `MetaMorpheus/Test/TestOGlyco.cs:26-29`, `namespace Test` / `class TestOGlyco`, the only occurrence of that string in the whole tree outside this yml). Rename that class, or typo either filter string, and `--filter "...&FullyQualifiedName~Test.TestOGlyco"` matches nothing. VSTest's default is to emit "No test matches the given testcase filter" as a *warning*: `TreatNoTestsAsError` defaults to `false`, so the step exits 0, the job is green, and the aggregator at line 78 is green too — it only inspects `needs.*.result`, never a test count.

**Impact:**
A future rename of `TestOGlyco` (or a bad edit to either filter string) silently stops running up to 50 tests, including the entire O-glyco search path, with all three required-check-relevant jobs reporting success. The complement half absorbs the difference, so total-count drift is invisible without manually reading two job logs.

**Suggested fix:**
Make an empty half a hard failure rather than a warning. Note that the obvious flags do **not** work — I checked all three against a throwaway NUnit project on the .NET 10 SDK:

| mechanism | zero-match filter | matching filter |
|---|---|---|
| plain `dotnet test --filter` | exit **0** (silently green) | 0 |
| `-p:TreatNoTestsAsError=true` | exit **0** - no effect | 0 |
| `--treat-no-tests-as-error` | exit 1, but only because MSBuild rejects the switch - it fails even when tests *do* run | **1** |
| `--settings` with `<TreatNoTestsAsError>true</TreatNoTestsAsError>` | exit **1** | exit **0** |

Only the runsettings form behaves correctly in both directions. So: add a small runsettings file with

```xml
<RunSettings><RunConfiguration><TreatNoTestsAsError>true</TreatNoTestsAsError></RunConfiguration></RunSettings>
```

and pass `--settings <that file>` on both `dotnet test` invocations (lines 39 and 66). That turns the exact drop scenario this PR describes into a red build, with no branch-protection or aggregator change.

**Affected test scenarios:**
- `Test on Windows (glyco)` — `TestOGlyco` fixture renamed to `OGlycoTests` — job must fail, not report success with 0 tests run
- `Test on Windows (glyco)` — filter string typo (`Test.TestOGlyc`) — job must fail
- `Test on Windows (core)` — both filters intact — 1712 tests run, job green
- `Test on Windows` aggregator — one half ran zero tests — must report failure

.github/workflows/Test.yml
```diff
@@ -11,10 +11,45 @@ env:
...
 
+      # Both halves upload. Codecov merges uploads for the same commit, so a single upload would
+      # report the suite as covering only its own half and sink the project/patch percentages.
+      - name: Upload coverage to Codecov
```
trishorts, line 46: **Minor** - Two Codecov uploads per commit with no after_n_builds and no fail_ci_if_error

**Problem:**
Both halves now call `codecov/codecov-action@v4` for the same commit (lines 43-48 and 68-73), each with `flags: unittests`. The merge claim in the comment is right — Codecov unions same-commit uploads — but there is no `codecov.yml` anywhere in the repo (confirmed: no `codecov.y*ml` file exists), so `after_n_builds` is unset and Codecov computes and posts a `codecov/project` status the moment the *first* report arrives. That first report covers only one half. Separately, `fail_ci_if_error` is not set, so it defaults to false: if one half's upload fails (network, token, rate limit), that job still goes green, the aggregator still goes green, and Codecov reports the surviving half's coverage as the whole suite.

**Impact:**
Two distinct symptoms. First, a transient red/incorrect `codecov/project` status on every PR between the two uploads, since the glyco half alone covers ~50 tests' worth of lines. Second — the more serious one — a silently dropped upload produces exactly the collapsed project/patch percentage the comment says this design prevents, with no red job anywhere to attribute it to.

**Suggested fix:**
Add a `codecov.yml` at the repo root setting `codecov.notify.after_n_builds: 2` (and `comment.after_n_builds: 2`) so no status is computed until both halves land, and set `fail_ci_if_error: true` on both upload steps so a lost upload reddens the half that lost it rather than silently halving the reported coverage.

**Affected test scenarios:**
- Codecov project status — glyco upload lands, core upload still running — no partial status posted
- Codecov project status — core upload fails entirely — `Test on Windows (core)` must fail rather than report success with half coverage
- Codecov project status — both uploads land — merged total matches the pre-split single-job total

.github/workflows/Test.yml
```diff
@@ -42,6 +72,23 @@ jobs:
...
           flags: unittests
 
+  # Aggregator so the required status check keeps its existing name. Branch protection can go on
+  # keeping "Test on Windows" required without a settings change; this fails if either half failed
+  # or was skipped.
+  test:
+    name: Test on Windows
+    runs-on: ubuntu-latest
+    needs: [ test-glyco, test-core ]
+    if: always()
```
trishorts: **Minor** - `if: always()` makes the aggregator red on cancelled runs and uncancellable

**Problem:**
`if: always()` on the `test` aggregator evaluates true even when the workflow run is *cancelled*, not just when a dependency failed or was skipped. On a cancelled run `needs.test-glyco.result` is `cancelled`, the `!= "success"` check at line 87 fires, and the required check `Test on Windows` reports failure with `::error::One or both halves of the Windows test suite did not pass.` — which is not what happened. `always()` also opts the job out of cancellation, so it schedules and occupies an Ubuntu runner after the user has already cancelled.

**Impact:**
Cancelling a superseded run (or hitting a workflow-level cancel) leaves a red required status on that commit and an incorrect error message claiming a test half failed. Low blast radius, but it misattributes cancellations as test failures on the one check branch protection cares about.

**Suggested fix:**
Use `if: ${{ !cancelled() }}` instead of `if: always()`. That still runs the aggregator when a half fails or is skipped — the two cases the comment above the job says it must catch — while letting a cancelled run stay cancelled.

**Affected test scenarios:**
- Aggregator — workflow cancelled mid-run — check reports cancelled, not failed
- Aggregator — `test-core` failed — check reports failure with the error annotation
- Aggregator — `test-glyco` skipped — check reports failure

.github/workflows/Test.yml
```diff
@@ -11,10 +11,45 @@ env:
...
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: ${{ env.DOTNET_VERSION }}
+      - name: Restore dependencies
+        run: dotnet restore ./MetaMorpheus/MetaMorpheus.sln
+
+      - name: Build
+        run: dotnet build --no-restore ./MetaMorpheus/MetaMorpheus.sln --configuration Release
+
+      - name: Run unit tests with coverage
```
trishorts, line 38: **Minor** - Rationale comment for the `Category!=ExternalService` exclusion was dropped

**Problem:**
The base file carried an explanatory comment on the test step: `# Exclude live external-service tests (UniProt, etc.) from the required run; they run in the / # separate, non-blocking external-service-tests job below.` Both new halves (lines 38-39 and 65-66) retain the `Category!=ExternalService` clause but neither carries the comment. The clause is now the only thing keeping third-party outages out of the required check, and the reason it exists is no longer recorded next to it — while the comment explaining the *other* pre-existing exclusion (the `external-service-tests` job, lines 92-95) survives, leaving the pairing half-documented.

**Impact:**
The next person reading `Category!=ExternalService` on either half has no local indication that removing it would let a UniProt outage redden a required check. The contract survives, its justification does not.

**Suggested fix:**
Restore the two-line rationale above the `Run unit tests with coverage` step in at least one half (or once above the pair, alongside the existing partition comment at lines 14-22).

**Affected test scenarios:**
- N/A — documentation only; no change to what CI executes

MetaMorpheus/CMD/CMD.csproj
```diff
@@ -1,11 +1,11 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net10.0</TargetFramework>
     <ApplicationManifest>app1.manifest</ApplicationManifest>
     <Configurations>Debug;Release</Configurations>
-    <DebugType>full</DebugType>
+    <DebugType>portable</DebugType>
```
trishorts, line 8: **Major** - The PR description disclaims two changes that are actually in this diff

**Problem:**
The description defers two changes to separate work, but both are in this diff:

1. On the coverlet access-violation flake it says *"`portable` avoids it. I have not proven that fixes it ... so it is left for a separate change rather than bundled in here."* The diff flips `<DebugType>full</DebugType>` to `<DebugType>portable</DebugType>` in all six csprojs (`CMD`, `EngineLayer`, `GUI`, `GuiFunctions`, `TaskLayer`, `Test`). PR #2746, "Portable PDBs so coverlet stops crashing the test job", is now closed - so this looks like it was folded in without the body being updated.
2. Under "Not addressed here, both worth doing separately" it lists *"moving `coverlet.collector` into `Test.csproj`"*. `Test.csproj:24` adds exactly that `PackageReference`, and the `dotnet add package coverlet.collector` step is correspondingly deleted from the workflow.

**Impact:**
The PDB format change is the one with a risk profile a reviewer would want to weigh separately - it is the stated remedy for an intermittent CI crash, bundled into a PR whose measured evidence is entirely about wall clock. A reviewer who reads the "left for a separate change" sentence will not scrutinise it. It also makes the PR's own careful framing - measured, not estimated; this is what I did *not* do - inaccurate, which is a shame because the rest of the writeup is unusually rigorous.

**Suggested fix:**
Update the description to state that both changes are included, or split them back out. No code change needed either way - this is about the diff and its description agreeing.

**Affected test scenarios:**
- N/A - description accuracy

MetaMorpheus/CMD/CMD.csproj
```diff
@@ -1,11 +1,11 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
trishorts: **Minor** - All six csproj files gain a UTF-8 BOM on line 1

**Problem:**
Every csproj touched here picks up a UTF-8 BOM (`EF BB BF`) before `<Project Sdk="Microsoft.NET.Sdk">` - `CMD`, `EngineLayer`, `GUI`, `GuiFunctions`, `TaskLayer` and `Test`. Confirmed with `git show <head>:... | cat -A`, which renders line 1 as `M-oM-;M-?<Project ...`. That is unrelated to both the job split and the `DebugType` change; it is an editor artifact.

**Impact:**
None functionally - MSBuild reads either encoding. But it makes line 1 of six files show as changed in this diff and in every future `git blame` on them, for no reason.

**Suggested fix:**
Re-save the six csprojs as UTF-8 without BOM so line 1 drops out of the diff. Raising this once rather than six times.

**Affected test scenarios:**
- N/A - cosmetic

trishorts (commented):
Automated code review - 6 finding(s) posted as inline comments.

acesnik:
🤖 All six verified and addressed in `575bff60`. I checked each rather than taking them as given; every one held up, including the parts I would have got wrong.

The zero-match table is the valuable one. I reproduced it on the .NET 10 SDK and then re-ran with the **production filter shapes** rather than simplified ones — the glyco half with its fixture renamed exits 1, and both halves with the real compound `Category!=ExternalService&FullyQualifiedName…` filters still pass. I would have reached for `-p:TreatNoTestsAsError=true` and shipped a guard that does nothing.

Also updated the description, which is finding 5's actual fix: it claimed the PDB change and the `coverlet.collector` move were deferred, and both are in the diff. #2746 was folded in here and the body never caught up. Your point that this matters stands — the PDB change is the unproven remedy for an intermittent crash sitting in a PR whose evidence is all wall clock, so the description now says that plainly instead of disclaiming it.

.github/workflows/Test.yml
acesnik: 🤖 Fixed, and thanks for testing the mechanisms — I would have used `-p:TreatNoTestsAsError=true` and shipped a no-op.

Reproduced your table on the .NET 10 SDK, then re-ran with the production filters rather than a simplified one, since a guard that reddens legitimate runs would be worse than none:

```
glyco half   (compound, matches)      exit=0   Passed: 1
core half    (compound, matches)      exit=0   Passed: 2
glyco half   (fixture RENAMED)        exit=1   no tests
```

Added `.github/no-empty-test-run.runsettings` and `--settings` on both invocations. The file carries a comment explaining why the other two mechanisms do not work, so the next person does not swap it for the simpler-looking flag.

.github/workflows/Test.yml
acesnik: 🤖 Both fixed. Added `codecov.yml` with `codecov.notify.after_n_builds: 2` and `comment.after_n_builds: 2`, and `fail_ci_if_error: true` on both uploads.

Checked that 2 is the right number rather than assuming it: `Test.yml` is the only workflow in the repo that uploads to Codecov, and after the split it has exactly two upload steps. If a third is ever added, that number has to move with it — worth knowing.

.github/workflows/Test.yml
acesnik: 🤖 Fixed — `if: ${{ !cancelled() }}`. It still fires when a half fails or is skipped, which is what the aggregator exists for, and a cancelled run now reports cancelled instead of claiming a test half failed.

The second effect you noted is the better argument for the change: `always()` opts the job out of cancellation, so it was scheduling a runner after the user had already cancelled the run.

.github/workflows/Test.yml
acesnik: 🤖 Restored, above both halves rather than once, since each is now a standalone `dotnet test` invocation and the clause appears in both.

Extended it to cover `--settings` as well, so the two non-obvious parts of that command line are explained where they are rather than only in the runsettings file.

MetaMorpheus/CMD/CMD.csproj
acesnik: 🤖 Correct on both counts, and the description is the fix — no code change. #2746 was folded in here and the body was never updated, so it still said the PDB change was "left for a separate change" while the diff flips `DebugType` in all six csprojs, and still listed moving `coverlet.collector` under "not addressed here" while `Test.csproj` adds it and the workflow's `dotnet add package` step is gone.

Updated to state both are included, and to keep the PDB change's caveat visible rather than buried: it is the suspected remedy for the coverlet access violation, it is **not** proven to fix it, and it is here because the flake is worse with two jobs than with one. The description now says to weigh it on that basis rather than on the wall-clock measurements, which say nothing about it.

MetaMorpheus/CMD/CMD.csproj
acesnik: 🤖 Fixed, and thanks for raising it once rather than six times.

`head -c 3 | xxd` gives `efbbbf` on the branch against `3c5072` on master for all six. Stripped, so line 1 drops out of the diff — `CMD.csproj` is back to a one-line change.

trishorts (approved):

RayMSMS (approved):

nbollis (approved):

PR #2745, opened 2026-08-18
Add configurable retention time predictor for search and glyco search

RayMSMS:
## Summary

  - Glyco search's post-search PEP/FDR analysis always used the Chronologer retention-time (RT) predictor, a heavier deep-learning model, which noticeably increased search time. This PR
  makes the RT predictor configurable instead of hardcoded.
  - Adds a "Retention Time Predictor" option (Chronologer / SSRCalc / Prosit2019iRT / Prosit2020iRTTMT) to the Post-Search Analysis section of both `SearchTaskWindow` and
  `GlycoSearchTaskWindow`, backed by the existing `CommonParameters.RTPredictorName` field and `FdrAnalysisEngine.GetRTPredictor` switch (Prosit options already existed engine-side;
  SSRCalc is newly added).
  - Default stays **Chronologer** for both task types — fully backward compatible, purely opt-in.
  - The picker is hidden in **RNA mode** on `SearchTaskWindow`, since `FdrAnalysisEngine.GetRTPredictor` short-circuits to `null` before ever reading the name whenever `searchType !=
  "standard"` — the choice is genuinely inert there, unlike CZE (see below). `GlycoSearchTaskWindow` has no RNA mode, so no change was needed there.
  - The picker is **not** hidden for CZE separation type. An earlier version of this PR hid it there too, on the assumption CZE never uses `IRetentionTimePredictor`. That assumption was
  wrong: `PEPAnalysisEngine.BuildFileSpecificDictionaries` calls `ComputeRetentionTimeEquivalentValues(..., RetentionTimePredictor)` unconditionally when building the training
  dictionaries, so the selected predictor is invoked for CZE runs too — only the later per-PSM z-score step branches to `GetMobilityZScore` instead of using that result. Hiding the picker
  for CZE would have removed the user's only way to avoid paying for the expensive Chronologer path on CZE data, while doing nothing to reduce that cost. The hide was reverted.
  - Fixed a separate, pre-existing bug this PR's new setting made observable: `MetaMorpheusTask.SetAllFileSpecificCommonParams` rebuilds `CommonParameters` from scratch whenever a
  file-specific `<basename>.toml` exists (e.g. the `<file>-calib.toml` that `CalibrationTask` writes automatically), and did not forward `RTPredictorName` into the rebuilt object — so any
  pipeline that includes a Calibration step before Search/GlycoSearch silently reverted the user's predictor choice back to Chronologer. Now fixed by forwarding `rtPredictorName` through
  that rebuild.
  - The four predictor-name strings (`"Chronologer"`, `"SSRCalc"`, `"Prosit2019iRT"`, `"Prosit2020iRTTMT"`) were previously duplicated as literals in six places across two assemblies (both
  GUI windows and the engine switch). Consolidated into a single `EngineLayer.FdrAnalysis.RTPredictorNames` constants class referenced everywhere, so a typo is now a compile error instead
  of a silent fallback to no predictor.

  ## Changes

  - `EngineLayer/FdrAnalysis/RTPredictorNames.cs` (new): shared `const string` names for the four predictors.
  - `EngineLayer/CommonParameters.cs`: default `rtPredictorName` now references `RTPredictorNames.Chronologer`.
  - `EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs`: `GetRTPredictor`'s switch gains `RTPredictorNames.SSRCalc => new SSRCalc3RetentionTimePredictor()`; all four arms use the shared
  constants.
  - `TaskLayer/MetaMorpheusTask.cs`: `SetAllFileSpecificCommonParams` now forwards `rtPredictorName: commonParams.RTPredictorName` into the rebuilt `CommonParameters` — fixes the
  file-specific-toml bug described above.
  - `GUI/TaskWindows/SearchTaskWindow.xaml(.cs)`: new "Retention Time Predictor" `GroupBox` (4 radio buttons); save/restore wiring to `CommonParameters.RTPredictorName` via the shared
  constants; `Visibility` bound to the existing `CollapseOnRnaModeConverter` (already used 7 other places in this file) so the group collapses in RNA mode only.
  - `GUI/TaskWindows/GlycoSearchTaskWindow.xaml(.cs)`: same 4-option predictor `GroupBox` and save/restore wiring — no visibility toggle, not applicable to this window.
  - `Test/FdrTest.cs`:
    - `FdrAnalysisEngine_GetRTPredictor_ReturnsExpectedPrositPredictor` — covers `Prosit2019iRT`/`Prosit2020iRTTMT` (network-dependent, sits behind a currently-inactive
  `Explicit`/`Category("Koina")` gate).
    - `FdrAnalysisEngine_GetRTPredictor_ReturnsExpectedLocalPredictor` (new) — covers `Chronologer`/`SSRCalc`, deliberately kept in a separate, ungated method so this coverage can't be
  silently excluded if the Koina gate above is ever turned on.
    - `SearchTask_DefaultRTPredictor_ResolvedToChronologer` — proves both `SearchTask` and `GlycoSearchTask` default to Chronologer; uses an explicit `case`/failing `default` so a typo'd
  `[TestCase]` label fails loudly instead of silently testing the wrong task type twice.
    - `SetAllFileSpecificCommonParams_PreservesRTPredictorName` (new) — regression test for the `MetaMorpheusTask.cs` fix above.

  ## Test plan

  - [x] `dotnet build` on the full solution — succeeds with 0 errors.
  - [x] `Test/FdrTest.cs`, full class (24 tests) — all passing. Coverage of the four predictor names is split across two tests by network dependency: `Chronologer`/`SSRCalc` in
  `FdrAnalysisEngine_GetRTPredictor_ReturnsExpectedLocalPredictor`, `Prosit2019iRT`/`Prosit2020iRTTMT` in `FdrAnalysisEngine_GetRTPredictor_ReturnsExpectedPrositPredictor`; Chronologer's
  role as the *default* (as opposed to just a valid selection) is separately covered by `SearchTask_DefaultRTPredictor_ResolvedToChronologer`.
  - [x] Manual GUI verification (not covered by automated tests — `Test.csproj` doesn't reference `GUI.csproj`): opened both task windows, confirmed each of the 4 radio options saves and
  restores correctly on reopen for both `SearchTask` and `GlycoSearchTask`; confirmed the predictor group box disappears in `SearchTaskWindow` when RNA mode is active and stays visible for
  CZE.
  - [ ] No behavior change for existing saved `.toml` task files without an explicit `RTPredictorName` — they continue to default to Chronologer.


MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml.cs
```diff
@@ -649,7 +670,7 @@ private void SaveButton_Click(object sender, RoutedEventArgs e)
                 maxHeterozygousVariants: MaxHeterozygousVariants,
                 precursorDeconParams: precursorDeconvolutionParameters,
                 productDeconParams: productDeconvolutionParameters,
-                precursorMassMatchMode: _massDifferenceAcceptorViewModel.PrecursorMassMatchMode,
+                precursorMassMatchMode: DeconHostViewModel.PrecursorMassMatchMode,
```
RayMSMS: I don't do that change. It was automatically added. I will revise it!

MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml.cs
```diff
@@ -435,7 +435,22 @@ private void UpdateFieldsFromTask(SearchTask task)
                 ye.VerifyCheckState();
             }
 
-            _massDifferenceAcceptorViewModel = new(task.SearchParameters.MassDiffAcceptorType, task.SearchParameters.CustomMdac, task.CommonParameters.PrecursorMassTolerance.Value, task.CommonParameters.PrecursorMassMatchMode);
+            _massDifferenceAcceptorViewModel = new(task.SearchParameters.MassDiffAcceptorType, task.SearchParameters.CustomMdac, task.CommonParameters.PrecursorMassTolerance.Value);
```
nbollis: It appears as if a parameter was removed from the constructor. If so, add back the precursor mass match mode to the constructor.

MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml.cs
nbollis: Go ahead and switch this back as well, it is out of scope for this PR.

nbollis (changes requested):

MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml.cs
```diff
@@ -617,7 +632,13 @@ private void SaveButton_Click(object sender, RoutedEventArgs e)
             DeconvolutionParameters productDeconvolutionParameters = DeconHostViewModel.ProductDeconvolutionParameters.Parameters;
             bool useProvidedPrecursorInfo = DeconHostViewModel.UseProvidedPrecursors;
             bool doPrecursorDeconvolution = DeconHostViewModel.DoPrecursorDeconvolution;
-            
+
+            string rtPredictorModelName;
```
trishorts, line 637: **Critical** - SearchTaskWindow computes rtPredictorModelName but never saves it

**Problem:**
`SaveButton_Click` builds `rtPredictorModelName` from the four new radio buttons (lines 636-640), but the `CommonParameters commonParamsToSave = new CommonParameters(...)` call that immediately follows (lines 642-674) never passes a `rtPredictorName:` argument. The variable is written and never read — it is dead code. `CommonParameters`' constructor defaults `rtPredictorName` to `"Chronologer"` (`EngineLayer/CommonParameters.cs:67`), so the saved task always carries `RTPredictorName == "Chronologer"` no matter which radio button the user selected. The sibling `GlycoSearchTaskWindow.xaml.cs:465` does pass `rtPredictorName: rtPredictorModelName`; the classic search window does not.

**Impact:**
The entire feature is inert for `SearchTask` — the primary surface this PR adds. A user who picks SSRCalc, Prosit2019iRT, or Prosit2020iRTTMT in the Search Task window silently gets Chronologer, with no error and no indication the choice was ignored. The restore switch at line 439 then dutifully re-checks Chronologer on reopen, so the setting appears to "not stick" with no diagnosable cause. This also falsifies the PR's manual-test claim that "each of the 4 radio options saves and restores correctly ... for both `SearchTask` and `GlycoSearchTask`".

**Suggested fix:**
Add `rtPredictorName: rtPredictorModelName` to the `CommonParameters` constructor call in `SearchTaskWindow.SaveButton_Click`, mirroring the glyco window. Then add a round-trip assertion that a `SearchTask` whose `CommonParameters` was built with a non-default predictor still reports that predictor after save, so a dropped argument cannot regress silently again.

**Affected test scenarios:**
- SearchTaskWindow save/restore - SSRCalc radio checked, task saved - `TheTask.CommonParameters.RTPredictorName` equals `"SSRCalc"`, not `"Chronologer"`
- SearchTaskWindow save/restore - Prosit2019iRT radio checked, task saved then reloaded into a new window - `Prosit2019iRTRadioButton.IsChecked` is true
- SearchTaskWindow save/restore - no radio changed from default - `RTPredictorName` equals `"Chronologer"` (backward-compat invariant)

MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml.cs
trishorts: **Critical** - DeconHostViewModel has no PrecursorMassMatchMode member — GUI does not compile

**Problem:**
The diff changed `precursorMassMatchMode: _massDifferenceAcceptorViewModel.PrecursorMassMatchMode` to `precursorMassMatchMode: DeconHostViewModel.PrecursorMassMatchMode`. The field declared at line 44 is `private DeconHostViewModel DeconHostViewModel;`, and `GuiFunctions/ViewModels/Deconvolution/DeconHostViewModel.cs` declares no `PrecursorMassMatchMode` member; neither does its base `GuiFunctions/ViewModels/BaseViewModel.cs`. A repo-wide search finds exactly one declaration of that property, on `MassDifferenceAcceptorSelectionViewModel` (`GuiFunctions/ViewModels/MassDifferenceAcceptorSelectionViewModel.cs:114`) — the type of `_massDifferenceAcceptorViewModel`, i.e. the expression the diff removed.

**Impact:**
`GUI.csproj` fails to build with CS1061 ("'DeconHostViewModel' does not contain a definition for 'PrecursorMassMatchMode'"). Nothing in this PR ships until this is corrected, and the PR body's "`dotnet build` on `EngineLayer`, `GUI`, and `Test` — all succeed with 0 errors" does not hold for the branch as committed.

**Suggested fix:**
Restore the source to `_massDifferenceAcceptorViewModel.PrecursorMassMatchMode`. If the intent of the merge reconciliation really was to relocate this setting onto the deconvolution view model, then that property has to actually be added to `DeconHostViewModel` and populated from `TheTask.CommonParameters.PrecursorMassMatchMode` — but that is a separate change and does not belong in an RT-predictor PR.

**Affected test scenarios:**
- Build - `dotnet build MetaMorpheus/GUI/GUI.csproj` on this branch - compiles with 0 errors
- SearchTaskWindow save - "Most Abundant Mass" selected in the mass-difference-acceptor control - `commonParamsToSave.PrecursorMassMatchMode` is `MostAbundant`

MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml.cs
trishorts: **Major** - Dropped ctor arg silently resets a saved task's PrecursorMassMatchMode to Monoisotopic

**Problem:**
`_massDifferenceAcceptorViewModel = new(task.SearchParameters.MassDiffAcceptorType, task.SearchParameters.CustomMdac, task.CommonParameters.PrecursorMassTolerance.Value)` lost its fourth argument, `task.CommonParameters.PrecursorMassMatchMode`. That parameter is optional and defaults to `PrecursorMassMatchMode.Monoisotopic` (`MassDifferenceAcceptorSelectionViewModel.cs:41-42`), and line 76 of that constructor assigns it to the view model's `PrecursorMassMatchMode`, which in turn drives `UseMostAbundantMass` / `ActiveMassDiffAcceptorTypes` and the radio buttons in `MassDiffAcceptorControl.xaml:48-49`. So the window no longer restores the task's saved match mode — it always initializes to Monoisotopic.

**Impact:**
Opening a previously saved search task that used most-abundant precursor mass matching shows Monoisotopic in the UI; saving without touching anything writes Monoisotopic back into the task, changing which precursor mass the search matches against and hence which PSMs are found. Silent, and invisible unless the user notices the radio flipped.

**Suggested fix:**
Pass `task.CommonParameters.PrecursorMassMatchMode` back as the fourth constructor argument so the view model is seeded from the task being loaded. This is orthogonal to the RT-predictor feature and should not have changed in this PR.

**Affected test scenarios:**
- SearchTaskWindow load - task with `PrecursorMassMatchMode.MostAbundant` - `_massDifferenceAcceptorViewModel.UseMostAbundantMass` is true
- SearchTaskWindow load-then-save with no user edits - task with `MostAbundant` - saved `CommonParameters.PrecursorMassMatchMode` is still `MostAbundant`

MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml
```diff
@@ -1057,6 +1057,31 @@
                                     </Style>
                                 </Expander.Style>
                                 <StackPanel>
+                                    <GroupBox Header="Retention Time Predictor"
+                                              Visibility="{Binding ElementName=SeparationTypeComboBox, Path=SelectedItem, Converter={local:SeparationTypeToVisibilityConverter}}">
```
trishorts: **Major** - CZE hides the RT-predictor picker, but the predictor still runs for CZE files

**Problem:**
The new `Visibility="{Binding ElementName=SeparationTypeComboBox, Path=SelectedItem, Converter={local:SeparationTypeToVisibilityConverter}}"` collapses the picker whenever separation type is `CZE`, on the stated premise that CZE PSMs "don't use an `IRetentionTimePredictor` at all". That premise is false. `PepAnalysisEngine.BuildFileSpecificDictionaries` (`PEPAnalysisEngine.cs:189-193`) calls `ComputeRetentionTimeEquivalentValues(trainingData, false, RetentionTimePredictor)` and `ComputeRetentionTimeEquivalentValues(trainingData, true, RetentionTimePredictor)` unconditionally for any `searchType == "standard"` — there is no separation-type guard there. Inside that method (`PEPAnalysisEngine.cs:716`) `predictor.PredictRetentionTimeEquivalent(pep, out _)` is invoked for every distinct full sequence of every target PSM with q ≤ 0.01, regardless of `SeparationType`. Only the per-PSM *feature* assignment at `PEPAnalysisEngine.cs:535-550` branches to `GetMobilityZScore` for CZE; the predictor's full cost has already been paid by then and the results are simply discarded.

**Impact:**
A CZE user pays the entire Chronologer inference cost — the exact runtime cost this PR exists to let users avoid — and the PR removes the only control that would let them switch to SSRCalc. The hidden radio buttons also keep whatever value was last restored and silently write it back on save, so a task can persist a predictor the user can no longer see or change.

**Suggested fix:**
Drop the CZE visibility binding (and, if nothing else consumes it, the new converter) so the picker stays available for CZE, since the predictor is still executed there. If the intent is genuinely to make the predictor inert for CZE, that has to be done engine-side by skipping the `ComputeRetentionTimeEquivalentValues` calls when every file is CZE — hiding the UI control alone changes nothing about the work performed.

**Affected test scenarios:**
- PepAnalysisEngine constructed with a CZE-only `fileSpecificParameters` set and a counting/stub `IRetentionTimePredictor` - `PredictRetentionTimeEquivalent` call count - currently non-zero, proving the predictor is exercised under CZE
- SearchTaskWindow save/restore with separationType CZE and RTPredictorName "SSRCalc" - reopen and save - value must remain user-visible and changeable

trishorts (commented):
Automated code review - 4 finding(s) posted as inline comments.

nbollis (dismissed):

MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml.cs
```diff
@@ -650,7 +671,8 @@ private void SaveButton_Click(object sender, RoutedEventArgs e)
                 precursorDeconParams: precursorDeconvolutionParameters,
                 productDeconParams: productDeconvolutionParameters,
                 precursorMassMatchMode: _massDifferenceAcceptorViewModel.PrecursorMassMatchMode,
-                fragmentationParams: _fragmentationParamsViewModel.ToFragmentationParams() );
+                fragmentationParams: _fragmentationParamsViewModel.ToFragmentationParams(),
+                rtPredictorName: rtPredictorModelName);
```
trishorts, line 676: **Major** — the selected predictor never reaches the engine when a file-specific `.toml` exists

This line correctly stores the choice on the task's `CommonParameters`, but `FdrAnalysisEngine.GetRTPredictor` doesn't read the task's `CommonParameters` — it reads the *file-specific* one:

```csharp
// FdrAnalysisEngine.cs:423
var predictorName = fileSpecificParameters.FirstOrDefault().fileSpecificParameters?.RTPredictorName;
```

and that list is built in `MetaMorpheusTask.RunTask`. When a `<spectrafile>.toml` sits next to the data file, the entry comes from `SetAllFileSpecificCommonParams(CommonParameters, ...)` (`MetaMorpheusTask.cs:652`), which rebuilds `CommonParameters` from scratch (`MetaMorpheusTask.cs:576-616`) and **never forwards `rtPredictorName`** — the call ends at `precursorMassMatchMode: commonParams.PrecursorMassMatchMode);`. So the rebuilt object falls back to the constructor default `"Chronologer"` (`CommonParameters.cs:67`). Only the no-toml `else` branch (`:667`) passes `CommonParameters` through intact.

This is exactly the contract the comment above the property warns about:

> `CommonParameters.cs:150-151` — *"If you add a new property here, you must add it to MetaMorpheusTask.cs/SetAllFileSpecificCommonParams !! If you forget this, and the user uses file-specific parameters, your settings will be overwritten by default values!"*

And it isn't an edge case: `CalibrationTask` writes `<file>-calib.toml` beside every calibrated mzML (`CalibrationTask.cs:166` → `:528`), matching the `<basename>.toml` probe exactly, so **every Calibrate → GPTMD → Search run takes the dropping branch**. A glyco user who picks SSRCalc to escape the slow Chronologer path still gets Chronologer — silently, and with the task's own settings toml still showing SSRCalc.

Fix is one line in `MetaMorpheus/TaskLayer/MetaMorpheusTask.cs:616`:

```diff
                 fragmentationParams: commonParams.FragmentationParameters,
-                precursorMassMatchMode: commonParams.PrecursorMassMatchMode);
+                precursorMassMatchMode: commonParams.PrecursorMassMatchMode,
+                rtPredictorName: commonParams.RTPredictorName);
```

Worth a test that pushes a non-default `RTPredictorName` through `SetAllFileSpecificCommonParams` with a non-null `FileSpecificParameters` and asserts it survives — that's the one link in the chain the new tests don't cover (they hand-build the `fsp` list from a `CommonParameters` that already carries the name).

`GlycoSearchTaskWindow.xaml.cs:466` has the identical exposure; the same one-line fix covers both.

I know `SetAllFileSpecificCommonParams` is outside this PR's six files and the omission pre-dates it — but this PR is what makes `RTPredictorName` user-settable, and therefore what makes the omission observable. Fixing it inside `GetRTPredictor` instead would mean changing that method's pre-existing shape, so the forward above seems the smaller change.

MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml.cs
```diff
@@ -617,7 +632,13 @@ private void SaveButton_Click(object sender, RoutedEventArgs e)
             DeconvolutionParameters productDeconvolutionParameters = DeconHostViewModel.ProductDeconvolutionParameters.Parameters;
             bool useProvidedPrecursorInfo = DeconHostViewModel.UseProvidedPrecursors;
             bool doPrecursorDeconvolution = DeconHostViewModel.DoPrecursorDeconvolution;
-            
+
+            string rtPredictorModelName;
+            if (SSRCalcRadioButton.IsChecked == true) rtPredictorModelName = "SSRCalc";
```
trishorts: **Minor** — the four predictor names are now magic strings in five places

`"Chronologer"` / `"SSRCalc"` / `"Prosit2019iRT"` / `"Prosit2020iRTTMT"` are spelled out at `SearchTaskWindow.xaml.cs:441-451` and `:637-640`, `GlycoSearchTaskWindow.xaml.cs:144-154` and `:439-442`, and again in the engine switch at `FdrAnalysisEngine.cs:425-432` — writer and reader in different assemblies, with nothing tying them together.

A typo in any of the four save-side literals produces `predictorName switch { ... _ => null }` → no predictor, no exception, no log line. The two layers also disagree about unknown names: the GUI `default:` shows Chronologer, the engine yields `null`.

A shared `const` set (or an enum with the names on it) referenced by both windows, the engine switch, and the tests would make a typo a compile error instead of a silent behaviour change. The `_ => null` fallthrough itself is pre-existing — it's the new duplication I'd like to see pinned down.

MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml
```diff
@@ -1057,6 +1057,30 @@
                                     </Style>
                                 </Expander.Style>
                                 <StackPanel>
+                                    <GroupBox Header="Retention Time Predictor">
```
trishorts: **Minor** — the picker is shown in RNA mode, where it can't do anything

This window doubles as the RNA Search Task window (`SearchTaskWindow.xaml.cs:53` sets `Title = "RNA Search Task"`). `GetRTPredictor` short-circuits before it ever reads the name:

```csharp
// FdrAnalysisEngine.cs:419-420
if (searchType != "standard")
    return null;
```

and `Compute_PEPValue` classifies `OligoSpectralMatch` as `"RNA"` and a top-down protease as `"top-down"`. All four offered models are peptide-only, so the whole `GroupBox` is inert in those modes — but it carries no `Visibility` binding, unlike the seven neighbouring analyte-specific groups in this same file that use `CollapseOnRnaModeConverter`.

Adding that binding (or disabling the group in RNA mode) would keep it honest. `GlycoSearchTaskWindow` has no RNA mode and needs no change.

This is the same question the CZE feedback raised, just for a mode where the answer really is "inert" — the CZE removal was correct because the predictor genuinely still runs there; here it genuinely doesn't.

trishorts (commented):
Second review round. The four findings from my previous round are all resolved in `c86cd3bb` — `rtPredictorName` now reaches `commonParamsToSave`, the `PrecursorMassMatchMode` ctor arg is restored in both places, and the CZE auto-hide is gone. Thanks.

One new **Major** below (the selected predictor is silently dropped before it reaches the engine), plus five smaller items on the new tests and the picker's visibility.

**Also: the PR description is now stale.** It still documents `GUI/Util/Converters/SeparationTypeToVisibilityConverter.cs` and a `Visibility` binding that collapses the picker under CZE — `c86cd3bb` removed both, and no such converter file exists in the branch. Removing it was the right call: `PEPAnalysisEngine.BuildFileSpecificDictionaries` calls `ComputeRetentionTimeEquivalentValues(..., RetentionTimePredictor)` unconditionally, so the predictor *is* invoked for CZE runs when building the training dictionaries; only the per-PSM z-score branches to `GetMobilityZScore`. The description just needs to catch up with the diff. The Test-plan line claiming the parameterized test "now covers Chronologer/SSRCalc/Prosit2019iRT/Prosit2020iRTTMT" is also inaccurate — there is no `[TestCase("Chronologer", ...)]`; Chronologer is covered by the separate `SearchTask_DefaultRTPredictor_ResolvedToChronologer` test instead.

MetaMorpheus/Test/FdrTest.cs
```diff
@@ -913,5 +916,59 @@ public static void FdrAnalysisEngine_GetRTPredictor_ReturnsExpectedPrositPredict
...
+            switch (searchType) // The default setting both works for GlycoSearch and SearchTask
+            {
+                case "GlycoSearch":
+                    task = new GlycoSearchTask();
+                    break;
+                default:
+                    task = new SearchTask();
+                    break;
+            }
+            task.CommonParameters = new CommonParameters(rtPredictorName: "SSRCalc");
```
trishorts: **Minor** — this line discards the task the test just built, so both `[TestCase]`s are identical

`switch (searchType)` above constructs a `GlycoSearchTask` or a `SearchTask`, and then this line throws that away:

```csharp
task.CommonParameters = new CommonParameters(rtPredictorName: "SSRCalc");
```

From here on both `[TestCase]`s execute byte-identical code — the `task` variable contributes nothing, and the assertion just re-tests the `[TestCase("SSRCalc", typeof(SSRCalc3RetentionTimePredictor))]` arm already added at line 899. If `GlycoSearchTask` stopped honouring the setting entirely, this test would stay green.

To actually test the two task types, clone the task's own `CommonParameters` with only the name overridden, so each case still runs its own defaults. Otherwise the test is redundant with line 899 and could be dropped.

Same shape in `SearchTask_DefaultRTPredictor_ResolvedToChronologer`: the `default:` arm silently builds a `SearchTask`, so a typo in a `[TestCase]` label gives you two `SearchTask` runs, zero glyco runs, and a green suite. An explicit `case "SearchTask":` with a failing `default:` (or parameterizing by `Type`) would make that visible.

MetaMorpheus/Test/FdrTest.cs
```diff
@@ -913,5 +916,59 @@ public static void FdrAnalysisEngine_GetRTPredictor_ReturnsExpectedPrositPredict
...
+            Assert.That(result, Is.Not.Null);
+            Assert.That(result, Is.InstanceOf<ChronologerRetentionTimePredictor>());
+        }
+
+        [Test]
+        [TestCase("GlycoSearch")]
+        [TestCase("SearchTask")]
+        public static void TaskRTPredictor_WhenSSRCalcSelected_ResolvedToSSRCalc3(string searchType) 
+        {
+            var method = typeof(FdrAnalysisEngine).GetMethod("GetRTPredictor",
```
trishorts, line 947: **Minor** — both new tests dropped the reflection guard their siblings carry

The two pre-existing reflection tests assert the lookup succeeded before invoking:

```csharp
// FdrTest.cs:883 and :908
Assert.That(method, Is.Not.Null, "GetRTPredictor not found via reflection — signature changed?");
```

`SearchTask_DefaultRTPredictor_ResolvedToChronologer` (line 926) and `TaskRTPredictor_WhenSSRCalcSelected_ResolvedToSSRCalc3` (this one) both omit it and go straight to `method.Invoke(...)`. If `GetRTPredictor` is ever renamed or its signature changes, these two fail with a bare `NullReferenceException` pointing at the invoke, while the other two tell you exactly what happened. Worth copying the guard across for consistency.

MetaMorpheus/Test/FdrTest.cs
```diff
@@ -894,6 +896,7 @@ public static void FdrAnalysisEngine_GetRTPredictor_UnknownName_ReturnsNull()
         //[Category("Koina")]
         [TestCase("Prosit2019iRT", typeof(Prosit2019iRT))]
         [TestCase("Prosit2020iRTTMT", typeof(Prosit2020iRTTMT))]
+        [TestCase("SSRCalc", typeof(SSRCalc3RetentionTimePredictor))]
```
trishorts: **Minor** — SSRCalc's only engine-side coverage is parked behind the Koina gate

The new `[TestCase]` lands on a method whose network gating sits two lines above it:

```csharp
//[Explicit("Constructs Koina-backed Prosit predictors. Excluded from normal CI; run with: dotnet test --filter Category=Koina")]
//[Category("Koina")]
```

Those are commented out today, so the case does run — but the moment anyone uncomments them (which is what they're there for), the *only* direct test of this PR's single engine change quietly leaves CI, even though `SSRCalc3RetentionTimePredictor` is a local hydrophobicity calculation that needs no network at all. Moving the SSRCalc case (plus an explicit Chronologer case) to a non-network test method would keep it in the default run.

Two naming leftovers while you're in here: the method is still `FdrAnalysisEngine_GetRTPredictor_ReturnsExpectedPrositPredictor` and its comment still reads *"Covers the two Prosit arms of GetRTPredictor's switch"* — it now covers three arms, one of which isn't Prosit. `..._ReturnsExpectedPredictor` would match.

MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml.cs
trishorts: Addressed in `c86cd3bb` — `SaveButton_Click` now passes `rtPredictorName: rtPredictorModelName` to the `CommonParameters` ctor (`SearchTaskWindow.xaml.cs:675`), so the computed value is no longer dead. Resolving.

MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml.cs
trishorts: Addressed in `c86cd3bb` — line 673 is back to `precursorMassMatchMode: _massDifferenceAcceptorViewModel.PrecursorMassMatchMode`. The GUI-building jobs (Create Installer Artifact, Test on Windows) are green, confirming the compile break is gone. Resolving.

MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml.cs
trishorts: Addressed in `c86cd3bb` — the fourth argument is restored: `_massDifferenceAcceptorViewModel = new(..., task.CommonParameters.PrecursorMassTolerance.Value, task.CommonParameters.PrecursorMassMatchMode)` (`SearchTaskWindow.xaml.cs:438`), so a saved task's match mode survives a round-trip through the window. Resolving.

MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml
trishorts: Addressed in `c86cd3bb` — the `Visibility` binding is gone from the `GroupBox` and `SeparationTypeToVisibilityConverter.cs` is no longer in the branch, so the picker stays visible under CZE. That is the correct outcome: `BuildFileSpecificDictionaries` invokes the predictor unconditionally when building the training dictionaries. Resolving.

MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml.cs
trishorts: Fixed in `f2ecc7298` — `SetAllFileSpecificCommonParams` now forwards `rtPredictorName: commonParams.RTPredictorName` (`MetaMorpheusTask.cs:617`), so the selection survives the file-specific merge and reaches `GetRTPredictor`. The new `SetAllFileSpecificCommonParams_PreservesRTPredictorName` test pins it against a non-null `FileSpecificParameters`, which is exactly the path that was dropping it. Resolving.

MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml.cs
trishorts: Fixed in `f2ecc7298` — new `EngineLayer/FdrAnalysis/RTPredictorNames.cs` holds the four names as consts, and I confirmed all five sites now reference it rather than literals: both windows' restore switches and save blocks, and the `GetRTPredictor` switch arms (`FdrAnalysisEngine.cs:427-430`). A typo is a compile error now. Resolving.

MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml
trishorts: Fixed in `f2ecc7298` — the `GroupBox` now carries `Visibility="{Binding Converter={StaticResource CollapseOnRnaModeConverter}}"`, matching the neighbouring analyte-specific groups in the same file. The picker no longer shows in a mode where `GetRTPredictor` short-circuits to null. Resolving.

MetaMorpheus/Test/FdrTest.cs
trishorts: Fixed in `f2ecc7298` — the redundant `TaskRTPredictor_WhenSSRCalcSelected_ResolvedToSSRCalc3` is gone (it no longer appears in the file), which was the cleaner of the two options since line 899 already covered that assertion. The `default:` concern is handled too: `SearchTask_DefaultRTPredictor_ResolvedToChronologer` now has an explicit `case "SearchTask":` and a `default: Assert.Fail($"Unrecognized searchType: {searchType}")`, so a mistyped label fails instead of silently running SearchTask twice. Resolving.

MetaMorpheus/Test/FdrTest.cs
trishorts: Fixed in `f2ecc7298`, and neatly — SSRCalc moved out to a new `FdrAnalysisEngine_GetRTPredictor_ReturnsExpectedLocalPredictor` alongside an explicit Chronologer case, with a comment saying it is kept separate so the coverage can never be excluded by the Koina gate. That also makes the old method honest again: it now carries only the two Prosit cases, so `...ReturnsExpectedPrositPredictor` and its "two Prosit arms" comment are accurate as-is. Resolving.

MetaMorpheus/Test/FdrTest.cs
trishorts: Mostly fixed in `f2ecc7298` — both tests I named now carry the guard (`SearchTask_DefaultRTPredictor_ResolvedToChronologer` at :948, and the reflection lookup in the SSRCalc path).

Leaving this open for one straggler: the **new** `FdrAnalysisEngine_GetRTPredictor_ReturnsExpectedLocalPredictor` added in the same commit does `GetMethod(...)` and goes straight to `method.Invoke(...)` with no `Assert.That(method, Is.Not.Null, ...)`. Same one-line addition as the others and it would be consistent across all four reflection tests. Not blocking.

trishorts: Closed out — I pushed the one-line addition directly to the branch as `4102cb6ec` rather than leave it hanging over an otherwise-finished PR.

```csharp
Assert.That(method, Is.Not.Null, "GetRTPredictor not found via reflection — signature changed?");
```

Wording copied verbatim from the three sibling tests, so all four reflection tests in `FdrTest.cs` are now consistent. I built `Test.csproj` and ran the RT-predictor tests locally before pushing: 8 passed, 0 failed. Full CI is re-running on the new head.

@RayMSMS — apologies for touching your branch directly; it was a test-only one-liner and everything else in this review round is already resolved. Shout if you would rather I had left it to you. Resolving.

trishorts (approved):

zhuoxinshi (approved):

nbollis (approved):

PR #2747, opened 2026-08-19
Three small cleanups: glyco output paths, a duplicated default, a duplicated test file

acesnik:
## Three small cleanups

🤖 Low severity, no behaviour change on Windows. Split into three commits so they can be taken separately.

### 1. Nine glyco output paths

They were written as `Path.Combine(OutputFolder + "\\oglyco" + ".psmtsv")` — `Path.Combine` with a single argument, which returns the string unchanged, so the call looks like it handles the separator and doesn't.

A backslash separates paths only on Windows; elsewhere it is an ordinary filename character, so these files land beside the output folder instead of inside it. Windows behaviour is identical either way.

No new test: `TestOGlyco` and `OutputTest` already read `Path.Combine(outputFolder, "Task", "oglyco.psmtsv")` after a real glyco run, so they assert the intended location — they just cannot fail on a Windows-only CI. This makes them pass off Windows too.

The two equivalent sites in `MetaMorpheusTask` are **not** included: #2721 already fixes them.

### 2. A duplicated default

`SearchParameters` and `GlycoSearchParameters` held byte-identical copies of the same nine-entry `ModsToWriteSelection` dictionary, comment block included. It decides which modification types reach a pruned database, so drift between them would change output for one task and not the other.

Extracted as `SearchParameters.DefaultModsToWriteSelection()` — a factory, not a static instance, because the dictionary is mutable and the task windows assign into it per modification type, so a shared instance would let one task's edits leak into another's. `RnaSearchParameters` deliberately overrides it with an RNA-specific set; left alone and now pinned.

The accompanying test **passes before and after** — the dictionaries were already identical. It guards against future drift and against the factory becoming a cached static; it is not evidence of a fix.

### 3. A duplicated test file

`Test/BioPolymerNotchFragmentIonComparerTest.cs` and `Test/UtilitiesTest/BioPolymerNotchFragmentIonComparerTest.cs` declared the same class name with the same eleven tests in two namespaces. The bodies differ only by using order, one trailing space, and a byte-order mark — an invisible marker some editors write at the start of a UTF-8 file, which makes two files look identical on screen while differing byte for byte. Kept the `UtilitiesTest` copy.

`...FromFromSpectralMatchHypothesisTest.cs` is **not** a duplicate — it exercises the `SpectralMatchHypothesis` overload of `Compare` rather than the tuple overload, and adds a `Compare_DifferentScores` case. Only its filename was wrong (`FromFrom`, while declaring `FromSpectralMatchHypothesisTest`), so it is renamed to match. No coverage lost; the eleven duplicated cases now run once rather than twice.

### Verification

Verified on macOS via a throwaway `net10.0` harness, since `Test.csproj` targets `net10.0-windows`; `Test.csproj` and `TaskLayer` compile clean. The full Windows suite has not been run locally.


nbollis (approved):

RayMSMS (approved):

MetaMorpheus/TaskLayer/GlycoSearchTask/PostGlycoSearchAnalysisTask.cs
```diff
@@ -132,13 +132,13 @@ public MyTaskResults Run(string OutputFolder, List<DbForTask> dbFilenameList, Li
...
-                        var writtenFileNOGlyco = Path.Combine(OutputFolder + "\\no_glyco" + ".psmtsv");
+                        var writtenFileNOGlyco = Path.Combine(OutputFolder, "no_glyco.psmtsv");
 
                         var ProteinLevelLocalization = GlycoProteinParsimony.ProteinLevelGlycoParsimony(allPsmsgly.Where(p => p.Accession != null && p.OneBasedStartResidue.HasValue).ToList());
-                        var seen_no_glyco_localization_file = Path.Combine(OutputFolder + "\\seen_no_glyco_localization" + ".tsv");
+                        var seen_no_glyco_localization_file = Path.Combine(OutputFolder, "seen_no_glyco_localization.tsv");
                         WriteGlycoFile.WriteSeenProteinGlycoLocalization(ProteinLevelLocalization, seen_no_glyco_localization_file);
 
-                        var protein_no_glyco_localization_file = Path.Combine(OutputFolder + "\\protein_no_glyco_localization" + ".tsv");
+                        var protein_no_glyco_localization_file = Path.Combine(OutputFolder, "protein_no_glyco_localization.tsv");
```
trishorts, line 141: Non-blocking, and pre-existing - not asking you to fix it here.

The nine individual-file equivalents further down (lines ~239-278) already use the two-argument form, so they were never broken by the bug this PR fixes. But they build the filename as `individualFileFolder + "oglyco" + ".psmtsv"` with no separator, so a run of `MyRun.raw` writes `MyRunoglyco.psmtsv` - where the sibling call at line 82 writes `MyRun_AllPSMs.psmtsv`. Same nine names, missing the underscore.

Happy to file it as its own issue rather than widen this PR beyond the three cleanups it advertises.

trishorts (commented):
Reviewed this one by hand rather than through the usual engine - at six files and a ~60-line source payload the engine costs more than it returns.

I verified all three claims in the description against the branch rather than taking them on trust:

1. **Path fix.** `Path.Combine` with a single argument does return the string unchanged, so those nine sites were a no-op wrapped around a hard-coded backslash. The replacements are correct, and `git grep` finds no remaining single-argument `Path.Combine` anywhere in TaskLayer. I also checked #2721's diff: it really does fix the two deferred sites at `MetaMorpheusTask.cs:1114` and `:1128`, so nothing falls between the two PRs.

2. **Shared default.** `GlycoSearchParameters` does *not* derive from `SearchParameters`, so this is a genuine cross-class share rather than a redundant re-assignment over a base-class value. Factory over cached static instance is the right call for the reason given - `RnaSearchParameters`, which does derive, replaces the dictionary wholesale, and the task windows assign into the instance per modification type.

3. **Duplicate test file.** Diffed the two copies with line endings normalized: they differ only by the namespace line and one trailing space. The kept `UtilitiesTest` copy carries all ten `[Test]` methods (the description says eleven - that count includes `Setup`). No coverage lost, and `BioPolymerNotchFragmentIonComparerFromSpectralMatchHypothesisTest` is correctly kept as a distinct file.

Build is clean and 164 targeted tests are green: 23 comparer plus the new defaults test, 101 glyco, 40 TOML/task round-trip (the last because `SearchParameters` is Nett-serialized and I wanted to confirm a new public static member is ignored).

One non-blocking note inline. Nothing blocking from me.

trishorts (approved):

PR #2748, opened 2026-08-20
Rebuild the peptide-only index instead of returning null (#2412)

### LINKED ISSUE #2412: XLMS search run failed: HE:P
**Describing the issue**

**Reproducing the error**

- Protein database

- Mass spectrometry file types

- Settings

- Example datafiles

**Expected behavior for GUI**

**Screenshots**

**Desktop information**

**Additional context**

[https://uploadnow.io/en/share?utm_source=fQPF0YW](url)

### LINKED ISSUE #2513: XL search crash
**Describing the issue**
 Search crashes reporting value can't be null parameter: "source"

**Reproducing the error**
 Same error using version 1.0.5 and 1.0.8

- Protein database
 HIV and Human

- Mass spectrometry file types
 Tried multi protease and just trypsin

- Settings

- Example datafiles

**Expected behavior for GUI**

**Screenshots**

**Desktop information**

**Additional context**

Image

[results.txt](https://github.com/user-attachments/files/20175994/results.txt)

acesnik:
🤖 Fixes #2412. Closes #2513.

`GenerateIndexes_PeptideOnly` ended with an empty branch:

```csharp
if (!successfullyReadIndices)
{
    // This is for the second round search, so successfullyReadIndices must be true, otherwise there are problems.
}
```

When the index could not be read, the method returned with the caller's `ref` parameter still `null`. XLSearchTask's second round passes that straight into the `CrosslinkSearchEngine` constructor, which does `peptideIndex.Select(...)`, so it reached the user as

```
Value cannot be null. (Parameter 'source')
```

with a stack trace naming the crosslink engine rather than the missing index.

Round one is unaffected: it calls `GenerateIndexes`, which does rebuild after a failed read. So round one masks the condition and round two crashes on it, which is why the reported frame is `XLSearchTask.cs:191` — the second-round engine construction — and not the first.

There are two ways into that branch, and the issue reports both: no index folder matched, or the read threw. `ReadPeptideIndex` throws `MetaMorpheusException` on a database carrying one accession twice with different sequences, and the `catch` was bare, so that message never reached the user who needed it.

## The change

The branch rebuilds the index, and reports why the read failed rather than discarding the reason. The rebuild is in memory on purpose: `CheckFiles` requires a fragment index before it will match a folder, so a peptide-only folder could never be reused, and writing one would leave a dead timestamped directory behind on every run.

## Verification

Two regression tests, each confirmed to fail before the change and pass after — one with no index on disk, one where an index exists but cannot be read:

```
Failed GenerateIndexesPeptideOnly_WithNoIndexOnDisk_StillReturnsAPeptideIndex
   a null peptide index reaches CrosslinkSearchEngine as ArgumentNullException(source)
Failed GenerateIndexesPeptideOnly_WhenTheIndexCannotBeRead_RebuildsAndWarns
   the unreadable index must be rebuilt, not left null
```

Run through a throwaway net10.0 harness referencing `EngineLayer`/`TaskLayer`, since `Test.csproj` is `net10.0-windows` and does not run on macOS. `Test.csproj` itself compiles clean. Windows CI is the first real run of these.

**Reproduced end to end.** The crash reproduces on `upstream/master` without either reporter's data: two FASTAs plus `Test/XlTestData/BSA_DSS_23747.mzML`, run the XL search once so `DatabaseIndex/<timestamp>/` is written, then truncate `peptideIndex.ind` to stand in for an index the current build cannot deserialize. Rerunning gives `Run failed, Exception: Value cannot be null. (Parameter 'source')` at `CrosslinkSearchEngine.cs:67` → `XLSearchTask.RunSpecific` — the frames from both issues. **9 of 9 runs crash before this change, 0 of 9 after**, and the new warning names the reason (`Attempted to read past the end of the stream`). Run through `CMD.dll` on macOS/arm64, .NET 10.

Two properties the reproduction exposed, which are what the issue reporters were describing:

- The failure is self-perpetuating. Round one hits the same unreadable folder, rebuilds silently and writes a *new* timestamped folder; round two re-selects a stale one and dies. Every run adds another dead index folder and the crash recurs indefinitely — which is why #2513 saw it identically on 1.0.5 and 1.0.8, and why deleting `DatabaseIndex` is the workaround that works.
- Which stale folder round two selects is unordered — `DirectoryInfo.GetDirectories()` returns filesystem order. With one stale folder the crash is intermittent (1 of 10 runs); with 20 it is 20 of 20. So it can come and go with nothing changed.

Also settled by those runs: `DatabaseLoadingEngine.SanitizeBioPolymerDatabase` collapses duplicate accessions to a single entry before indexing, for both a single FASTA and two overlapping ones. So on current master the duplicate-accession route into `ReadPeptideIndex`'s throw is already closed upstream of it, and the route exercised here is an index that cannot be deserialized. The fix covers either.

**Merge note:** #2742 edits `GenerateIndexes_PeptideOnly` a few lines from this change (`ReadFragmentIndex`→`ReadPrecursorIndex`) without filling the empty branch, so the two will conflict and whichever lands second needs a rebase. #2759 rewrites the same `GenerateIndexes` signature region.

**Still not verified:** neither reporter's own run, since both databases are private; and this has only been exercised on macOS — the folder-ordering non-determinism above is APFS behaviour, and NTFS enumerating alphabetically would make the crash *more* consistent on Windows, not less.

**Behaviour change:** a run that previously crashed here will now rebuild the index and continue, emitting a warning that names the reason. Searches that already worked are untouched — the new code only runs where the old code returned null.


MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -1475,20 +1476,34 @@ public void GenerateIndexes_PeptideOnly(IndexingEngine indexEngine, List<DbForTa
...
-                // This is for the second round search, so successfullyReadIndices must be true, otherwise there are problems.
+                if (readFailure != null)
+                {
+                    Warn("Could not read the existing peptide index, so it is being rebuilt. Reason: " + readFailure.Message);
+                }
+
+                // Rebuilt in memory and deliberately not written to disk: a folder holding a peptide
+                // index but no fragment index is never matched by CheckFiles, so writing one would only
+                // leave an unusable directory behind on every run.
+                Status("Running Index Engine...", new List<string> { taskId });
```
trishorts, line 1511: The fix is right and the diagnosis of *why round one masks it* is the part I'd have missed. One thing about where the rebuild lands, which I don't think blocks it.

**The rebuild is inside XLSearchTask's N-squared partition loop.**

The second-round pairing loop is nested (`XLSearchTask.cs:158` outer, `:173` inner), so `GenerateIndexes_PeptideOnly` is called `N + N(N-1) = N^2` times per spectra file: once as `indexEngine_a` per outer iteration, and once as `indexEngine_b` for every inner iteration where `nextPartition != currentPartition`. That whole block is itself inside the per-spectra-file loop at `:93`.

Only `N` distinct indexes exist. `indexEngine_b` is constructed from `proteinListSubset_b`, which depends on `nextPartition` alone, so partition `q`'s index is rebuilt once for every outer partition `p`.

That was cheap when the branch was empty and merely produced a null, and it stays cheap in the normal path because a successful read is a disk read. It is not cheap now:

- Each `indexEngine.Run()` builds the **fragment** index too. `IndexingEngine.cs:145-149` allocates and fills it unconditionally, with no peptide-only mode, so a method called `GenerateIndexes_PeptideOnly` now pays the full fragment-index cost and discards it. On a mammalian database that is the dominant term.
- The duplicate-accession failure is not transient. `ReadPeptideIndex` rejects that database every time, so all `N^2` calls take the rebuild branch, on every run, forever — not once while a stale index is refreshed.

So for a 4-partition crosslink search over 10 files, the reported crash becomes 160 full index builds instead of 16.

Caching the rebuilt index by partition number — the outer `peptideIndex_a` values are exactly the ones the inner loop re-derives — would make it `N` per file and would also collapse the `N^2` disk reads on the healthy path. That is a change in `XLSearchTask` rather than here, so it may well belong in its own PR; I'd rather it were written down than discovered later as "crosslink search got slow".

Two smaller things:

- **Pre-existing, not yours, and I checked the diff before saying so:** for a duplicate-accession database `GenerateIndexes` fails its read, rebuilds, and writes a *new* timestamped folder under `DatabaseIndex` (`GenerateOutputFolderForIndices`) that can never be read back. Every run leaves another one. Worth an issue on its own.
- `#2742` edits this same method (the precursor read becomes `ReadPrecursorIndex`) and changes `precursorIndex` from `List<int>[]` to a `FragmentIndex`. Whichever of the two merges second will conflict here — small, but the `ref` parameter type in the rebuild you added moves with it.

trishorts (commented):
Reviewed at `6a602003`. Green on all jobs; the change is small, the tests are real regression tests (both confirmed red first), and the reasoning about why the crash surfaces in round two rather than round one matches what I read in `XLSearchTask`.

Verified rather than assumed:
- The rebuilt peptide index points at the same `Protein` objects the read path would have produced — `proteinListSubset` is a `GetRange` over the same `proteinList` — so nothing downstream sees a different identity.
- The precursor half is symmetric with the read path, and `GenerateIndexesPeptideOnly_WhenAPrecursorIndexIsRequested_RebuildsThatToo` pins it with an assertion that a fragment index could not satisfy.
- `NUnit` runs fixtures serially here (no assembly-level `Parallelizable`, no runsettings), so the static `WarnHandler` subscription in the third test is not a flakiness risk.

One non-blocking comment inline about where the rebuild lands in the caller's loop, plus a pre-existing issue worth filing separately. Also note the PR is behind master.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
trishorts: Correcting myself on one premise above, since it changes what is worth chasing.

I justified "all N^2 calls take the rebuild branch, on every run" with the duplicate-accession route. I then checked whether that route is reachable, and I don't think it is through normal loading: `SanitizeBioPolymerDatabase` ends with an unconditional

```csharp
for (int i = 0; i < accessionGroup.Count - 1; i++) //-1 to keep the last one (most mods)
    bioPolymers.RemoveAll(p => ReferenceEquals(p, accessionGroup[i]));
```

which collapses every accession group to a single entry regardless of `TargetContaminantAmbiguity`. So `ReadPeptideIndex`'s duplicate-accession throw is not reachable from a sanitized `proteinList` — which is the same reservation you put in the PR body, and you were right to put it there.

**What that changes:**

- **Withdraw the drive-by** about a new dead `DatabaseIndex` folder on every run. It rested entirely on that route. I'm not filing an issue for it.
- **The N^2 point stands, on a better argument.** It doesn't need a persistent read failure at all: the rebuild is deliberately in-memory, so nothing it produces is ever cached — within one run, the same `N` indexes are rebuilt `N^2` times, and the condition recurs identically on the next run. The realistic route is the one the original comment named, a stale index from an older MetaMorpheus that `NetSerializer` cannot deserialize, and `GetExistingFolderWithIndices` will keep handing back a matching stale folder.
- **The healthy path is confirmed unaffected**, which I had not stated: round one's `indexEngine` (`XLSearchTask.cs:124`) is constructed with the same arguments and the same `currentPartition` as round two's `indexEngine_a`, so the folder round one writes does match in round two, for every partition. The branch really is an edge case.

Nothing here changes my read of the fix itself.

nbollis (approved):

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
trishorts: Written down, so resolving this thread: #2768. It carries the corrected argument — the N-squared holds because the rebuild is deliberately in-memory and never cached, not because of the duplicate-accession route I withdrew — plus the fragment-index cost and the per-partition cache as the fix. It belongs in `XLSearchTask`, not here.

Nothing in it touches this PR, which replaces a branch that returned `null` with one that returns an index: strictly better at every call count.

trishorts (approved):
🤖 Approving at `6a602003`. Nothing has been pushed since I reviewed, all jobs are green, and my one open thread was an observation about `XLSearchTask`'s pairing loop rather than a request against this diff — it is now #2768, and the thread is resolved.

Re-checked the fix itself rather than taking my earlier read on trust:

- The rebuild uses `indexEngine.Run()`, and in `XLSearchTask` that engine is constructed from `proteinListSubset_a` / `proteinListSubset_b` — the same subset round one indexed and wrote. So the rebuilt `peptideIndex` covers exactly what a successful read would have returned, not the full `proteinList`.
- Not writing the rebuilt index to disk is the right call, and for the reason the comment gives: `CheckFiles` never matches a folder holding a peptide index without a fragment index, so a written one would be unreadable litter.
- `GeneratePrecursorIndex` is honoured on the rebuild path, matching the read path above it. `XLSearchTask` does not ask for one today, which is exactly why the second test earns its place.
- All three tests fail on master for the right reason and pass here.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -1487,20 +1488,34 @@ public void GenerateIndexes_PeptideOnly(IndexingEngine indexEngine, List<DbForTa
...
+                    readFailure = e;
                 }
             }
 
             if (!successfullyReadIndices) //if we didn't find indexes with the same params
             {
-                // This is for the second round search, so successfullyReadIndices must be true, otherwise there are problems.
+                if (readFailure != null)
+                {
+                    Warn("Could not read the existing peptide index, so it is being rebuilt. Reason: " + readFailure.Message);
```
trishorts, line 1505: This makes the two siblings disagree about whether an unreadable index is worth mentioning.

`GenerateIndexes` still has the byte-identical swallow at `:1424-1432`, comment and all -- including the line that reads like a request for exactly this PR ("could put something here..."). After this change, an unreadable index is announced on the peptide-only path and silent on the main path, for the same underlying cause.

Since the diagnostic value is the same in both places and the code is now written, I would lift the same three lines into `GenerateIndexes` too. It is a small addition and it stops the inconsistency being something a future reader has to explain.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -1487,20 +1488,34 @@ public void GenerateIndexes_PeptideOnly(IndexingEngine indexEngine, List<DbForTa
...
+                }
+
+                // Rebuilt in memory and deliberately not written to disk: a folder holding a peptide
+                // index but no fragment index is never matched by CheckFiles, so writing one would only
+                // leave an unusable directory behind on every run.
+                Status("Running Index Engine...", new List<string> { taskId });
+                var indexResults = (IndexingResults)indexEngine.Run();
+                peptideIndex = indexResults.PeptideIndex;
+
+                if (indexEngine.GeneratePrecursorIndex)
```
trishorts, line 1515: Minor: this branch is unreachable today.

`XLSearchTask` is the only caller of `GenerateIndexes_PeptideOnly`, and both of its index engines are constructed with `generatePrecursorIndex: false` (`XLSearchTask.cs:166-168` and `:187-189`). So `indexEngine.GeneratePrecursorIndex` is always false on this path and the assignment never runs.

I think keeping it is the right call -- it is symmetric with the read path just above at `:1483` and with `GenerateIndexes:1460`, and the asymmetry would be the more surprising thing to find later. Your test comment says as much ("leaving the rebuild's half of that pair untested is how the two halves drift apart"), and `GenerateIndexesPeptideOnly_WhenAPrecursorIndexIsRequested_RebuildsThatToo` exercises it directly, so it is covered even though no production caller reaches it.

Just flagging it so it is a known-dead branch rather than an assumed-live one -- and the assertion in that test that every peptide appears *exactly once* across the bins is a nice touch, since it would fail against a fragment index that repeats ids.

trishorts (commented):
Good fix, and the reasoning for **not** persisting the rebuilt index is correct -- I checked it rather than taking it on trust.

`CheckFiles` (`MetaMorpheusTask.cs:1366-1372`) requires `fragmentIndex.ind` unconditionally at `:1370`, and it is the only matcher -- `GetExistingFolderWithIndices:1354` is its sole caller, and is itself the only entry point used by all three `Generate*Indexes` methods. So a peptide-only folder genuinely could never be matched. The reason is actually stronger than you state it: `GenerateOutputFolderForIndices` mints a *new timestamped* subfolder every time, so persisting would accrete one permanently-unusable directory per affected run, forever.

I also confirmed the failure chain in the issue: `XLSearchTask.cs:171,192` are the only callers, both index engines are built with the same arguments as round one's, so round one writes the folder round two reads -- and the assumption breaks exactly when `ReadPeptideIndex` throws on duplicate accessions (`MetaMorpheusTask.cs:1291`), a data condition the *write* path never hits. Round 1 succeeds, round 2's read throws, `peptideIndex` stays null, `CrosslinkSearchEngine` does `peptideIndex.Select(...)` and you get `ArgumentNullException(source)` pointing at the wrong place. That matches #2412 precisely.

`Warn` is the right channel (7/7 comparable sites in this file use it for recoverable degradation; `:263` and `:670` set the precedent for appending `e.Message` from a caught exception), and `Status("Running Index Engine...")` correctly mirrors `:1443`.

Two notes below.

Issue #1185, opened 2018-05-25
IO.Thermo Data from fusion lumos


lonelu:
e: System.AggregateException: One or more errors occurred. ---> System.Runtime.InteropServices.COMException: The server threw an exception. (Exception from HRESULT: 0x80010105 (RPC_E_SERVERFAULT))
   at MSFileReaderLib.IXRawfile5.GetNoiseData(Object& pvarNoisePacket, Int32& pnScanNumber)
   at IO.Thermo.ThermoFile.GetMsDataOneBasedScanFromThermoFile(Int32 nScanNumber, IXRawfile5 theConnection, ThermoGlobalParams globalParams, IFilteringParams filterParams)
   at IO.Thermo.ThermoStaticData.LoadAllStaticData(String filePath, IFilteringParams filterParams)
   at TaskLayer.MyFileManager.LoadFile(String origDataFile, Nullable`1 topNpeaks, Nullable`1 minRatio, Boolean trimMs1Peaks, Boolean trimMsMsPeaks) in C:\projects\metamorpheus\TaskLayer\MyFileManager.cs:line 105
   at TaskLayer.GptmdTask.<>c__DisplayClass6_0.<RunSpecific>b__8(Int32 spectraFileIndex) in C:\projects\metamorpheus\TaskLayer\GPTMDTask\GPTMDTask.cs:line 148
   at System.Threading.Tasks.Parallel.<>c__DisplayClass17_0`1.<ForWorker>b__1()
   at System.Threading.Tasks.Task.InnerInvokeWithArg(Task childTask)
   at System.Threading.Tasks.Task.<>c__DisplayClass176_0.<ExecuteSelfReplicating>b__0(Object )
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.Tasks.Parallel.ForWorker[TLocal](Int32 fromInclusive, Int32 toExclusive, ParallelOptions parallelOptions, Action`1 body, Action`2 bodyWithState, Func`4 bodyWithLocal, Func`1 localInit, Action`1 localFinally)
   at System.Threading.Tasks.Parallel.For(Int32 fromInclusive, Int32 toExclusive, ParallelOptions parallelOptions, Action`1 body)
[...]

PR #2749, opened 2026-08-20
refactor(mods): take the modification data files from the mzLib package

acesnik:
🤖 Draft — depends on smith-chem-wisc/mzLib#1185 being merged and released. The `1.0.586` pin here is a placeholder for whatever that release number turns out to be, so CI will fail to restore until it exists. Deletes 206,417 lines.

Nine modification data files are maintained here and in mzLib, or live here despite not being MetaMorpheus concepts. This takes them from mzLib instead. A tenth file is deleted outright.

## Ownership

mzLib is the owner. Its `Mods` registry backs `PeptideWithSetModifications`, `OligoWithSetMods`, the ProForma converter and the external-result readers, so it needs a self-contained set and can't wait for us to supply one. We don't use that registry at all — `GlobalVariables` has its own — and only need the files on disk.

Cost is one extra release hop when a mod changes. Sizing that against history: `EngineLayer/Mods/Mods.txt` has **3 commits ever**, last touched 2023-12-14.

## Why nothing else has to move

- `GlobalVariables.cs:411` enumerates `Mods\` with `Directory.GetFiles`, so a package-delivered file is picked up with no code change.
- `Product.wxs` sources every one of these from `$(var.GUI_TargetDir)…` — the **build output**, not the source tree. The components and their GUIDs already exist, so the installer is unchanged apart from the three items below.

## RnaMods.txt stops being an embedded resource

It was embedded to avoid adding an installer component (the comment at `GlobalVariables.cs:442` says so). There's no functional reason for it, and the disk path was already half-built:

- the protein-loading loop already skips files matching `"Rna"`, so a `Mods/RnaMods.txt` can't pollute the protein registry;
- `LoadRnaModifications` already reads `Mods/RnaCustomModifications.txt` from disk — only the built-in list was embedded.

It now loads like every other modification file, users can edit it, and it costs one new WiX component. `tmt.txt` becomes `TMT.txt` to match the payload's name.

Also in that loop: the `glyco.txt` and `Rna` skips now match on `Path.GetFileName(modFile)` rather than the whole path. As written, an install directory containing `Rna` anywhere in it would have skipped **every** modification file.

## substitutions.txt and surfactants.txt move too

Same reason as the rest: plain modification lists in mzLib's format, read by mzLib's `ModificationLoader` through the generic `Mods\` loop, with no MetaMorpheus-side parser. Their installer components already source from the build output, so `Product.wxs` is unchanged for them.

**`glyco.txt` deliberately stays.** It looks like the same kind of file and it is not — `LoadGlycans` reads it at `GlobalVariables.cs:559` and converts it into `Glycan` objects, and both `Glycan` and `GlycanDatabase` are MetaMorpheus types. Same goes for `Crosslinkers.tsv` and the `Glycan_Mods/` databases. Those can't move until the types do, which is a library design question rather than a packaging one. `aListOfmods.txt` (which mods are pre-selected) and `combos.txt`/`rnaCombos.txt` (read by `GPTMDTask`) are genuinely ours and stay.

## Protease cleavage mods come from mzLib too

`Mods\ProteaseMods.txt` duplicated mzLib's `protease_mods.txt` in content, though not in spelling — this file wrote `ID Homoserine lactone` + `TG M` where mzLib writes `ID Homoserine lactone on M`. `Modification`'s constructor (`Modification.cs:64-80`) normalises both:

| source | input | `IdWithMotif` | `OriginalId` |
|---|---|---|---|
| mzLib | `ID Homoserine lactone on M` | `Homoserine lactone on M` | `Homoserine lactone` |
| here | `ID Homoserine lactone` + `TG M` | `Homoserine lactone on M` | `Homoserine lactone` |

Identical objects, so nothing that names the modification in a toml or in MetaDraw settings has to change.

mzLib's copy could not be used before because `LoadDigestionAgents` assigned `LoadEmbeddedProteaseMods()` to a **discarded local** (`GlobalVariables.cs:593`), so those modifications never reached `AllModsKnown` — and `GlobalVariables.ProteaseMods`, declared public at line 38 for exactly this, was never populated. `LoadModifications` now fills it and folds the result in, `LoadDigestionAgents` reuses it rather than re-reading, and the local file is deleted.

**One asymmetry this removes:** unlike every other file in `Mods\`, `ProteaseMods.txt` had no installer component. It reached developer builds and the command-line zip but was not among the files the MSI writes to the Mods folder. Sourcing the modifications from the assembly makes every install agree. I can't run WiX here, so I'm reporting the missing component as fact and the install-time consequence as the reading that follows from it.

Net effect on the modification list is **one entry**: mzLib defines a second cleavage mod, for `CNBr_N`, renamed to `Homoserine lactone N-terminal on M` in smith-chem-wisc/mzLib#1185.

Verified by running `SetUpGlobalVariables` against the built output:

```
AllModsKnown: 4417
GlobalVariables.ProteaseMods: 2
  Homoserine lactone on M              (Peptide C-terminal.)
  Homoserine lactone N-terminal on M   (Peptide N-terminal.)
AllModsKnownDictionary has 'Homoserine lactone on M': True
AllModsKnownDictionary has 'Homoserine lactone N-terminal on M': True
any mod named Test*: False
errors reading mods: 0
```

## elements.dat is deleted, not moved

Nothing reads it. `Loaders.LoadElements()` is `[Obsolete]` with an **empty body** — `PeriodicTable` has been self-populating from a static constructor for some time. Its only references here were the copy rule, an installer component, and two test files calling the empty method, plus an unused `elementsLocation` const in `SetUpTests.cs`. That is 63 KB shipped to every user for nothing, and one fewer installer component now.

## This changes output — read this part

Our UniProt PTM list jumps eight releases, **2018_01 (31-Jan-2018) → 2026_01 (28-Jan-2026)**:

- **+129** entries, **−7** removed (`3-hydroxyvaline`, `Lactic acid`, `Hydroxylated arginine`, …)
- **3 existing entries changed monoisotopic mass.** One substantively: `N6-(beta-hydroxybutyryl)lysine` **87.044604 → 86.036780**. The other two gained a mass where they had none.
- 0 target-residue changes.

`Mods.txt` also picks up mzLib's `Water Loss on E` cross-reference (`Unimod; 27`, Glu→pyro-Glu) in place of ours (`Unimod; 23`, the generic Dehydrated) — same formula, more specific accession, but it is a change and someone should confirm which we want. `Carbamidomethyl on U` is new. `Carbamidomethyl on C` and `Oxidation on M` appear in the incoming `Mods.txt` and are byte-identical to the copies in `aListOfmods.txt`, so `AddMods` skips them rather than reporting a duplicate.

`substitutions.txt` and `surfactants.txt` are byte-identical to the copies being deleted, so they contribute nothing to this.

**What I measured:** ran the bundled `--test` (Calibrate → GPTMD → Search) twice, swapping only `ptmlist.txt`. `AllPSMs.psmtsv`, `AllPeptides.psmtsv`, `AllQuantifiedPeptides.tsv` and the GPTMD database are byte-identical. Re-ran it after the substitutions/surfactants/elements.dat changes and again after the protease-mod change; both are identical to the earlier build once the embedded date stamps are normalised.

**What that does not cover:** `--test` uses a plain FASTA, so it never exercises UniProt XML modified-residue annotations — which is exactly where 129 new entries would surface. I would not call this inert. A vignette run against the UniProt XML database is the way to close it.

## Verification

Built against a locally-packed mzLib carrying the payload: `dotnet build MetaMorpheus.sln -c UbuntuMac -p:EnableWindowsTargeting=true` — 0 errors. All eight files arrive in `CMD/bin/Release/net10.0/{Mods,Data}/`, including the transitive hop through `EngineLayer` that `Product.wxs` relies on, and `elements.dat` is gone. `CMD.dll -g` and `CMD.dll --test` both run clean, so startup mod loading works off the package files.

Not verified: `Test/Test.csproj` (needs `Microsoft.WindowsDesktop.App`) and the WiX change (MSBuild/Windows-only). Both are CI-only from here.


trishorts:
Early feedback on the draft, since the blocker is cross-repo and easy to miss.

**The version this bumps to does not carry the files it deletes yet.** The pin moves `1.0.585` → `1.0.586`, and the diff removes `ptmlist.txt`, `unimod.xml`, `PSI-MOD.obo.xml`, `elements.dat`, `Mods.txt`, `RnaMods.txt`, `substitutions.txt`, `surfactants.txt` and `ProteaseMods.txt` on the strength of getting them from the package instead. But smith-chem-wisc/mzLib#1185 — the PR that adds them to the nuspec — is still open, and `1.0.586`'s nuspec has no data-file entries. So this needs #1185 merged **and** a release cut after it, not just the bump.

That puts it in the same queue as #2718 and #2755, both of which are drafts waiting on an mzLib release for the same structural reason.

**One consumer worth checking when you pick this back up.** `EngineLayer/Util/IsobaricMassTag.cs` resolves reporter-ion m/z at runtime out of `GlobalVariables.AllModsKnown`, matching on `ModificationType == "Multiplex Label"` — so the TMT/iTRAQ/DiLeu channel masses come from `ptmlist.txt`/`unimod.xml` by way of whatever `GlobalVariables` loaded. If the load path or the ordering changes when these move to the package, isobaric quantification changes with it, silently and numerically. Worth a TMT search before and after as part of this, and it is the reason mzLib#1213 lists the mod-file source as an open question for the isobaric channel table.

Nothing else from me while it is a draft — the mechanical part of the change looks like what it says it is.

trishorts (commented):
Deleting MetaMorpheus's duplicate copies of the modification data is the right direction -- the asymmetry where an mzLib mod fix lands in `Omics/Resources/Mods.txt` but MetaMorpheus keeps loading its own `EngineLayer/Mods/*.txt` is a real and recurring trap, and this closes it. The `ProteaseDictionary.LoadEmbeddedProteaseMods()` / `LoadAndMergeCustomProteases(...)` calls are good: I confirmed both exist in mzLib master (`Proteomics/ProteolyticDigestion/ProteaseDictionary.cs:82` and `:249`).

**But the mechanism this PR depends on is not in any published mzLib, and not in mzLib master either.** Details inline on the csproj. As written I believe this ships a MetaMorpheus that silently loses most of its modifications, so please don't merge it on the current pin. The PR also has no CI runs at all and is currently `CONFLICTING` with master, which is probably why this hasn't surfaced as a red build yet.

One smaller thing, on the file-name filtering in `GlobalVariables.LoadModifications` (`:414`, `:419`): both tests are `Contains(...)` rather than an exact match. Now that the directory's contents come from a package rather than from files in this repo, a loose substring test is riskier than it was -- a future mzLib file, or a user's own `MyGlycoScratch.txt`, or anything with `rna` in the name, gets silently skipped with no diagnostic. Exact file-name comparisons against a known list would say what is meant and fail loudly when the package changes underneath.

trishorts:
**Retracting my review from earlier today — it was wrong, and I have deleted the inline comment.**

I claimed `MzLibIncludeModificationData` and the `build/mzLib.targets` mechanism did not exist in mzLib 1.0.588, and that this PR therefore depended on an unreleased local build. That is false. smith-chem-wisc/mzLib#1185 (*feat(nuget): ship the modification data files alongside the assemblies*) is merged, is an ancestor of tag `1.0.588`, and is documented in `release-notes/1.0.588.md` by name. mzLib master carries `mzLib/nuget/mzLib.targets`, and the nuspec ships it as `build\mzLib.targets` along with all eight data files under `modificationData\`. The pin in this PR is correct.

Two mistakes produced that, both mine and both avoidable:

- I searched the mzLib clone's **working tree**, which is checked out on a branch 48 commits behind master. My own preflight printed that fact and I read past it.
- The local NuGet cache had 1.0.587 but not 1.0.588. 1.0.587 genuinely does not contain #1185 — it landed in 1.0.588 — so the absence I saw was real for the version I looked at and I generalised it to the one you actually pinned.

The `9.9.1186` dev package I cited as evidence of unreleased work was simply a local build of the same merged change.

Sorry for the noise — that finding would have sent you chasing an mzLib PR that was already done. The rest of my review stands: the smaller point about `Contains(...)` rather than exact file-name matching in `GlobalVariables.LoadModifications` (`:414`, `:419`) is unaffected, and is a nit rather than a blocker. This PR is `CONFLICTING` with master and has no CI runs, so it does still need a merge from master before it can be judged green.

trishorts (approved):
Approving at `deb49a64b`. All 11 checks green, including `Validate Installer Contents` and both `Search with Built` / `Search with Installed`.

I verified the premise myself rather than taking the pin on trust, having got it wrong earlier in the day: smith-chem-wisc/mzLib#1185 is merged, is an ancestor of tag `1.0.588`, and `release-notes/1.0.588.md` documents `MzLibIncludeModificationData` by name. mzLib master carries `nuget/mzLib.targets` and the nuspec ships it as `build\mzLib.targets` alongside all eight data files under `modificationData\`. The pin in this PR is correct.

What I checked beyond CI:

- **Nothing matching `*custom*` reaches the build output.** The conflict resolution kept master's `EmbeddedResource` + `EnsureCustomMonosaccharideFileExists` seeding and dropped the `<None Update ... CopyToOutputDirectory>PreserveNewest</None>` rule this branch was still carrying for `MonosaccharidesCustom.tsv` — which is the exact mechanism #2752 removed because it overwrote users' edited files. Resolving that conflict the other way would have reintroduced the bug.
- **`proteases_custom.tsv` / `rnase_custom.tsv` seeding still happens only on the file-absent branch**, writing a header template.
- **RNA mods now come from `Omics.Mods.MetaMorpheusRnaModifications`**, mzLib's embedded copy. That keeps the property #2752 established here — carried in an assembly, so no installer, repair or upgrade can lose them — while giving the single source of truth this PR is for.
- A probe against the built output: 43 RNA modifications, 4417 protein modifications, 2 protease mods, **0 RNA mods leaked into the protein collection**, **0 errors reading mods**.

**Two results changes that should go in the PR body before merge.** Neither is a defect — both are the drift this PR exists to end, becoming visible — but neither is currently written down, and a reader of the description would not expect them:

1. **RNA modifications gain Potassium, and Sodium becomes well-formed.** MetaMorpheus's `RnaMods.txt` had the two merged into one malformed entry: Sodium carried no `MT` and no `CF`, and took Potassium's Unimod accession (530; Sodium is 30). Potassium was absent entirely. mzLib defines both correctly.
2. **Some modifications now write `UNIMOD:` into .mzID where they wrote `RESID:`.** mzLib's `ptmlist.txt` is a newer UniProt release (10,787 lines vs 8,608) that adds Unimod cross-references to entries that previously had only RESID, and `GetUnimodCvParam` (`MzIdentMLWriter.cs:767`) checks Unimod first. This is what took `TestMzIdentMlWriterWithUniprotResId` red; it is now pinned on a modification that still has RESID and no Unimod, so the branch stays covered.

Both are arguably improvements. They are still output changes, and worth a line each so nobody rediscovers them from a diff of results.

One non-blocking note: `Appsrc_Mods_RnaMods.txt` installs a file that nothing now reads, since the RNA mods come from the assembly. The component is nonetheless required — mzLib's opt-in ships `Mods\` as one group, so the file is in the build output, and `Validate Installer Contents` demands every output file be listed. Correct as it stands; just worth knowing it is deliberate rather than leftover.

This is 1 of the 3 approvals master requires.

nbollis (approved):

MetaMorpheus/EngineLayer/GlobalVariables.cs
```diff
@@ -37,6 +37,25 @@ public static class GlobalVariables
...
+        /// protein collection.
+        /// Matched by whole file name rather than by substring. The folder's contents now arrive from the
+        /// mzLib package rather than from this repository, so a file added upstream -- or a user's own
+        /// "MyGlycoScratch.txt" dropped in beside them -- must not be skipped silently on the strength of
+        /// containing "glyco" or "rna" somewhere in its name.
+        /// </summary>
+        private static readonly HashSet<string> ModFilesLoadedElsewhere = new(StringComparer.OrdinalIgnoreCase)
+        {
+            "glyco.txt",
+            "RnaMods.txt",
```
trishorts, line 55: This is the finding worth resolving before merge.

`RnaMods.txt` is in the skip set, and `LoadRnaModifications` at line 474 takes the RNA mods from `Omics.Modifications.Mods.MetaMorpheusRnaModifications` instead. I grepped the head revision: the only disk read left under `Mods\` for RNA is `RnaCustomModifications.txt` at line 476. Nothing reads `Mods\RnaMods.txt`.

But `Product.wxs` gains a brand-new component that installs it (note there too). So the MSI writes a 196-line modification file into the user's `Mods\` folder, beside `Mods.txt`, `TMT.txt` and `substitutions.txt` -- all of which *are* live -- and edits to it do nothing, silently. That is a worse outcome than not shipping it, because the folder's whole convention is that files in it get read.

The body says the opposite:

> It now loads like every other modification file, users can edit it, and it costs one new WiX component.

The code comment right here is the newer and correct account; that body section describes an earlier design.

Two coherent endings, and I would rather you pick than have me pick:

1. **Drop the new WiX component.** The file still lands in the build output because `mzLib.targets` globs `Mods\*`, so developer builds keep it, but the MSI stops writing a file nothing reads. Costs an install-vs-build asymmetry, which is the kind of thing this PR elsewhere works to remove.
2. **Read it from disk after all**, as the body describes -- but then the embedded `AddMods` at 474 has to go, or the RNA mods double.

Either way the body needs to match whichever you choose.

trishorts (commented):
Read this one by hand rather than running the engine -- the 206k-line count is data deletion, the actual payload is `GlobalVariables.cs`, `Product.wxs`, `EngineLayer.csproj` and three test files, and the engine has no model of the WiX/nuspec seam that carries most of the risk here.

**What I checked and found sound:**

- **Load order holds.** `ProteaseMods` is now assigned in `LoadModifications` and consumed in `LoadDigestionAgents`. `SetUpGlobalVariables` calls them at lines 92 and 99 respectively, so the field is populated before it is reused. This was the thing most likely to be quietly broken and it is not.
- **The payload matches the WiX `Source=` paths exactly.** mzLib's nuspec ships eight files -- `modificationData\Mods\{Mods.txt, RnaMods.txt, TMT.txt, substitutions.txt, surfactants.txt}` and `modificationData\Data\{ptmlist.txt, PSI-MOD.obo.xml, unimod.xml}` -- and `mzLib.targets` links them into `Mods\` and `Data\` with `CopyToOutputDirectory`, which is what every `$(var.GUI_TargetDir)` source in `Product.wxs` reads from. `protease_mods.txt` is deliberately not in that list, consistent with taking the cleavage mods from the assembly instead.
- **`MzLibIncludeModificationData` and `Omics.Modifications.Mods.MetaMorpheusRnaModifications` both exist on mzLib master**, so neither reference is speculative.
- **The `SearchTaskTest` swap is exactly right, and I checked it against the incoming file rather than the reasoning.** In mzLib's `Omics/Resources/ptmlist.txt`, `FMN phosphoryl threonine` now carries `DR   Unimod; 442.` -- so it takes the Unimod branch and the RESID branch goes untested, precisely as described. `(E)-2,3-didehydrobutyrine` has `TG   Threonine.`, `DR   RESID; AA0547.`, `DR   PSI-MOD; MOD:01470.` and no Unimod line. Correct diagnosis, correct replacement.

**One thing to decide before merge**, on `Mods\RnaMods.txt` -- inline below. Short version: the PR adds a new installer component for a file that nothing reads, and the body's description of that file is the opposite of what the code does.

**The body is stale in its first paragraph.** It opens "Draft -- depends on smith-chem-wisc/mzLib#1185 being merged and released. The `1.0.586` pin here is a placeholder ... so CI will fail to restore until it exists." All of that has moved on: #1185 merged 2026-08-27, master is already on 1.0.588, the branch carries **no pin change at all** (every `.csproj` in this diff is touched only for a byte-order mark), the PR is not a draft, and all twelve checks pass. Worth rewriting so a reviewer does not open it braced for red CI.

**Two smaller things:**

- `EngineLayer/Util/IsobaricMassTag.cs:201` still reads "the `# 120` comment on that DI line in `Mods/tmt.txt`". The incoming file is `TMT.txt`, and the CLI runs on Linux and macOS where that difference is real. That file is not otherwise in this diff, so it needs a deliberate touch.
- Worth being clear about what the green **Validate Installer Contents** check covers here: it diffs *DLLs* between the build output and the MSI, at the top level and `runtimes\win-x64\native`. It does not look at `Mods\` or `Data\` at all, so it is not evidence that the eight package files reached the installed layout or that dropping `elements.dat` was clean. **Search with Installed** is the check that actually carries that, since startup mod loading has to succeed against the installed folder -- and it passes. Fine as-is, just not the coverage the job name suggests.

MetaMorpheus/EngineLayer/GlobalVariables.cs
```diff
@@ -445,17 +466,12 @@ private static void LoadRnaModifications()
...
-                string fileContent = reader.ReadToEnd();
-                var mods = ModificationLoader.ReadModsFromString(fileContent, out var errors);
-                AddMods(mods, false, true);
-            }
+            // The RNA modifications come from mzLib's embedded copy, the same one this branch already
+            // takes the protease cleavage mods from. That keeps a single source of truth with
+            // Omics.dll while preserving the property #2752 established when it made these an
+            // embedded resource here: they are carried in an assembly, so no installer, repair or
+            // upgrade can leave them missing. A file in Mods\ could.
+            AddMods(Omics.Modifications.Mods.MetaMorpheusRnaModifications, false, true);
```
trishorts, line 474: Small one, on the citation rather than the code: #2752 is an open *issue* -- "Custom monosaccharides are lost on install/upgrade: MonosaccharidesCustom.tsv is installer-managed under OVERWRITEAPPDATA" -- not the change that made the RNA mods an embedded resource here.

The property the comment describes is real, and it is the right reason to keep them in the assembly. It just is not what #2752 established. Point it at whatever actually made them embedded, or drop the number and keep the sentence.

MetaMorpheus/MetaMorpheusSetup/Product.wxs
```diff
@@ -582,7 +582,14 @@
       <Component Id="Appsrc_Mods_tmt.txt" Guid="f064b7a0-c650-4a40-8a9f-301cc6cb946e" Condition="OVERWRITEAPPDATA">
         
         <RegistryValue Root="HKCU" Key="Software\SmithGroup\MetaMorpheus" Name="installed" Type="integer" Value="1" KeyPath="yes" />
-        <File ReadOnly="no" Id="Appsrc_Mods_tmt.txt" Name="tmt.txt" Source="$(var.GUI_TargetDir)Mods\tmt.txt" />
+        <File ReadOnly="no" Id="Appsrc_Mods_tmt.txt" Name="TMT.txt" Source="$(var.GUI_TargetDir)Mods\TMT.txt" />
+      </Component>
+
+      <!-- RnaMods.txt -->
+      <Component Id="Appsrc_Mods_RnaMods.txt" Guid="8BC69D83-0C35-49A1-A907-A30CFE46190D" Condition="OVERWRITEAPPDATA">
```
trishorts, line 589: This is the component the `ModFilesLoadedElsewhere` note refers to: it installs a file no code path reads.

Worth saying that the GUID is new, so dropping it costs nothing -- no shipped MSI has ever carried `8BC69D83-0C35-49A1-A907-A30CFE46190D`, so there is no upgrade row to orphan.

The `tmt.txt` to `TMT.txt` rename directly above is right and matches the nuspec (`Omics\Resources\TMT.txt`). Note that the component `Id` and the `<!-- tmt.txt -->` comment on line 581 still say `tmt`. The `Id` is opaque so the installer does not care, but the comment now mislabels the file it sits on.

MetaMorpheus/CMD/CMD.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
trishorts, line 1: This file is in the changed-file list for a byte-order mark and nothing else. Same for `GUI.csproj`, `GuiFunctions.csproj`, `TaskLayer.csproj`, `Test.csproj` and line 1 of `SearchTaskTest.cs` -- six files whose entire diff is `-<Project` / `+﻿<Project`.

On a PR that already deletes 206k lines and touches the installer, six files of pure encoding change are six extra conflict surfaces and six files a reviewer opens for nothing. I would revert them and keep the diff to files with a reason to be in it.

It also cost me a detour: the body says the `1.0.586` pin is a placeholder, so I opened these expecting a version bump and found a BOM. The pin turned out to be moot -- master is already on 1.0.588.

Alexander-Sol (approved):

PR #2751, opened 2026-08-21
 Add GUI for creating custom monosaccharides (glycan customization, Part A)

RayMSMS:
## Summary

Custom monosaccharides (name, mass, diagnostic ions) previously had to be hand-edited into `Glycan_Mods/MonosaccharidesCustom.tsv`, a tab-separated file guarded only by a comment-based instruction banner. Mistakes (bad formatting, or a glycan database referencing a monosaccharide that wasn't registered yet) surfaced as an unhandled exception at the next app launch, with no indication of what went wrong or why.

This PR adds a GUI window for defining custom monosaccharides in place of direct file editing, as the first step (Part A) of a larger glycan customization workflow.

## What's included

* **`CustomMonosaccharideWindow`** (`GUI/Views`): defines a custom monosaccharide via name, single-character code, chemical formula (parsed with mzLib's `ChemicalFormula`, following the same pattern as `CustomModWindow`/`CustomAminoAcidWindow`) or manual mass, diagnostic ions, and a description.

  * Validates inputs before committing: name/code uniqueness, `"Name"` is rejected as reserved (otherwise it would be silently treated as the TSV header row and disappear on the next launch), and mass must be positive and below 20,000 Da.
  * If both a chemical formula and a mass are entered and they disagree by more than a rounding step, the user is warned which value will actually be used before saving, rather than silently discarding the typed value.
  * Registers the monosaccharide live via `Glycan.RegisterCustomMonosaccharide` so it is usable immediately, and persists it via `GlycanDatabase.PersistCustomMonosaccharide` for future sessions.

* **`GlycanDatabase.PersistCustomMonosaccharide`**: implements the write side of the exact TSV format consumed by `LoadCustomMonosaccharides` (column order, header row, and invariant-culture decimal formatting). Keeping this logic in one place prevents the reader and writer formats from drifting apart. It also guards against appending to a hand-edited file that does not end with a newline, which would otherwise concatenate the new row with the previous line and fail to parse on the next launch.

* **Entry point**: adds "Create new monosaccharide" to the Settings tab, directly beneath "Create new modification" — a global registry action consistent with how custom mods, amino acids, and crosslinkers are already exposed. A "Build My Own Database" stub is also included in `GlycoSearchTaskWindow` as a placeholder for the glycan database designer (Part B, tracked separately).

* **Test coverage**: `PersistCustomMonosaccharide_RoundTripsThroughLoadCustomMonosaccharides` in `Test/TestNGlyco.cs` exercises the writer directly against the reader without involving the GUI, providing round-trip protection for the persisted TSV format.

## Known limitation

Custom monosaccharides added through this window are persisted to `Glycan_Mods/MonosaccharidesCustom.tsv`. However, this file is also managed by the Windows installer (`MetaMorpheusSetup/Product.wxs`, `Appsrc_Data_MonosaccharidesCustom.tsv`) and is overwritten when `OVERWRITEAPPDATA` is enabled (default: `1`). As a result, user-added monosaccharides persist across application sessions but may be lost after an install, repair, or upgrade.

Other custom-registry files (`CustomModifications.txt`, custom amino acids, and custom ion losses) are not installer-managed and therefore do not have this issue.

This is not addressed in this PR. A planned follow-up is to store user-defined monosaccharides in a separate, non-installer-managed file (e.g. `MonosaccharidesCustom.user.tsv`) and update `GlobalVariables.LoadGlycans()` to load both files at startup.

## Out of scope for this PR

* The glycan database designer itself (Part B).
* Hardening the composition-format `.gdb` loader (`String2Kind` in `GlycanDatabase.cs`), which currently throws a raw `KeyNotFoundException` for an unrecognized monosaccharide name without file/line context.
* Cleanup of the custom-monosaccharide registry's internal data structures in `Glycan.cs` — a separate maintainability pass.

## Testing

* `dotnet build GUI/GUI.csproj` and `dotnet build Test/Test.csproj` — 0 errors.
* `dotnet test` on `Test/TestNGlyco.cs` — all 24 tests pass, including the new round-trip test.
* No changes to any per-scan search-loop code (`GlycoSearchEngine.RunSpecific` and everything it calls); all new/changed code runs at load time or in response to a UI action, so there is no impact on glyco search performance.


trishorts (commented):
Reviewed the full diff against `master` (`2795a3ffc`), with every sibling quoted from `git show upstream/master:` rather than a local branch. Nice, well-scoped Part A — the entry point matches how custom mods/amino acids/crosslinkers are already exposed, `Views/` is the right home, no `.csproj` change is needed, and the `MainWindow.xaml` row renumbering is correct (rows shift 4→5, 5→6, 6→7, 8→9, 9→10, 10→11, spacers preserved at 4 and 8).

**Things I checked and found clean**, so nobody re-litigates them:

* **The TSV round-trip is exact.** The written line matches `LoadCustomMonosaccharides` column-for-column: header skipped at `GlycanDatabase.cs:87`, ≥3 columns at `:92`, `NumberStyles.Float`/invariant mass parse at `:111`, same `* 1E5` scaling at `:116`. Column 5 (Description) is read and ignored at `:102` — fine. The write path also targets the identical `DataDir\Glycan_Mods` path that `GlobalVariables.LoadGlycans` reads at `:473`.
* **This composes with #2665.** That PR adds a 0.01 Da diagnostic-ion collision rule inside `RegisterCustomMonosaccharide` which throws `ArgumentException` — already caught at line 64 here, so a colliding ion produces a dialog and no file write, rather than a file that fails to load at next launch. No change needed on either side.
* **The bare `catch` blocks in `ErrorsDetected` are correct**, not lazy: `ChemicalFormula.ParseFormula` throws `MzLibException` on a bad format (`mzLib/Chemistry/ChemicalFormula.cs:188`) but `KeyNotFoundException` from `PeriodicTable.GetElement` on an unknown symbol, so catching narrowly would leak one of them.
* **Culture handling is right** — every parse passes `CultureInfo.InvariantCulture`, which matters here because the comma is the diagnostic-ion separator.
* **Validation style, `DialogResult` without `Close()`, the duplicated CueBanner styles, no `IsDefault`/`IsCancel`, and no test for the window** all match the house pattern as it actually is; none of those are worth changing in this PR.

Ten comments below. Nothing here is a merge blocker on its own — the two I'd most want addressed are the stray `using` block and the required-field asterisks, both one-line fixes. The persistence-ownership and installer-overwrite notes are design questions worth an answer before Part B builds on this, but they don't have to be solved here.

<sub>Reviewed with Claude Code assistance; every finding was confirmed against the source at `2795a3ffc` before posting.</sub>

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml.cs
```diff
@@ -0,0 +1,171 @@
...
+using EngineLayer.GlycoSearch;
+using Google.Protobuf.Collections;
+using Org.BouncyCastle.Asn1.BC;
+using System;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Security.Policy;
+using System.Windows;
+using static System.Runtime.InteropServices.JavaScript.JSType;
```
trishorts: **Four unused `using` directives — IDE autocomplete accidents.** None of these types are referenced anywhere in the file:

```csharp
using Google.Protobuf.Collections;   // RepeatedField/MapField — unused
using Org.BouncyCastle.Asn1.BC;      // unused
using System.Security.Policy;        // Evidence/Url/Site/Zone — unused
using static System.Runtime.InteropServices.JavaScript.JSType;   // unused
```

The last one is more than cosmetic: `using static` on `JSType` imports its nested types `String`, `Object`, `Number`, `Array`, `Boolean`, `Function`, `Error`, `Date` into this file's scope. It compiles today only because the file consistently uses the lowercase `string` keyword — the first time anyone writes `String.Join(...)` here it becomes an ambiguous reference. It also drags a `[SupportedOSPlatform("browser")]` type into a `net10.0-windows` WPF file.

This is not a repo habit (the only comparable case anywhere is `GUI/Util/TaskValidator.cs:1`), so it reads as a slip rather than a convention. Please delete all four lines.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml
```diff
@@ -0,0 +1,187 @@
...
+                                </Trigger>
+                            </Style.Triggers>
+                        </Style>
+                    </TextBox.Style>
+                </TextBox>
+            </StackPanel>
+            <!--Diagnostic ions-->
+            <StackPanel Orientation="Horizontal" Margin="2">
+                <Label Content="Diagnostic Ions (m/z)" Width="175"></Label>
+                <Label Content="*" Foreground="Red" FontWeight="Bold"/>
```
trishorts: **The red `*` markers contradict this window's own tooltips and the house convention.** `*` is on all five inputs, but the tooltips right beside them say:

- Chemical Formula — *"optional if monoisotopic mass is specified"*
- Monoisotopic Mass — *"optional, if chemical formula is specified"*
- Diagnostic Ions — *"optional, separate with ','"*

`ErrorsDetected` agrees: it never requires ions, and accepts formula-**or**-mass.

`CustomModWindow.xaml` already solved exactly this "one of two" case — the asterisk goes on a shared `and/or` row, not on either field, and genuinely-optional fields carry none (verified on master):

```xml
<!--Chemical formula-->
<StackPanel Orientation="Vertical" Background="LightGray" Width="350">
    <StackPanel Orientation="Horizontal">
        <Label Content="Chemical Formula" Width="190"></Label>   <!-- no asterisk -->
        ...
    <StackPanel Orientation="Horizontal">
        <Label Content="and/or" Height="20" FontSize="9" Width="175"/>
        <Label Content="*" Foreground="Red" FontWeight="Bold" />
    </StackPanel>
    <!--Monoisotopic mass-->
    <StackPanel Orientation="Horizontal">
        <Label Content="Monoisotopic Mass" Width="190"></Label>   <!-- no asterisk -->
```

(`CustomModWindow.xaml:107-146`; its optional Neutral Losses `:192` and Diagnostic Ions `:222` carry no asterisk.)

Suggested: drop the `*` on **Diagnostic Ions** (line 142) outright, and group **Chemical Formula** + **Monoisotopic Mass** under a single `and/or` + `*` row copied from `CustomModWindow.xaml:139-142`. Keep `*` on Name and Single-Char Code.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml.cs
```diff
@@ -0,0 +1,171 @@
...
+                    MessageBox.Show("Could not parse chemical formula", "Error", MessageBoxButton.OK, MessageBoxImage.Hand);
+                    return true;
+                } 
+            }
+
+            if (!string.IsNullOrEmpty(mass) && !double.TryParse(mass, NumberStyles.Float, CultureInfo.InvariantCulture, out _))
+            {
+                MessageBox.Show("Could not parse monoisotopic mass", "Error", MessageBoxButton.OK, MessageBoxImage.Hand);
+                return true;
+            }
```
trishorts: **No range or sign check on the mass.** `double.TryParse` only proves the text is a number, and that value goes straight into `(int)Math.Round(massDa * 1E5)` at line 50.

Two reachable consequences:

1. **Negative / zero mass.** `-176.03` and `0` both parse, both register. A monosaccharide with a negative residue mass is silently accepted into `Glycan`'s kind table.
2. **Silent saturation above ~21474.8 Da.** `int.MaxValue / 1E5 = 21474.83647`. A realistic typo — entering the already-scaled integer `17603209`, or a mass in the wrong unit — yields `1.76e12`, which since .NET Core 3.0 saturates to `int.MaxValue` on the cast rather than throwing. The user is told nothing, a monosaccharide of 21474.83647 Da is registered, and that wrong value is then persisted to the TSV so it comes back on every subsequent launch.

Residue masses here run from ~14 Da (`Me`) to ~310 Da (`NeuGc`), so a cheap guard covers it:

```csharp
if (!double.TryParse(mass, NumberStyles.Float, CultureInfo.InvariantCulture, out double massDa)
    || massDa <= 0 || massDa > 20000)
{
    MessageBox.Show("Monoisotopic mass must be a positive number below 20000 Da", "Error", MessageBoxButton.OK, MessageBoxImage.Hand);
    return true;
}
```

The same applies to the diagnostic-ion values at line 159 — they take the identical `* 1E5` cast at line 56 with no bound. Given the stated purpose of this PR is to stop bad values reaching the file, catching them here rather than at the next launch seems in scope.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml.cs
```diff
@@ -0,0 +1,171 @@
...
+            if (string.IsNullOrWhiteSpace(name))
+            {
+                MessageBox.Show("The monosaccharide name needs to be specified", "Error", MessageBoxButton.OK, MessageBoxImage.Hand);
+                return true;
+            }
+            else if(Glycan.NameCharDic.ContainsKey(name))
+            {
+                MessageBox.Show("A monosaccharide already exists with the name: " + name, "Error", MessageBoxButton.OK, MessageBoxImage.Hand);
+                return true;
+            }
```
trishorts: **A monosaccharide named `Name` is silently discarded on the next launch.** The reader treats any line beginning with `Name` + tab as the optional column-header row and skips it:

```csharp
// GlycanDatabase.cs:87
if (line.TrimStart().StartsWith("Name\t", StringComparison.OrdinalIgnoreCase))
{
    continue;
}
```

`Glycan.NameCharDic` has no `Name` key, so this validator accepts it, `RegisterCustomMonosaccharide` accepts it, the window reports success, and the row is written — then vanishes at startup with no error. Same for `name`/`NAME` (the comparison is `OrdinalIgnoreCase`).

One extra clause here fixes it:

```csharp
else if (name.Equals("Name", StringComparison.OrdinalIgnoreCase))
{
    MessageBox.Show("'Name' is reserved for the column-header row and cannot be used as a monosaccharide name", "Error", MessageBoxButton.OK, MessageBoxImage.Hand);
    return true;
}
```

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml.cs
```diff
@@ -0,0 +1,171 @@
...
+            catch (ArgumentException ex)
+            {
+                MessageBox.Show(ex.Message, "Error", MessageBoxButton.OK, MessageBoxImage.Hand);
+                return;
+            }
+
+            //persist to file so the monosaccharide is still recognized on the next launch
+            string glycanModsDirectory = Path.Combine(GlobalVariables.DataDir, @"Glycan_Mods");
+            string customMonosaccharidePath = Path.Combine(glycanModsDirectory, "MonosaccharidesCustom.tsv");
+            string line = string.Join("\t", name, code.ToString(), massDa.ToString(CultureInfo.InvariantCulture), ionsText, descriptionText);
```
trishorts: **The TSV format is now defined in two places that must agree, and only one of them can be tested.**

The reader owns this format today — column order, the `Name\t` header rule, the `#`-comment convention, the `NumberStyles.Float`/invariant parse and the `* 1E5` scaling all live in `GlycanDatabase.LoadCustomMonosaccharides` (`GlycanDatabase.cs:68-146`). This line re-derives the writing half of that contract in GUI code-behind. That asymmetry has teeth: a divergence here does not surface as a dialog error, it surfaces as a `MetaMorpheusException` thrown from `GlobalVariables.LoadGlycans` at **application startup** (`GlobalVariables.cs:473-474`).

And it cannot be covered by a test — `Test/Test.csproj:37-42` references CMD, EngineLayer, GuiFunctions and TaskLayer, but **not** `GUI.csproj`, so nothing under `MetaMorpheus/GUI/**` is reachable from the test project.

There is a newer sibling that solves exactly this. `CustomMIonLossWindow.xaml.cs:97` persists in one line — `CustomMIonLossManager.AddCustomMIonLoss(customLoss);` — with no `File.*` call and no string formatting in the code-behind at all. The format lives on the model (`GuiFunctions/Models/CustomMIonLoss.cs:45` `Serialize()` / `:53` `Deserialize()`), file I/O and dedupe live in `GuiFunctions/Util/CustomMIonLossManager.cs`, and the whole thing is tested in `Test/GuiTests/CustomMIonLossTests.cs`. `CustomAminoAcidWindow.xaml.cs:82` does hand-build its line the way this PR does — but that code is from 2023, and the M-ion-loss pattern is from 2026 (#2627).

Suggested: put an `AppendCustomMonosaccharide(...)` next to `LoadCustomMonosaccharides` in `GlycanDatabase` so reader and writer share the format constants, and let this window shrink to `InitializeComponent`, event wiring, `MessageBox` on the typed exception, and `DialogResult`. A round-trip test (`write → LoadCustomMonosaccharides → assert registered`) then becomes possible, which is the one test that would actually protect this feature.

Small related point: `(int)Math.Round(massDa * 1E5)` at line 50 and line 56 duplicates `GlycanDatabase.cs:116` and `:130`. `1E5` is already an unnamed literal in 35 places across 7 glyco files, so this is not a problem this PR created — but if a writer lands on `GlycanDatabase`, a single `Glycan.ToScaledMass(double)` used by both sides is nearly free at that point. Converting the other call sites is separate cleanup and shouldn't be asked of this PR.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml.cs
```diff
@@ -0,0 +1,171 @@
...
+            try
+            {
+                Directory.CreateDirectory(glycanModsDirectory);
+                if (!File.Exists(customMonosaccharidePath))
+                {
+                    File.WriteAllLines(customMonosaccharidePath, new[] { "Name\tSingleCharCode\tMonoisotopicMass\tDiagnosticIonMasses\tDescription", line });
```
trishorts: **This branch is effectively unreachable, and the file it targets is installer-managed — which puts the persisted entries at risk.**

`MonosaccharidesCustom.tsv` is not a create-on-demand file like `CustomModifications.txt` or `CustomMIonLosses.txt`. It ships: `EngineLayer.csproj` copies it to output, and the installer places it explicitly —

```xml
<!-- Product.wxs:649-653 -->
<Component Id="Appsrc_Data_MonosaccharidesCustom.tsv" Guid="8233..." Condition="OVERWRITEAPPDATA">
  <File ReadOnly="no" Id="Appsrc_MonosaccharidesCustom.tsv" Name="MonosaccharidesCustom.tsv"
        Source="$(var.GUI_TargetDir)Glycan_Mods\MonosaccharidesCustom.tsv" />
</Component>
```

with `<Property Id="OVERWRITEAPPDATA" Value="1" />` at `Product.wxs:61` — so it is **on by default**, alongside a `RemoveGlycanModsData` component under the same condition (`:644`).

Two consequences:

1. This `!File.Exists` branch never fires on a real install, and if it ever did it would write a header-only file that differs from the shipped one (which carries the long `#` documentation banner).
2. More importantly, a default install or upgrade replaces this file with the shipped copy — discarding every monosaccharide the user added through this window. None of the sibling custom files have that exposure; grepping `Product.wxs` for `CustomModifications`, `CustomAminoAcids` and `CustomMIonLoss` returns nothing, because those are generated at runtime and the installer never touches them.

That doesn't have to be solved in this PR, but it does mean "persists for future sessions" is only true until the next upgrade. Worth either writing user entries to a separate non-shipped file (matching how the other custom registries behave), or calling the limitation out explicitly.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml.cs
```diff
@@ -0,0 +1,171 @@
...
+                }
+                else
+                {
+                    File.AppendAllLines(customMonosaccharidePath, new[] { line });
```
trishorts: **`File.AppendAllLines` does not check for a trailing newline on the existing file.** It appends each line *followed by* a terminator; it never prepends one. If the file's last byte is not a newline — which happens as soon as a user hand-edits the TSV, exactly the audience this window targets — the new row is concatenated onto their last line, producing something like

```
Pent<TAB>T<TAB>132.04226<TAB><TAB>Generic pentoseHexA<TAB>U<TAB>176.03209<TAB>...
```

which the loader then rejects at startup with a `MetaMorpheusException`. That is the failure mode the PR summary says this window exists to prevent, reintroduced through the write path.

The shipped file does end in CRLF, so a default install is fine — it is the hand-edited case that breaks. A guard costs two lines:

```csharp
string existing = File.ReadAllText(customMonosaccharidePath);
if (existing.Length > 0 && !existing.EndsWith("\n"))
{
    File.AppendAllText(customMonosaccharidePath, Environment.NewLine);
}
File.AppendAllLines(customMonosaccharidePath, new[] { line });
```

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml.cs
```diff
@@ -0,0 +1,171 @@
...
+            char code = codeText[0];
+            double massDa;
+            if (!string.IsNullOrEmpty(formulaText))
+            {
+                massDa = ChemicalFormula.ParseFormula(formulaText).MonoisotopicMass;
+            }
+            else 
+            {
+                massDa = double.Parse(massText, NumberStyles.Float, CultureInfo.InvariantCulture);
+            }
```
trishorts: **When both a formula and a mass are entered, the formula silently wins.** The typed mass is discarded with no indication, and the *derived* mass is what gets persisted at line 73 — so the user reopens the TSV later and finds a number they never typed.

`CustomModWindow` handles the same formula-or-mass pair differently: it keeps both and hands both to the `Modification` constructor (`CustomModWindow.xaml.cs:111-121`).

Either behaviour is defensible; the issue is that it is undocumented. Simplest fix is to make the precedence visible — either say so in the Chemical Formula tooltip ("takes precedence over Monoisotopic Mass if both are given"), or warn when the two disagree by more than a rounding step.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml
```diff
@@ -0,0 +1,187 @@
...
+            <StackPanel Orientation="Horizontal" Margin="2">
+                <Label Content="Description" Width="175" HorizontalAlignment="Center" FontSize="10"></Label>
+                <TextBox Name="descriptionTextBox" Width="155" Height="20">
+                    <ToolTipService.ToolTip>
+                        <ToolTip Content="Free text for human readers (optional), e.g. source or formula note"/>
+                    </ToolTipService.ToolTip>
+                </TextBox>
+            </StackPanel>
+            <!--Save and cancel buttons-->
+            <StackPanel Orientation="Horizontal" Height="35" VerticalAlignment="Bottom" HorizontalAlignment="Center" Margin="15">
```
trishorts: **`VerticalAlignment="Bottom"` is a no-op here** — this `StackPanel` is nested *inside* the outer `Orientation="Vertical"` StackPanel, which stacks its children top-down and ignores their vertical alignment. `Margin="15"` is compensating for that.

Every sibling makes the button bar a direct child of the root `Grid`, sibling to the input panel, where the alignment actually binds (verified on master, `CustomModWindow.xaml:252-257`):

```xml
        </StackPanel>          <!-- input panel closes -->
    <!--Save and cancel buttons-->
    <StackPanel Orientation="Horizontal" Height="35" VerticalAlignment="Bottom" HorizontalAlignment="Center">
        <Button Name="saveButton" Content="Save Mod" FontSize="13" Margin="5" Width="100" Click="SaveCustomMod_Click" />
        <Button Name="cancelButton" Content="Cancel" FontSize="13" Margin="5" Width="100" Click="CancelCustomMod_Click" />
    </StackPanel>
</Grid>
```

Moving the panel out one level and dropping `Margin="15"` matches. Related: `Height="270"` is *shorter* than `CustomModWindow`'s `330` despite six input rows plus a 35px button bar with 30px of margin, and `ResizeMode="NoResize"` means anything that overflows is unreachable — worth confirming it renders un-clipped before merge.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml
```diff
@@ -0,0 +1,187 @@
...
+                                    <Setter Property="Background" Value="White" />
+                                </Trigger>
+                            </Style.Triggers>
+                        </Style>
+                    </TextBox.Style>
+                </TextBox>
+            </StackPanel>
+            <!--Description-->
+            <StackPanel Orientation="Horizontal" Margin="2">
+                <Label Content="Description" Width="175" HorizontalAlignment="Center" FontSize="10"></Label>
```
trishorts: Minor: `HorizontalAlignment` aligns the Label *within its parent*, which is a no-op given the fixed `Width="175"`. The sibling small-label pattern uses `HorizontalContentAlignment`, which aligns the text inside the label — `CustomModWindow.xaml:192` and `:222`:

```xml
<Label Content="Neutral Losses (mass)" Width="190" HorizontalContentAlignment="Center" FontSize="10"/>
```

Also, this is the only one of the six TextBoxes without the CueBanner watermark style, which looks unintentional given the tooltip already suggests example text.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml.cs
trishorts: Fixed — all four removed at `0cff66f5`; the remaining eight usings are all live. Thanks.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml
trishorts: Fixed — Formula and Mass are now grouped in the `LightGray` panel under a single `and/or` + `*` row, and the Diagnostic Ions asterisk is gone. Matches `CustomModWindow` exactly.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml.cs
trishorts: Fixed — the `name.Equals("Name", OrdinalIgnoreCase)` guard is in, so the header-row collision is now impossible from the GUI.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml.cs
trishorts: Fixed, and better than I asked for. `GlycanDatabase.PersistCustomMonosaccharide` now owns the write side next to the reader, and `PersistCustomMonosaccharide_RoundTripsThroughLoadCustomMonosaccharides` exercises write -> reset -> `LoadCustomMonosaccharides` -> assert, which is exactly the test that keeps the two halves from drifting. I ran the fixture locally: 24/24 in `TestNGlyco`, 101/101 across all glyco tests, including the pre-existing `KindCapacity == 12` cases — so the `ResetCustomMonosaccharides` calls and the file save/restore in the `finally` are not leaking state into neighbouring tests.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml.cs
trishorts: Fixed — the `EndsWith("\n")` check plus `AppendAllText(Environment.NewLine)` is in `PersistCustomMonosaccharide`, and the comment explaining why is a nice touch.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml.cs
trishorts: Fixed, and you went further than the suggestion — the tooltip now states the precedence *and* a >0.01 Da disagreement raises a Yes/No confirmation naming both masses. That is the better answer; the user can no longer lose a typed mass without seeing it happen.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml
trishorts: Fixed — the button panel is now a direct child of the root `Grid`, `Margin="15"` is gone, and `Height` went 270 -> 320, which also settles the clipping question I raised.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml
trishorts: Fixed — dropping `HorizontalAlignment` outright works as well as converting it, and adding the CueBanner watermark to this TextBox makes the six inputs internally consistent.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml.cs
trishorts: Half of this is fixed — thanks. The typed mass is now bounded (`parsed <= 0 || parsed > 20000`) and rejected with a clear message, which kills both the negative-mass and the `17603209` saturation cases for that input.

Two paths into the same `* 1E5` cast are still unguarded, so I'll leave this thread open:

1. **Diagnostic ions.** `ErrorsDetected` still only proves they parse (`CustomMonosaccharideWindow.xaml.cs:130-141`), and `PersistCustomMonosaccharide` then does `(int)Math.Round(... * 1E5)` on each (`GlycanDatabase.cs:163`). `-175.02` and `50000` both sail through; the latter saturates to `int.MaxValue` exactly like the mass case did.
2. **Formula-derived mass.** The new bound is applied to `typedMassDa` only. `formulaMassDa` never gets one, so a formula like `C10000` yields ~120 000 Da and saturates on the same cast at `GlycanDatabase.cs:159`.

Both are the same two-line shape as the fix you already wrote. Neither is a merge blocker as far as I'm concerned — flagging so it's a deliberate call rather than an oversight.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml.cs
trishorts: Leaving this one open as a tracked design question, not as a request for this PR — the `!File.Exists` header branch moved into `PersistCustomMonosaccharide` unchanged, which is fine.

The point that still stands is the second one: `Product.wxs:649-653` installs the shipped `MonosaccharidesCustom.tsv` under `Condition="OVERWRITEAPPDATA"`, and `Product.wxs:61` sets that property to `1` by default, so a routine upgrade replaces the file and discards everything this window appended. The sibling registries don't have that exposure — `CustomModifications.txt`, `CustomAminoAcids.txt` and `CustomMIonLosses.txt` appear nowhere in `Product.wxs` because they're generated at runtime.

Worth deciding before Part B builds on this persistence path, since the fix (write user entries to a separate non-shipped file) gets more expensive once a glycan-database designer depends on the current layout.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml.cs
```diff
@@ -0,0 +1,146 @@
+﻿using Chemistry;
+using EngineLayer;
+using EngineLayer.GlycoSearch;
+using System;
+using System.Globalization;
+using System.IO;
```
trishorts: Nit, and slightly funny given the thread just above: moving the file I/O into `GlycanDatabase` made this one unused. Nothing in the file now references `Path`, `File`, `Directory` or any other `System.IO` type — the last use went with the `PersistCustomMonosaccharide` extraction. Safe to delete.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
```diff
@@ -145,6 +144,70 @@ public static void LoadCustomMonosaccharides(string filePath)
             }
         }
 
+        public static string PersistCustomMonosaccharide(string name, string codeText, string formulaText, string massText, string ionsText, string descriptionText) 
+        {
+            char code = codeText[0];
```
trishorts: Nice — putting the writer next to the reader is exactly right, and the round-trip test is the part that makes it stick. Three small things now that this is public EngineLayer API rather than GUI code-behind:

1. **`codeText[0]` will throw `IndexOutOfRangeException` on an empty string.** The GUI can't reach it (`ErrorsDetected` guarantees length 1 first), but this method is now callable from anywhere, and the test already calls it directly. A `string.IsNullOrEmpty(codeText)` guard throwing `MetaMorpheusException` with the same message shape as the rest of the file would make it self-defending.

2. **The `> 20000` mass bound lives only in the GUI validator.** Now that this is the shared entry point, that bound arguably belongs here too — otherwise the one caller that isn't the dialog can still register a saturating mass. Same for the diagnostic-ion values.

3. **No XML doc**, while its neighbour `LoadCustomMonosaccharides` has one. Worth a `<summary>` noting that it both *registers* and *persists*, and — importantly — what the `string` return means: `null` on success, a user-facing warning when registration succeeded but the file write did not. That two-channel contract (throw for hard failure, return-string for soft) is reasonable, but it's invisible from the signature.

None of these block; they're the difference between a method that happens to work for its one caller and one that reads as EngineLayer API.

trishorts (commented):
Re-reviewed at `b41b2075a`. **8 of the 10 threads are fixed and resolved** — the four stray usings, the required-field asterisks, the `Name` collision, the `AppendAllLines` newline, the formula/mass precedence, the button-panel nesting and the Description label are all done, several of them better than what I suggested.

The structural one is the standout: `GlycanDatabase.PersistCustomMonosaccharide` now owns the write side of the format next to the reader, and `PersistCustomMonosaccharide_RoundTripsThroughLoadCustomMonosaccharides` actually exercises write -> reset -> load -> assert. That is the test that keeps the two halves from drifting, and it was the main thing I wanted.

**Verified locally**, not just read:

* `dotnet build GUI/GUI.csproj` — 0 errors (the 103 warnings are all pre-existing MetaDraw/MainWindow noise, none in the new files).
* `dotnet test --filter TestNGlyco` — **24/24 pass**, including the new round-trip test.
* `dotnet test --filter Glyco` — **101/101 pass**. Worth calling out because the new test mutates global `Glycan` state and the real `MonosaccharidesCustom.tsv`; the `ResetCustomMonosaccharides` calls and the save/restore in the `finally` hold up, and the pre-existing `KindCapacity == 12` assertions still pass alongside it.

Two threads stay open, with detail in each: the diagnostic-ion and formula-derived-mass halves of the range check, and the installer `OVERWRITEAPPDATA` question (a design note for Part B, not a request here). Plus two small new nits below. Nothing here is a merge blocker.

<sub>Re-reviewed with Claude Code assistance; build and test results above were run locally against `b41b2075a`.</sub>

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml.cs
trishorts: Fully fixed at `e5eac9768` — thanks for taking both halves. Diagnostic ions are now bounded in the GUI validator (`parsedIons.Any(ionDa => ionDa <= 0 || ionDa > 20000)`) and again in `PersistCustomMonosaccharide`, and the formula-derived mass gets the same treatment on both sides. Bounding it in the writer as well as the dialog is the right call — the saturating-cast path is now closed to every caller, not just this one. Resolving.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml.cs
trishorts: Fixed — and you caught that `using EngineLayer.GlycoSearch;` was dead too. I double-checked that one before believing it: `Glycan` is declared in `namespace EngineLayer` (the `GlycoSearch/` folder doesn't map to a namespace segment), so `using EngineLayer;` still covers the `Glycan.NameCharDic` / `Glycan.CharMassDic` calls in `ErrorsDetected`. Builds clean. Resolving.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
trishorts: All three addressed at `e5eac9768` — the `codeText` length guard throwing `MetaMorpheusException` in the file's own message style, the mass and diagnostic-ion bounds moved into the writer so they apply to every caller rather than just the dialog, and an XML doc that spells out the three-way return contract (throw = nothing happened / `null` = registered and persisted / non-null string = registered but not persisted). That last one is the part that makes this read as EngineLayer API. Resolving.

trishorts (dismissed):
## Approved

**11 of 12 threads resolved.** Everything raised across both review rounds is addressed, and several came back better than what I proposed.

**The structural one.** `GlycanDatabase.PersistCustomMonosaccharide` now owns the write side of the TSV format directly beneath the reader that consumes it, with an XML doc spelling out the three-way return contract (throw = nothing happened / `null` = registered and persisted / non-null string = registered but not persisted). `PersistCustomMonosaccharide_RoundTripsThroughLoadCustomMonosaccharides` exercises write -> reset -> `LoadCustomMonosaccharides` -> assert with no GUI in the loop, which is what keeps the two halves from drifting into a startup exception nobody can debug.

**Where you went past the ask.** I wanted the formula/mass precedence *documented*; you made it visible — the tooltip states it and a >0.01 Da disagreement raises a Yes/No naming both masses. And I suggested bounding masses in the dialog; you put the bounds in the writer as well, which closes the saturating-cast path for every caller rather than just this one. That is the more durable fix.

**Verified locally at `e5eac9768`**, not just read:

* `dotnet build GUI/GUI.csproj` — 0 errors, no new warnings from these files.
* `dotnet test --filter Glyco` — **101/101 pass**, including the new round-trip test. Worth stating explicitly: that test mutates global `Glycan` registry state and the real `MonosaccharidesCustom.tsv`, and the `ResetCustomMonosaccharides` calls plus the save/restore in the `finally` hold up — the pre-existing `KindCapacity == 12` assertions still pass alongside it, so there is no cross-test leakage.
* I specifically re-checked that dropping `using EngineLayer.GlycoSearch;` was safe rather than assuming it: `Glycan` is declared in `namespace EngineLayer`, so the `Glycan.NameCharDic` / `Glycan.CharMassDic` calls in `ErrorsDetected` still resolve. The `GlycoSearch/` folder is not a namespace segment.

**One thread left open on purpose** — the installer question, which is not a request against this PR. `Product.wxs:649-653` installs the shipped `MonosaccharidesCustom.tsv` under `Condition="OVERWRITEAPPDATA"`, and `Product.wxs:61` defaults that property to `1`, so an upgrade can discard entries this window persisted. The sibling registries don't have that exposure — `CustomModifications.txt`, `CustomAminoAcids.txt` and `CustomMIonLosses.txt` appear nowhere in `Product.wxs` because they're generated at runtime. I'm leaving it open rather than resolving it so it stays visible: it's worth settling before Part B's database designer depends on this persistence layout, and the fix gets more expensive later.

Nice, well-scoped Part A. Approving.

<sub>Reviewed with Claude Code assistance; build and test results above were run locally against `e5eac9768`, and every finding was confirmed against the source before posting.</sub>

trishorts (dismissed):
## Approved (re-approving at `76a0dd9d3`)

My earlier approval was dismissed by the coverage push, so this re-approves. The only change since then is `76a0dd9d3`, and it is **test-only** — `Test/TestNGlyco.cs`, +95/−0, no production code touched.

**What the new tests buy.** Seven of them close the throw paths on `PersistCustomMonosaccharide` — empty and multi-character code, mass out of range, diagnostic ion out of range, name collision, code collision, and the single-character-but-not-a-letter case that clears the writer's own `length == 1` guard only to be rejected by `RegisterCustomMonosaccharide`. That last one is the interesting one; it is the seam between the two validators, and it now has a test.

The eighth is the one that matters most: `PersistCustomMonosaccharide_AppendsOnOwnLineWhenFileMissingTrailingNewline` stages a TSV whose last byte is not a newline and asserts the new row lands on its own line. That is exactly the concatenation bug from round one, now regression-protected rather than just fixed.

Pinning the collision tests to `HexNAc` / `'N'` with a comment about built-ins surviving `ResetCustomMonosaccharides` is the right call — those assertions hold regardless of test order.

**Verified locally at `76a0dd9d3`**, not read:

* `dotnet test --filter "FullyQualifiedName~CustomMonosaccharide"` — **20/20 pass**.
* `dotnet test --filter "FullyQualifiedName~Glyco"` — **109/109 pass**, up from the 101 I ran at `e5eac9768`.

Worth stating explicitly, since it is the one thing that could have gone wrong here: the new append test mutates the real `MonosaccharidesCustom.tsv` and global `Glycan` registry state, the same as the existing round-trip test. Its save/restore `finally` holds, the pre-existing `KindCapacity` assertions still pass alongside it, and there is no assembly-level `Parallelizable` attribute or `.runsettings` in the Test project, so the two file-mutating tests cannot race each other.

`codecov/patch` is green at **94.28%** (33/35), `codecov/project` green, all 13 checks pass.

**One optional nicety, not a request.** Per the Codecov line map, the two uncovered patch lines are `GlycanDatabase.cs:228,230` — the `catch (Exception)` that returns the "available for this session, but could not be saved" warning. That is the third leg of the return contract the XML doc describes (throw = nothing happened / `null` = registered and persisted / non-null string = registered but not persisted), and it is the only leg without a test. It is reachable without mocking: mark the file read-only, call `PersistCustomMonosaccharide`, assert the return is non-null, restore the attribute in a `finally`. Cheap, and it would make the doc comment fully executable. Codecov is green either way, so take it or leave it.

**Two housekeeping items before this can merge, neither of them yours to fix in code:**

1. The branch is `BEHIND` master and needs an update. A base-only merge does not trip stale-dismissal here, so it will not cost this approval.
2. Master requires three approvals; this is one.

**And the last open thread is now tracked properly.** I filed #2752 for the `OVERWRITEAPPDATA` problem — `Product.wxs:644-653` clears the whole `Glycan_Mods` folder (`RemoveFolder ... On="both"`) and re-lays the shipped `MonosaccharidesCustom.tsv` over it, with `Product.wxs:61` defaulting the property to `1`, so GUI-added monosaccharides survive sessions but not an install, repair, or upgrade. I confirmed while writing it up that `CustomModifications.txt`, the custom amino acid file and `CustomMIonLosses.txt` appear nowhere in `Product.wxs` — this file is the only user-writable registry the installer manages, which is what makes it the odd one out. That is pre-existing behavior this PR surfaces rather than causes, it stays out of scope here, and it is a natural fold-in for Part B. I am resolving that thread now that the issue carries it.

Nice work across all three rounds — the writer moving into `GlycanDatabase` next to its reader, the bounds landing in the writer rather than the dialog, and now the round-trip and throw-path tests, add up to a format that cannot quietly drift into a startup exception nobody can debug.

MetaMorpheus/GUI/Views/CustomMonosaccharideWindow.xaml.cs
trishorts: Tracked as #2752, so closing this thread out.

The issue carries the detail: `Product.wxs:644-653` clears the entire `Glycan_Mods` folder (`RemoveFolder ... On="both"`) and re-lays the shipped `MonosaccharidesCustom.tsv` over it, and `Product.wxs:61` defaults `OVERWRITEAPPDATA` to `1` — so monosaccharides added through this window survive sessions but not an install, repair, or upgrade.

One thing I confirmed while writing it up that sharpens the point: `CustomModifications.txt`, the custom amino acid file and `CustomMIonLosses.txt` appear nowhere in `Product.wxs`. This is the only user-writable custom registry the installer manages, which is why it is the one with the problem. The proposed fix in #2752 needs no installer change — persist to a sibling `MonosaccharidesCustom.user.tsv` and have `GlobalVariables.LoadGlycans()` load both, shipped first then user.

Still out of scope here; this PR surfaces the behavior rather than causing it. Natural fold-in for Part B.

nbollis (dismissed):

Alexander-Sol (approved):

nbollis (approved):

pcruzparri (approved):

PR #2753, opened 2026-08-24
Fix MetaDraw export crashes and rendering artifacts

nbollis:
## Motivation
MetaDraw exports could crash for ambiguous or partially rendered selections, produce shifted sequence and legend placement, and generate black or white-background artifacts in coverage and annotation exports.

## Summary
- Add regression coverage for ambiguous PSMs, disabled stationary sequence rendering, zero-size canvases, and unmeasured annotation surfaces.
- Guard export preparation and preserve the original sequence layout geometry while rendering.
- Use valid render dimensions and robust optional legend handling.
- Preserve transparent export backgrounds and use an in-memory PNG conversion path to avoid BMP text-rendering artifacts.
- Preserve cumulative display errors without self-appending the error list.

## Testing
- MetaDraw export regression tests: 4 passed.
- Spectrum plot tests: 146 passed.
- BioPolymer coverage and export tests: 28 passed.
- GUI project build completed with zero errors.

## Migration Steps
None. Export file formats and user workflows remain unchanged.

## Screenshots
- Before: [attach MetaDraw export crash or artifact screenshot]
- After: [attach transparent sequence annotation export screenshot]
- After: [attach transparent protein coverage export screenshot]


Alexander-Sol (dismissed):

RayMSMS (dismissed):

zhuoxinshi (dismissed):

trishorts (commented):
Reviewed the diff by hand (8 files, ~320 lines; small enough that a targeted read beat the review engine). The direction is right — the `stationarySequence.Height +=` accumulation, the `Annotations.First()`, the zero-size render targets, and the self-appending `errors.AddRange(errors)` were all real bugs, and the in-memory PNG path matches what `BioPolymerTabViewModel.ExportImage` already does. CI is green across the board.

Two things I would like resolved before merge, inline. Minor notes below:

- `GetCanvasDimension` is now copy-pasted verbatim into `MetaDrawLogic`, `PeptideSpectrumMatchPlot`, and `CrosslinkSpectrumMatchPlot`. Worth a single `internal static` helper.
- `errors ??= new List<string>()` is dead — `ExportPlot` already does `errors = new List<string>()` on entry, so it can never be null at those points.
- `else if (errors?.Count == 0) { errors = null; }` makes the `out` contract inconsistent: callers get an empty list on the `skipPlotRegeneration` path and `null` otherwise, for no apparent gain. The GUI handles both, so this looks like it can just go.
- The summary says "Preserve transparent export backgrounds ... in coverage and annotation exports", but both of those exports go through `CombineBitmap`, which does `g.Clear(System.Drawing.Color.White)`. The transparent-background behaviour lives in `BioPolymerTabViewModel.ExportImage`, which this PR does not change — the new test there is welcome regression coverage, but the description overclaims what changed.

MetaMorpheus/GuiFunctions/MetaDraw/MetaDrawLogic.cs
```diff
@@ -845,28 +866,36 @@ public void CleanUpCurrentlyDisplayedPlots()
...
-            renderCanvasBitmap.Render(canvas);
-
-            BmpBitmapEncoder encoder = new BmpBitmapEncoder();
-            encoder.Frames.Add(BitmapFrame.Create(renderCanvasBitmap));
-            using (FileStream file = File.Create(tempBitmapPath))
+            DrawingVisual drawingVisual = new();
+            using (DrawingContext drawingContext = drawingVisual.RenderOpen())
             {
-                encoder.Save(file);
+                drawingContext.DrawRectangle(new VisualBrush(canvas), null, new Rect(0, 0, width, height));
```
trishorts: **Major** — this turns `ConvertCanvasToBitmap` from a 1:1 render into a *stretch to fill*, so any canvas whose content doesn't exactly fill its layout rect gets magnified.

`VisualBrush` defaults to `Stretch=Fill` with `Viewbox=(0,0,1,1)` relative to the visual's **content bounds**, so `DrawRectangle(brush, null, new Rect(0, 0, width, height))` scales the canvas's rendered content up to `width x height`, independently on each axis. `renderCanvasBitmap.Render(canvas)` (what master did) rendered at true scale.

Verified with a standalone WPF probe — a 400x400 transparent `Canvas` whose only child is a 400x40 red bar, rendered both ways at 96 dpi:

| row | `Render(canvas)` (master) | `VisualBrush` (this PR) |
|---|---|---|
| y=39 | red | red |
| y=45 | transparent | **red** |
| y=395 | transparent | **red** |

Canvases with an opaque `Background` are immune, because the background pins the content bounds to the layout rect — `sequenceText` and `map` both set `Background="White"`, so `ExportSequenceCoverage` is unaffected. But `sequenceAnnotationCanvas` (MetaDraw.xaml:482) has no `Background`, and it is exactly the canvas `ExportAnnotatedSequence` passes in here. `DrawnSequence` sizes that canvas to `42 * rows + 10` while the residue `TextBlock`s are 30 px tall, so a one-row annotation export comes out stretched ~1.7x vertically.

The `BmpBitmapEncoder` -> `PngBitmapEncoder` + `MemoryStream` change is the part that fixes the BMP text artifacts; the `DrawingVisual`/`VisualBrush` detour is orthogonal to it. Simplest fix is to drop the detour and keep `renderCanvasBitmap.Render(canvas)`. If you want to keep it, pin the brush:

```csharp
new VisualBrush(canvas) { Stretch = Stretch.None, AlignmentX = AlignmentX.Left, AlignmentY = AlignmentY.Top }
```

I re-ran the probe with that and it matches `Render(canvas)` pixel for pixel.

MetaMorpheus/GuiFunctions/MetaDraw/SpectrumMatch/PeptideSpectrumMatchPlot.cs
```diff
@@ -82,10 +98,9 @@ public PeptideSpectrumMatchPlot(OxyPlot.Wpf.PlotView plotView, SpectrumMatchFrom
             points.Add(new Point(0, 0));
 
             var tempStatSequenceBitmap = new System.Drawing.Bitmap(tempStationarySequencePngPath);
-            System.Drawing.Bitmap stationarySequenceBitmap = new System.Drawing.Bitmap(tempStatSequenceBitmap, new System.Drawing.Size((int)stationarySequence.Width, (int)stationarySequence.Height));
+            System.Drawing.Bitmap stationarySequenceBitmap = new System.Drawing.Bitmap(tempStatSequenceBitmap, new System.Drawing.Size(stationarySequenceWidth, stationarySequenceHeight));
             bitmaps.Add(stationarySequenceBitmap);
-            var stationarySequenceLocationVector = (Vector)stationarySequence.GetType().GetProperty("VisualOffset", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(stationarySequence);
-            Point stationarySequencePoint = new Point(stationarySequenceLocationVector.X, stationarySequenceLocationVector.Y);
+            Point stationarySequencePoint = new Point(stationarySequenceLocationVector.X + 50, stationarySequenceLocationVector.Y);
```
trishorts, line 105: **Major** — where does the `+ 50` come from, and the `- 14, - 20` on the legend point at line 128? Both are new, uncommented, and `CrosslinkSpectrumMatchPlot.ExportPlot` — otherwise the same routine, changed in the same way by this PR — keeps the unshifted `(vector.X, vector.Y)` at its lines 102 and 127.

Both paths render the same stationary-sequence canvas through the same `Measure`/`Arrange`/`VisualOffset` sequence, and both now receive `stationaryCanvas` instead of `StationarySequence.SequenceDrawingCanvas`, so I would expect the same correction on both. As it stands either the crosslink export is left misaligned or the peptide export is over-corrected.

Could you say what these constants compensate for (a comment at the call site would help the next reader), and mirror them onto the crosslink path if they are a real fix?

MetaMorpheus/GuiFunctions/MetaDraw/MetaDrawLogic.cs
```diff
@@ -499,6 +499,7 @@ public void ExportPlot(PlotView plotView, Canvas stationaryCanvas, List<Spectrum
                     // get the scan
                     if (!MsDataFiles.TryGetValue(psm.FileNameWithoutExtension, out MsDataFile spectraFile))
                     {
+                        errors ??= new List<string>();
```
trishorts: **Minor** — this is dead. `ExportPlot` opens with `errors = new List<string>();` (line 480), and nothing between there and here can null it, so `errors` is never null at this point. Same for the identical guards at lines 522 and 595. Plain `errors.Add(...)` / `errors.AddRange(...)` reads better and doesn't imply a nullable state that can't occur.

MetaMorpheus/GuiFunctions/MetaDraw/MetaDrawLogic.cs
```diff
@@ -515,13 +516,18 @@ public void ExportPlot(PlotView plotView, Canvas stationaryCanvas, List<Spectrum
...
-                        DisplaySpectrumMatch(plotView, psm, parentChildScanPlotsView, out errors);
+                        DisplaySpectrumMatch(plotView, psm, parentChildScanPlotsView, out var displayErrors);
+                        if (displayErrors?.Any() == true)
+                        {
+                            errors ??= new List<string>();
+                            errors.AddRange(displayErrors);
+                        }
+                        else if (errors?.Count == 0)
+                        {
+                            errors = null;
```
trishorts: **Minor** — this (and the copy at line 600) makes the `out` contract inconsistent for no gain: a caller gets an empty `List<string>` when `skipPlotRegeneration` is true or `plotView.Name != "plotView"`, and `null` otherwise. Nothing downstream benefits — `MetaDraw.xaml.cs:704` already does `if (errors != null && errors.Any())`, and `MetaDrawTest.cs:1495` does a bare `errors.Any()` that only survives because its `plotView` has no `Name` and so never reaches this branch.

The real bug here was the old `errors.AddRange(errors)` self-append, and swapping in the `displayErrors` local fixes that on its own. I'd drop both `else if` blocks and let `errors` stay a non-null (possibly empty) list throughout.

MetaMorpheus/GuiFunctions/MetaDraw/MetaDrawLogic.cs
```diff
@@ -845,28 +866,36 @@ public void CleanUpCurrentlyDisplayedPlots()
...
+            using System.Drawing.Bitmap sourceBitmap = new(stream);
+            return new System.Drawing.Bitmap(sourceBitmap, new System.Drawing.Size(width, height));
+        }
 
-            System.Drawing.Bitmap unformattedBitmap = new(tempBitmapPath);
-            System.Drawing.Bitmap bitmap = new(unformattedBitmap, new System.Drawing.Size(width, height));
-            unformattedBitmap.Dispose();
-            File.Delete(tempBitmapPath);
-            return bitmap;
+        private static int GetCanvasDimension(double requestedDimension, double actualDimension)
```
trishorts: **Minor** — this helper is now copy-pasted verbatim into three files: here, `PeptideSpectrumMatchPlot.cs:142`, and `CrosslinkSpectrumMatchPlot.cs:141`. All three bodies are byte-identical.

Worth hoisting to one `internal static int GetCanvasDimension(double requested, double actual)` — `MetaDrawLogic` is already the shared home for `CombineBitmap`, which both plot classes call, so it would fit there without a new type.

MetaMorpheus/Test/MetaDraw/BioPolymerTabViewModelTests.cs
```diff
@@ -417,6 +417,41 @@ public void ExportImage_ExportsImage_WhenCoverageDrawingIsPresent()
...
+            try
+            {
+                MetaDrawSettings.ExportType = "Png";
+                var method = typeof(BioPolymerTabViewModel)
+                    .GetMethod("ExportImage", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);
+                MessageBoxHelper.SuppressMessageBoxes = true;
+                method.Invoke(vm, null);
+
+                using var bitmap = new System.Drawing.Bitmap(expectedPath);
+                Assert.That(bitmap.GetPixel(0, 0).A, Is.EqualTo(0));
```
trishorts: **Minor**, on the PR description rather than this test — the summary says "Preserve transparent export backgrounds ... in coverage and annotation exports", but `ExportSequenceCoverage` and `ExportAnnotatedSequence` both finish through `CombineBitmap`, which does `g.Clear(System.Drawing.Color.White)`. Those two exports are still white-backed after this PR.

The transparency this test asserts comes from `BioPolymerTabViewModel.ExportImage`, whose only change in this diff is the trailing newline — so the test is new regression coverage for behaviour that already worked, which is a good thing to add, just not a fix. Worth rewording the summary so the changelog doesn't claim more than shipped.

One small robustness thought while you're here: the drawing is a 2 px pen from (20,20) to (80,80), so its bounds are roughly (19,19)-(81,81) and pixel (0,0) of the export lands right on the corner where the stroke's cap region is closest. Sampling somewhere unambiguously off the diagonal (say the bottom-left quadrant) would give the assertion more margin.

trishorts (commented):
Following up with the minor notes from my earlier review as inline comments, as requested. None of these are blocking — the two Majors above are the ones I'd want resolved before merge.

Alexander-Sol (dismissed):

MetaMorpheus/GuiFunctions/MetaDraw/MetaDrawLogic.cs
nbollis: Fixed in 6e27e9167. Canvas export now uses direct RenderTargetBitmap rendering followed by the in-memory PNG conversion, preserving 1:1 geometry instead of VisualBrush stretch-to-fill behavior. The focused MetaDraw review suite passes 173 tests.

MetaMorpheus/GuiFunctions/MetaDraw/SpectrumMatch/PeptideSpectrumMatchPlot.cs
nbollis: Addressed in 9b344ada8. The peptide-specific alignment offsets are intentional legacy corrections, so they were retained and documented at the call sites. The crosslink path remains unshifted because its export composition uses a different layout; focused export tests pass.

MetaMorpheus/GuiFunctions/MetaDraw/MetaDrawLogic.cs
nbollis: Fixed in 9b344ada8. GetCanvasDimension is now centralized as MetaDrawLogic.GetCanvasDimension, and the duplicate implementations were removed from both spectrum plot classes.

MetaMorpheus/GuiFunctions/MetaDraw/MetaDrawLogic.cs
nbollis: Fixed in 6e27e9167. The dead null-coalescing guards were removed, display errors are accumulated explicitly, and the historical null-versus-empty-list behavior is preserved with an explicit display-attempt flag. TestMetaDrawErrors and TestMetaDrawLoadingWithWeirdFileNames pass.

MetaMorpheus/Test/MetaDraw/BioPolymerTabViewModelTests.cs
nbollis: Fixed in 2a1dedaed. The transparency assertion now samples a pixel clearly away from the diagonal stroke, avoiding corner ambiguity while still verifying zero-alpha output.

MetaMorpheus/GuiFunctions/MetaDraw/MetaDrawLogic.cs
nbollis: Addressed in the PR description. It now accurately states that protein coverage, sequence coverage, and sequence annotation exports preserve transparent backgrounds, without claiming behavior that is not implemented.

RayMSMS (dismissed):

MetaMorpheus/GuiFunctions/MetaDraw/SpectrumMatch/PeptideSpectrumMatchPlot.cs
trishorts: Reopening — I don't think this one is settled yet.

> The peptide-specific alignment offsets are intentional legacy corrections

They aren't legacy. On `master`, `PeptideSpectrumMatchPlot.cs:88` is `new Point(stationarySequenceLocationVector.X, stationarySequenceLocationVector.Y)` and `:111` is `new Point(ptmLegendLocationVector.X, ptmLegendLocationVector.Y)` — both unshifted. The `+ 50` and the `- 14, - 20` are introduced by this PR, so "preserve the legacy correction" is describing a state that never existed.

That matters because the original question is still open: if the peptide export needs a 50 px horizontal shift after this PR's `Measure`/`Arrange`/`VisualOffset` changes, what is the 50 compensating for, and why doesn't `CrosslinkSpectrumMatchPlot` — same routine, same canvas plumbing, changed the same way in this PR — need it? "Its export composition uses a different layout" is plausible but I can't see the difference in the diff: both now receive `stationaryCanvas`, both call `MetaDrawLogic.CombineBitmap` with `overlap: true`, and the only structural divergence I can find is the `+ 30` vs `+ 80` on `stationarySequenceHeight`. If that height difference is what the `50` is absorbing, saying so in the comment would close this out.

Two smaller things while this is open:

- The comments are duplicated. `PeptideSpectrumMatchPlot.cs:103-104` are two near-identical lines, and `:130-131` are the same line twice verbatim. Both pairs would land in master as-is.
- Neither comment states what the constant corrects for, only that a correction exists. A reader hitting `X + 50` still can't tell whether it's safe to touch.

MetaMorpheus/GuiFunctions/MetaDraw/MetaDrawLogic.cs
trishorts: Reopening — this got two of the three, not three.

`MetaDrawLogic.GetCanvasDimension` is `internal static` now and `CrosslinkSpectrumMatchPlot` calls through to it at lines 62, 63, 111, 112. Good. But `PeptideSpectrumMatchPlot.cs:146` still carries its own byte-identical `private static int GetCanvasDimension(double, double)`, and lines 63, 64, 114, 115 still bind to that local copy rather than the shared one.

Deleting the `PeptideSpectrumMatchPlot` copy and qualifying those four call sites with `MetaDrawLogic.` — exactly what was already done in `CrosslinkSpectrumMatchPlot` — finishes it.

MetaMorpheus/GuiFunctions/MetaDraw/MetaDrawLogic.cs
```diff
@@ -699,14 +711,14 @@ public static System.Drawing.Bitmap CombineBitmap(List<System.Drawing.Bitmap> im
...
+                // Create a transparent bitmap to hold the combined image without changing source alpha.
+                finalImage = new System.Drawing.Bitmap(width, height, System.Drawing.Imaging.PixelFormat.Format32bppArgb);
 
                 //get a graphics object from the image so we can draw on it
                 using (System.Drawing.Graphics g = System.Drawing.Graphics.FromImage(finalImage))
                 {
-                    //set background color
-                    g.Clear(System.Drawing.Color.White);
+                    // Keep the output transparent wherever no source image has content.
+                    g.Clear(System.Drawing.Color.Transparent);
```
trishorts: **Major**, new observation on this commit rather than a follow-up to my earlier review — flipping this from `White` to `Transparent` is what makes the transparency claim in the summary actually true, so it's a real fix. But it fires for *every* export format, and only some of them have an alpha channel.

`MetaDrawSettings.ExportTypes` is `{ "Pdf", "Png", "Jpeg", "Tiff", "Wmf", "Bmp" }`. `CombineBitmap` builds a `Format32bppArgb` bitmap and now clears it to `Color.Transparent` (A=0, and `Color.Transparent`'s RGB is 0,0,0). `ExportSequenceCoverage` and `ExportAnnotatedSequence` both write through `MetaDrawSettings.ExportType`, so when the user has Jpeg or Bmp selected, that fully-transparent margin gets flattened by an encoder with nowhere to put the alpha — JPEG discards it and keeps the black RGB, and `BmpBitmapEncoder`/`Bitmap.Save` to BMP composites onto black rather than white.

So the coverage and annotation exports that used to come out white-backed in those two formats should now come out **black**-backed, which is the artifact the Motivation section says this PR fixes. `Pdf` is worth checking too depending on how it composites.

Worth confirming before merge — export a coverage map with `ExportType = "Jpeg"` and look at the margin. If it reproduces, the fix is probably to pick the clear colour by format rather than unconditionally, e.g. clear to `Transparent` only for the alpha-capable formats (Png, Tiff) and keep `White` otherwise.

trishorts: Reproduced locally — it's real, and it's Jpeg and Bmp only.

I built a standalone net10.0-windows probe that copies `CombineBitmap` and the non-Pdf branches of `ExportBitmap` verbatim out of this commit, parameterising only the clear colour so master (`White`) and this PR (`Transparent`) run side by side. Each case writes a real file through `Bitmap.Save`, re-reads it, and samples a pixel in a region no source bitmap draws on.

Bare case — stack a 200x50 row over a 100x50 row so `x` in [100,200) of the bottom row is never drawn on, sample (150,75):

| clear | Png | Jpeg | Tiff | Bmp | Wmf |
|---|---|---|---|---|---|
| `White` (master) | A=255 fff | A=255 fff | A=255 fff | A=255 fff | A=255 fff |
| `Transparent` (this PR) | A=0 | **A=255 R=0 G=0 B=0** | A=0 | **A=255 R=0 G=0 B=0** | A=0 |

Jpeg and Bmp come back fully opaque black. Png and Tiff keep the alpha, which is the intended win. Wmf also keeps it, because GDI+ has no WMF encoder and `Save(path, ImageFormat.Wmf)` quietly falls back to PNG bytes — pre-existing, just noting it so the table isn't misread.

I also ran the two real export geometries and got the same result at the sampled pixel:

- **`ExportAnnotatedSequence` with a PTM legend.** `CombineBitmap(overlap: false)` makes the final height `annotationH + legendH` and the width `max(...)` = the annotation width, and the legend is drawn at `(annotationWidth/2 - legendWidth/2 - 50, canvasHeight)`. So the whole bottom band outside the legend rectangle is uncovered, and the `combinedBitmap.Width - 140` crop only takes it off the right edge — most of the band survives. Sampling (60,150) of a 900x180 composite: Jpeg and Bmp black.
- **`ExportSequenceCoverage` with a text canvas narrower than the map.** Width is `max(textW, mapW)`, so the text row has an uncovered right margin. Sampling (750,40) of a 900x280 composite: same.

Without a legend and with equal-width coverage canvases there's no uncovered region and nothing changes, so this won't reproduce on every export — but the legend path is a normal one.

**`Pdf` is fine**, which matters since it's the default `ExportType`. That branch saves PNG first and hands it to `ImageDataFactory`; I confirmed the generated PDF carries an `/SMask`, so the alpha survives and the transparent area shows the white page.

So the fix is narrower than my first comment implied — the clear colour only needs to depend on format:

```csharp
bool alphaCapable = MetaDrawSettings.ExportType is "Png" or "Tiff" or "Pdf" or "Wmf";
g.Clear(alphaCapable ? System.Drawing.Color.Transparent : System.Drawing.Color.White);
```

Reading `MetaDrawSettings` from inside `CombineBitmap` isn't lovely — a `Color` parameter defaulting to `Transparent` with the two export methods passing the format-appropriate value would keep the static helper honest. Either way it's worth a regression test: the `BioPolymerTabViewModelTests` transparency test already flips `MetaDrawSettings.ExportType` to `"Png"` in a try/finally, so a `[TestCase("Png", 0)] [TestCase("Jpeg", 255)]` variant asserting the alpha of an uncovered pixel would pin this down.

Alexander-Sol (dismissed):

zhuoxinshi (dismissed):

MetaMorpheus/GuiFunctions/MetaDraw/MetaDrawLogic.cs
nbollis: Confirmed. JPEG and BMP now use an opaque white background, while PNG, TIFF, PDF, and WMF retain transparency. Added regression coverage for the format-specific behavior; the focused MetaDraw suite passes 552/552. The changes are in commits b5c31166a and fbd1affff.

MetaMorpheus/GuiFunctions/MetaDraw/SpectrumMatch/PeptideSpectrumMatchPlot.cs
nbollis: The legacy magic numbers may not have been on that exact line, but there were magic numbers designed to position the sequence and legend on the final spectrum image in their proper location. The numbers have changed due to processing changes. I added comments to state this explicitly.

MetaMorpheus/GuiFunctions/MetaDraw/MetaDrawLogic.cs
nbollis: Good catch. Removed the duplicated helper method

trishorts (approved):

RayMSMS (approved):

zhuoxinshi (approved):

PR #2754, opened 2026-08-25
ci: open the mzLib bump pull request automatically

trishorts:
Nothing connects an mzLib release to this repository. The pin moves when someone remembers, which is why this repository is the consumer that matters most: mzLib's own integration job clones this repository's master and runs its suite, so a stale or bad pairing reddens every open mzLib pull request.

This adds the watcher the Python bindings already use. On Mondays it checks mzLib's `releases/latest`, compares it with what this repository pins, and opens a pull request bumping every pin site.

### It opens a pull request; it does not push

The checks on that pull request are the point. **A version bump here is only sometimes a version string** — of the four consumers bumped for mzLib 1.0.585, exactly one was. The others needed a target-framework migration, a namespace fix after drifting nineteen releases, and thirteen test updates because mzLib had added file formats whose count the documentation asserted. Pushing the bump would skip the one check that makes it trustworthy.

So the workflow decides whether a bump is *available*, never whether it is *safe*. The pull request body says so, and points at the release notes.

### Details worth knowing

- **Pin sites are discovered, not listed.** It greps for the mzLib reference across `*.csproj` and `*.nuspec`, so it cannot go stale the first time someone adds a project — a hard-coded list would, silently.
- **Disagreement is reported, not smoothed over.** If pin sites hold different versions, the pull request says so before reconciling them.
- **A pin ahead of the release is left alone.** Taking an unreleased mzLib deliberately is a real thing; rolling it backwards onto an older package would be a regression opened as a chore.
- **It refuses to open a partial bump.** If any reference survives the rewrite, the run fails rather than proposing a half-updated tree.
- **One pull request per release**, updated in place, rather than a duplicate every Monday until it merges.
- Least-privilege `permissions`, and it touches nothing under `.github/workflows/`, which the default token cannot push to anyway.

### Verified before opening

I ran the detect-and-rewrite logic against this repository's actual files: found all 6 pin sites (`CMD`, `EngineLayer`, `GUI`, `GuiFunctions`, `TaskLayer`, `Test`), read the current pin as `1.0.585`, and rewrote every one to `1.0.586` with nothing left behind.

### The token

`UPSTREAM_WATCH_TOKEN` is set here. It is not what makes the checks run — a pull request authored by `github-actions[bot]` *does* queue workflows, but they land as `action_required` behind the first-time-contributor approval gate. The token makes the pull request come from a human account so CI starts without a click. Nothing breaks without it.


acesnik:
🤖 One thing before this runs unattended: the "a pin ahead of the release is left alone" rule only holds when *every* pin site is ahead.

The direction check compares the release against `versions[0]`, and `versions` comes out of `sort -uV`, so that is the **lowest** pinned version:

```bash
current=${versions[0]}
...
newest=$(printf '%s\n%s\n' "$current" "$NEW" | sort -V | tail -1)
if [ "$newest" != "$NEW" ]; then   # refuse
```

With mixed pins it inverts. Sites at `1.0.586` and `1.0.590`, latest release `1.0.586`:

- `current` = 1.0.586, `NEW` = 1.0.586
- the already-pinned exit is skipped, because `${#versions[@]}` is 2
- the direction check passes, because the lowest equals `NEW`
- the rewrite sets every site to 1.0.586, moving the 1.0.590 site backwards

Which is the regression-opened-as-a-chore the rule is there to prevent. Not hypothetical for this org either — the body cites ProteaseGuru holding its csprojs and its nuspec at different versions, so mixed pins are a state that happens here.

Two things keep it small: it opens a pull request rather than pushing, and the body prints `pin sites did not agree: ...`, so a reviewer sees it. But the guard is meant to stop that pull request being opened at all.

Checking direction against the highest pinned version instead looks like it covers it:

```bash
highest=${versions[-1]}
newest=$(printf '%s\n%s\n' "$highest" "$NEW" | sort -V | tail -1)
```

I have not run the workflow, so that is read off the script rather than observed.

trishorts:
You are right, and I ran it to be sure. Fixed in 3d69525.

I pulled the `pin` step's script straight out of the yaml and ran it against synthetic pin trees, so this is the shipped code and not a paraphrase of it. Your case reproduces exactly:

```
pinned: 1.0.586 1.0.590   release: 1.0.586   ->  action=bump      (before)
                                             ->  action=none      (after)
```

There is a second shape it also lets through, which the lowest-vs-highest framing makes obvious once you have it: sites at `1.0.580` and `1.0.590` with a release of `1.0.586`. The lowest is behind, so the guard sees an ordinary bump, and the `1.0.590` site still goes backwards. Same fix covers it.

Took your `${versions[-1]}` as written. The five cases where nothing moves backwards are unchanged — clean-behind, clean-equal, clean-ahead, mixed-both-behind, and this repository's real tree, which still reports its six sites at `1.0.585` and bumps.

Two things beyond the one-liner:

- The refusal notice now names the drift when there is drift, because "pins 1.0.586, which is newer than 1.0.586" would have been a confusing thing to find in a cron log.
- The drift `::warning::` no longer says "the bump reconciles them to one version." On the path you found, there is no bump, so it was promising something that then did not happen.

Worth being explicit about what the fix gives up: with mixed pins where the highest is ahead of the release, the workflow now does nothing at all, so it never reconciles that drift. That is the right trade — it is a bump watcher, not a drift fixer, and a drift where one site is deliberately ahead is not something to reconcile without a human deciding which direction is correct. The `::warning::` still fires, so the run says so.

acesnik (approved):
🤖 Verified — both shapes refuse now, and nothing rolls backwards across the seven pin trees I ran the fixed logic against:

```
pinned 1.0.585          release 1.0.586  -> bump
pinned 1.0.586          release 1.0.586  -> none  (already pinned)
pinned 1.0.590          release 1.0.586  -> none  (refused: highest is ahead)
pinned 1.0.580 1.0.585  release 1.0.586  -> bump
pinned 1.0.586 1.0.590  release 1.0.586  -> none  (refused: highest is ahead)
pinned 1.0.580 1.0.590  release 1.0.586  -> none  (refused: highest is ahead)
pinned 1.0.580 1.0.586  release 1.0.586  -> bump
```

Your `1.0.580` / `1.0.590` shape is the one I'd missed — the lowest being behind made it look like an ordinary bump.

Worth noting the last row still bumps: mixed pins whose highest equals the release still get reconciled upward, so the fix only gives up the case where a site is genuinely ahead. That is the trade you named and I think it is the right one here.

nbollis (approved):

acesnik (commented):
How about doing this from the mzLib release action?

trishorts:
🤖 Fair question, and it is the better-looking design on one axis: `release.yml` is the only thing that *knows* a release succeeded, and it knows at the moment it happens, whereas a Monday cron can lag by up to seven days.

I went and read mzLib's `.github/` before answering rather than arguing from taste. Three things decided it against push for me.

**1. There is a house precedent, and it is already the pull shape.**

`check-thermo-rawfilereader.yml` solves this exact problem — "did an upstream repo publish something new?" — for the Thermo RawFileReader. It is a `schedule:` cron in mzLib that polls `thermofisherlsms/RawFileReader` monthly and files an issue *in mzLib*. Consumer-side poll, upstream never notified. That is structurally the same choice this PR makes, one repo down.

Push has no precedent anywhere in mzLib: no `repository_dispatch`, no `gh workflow run`, no `gh pr create --repo`, no `peter-evans/create-pull-request`. `git log -S` finds zero commits ever adding one, so it is not something that was tried and abandoned either.

**2. Push needs a credential mzLib does not have, and it is the wrong credential to create.**

`release.yml:13-14` is `permissions: contents: write`, with a comment saying it is "least privilege, and enough to create the Release page." The default `GITHUB_TOKEN` cannot write to another repository under any `permissions:` block, so a push design means adding a fine-grained PAT or App token with **Contents: write + Pull requests: write on every consumer repo** to the repository that publishes the package.

That is a long-lived, near-org-wide write credential, tied to a person, whose rotation breaks fan-out to all four consumers at once — and it fails *on the release tag*, after the package has already shipped to nuget.org. This PR's design needs no shared secret at all: each consumer opens its own PR with its own default token. `UPSTREAM_WATCH_TOKEN` here is a convenience for the first-time-contributor approval gate, not a requirement; the workflow works without it.

**3. Push has to know four repositories' internals; pull only has to know one.**

There is no machine-readable consumer registry in mzLib — the list lives as prose in `release-notes/1.0.587.md`'s `## Downstream` section. A push design has to invent one and keep it current, then track each consumer's pin layout. mzLib already demonstrates the cost: `dotnet.yml:128-142` rewrites MetaMorpheus's six pins, and that recipe is MetaMorpheus-shaped — FlashLFQ, ProteaseGuru and pyMzLib each pin differently (ProteaseGuru's csprojs and nuspec are the ones that drift apart). Four consumers, four layouts, all maintained in mzLib and all silently wrong the day one of them moves. A consumer polling `releases/latest` only ever has to know its own tree, which is why the pin discovery here is a grep rather than a list.

**And one on blast radius.** `release.yml` puts the Release page last on purpose, because an earlier failure means deleting the tag and re-tagging. Appending fan-out to that job would put four cross-repo PR creations inside the blast radius of a release that may still be rolled back, and a fan-out failure would redden a release whose package already shipped.

**What your point does buy.** The lag is real, and it is not free — mzLib's `integration` job clones this repository's master, so a bad pairing reddens every open mzLib PR for up to a week. If that becomes the binding problem, the cheap fix is not to invert the direction but to add `workflow_dispatch:` to this workflow, so a release can be picked up on demand by anyone with a button, no new secret. Happy to add that here if you want it.

Worth noting the thing that made pull viable at all is recent: `releases/latest` only moves reliably because mzLib #1206 automated the Release page. Before that the tag existed and the page often did not, and polling would have missed releases.

RayMSMS (approved):

PR #2755, opened 2026-08-25
Call mzLib's QuantificationEngine for isobaric searches

trishorts:
Completes the MetaMorpheus side of mzLib #996. A TMT search now produces peptide- and protein-level quantification instead of stopping at the PSM.

This is the last piece of the TMT stack, and with #2774, #2775 and #2776 merged it is 3 files plus an mzLib pin bump — everything else this branch used to show has landed on master.

## The problem

Multiplex quantification stopped at the PSM level. Reporter ion intensities were extracted during the search (`IsobaricMassTag.GetReporterIonIntensities`) and written as extra columns in the .psmtsv (`WritePsmPlusMultiplexIons`) — and that was all. No roll-up to peptides, no roll-up to proteins, no channel-to-sample mapping. `QuantificationAnalysis()` picked the multiplex modification and returned.

mzLib has had the engine to do this for a while, and the three input contracts it needs are now all satisfied on master:

| Engine input | MetaMorpheus type | Where it landed |
|---|---|---|
| `List<ISpectralMatch>` | `SpectralMatch` | master, via #2475 |
| `List<IBioPolymerGroup>` | `ProteinGroup` (derives mzLib's `BioPolymerGroup`) | master |
| `IExperimentalDesign` | `TmtExperimentalDesign.ToMzLibDesign` | master, via #2774 |
| `List<IBioPolymerWithSetMods>` | `PeptideWithSetModifications` | already there |

So nothing stands between MetaMorpheus and the engine except the call itself. That call is this PR.

## What this does

**`PostSearchAnalysisTask.MultiplexQuantificationAnalysis`** reads `TmtDesign.txt` from beside the spectra files — the same idiom the label-free path uses for `ExperimentalDesign.tsv`, `Directory.GetParent(CurrentRawFileList.First())` — projects it through `TmtExperimentalDesign.ToMzLibDesign`, and runs `QuantificationEngine` over the PSMs that carry reporter ions.

Output: `RawQuantification.tsv`, `PeptideQuantification.tsv` and `ProteinGroupQuantification.tsv` in the task folder, one column per channel, plus a copy of the design file so a result folder records what it was quantified under. `AllProteinGroups.tsv` gains `Intensity_<file>_<channel>` beside the count columns it already had.

`PostSearchAnalysisParameters` gains `MultiplexQuantificationResults` to carry the result; the per-channel protein values also land on the real `ProteinGroup` objects the writers already hold.

## Things worth reviewing

**1. Nothing here is fatal, and that is the whole design.** Eight guards — no spectra files, no design file, an unreadable design, a design that names no file in this run, an unprojectable design, no PSM carrying reporter ions, no protein groups, no protein group that the protein table would show — each warn and leave the search exactly as it was, reporter ion columns included. A TMT search without a channel-to-sample mapping is a legitimate thing to run.

**2. The channel ordering is a two-sided contract that fails silently.** `IsobaricMassTag.ReporterIonMzs` is built with `.OrderBy(mz => mz)`, so `IsobaricMassTagReporterIonIntensities` is in ascending reporter m/z, and `ToMzLibDesign` emits its `IsobaricQuantSampleInfo` array in that same order. mzLib maps the two **positionally**. Change either side alone and every channel is mislabelled — the totals stay plausible, they just belong to other samples. Both sides carry a comment saying so, and the protein table's header is asserted **column by column** against the plex's channel order. @acesnik found the earlier version of that assertion wanting and he was right: counting the columns and checking they were distinct passed just as happily on a design rotated by one channel, because the roll-up check compares index *j* on one side to index *j* on the other and a rotation agrees with itself.

**3. The quantified set is the written set.** The PSMs handed to the engine are filtered to unambiguous targets below threshold; the protein groups now go through the same predicate `WriteProteinGroupsToTsv` uses, so decoys and above-threshold groups no longer appear as rows in `ProteinGroupQuantification.tsv` that `AllProteinGroups.tsv` suppresses. Groups held back are still given this run's channel columns with no values, so the protein table stays rectangular however the groups sort.

**4. Excluded matches are reported, not swallowed.** mzLib quantifies only a match identifying exactly one biopolymer, and one sequence found in two proteins is two unequal objects — `PeptideWithSetModifications.Equals` compares `Parent?.Accession` — so a peptide shared between two protein groups is dropped. `QuantificationResults.AmbiguousSpectralMatchesExcluded` counts them; this now warns whenever it is non-zero on a **successful** run, which is where the loss actually shows up. @acesnik measured the cost on a database with every entry duplicated: the raw table intact at 1639587.26 and both roll-up tables at zero. The exclusion itself is mzLib's to fix; making it audible is this PR's part.

**5. Strategies are all no-ops except the sum roll-ups.** No normalization. Reference-channel normalization exists in mzLib and `TmtDesign.txt` can already describe it through the `Sample Type` column, but it should be a user-facing setting rather than something that starts applying itself to everyone's data. Wiring it to a toml/GUI option is the obvious follow-up. `UseSharedPeptidesForProteinQuant` now reads `SearchParameters.UseSharedPeptidesForLFQ` — the same switch the label-free path reads and the same GUI checkbox, which is labelled for quantification rather than for LFQ — instead of being hardcoded off. `OutputDirectory` is passed explicitly as the task output folder rather than relying on mzLib's beside-the-data fallback, which is right for a caller with no output folder of its own and wrong for a task that has one.

## Tests

`Test/MultiplexQuantificationTests.cs`, twelve tests.

Four end-to-end searches over the existing TMT11 fixture (`VA084TQ_6.mzML` + `mouseTmt.fasta`):

- All three tables are written, with one column per channel — including channels the design left unannotated, which must be present or every later channel shifts by one — and **each column header names the channel it belongs to**, which is what a rotated design would fail.
- **Each peptide's per-channel value equals the sum of its PSMs' values in the matching `Reporter_n` column of the raw snapshot.** The roll-up arithmetic, checked against the pre-roll-up file.
- **`AllProteinGroups.tsv` keeps its `SpectralCount_` and `CountOccupancy_` columns and gains `Intensity_` ones**, and every row carries the columns the header advertises. This is the regression @acesnik caught: the engine's write-back clears each group's cached `SampleGroupResults`, and this repo's copy of the protein header reads that cache without rebuilding it, so the file silently went 24 columns to 22 with zero diffs on the 22 that remained.
- **The quantified groups are exactly the groups `AllProteinGroups.tsv` shows.**
- The .psmtsv reporter ion columns survive — this is additive, not a replacement.
- A search with **no** design file still completes and writes no quantification files.
- A search with **parsimony off** warns and writes no quantification files. This is the guard a user is likeliest to hit, because the stock `TMT-Task1-SearchTaskconfig.toml` leaves parsimony off and protein roll-up then has nothing to roll up onto.

Eight more drive `MultiplexQuantificationAnalysis` directly, the way `PostSearchAnalysisTaskTests` drives the other private members of this task, and assert the specific warning rather than merely reaching the line. Three of them — an unreadable design, an unknown multiplex label, a design naming no file in this run — cannot be provoked through the GUI or the toml at all, because both front ends validate the design before the search starts.

Two worth naming:

- **`DesignNamingNoFileInThisRun_WarnsAndSkips`.** Every row parses and every value is legal; they all just name some other file. Before #2774 added the no-match error, that read as success with an empty annotation set — the difference between a table of wrong numbers and a table of nothing. @acesnik singled it out in review as "the one that goes wrong quietly".
- **`PeptideSharedBetweenTwoProteins_WarnsThatMatchesWereExcluded`.** Two proteins with the same sequence, one PSM carrying real TMT11 reporter ions, one protein group. Neither TMT fixture database has a shared peptide — every `Shared Peptides` cell is empty — so this is the only place the exclusion path runs.

## mzLib pin

Bumped 588 → **1.0.589**, the release that carries the quant batch this PR consumes. Worth being precise: nothing here *needed* it. The branch built and the full suite passed on 588, because `QuantificationEngine` shipped in 1.0.587. The bump aligns the consumer with the release rather than fixing anything. #2789 and #2790 make the identical edit, so whichever lands first makes the others' a no-op.

## Verification

Against mzLib 1.0.589, with `upstream/master` merged in: the solution builds with **0 errors** and the full MetaMorpheus suite passes — ****1974 passed, 1 skipped, 0 failed** (1973 before the twelfth test)**.

https://claude.ai/code/session_01UXg2FFyBYx4zMxHjL5wJE4


trishorts:
**Reviewer note: merge #2598 first and this PR gets a lot smaller.**

Right now the diff here is 20 commits, 18 files, +3356/−71 — but only one commit is actually this PR:

| | commits | files | diff |
|---|---|---|---|
| What #2755 shows today | 20 | 18 | +3356 / −71 |
| **What #2755 is** — `2ed1d6a` "Call mzLib's QuantificationEngine for isobaric searches" | 1 | 5 | **+606 / −3** |

The other ~17 commits are #2598's — `TmtDesign.txt`, the `Sample Type` column, `ToMzLibDesign`, the GUI design window. They appear here only because #2598 is still a draft, so this branch had to stack on it rather than on master. GitHub recomputes the diff automatically once #2598 lands; nothing needs re-pushing here.

The five files that are genuinely this PR:

```
MetaMorpheus/EngineLayer/SpectralMatch.cs                        +75   ISpectralMatch
MetaMorpheus/EngineLayer/ProteinParsimony/ProteinGroup.cs       +128   IBioPolymerGroup
MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs     +152   the engine call
MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisParameters.cs +9   carries the results
MetaMorpheus/Test/MultiplexQuantificationTests.cs               +245   3 end-to-end tests
```

So the review order that costs the least: **#2598, then an mzLib release, then this.** Those two are independent of each other — #2598 needs no new mzLib, and the release needs no #2598 — so they can happen in either order or at once. This PR needs both.

For completeness on the second one: mzLib #1193 and #1194 are merged, but `1.0.586` was cut at 18:13, before #1194 merged at 19:51 and #1193 at 20:20. So nothing released carries them, this branch's pin is still `1.0.585`, and CI here cannot compile until there is a release that has them. That is expected and is why this is a draft — it is not a signal that anything is wrong with the code. It is built and tested locally against a package packed from mzLib `master` at 46f850b.

acesnik (commented):
🤖 This is the PR that makes the rest of the TMT work mean something — the design file has been readable and validated for a while, and until now nothing consumed it. Reviewed the ~410 lines that are actually new here (`MultiplexQuantificationAnalysis`, the parameters addition and `MultiplexQuantificationTests`), since the other ~2,000 are the design stack that #2774 carries.

**The blast radius is as small as it could be, and I checked rather than assuming.** On master, `QuantificationAnalysis` already had a `DoMultiplexQuantification` branch and that branch did *nothing*: it picked `MultiplexModification` from the fixed mods and returned. This PR inserts `MultiplexQuantificationAnalysis()` before that return. So for anyone who has not enabled multiplex quantification the branch is unreachable and no output moves, and for anyone who has, output goes from nothing at all to isobaric quantification files. Purely additive, which is the easiest kind of change to be confident about.

The guarding is careful in a way I want to name, because it is the difference between a no-op and a crash on someone's data. Seven early returns cover an empty file list, a missing design, design parse errors, a null projection, no reporter-ion-bearing PSMs, no protein groups, and an unsuccessful engine run. `GetIsobaricMassTag` returns null rather than throwing on an unrecognised id, and that null is checked. And `PopulateCountBasedOccupancy()` deliberately runs *before* every return with a comment explaining that spectral counts need no quantification — so the count columns are filled even on the paths that bail, which is exactly the sort of thing that gets forgotten and then reported as missing columns.

CI is green across the board — Windows tests, all three OS builds, the installer, search-with-installed — and since no csproj here bumps the pin, that also settles that `QuantificationEngine` is genuinely in the released 1.0.587 rather than only on mzLib master.

## The one thing I would change, and it is ordering rather than code

**This branch carries the oldest of five copies of `TmtExperimentalDesign.cs`.** Measured across the family:

```
#2774                    715 lines   (current — has both silent-data-loss fixes)
#2598, #2775, #2776      638 lines
#2755 (this)             579 lines
```

So it is ~136 lines behind, and specifically it lacks `ResolveAgainstDesignFile` and the no-match error — the two fixes #2598's review asked for, where a design written beside its raw files matched only when the process happened to be launched from that folder, and a design matching no file at all read as success with an empty annotation set.

That matters more here than on any of the others, and it is worth being blunt about why: **this is the only one of the five that consumes the design at runtime.** #2774 parses it, #2775 validates it for an exit code, #2776 shows it in a grid. This one hands the projection to `QuantificationEngine` and writes quantification files from it. A design that silently resolves to nothing produces an empty grid in #2776 and a wrong number here. Rebasing on a merged #2774 rather than merging this copy is the whole fix, and #2774 is at 2 of 3 now.

## Smaller

`codecov/patch` is the only red check. Given the shape of the method — seven guards, most of which need a specific malformed input to reach — I would expect that to be the guard paths rather than the engine call, and `MultiplexQuantificationTests` at 245 lines is a reasonable start. Worth a look at whether the untested branches are the ones a user would actually hit, particularly the "design matches no file" path, which is the one that goes wrong quietly.

Commenting rather than requesting changes deliberately: it is a draft, CI is green, and the one substantive item is a rebase ordering rather than a defect in the code you wrote. I would approve this on a rebase against #2774 with the design copy current.

## Not verified

`Test/Test.csproj` is `net*.0-windows`, so the 245 lines of `MultiplexQuantificationTests` are CI's rather than mine — and CI is green on them. I have not run an isobaric search end to end, so I am not making any claim about the quantification values themselves, only about when the path is entered and what happens when it declines to run.

trishorts:
Cherry-picked `795773d71` from #2776 as `93382174f` — the round-3 review fixes on the GUI layer this branch carries.

This branch held #2776's six GUI blobs byte-identically, so without the replay its merge would have resurrected the copies those fixes replace: a Save that wrote no file, a folder drop past the extension allowlist and the Thermo licence gate, plex annotations discarded on a drop-down change, and paste misaligning reporter channels. All six blobs are identical to `tmt/gui` again at this head, so whichever merges first makes the other's copy a no-op — the same way #2775's merge absorbed this branch's CMD copies.

Still draft, still 0/3, and the run wiring (`PostSearchAnalysisTask.cs`, `PostSearchAnalysisParameters.cs`) is unchanged by this. `GUI.csproj` builds with 0 errors.

trishorts:
**Round 1 answered, and the stack this was waiting on has landed.**

@acesnik's one substantive item was ordering, not code: this branch carried the oldest of five copies of `TmtExperimentalDesign.cs` (579 lines), missing `ResolveAgainstDesignFile` and the no-match error — and it mattered here more than anywhere else because this is the only one of the five that consumes the design at runtime. That is resolved the way he asked: with #2774, #2775 and #2776 merged, the branch now takes master's 715-line copy. `git diff upstream/master` shows zero lines of `TmtExperimentalDesign.cs`, and the branch is current with master.

The diff is down to **3 files, +406 / −0** — `PostSearchAnalysisTask.cs`, `PostSearchAnalysisParameters.cs`, and the tests — plus the pin.

**The `codecov/patch` note is addressed with eight tests**, and the one he flagged specifically is among them. `DesignNamingNoFileInThisRun_WarnsAndSkips` covers "the design matches no file" — the path he called the one that goes wrong quietly. Seven of the eight drive `MultiplexQuantificationAnalysis` directly (the pattern `PostSearchAnalysisTaskTests` already uses on this task) and assert the specific warning rather than just reaching the line; three of those paths cannot be provoked through the GUI or the toml at all, because both front ends validate the design before the search. The eighth stays end to end because it needs real reporter-ion PSMs: parsimony off, which is the stock TMT toml's default and so the guard a user is likeliest to hit.

**Pin bumped 588 → 1.0.589.** Being precise, since his review used the absence of a bump as evidence: nothing here needs it. It built and passed the full suite on 588, exactly because `QuantificationEngine` shipped in 1.0.587. The bump aligns the consumer with the release that carries the quant batch. #2789 and #2790 make the identical edit, so whichever lands first makes the others' a no-op.

Full suite against 1.0.589 with master merged: **1973 passed, 1 skipped, 0 failed** (1965 before the new tests). The PR description has been rewritten — it was still describing a draft stacked on #2598 pinned to 1.0.585.

This is 0/3, so nothing was dismissed by the push.

nbollis (dismissed):

acesnik (commented):
🤖 The rebase is what I asked for: zero lines of `TmtExperimentalDesign.cs` in `git diff upstream/master`, branch current with master, `codecov/patch` green. Your correction on the pin is right too — `QuantificationEngine` is in the 1.0.587 tree, so nothing here needed 589, and using the absent bump as evidence was the loose part of my last review. I did check what 589 carries, since this PR is a vehicle for it: the one change to the engine you call is the collapse stage moving into `RunPeptideQuant`, inert under `NoCollapse`, and the rewritten `UniprotAccessionRegex` parses identically on 17,078 real headers (vignette mouse-reviewed, cRAP, `mouseTmt.fasta`).

This time I built CMD and searched both TMT fixtures end to end with a design file, and the handoff itself works: for every peptide and every protein group the per-channel values are bit-exact sums of their PSM values — equal, not within a tolerance — and `RawQuantification.tsv` agrees with the `.psmtsv` reporter columns channel for channel. Two things came out of the runs, and the first is why I'm commenting rather than approving.

Neither fixture database has a peptide mapping to two proteins — every `Shared Peptides` cell is empty — so the fixtures can't reach that case and the tests inherit the blind spot. I duplicated every entry of `MUS_snip.fasta` under a second accession, same spectra, same toml, same design:

```
                          control      every peptide shared   partial sharing
RawQuantification.tsv     1639587.26   1639587.26             1639587.26
PeptideQuantification     1639587.26         0.00             1299147.11
ProteinGroupQuantific.    1639587.26         0.00             1102049.42
```

Both roll-up tables are entirely zero while the raw table is intact. Parsimony doesn't rescue it — it merges each pair into one indistinguishable group, `A0A2I3BPY7|A0A2I3BPY7DUP`, which is the ordinary isoform case, and that group quantifies to nothing. `PeptideQuantification.tsv` also emits 14 rows for 7 sequences with byte-identical `Peptide` labels, and `RawQuantification.tsv`'s `Accession` column is blank for every one of those PSMs.

`SingleIdentifiedBioPolymerOrNull` keeps a match only when `GetIdentifiedBioPolymersWithSetMods().Distinct()` yields one element, and `PeptideWithSetModifications.Equals` compares `Parent?.Accession` and `OneBasedStartResidue` as well as `FullSequence`, so one sequence from two proteins is two unequal objects. Its own remark intends the opposite, and the mzLib test covering it passes the same object twice, so `Distinct()` collapses by reference and the two-accession case is never exercised. `includeAmbiguous: false` doesn't screen these out either — it only tests whether `BaseSequence` is empty.

The partial-sharing arm separates a second loss from the first. A chimeric protein sharing one peptide with each of two others loses 340,440.15 (20.8%) between raw and peptide, and a further 197,097.69 between peptide and protein — exactly the peptide row of the group left with no unique peptides. That one is `UseSharedPeptidesForProteinQuant = false` routing through `GetUniquePeptideToProteinMap`, and it's hardcoded here while the label-free path reads `UseSharedPeptidesForLFQ`; your block comment justifies the normalization choices but is silent about this one.

The first fix is mzLib's. What belongs here is small: the engine already counts these into `AmbiguousSpectralMatchesExcluded` and says so in `Summary`, which this PR reads only in the failure branch — so a total loss produces no warning at all.

Separately, `AllProteinGroups.tsv` goes from 24 columns to 22, losing `SpectralCount_<file>` and `CountOccupancy_<file>` with zero diffs on the 22 that remain, reproduced on both fixtures with identical identifications and q-values. It's the engine call and not the pin: with no design file the output is byte-identical to the merge-base. `OverwriteSampleIntensities` nulls `SampleGroupResults` through the setters; mzLib's base header repopulates it lazily and MetaMorpheus's copy doesn't, so no sample column is emitted — which is also why the per-channel `Intensity_` columns your parameters comment promises never appear. #2789 makes that header `override` but keeps the body, so it won't restore them. The label-free path already handles this at line 793, "Populate SampleGroupResults AFTER all quant-state mutation"; one `PopulateSampleGroupResults()` loop at the end of the new method should do it.

Two smaller things. A design rotated by one channel produces md5-identical quant tables, because column labels are `<file>_<channel>` and sample names never reach the output — so `TmtSearchWithDesignFile_WritesPerChannelPeptideAndProteinTables` isn't what would catch an order desync, contrary to its docstring: the comparison is index *j* to index *j* on both sides and the header assertions are length and distinctness. Asserting column `1+j` ends in `Tmt11Channels[j]` closes it. And `ProteinGroups` reaches the engine unfiltered, so decoys and above-threshold groups become rows in `ProteinGroupQuantification.tsv` that `AllProteinGroups.tsv` suppresses — the PSMs are filtered but the groups aren't.

`Test.csproj` is Windows-only, so the eleven tests are CI's rather than mine, and CI is green. My runs are 4- and 8-PSM snips against 5-to-10-entry databases: they establish that these mechanisms fire and by how much on those constructs, not what fraction of a real TMT experiment is affected. I didn't measure the shared-peptide loss on an isoform-rich proteome, which is where it would be worst.

trishorts:
🤖 Five findings, five changes, in `bdd13d200`. Building CMD and searching both fixtures with a design file is what made all of them findable — none of the eleven tests could see any of them, and I'd have shipped every one.

## The shared-peptide loss

Confirmed, and your reading of the mechanism is exactly right. `PeptideWithSetModifications.Equals` compares `FullSequence`, the digestion agent, `OneBasedStartResidue` **and `Parent?.Accession`** — deliberately, the comment says "important for parsimony in MetaMorpheus" — so `Distinct()` in `SingleIdentifiedBioPolymerOrNull` cannot collapse one sequence from two proteins. Its remark directly above describes the opposite:

> A match that names the same biopolymer more than once -- once per protein it maps to, for instance -- identifies one peptide in substance, and dropping it as ambiguous would discard a perfectly good measurement.

And `AmbiguousMatchFilterTests.cs:180` is the case that would have caught it:

```csharp
Match(2, pepA, pepA),   // one peptide, named twice -- kept
```

The same object twice, so `Distinct()` collapses it by reference and the assertion holds for a reason unrelated to the one the remark gives. Two `Peptide("PEPTIDEK", "P1")` / `("PEPTIDEK", "P2")` objects — the real shape — have never been tried.

That fix is mzLib's, as you said. Filed as smith-chem-wisc/mzLib#1280, with your three-column table. What belongs here is the part you named: `AmbiguousSpectralMatchesExcluded` was read only in the failure branch, so a total loss produced a clean search and silence. It now warns whenever the count is non-zero on a **successful** run, naming the symptom (the roll-ups total less than the .psmtsv reporter columns) and the usual cause.

The twelfth test is `PeptideSharedBetweenTwoProteins_WarnsThatMatchesWereExcluded` — two proteins with one sequence, one PSM carrying real TMT11 reporter ions, one protein group, asserting both the count and the warning text. It's a unit test rather than a duplicated-FASTA search, so it costs no CI time, but it's the only place in the file the exclusion path runs at all.

## The missing columns

Reproduced and fixed. Your diagnosis holds all the way down: `OverwriteSampleIntensities` assigns `IntensitiesBySample` then `SamplesForQuantification`, both setters null `SampleGroupResults`, and this repo's `ProteinGroup.GetTabSeparatedHeader` guards on `SampleGroupResults != null` where mzLib's base does `if (SampleGroupResults is null) PopulateSampleGroupResults()`. So the whole quantification block emitted nothing — which is why `Intensity_` never appeared either. One `PopulateSampleGroupResults()` loop at the end, exactly the label-free path's line 793 and for the same reason.

I checked that the new assertion actually catches it rather than merely passing: with the loop commented out, `AllProteinGroups.tsv` comes back with your 22 columns and the test fails naming `SpectralCount_VA084TQ_6_126`. And you're right that #2789 wouldn't have restored them — it makes the header `override` but keeps the body's `!= null` guard, so this fix is needed either way and survives that merge.

## Decoys and above-threshold groups

Fixed by sharing the predicate rather than restating it: `WriteProteinGroupsToTsv`'s decoy/contaminant/q-value test is lifted into `ProteinGroupIsWritten`, and the quantification path calls the same thing, so the two files cannot drift apart the way they just did.

One consequence worth flagging, because it isn't obvious: the header of `AllProteinGroups.tsv` comes from `proteinGroups.First()`, which need not be a group that survives the row filter. Filtering the engine's input alone would therefore have made the file ragged whenever a suppressed group sorted first. So the held-back groups are given this run's channel columns with an empty intensity dictionary — same schema, blank intensity cells, which is also the honest reading: they exist in this run's channels and were not quantified.

## The rotated design

You're right, and the docstring that claimed otherwise is corrected rather than left standing — the roll-up check compares index *j* to index *j*, so a rotation agrees with itself. Each column header is now asserted to equal `VA084TQ_6_<channel>` for the channel it should be, which is the assertion you described.

## The shared-peptide switch

`UseSharedPeptidesForProteinQuant` now reads `SearchParameters.UseSharedPeptidesForLFQ` — the same switch line 740 reads, and the same GUI checkbox, which is named `CheckBoxUseSharedPeptidesForQuant`, i.e. for quantification rather than for LFQ specifically. Off by default so nothing changes out of the box, but a user who ticks it gets it on both paths, and the isobaric path no longer decides this on its own. The block comment covers it now.

## Not verified

Same limits as before: `Test.csproj` is Windows-only, so the twelve tests are CI's rather than mine, and the full suite is 1974 passed / 0 failed / 1 skipped locally. I have not re-run your duplicated-FASTA construct end to end — the new unit test reaches the same predicate, but it does not measure the loss on a real isoform-rich proteome, which is your open question and mzLib's to answer.

https://claude.ai/code/session_01UXg2FFyBYx4zMxHjL5wJE4

trishorts:
🤖 @nbollis — your approval was dismissed by a push, sorry. Re-requesting.

You approved `87e0600d9`. The head is now `a46d9e62a`; two commits, both answering @acesnik's second review, which he did by building CMD and searching both TMT fixtures end to end. **Nothing you already read has changed** — the design discovery, the guards, the engine call and the strategy choices are all as they were. The delta is `git diff 87e0600d9 a46d9e62a`, two files, and it is five additions:

1. **`PopulateSampleGroupResults()` loop after the engine run.** `AllProteinGroups.tsv` had gone from 24 columns to 22 with zero diffs on the 22 that remained: the engine's write-back nulls each group's cached `SampleGroupResults` through both setters, and this repo's `ProteinGroup.GetTabSeparatedHeader` guards on `!= null` where mzLib's base repopulates lazily. Same fix, same reason, as the label-free path's line 793.
2. **Protein groups filtered through `WriteProteinGroupsToTsv`'s own predicate**, lifted out as `ProteinGroupIsWritten` and shared, so `ProteinGroupQuantification.tsv` and `AllProteinGroups.tsv` cannot disagree about which groups exist. Held-back groups still get the run's channel columns with an empty intensity dictionary, because the protein table's header comes from `proteinGroups.First()` and that need not be a row that survives.
3. **`UseSharedPeptidesForProteinQuant` reads `SearchParameters.UseSharedPeptidesForLFQ`** instead of being hardcoded `false`. Same switch the label-free path reads at :740, same GUI checkbox. `false` by default, so no behaviour change out of the box.
4. **A warning when `AmbiguousSpectralMatchesExcluded` is non-zero on a *successful* run.** It was read only in the failure branch, so a peptide shared between two proteins — dropped by mzLib, tracked as smith-chem-wisc/mzLib#1280 — could zero both roll-up tables while the search reported clean.
5. **Test additions**: channel columns asserted by identity rather than by count and distinctness, the protein table's columns asserted, the quantified set asserted equal to the written set, and a twelfth test covering the shared-peptide path that neither fixture database can reach.

Full suite 1974 passed / 0 failed / 1 skipped. The narrated version is in [my reply to @acesnik above](https://github.com/smith-chem-wisc/MetaMorpheus/pull/2755#issuecomment-5552453405).

https://claude.ai/code/session_01UXg2FFyBYx4zMxHjL5wJE4

nbollis (dismissed):

nbollis (approved):

Alexander-Sol (approved):

Issue #2752, opened 2026-08-24
Custom monosaccharides are lost on install/upgrade: MonosaccharidesCustom.tsv is installer-managed under OVERWRITEAPPDATA


trishorts:
### Summary

`Glycan_Mods/MonosaccharidesCustom.tsv` is the only user-writable custom registry in MetaMorpheus that the Windows installer also manages. Anything a user adds to it is destroyed by the next install, repair, or upgrade.

This is a pre-existing condition of the file, not a regression. It becomes user-visible with #2751, which adds a GUI ("Create new monosaccharide" on the Settings tab) that writes to this file — so users will now be putting data there through a supported path rather than by hand-editing.

### Why it happens

`MetaMorpheusSetup/Product.wxs` sets the overwrite property on by default:

```xml
<!-- Product.wxs:61 -->
<Property Id="OVERWRITEAPPDATA" Value="1" />
```

and the `Glycan_Mods` component group removes the whole folder and then re-lays the shipped TSV over it:

```xml
<!-- Product.wxs:644-653 -->
<Component Id="RemoveGlycanModsData" Guid="40090707-..." Condition="OVERWRITEAPPDATA">
  <RemoveFolder Id="MetaAppGlycan_ModsFolder" On="both" />
  ...
</Component>
<Component Id="Appsrc_Data_MonosaccharidesCustom.tsv" Guid="82338312-..." Condition="OVERWRITEAPPDATA">
  <File ReadOnly="no" Id="Appsrc_MonosaccharidesCustom.tsv" Name="MonosaccharidesCustom.tsv"
        Source="$(var.GUI_TargetDir)Glycan_Mods\MonosaccharidesCustom.tsv" />
</Component>
```

`RemoveFolder ... On="both"` means this fires on install *and* uninstall, so it is not only an overwrite — the folder is cleared first.

### Why the other custom registries are fine

`CustomModifications.txt`, the custom amino acid file, and `CustomMIonLosses.txt` do not appear anywhere in `Product.wxs`. They are create-on-demand and never installer-managed, so user additions survive an upgrade. `MonosaccharidesCustom.tsv` is the odd one out because it ships as a template with an instruction banner.

### Effect

Custom monosaccharides persist across application sessions but are silently lost on install/repair/upgrade. A glycan database that references a lost monosaccharide then fails to load at the next launch.

### Proposed fix

Split the shipped template from user data:

[...]

PR #2756, opened 2026-08-25
Fix: custom monosaccharides lost on install/repair/upgrade (#2752)

RayMSMS:
## Problem

  Custom monosaccharides added via the GUI (`PersistCustomMonosaccharide`, from #2751) were silently wiped on every install, repair, or version upgrade of MetaMorpheus.

  **Root cause:** `Glycan_Mods\MonosaccharidesCustom.tsv` was declared in `Product.wxs` as an installer-managed file *and* was the exact path the GUI wrote user data to. Every
  install/repair re-laid the installer's packaged copy on top of the user's edits, and `MajorUpgrade Schedule="afterInstallValidate"` fully uninstalls the old version — deleting the file —
  before the new version's files are laid down.

  ## Fix

  Brought this file in line with the pattern already used for every other user-editable registry file in the codebase (`proteases_custom.tsv`, `rnase_custom.tsv`): never declare it in the
  installer, and create it at runtime from a bundled default if missing.

  - **`Product.wxs`** — removed the `Appsrc_Data_MonosaccharidesCustom.tsv` component entirely. The file is no longer installer-tracked in any way.
  - **`GlobalVariables.cs`** — added `CustomMonosaccharidePath` (`DataDir\MonosaccharidesCustom.tsv`, matching `CustomProteasePath`/`CustomRnasePath`). `LoadGlycans()` now seeds the file
  before loading it.
  - **`GlycanDatabase.cs`**
    - `EnsureCustomMonosaccharideFileExists(path)`: if the file is missing, seeds it from an embedded resource (the full instructional template — column spec, built-in name/code table,
  worked examples) rather than a hand-typed string, so the shipped `.tsv` is always the single source of truth.
    - **One-time legacy migration**: if an old pre-fix file exists at `Glycan_Mods\MonosaccharidesCustom.tsv` next to the new target path (relevant for portable/source runs or a
  user-specified `DataDir`), it's copied to the new location instead of being silently ignored, and the old copy is left untouched.
    - If seeding the file fails for any reason (missing/renamed embedded resource, unwritable target directory, etc.), a `MetaMorpheusException` is thrown rather than being silently
  caught. An unwritable-target failure is a rare, environment-specific condition — not worth a dedicated warning/collection mechanism — so it surfaces loudly instead of disabling custom
  monosaccharides for the session with no visible indication why.
    - `PersistCustomMonosaccharide` reuses `GlobalVariables.CustomMonosaccharidePath` and the shared `MonoSaccharidesHeader` constant instead of independently-built literals.
  - **`EngineLayer.csproj`** — `Glycan_Mods\MonosaccharidesCustom.tsv` changed from a copied `<None>` item to an `<EmbeddedResource>`, matching the existing `RnaMods.txt` pattern.
  ## Testing

  Added to `TestNGlyco.cs`:

  - [x] `EmbeddedMonosaccharideTemplate_EveryLineIsCommentBlankOrHeader` — guards the embedded template against a stray non-comment line breaking startup parsing
  - [x] `EnsureCustomMonosaccharideFileExists_CreatesHeaderedFileWhenMissing`
  - [x] `EnsureCustomMonosaccharideFileExists_ThenLoadCustomMonosaccharides_DoesNotThrow`
  - [x] `EnsureCustomMonosaccharideFileExists_DoesNotTouchExistingFile`
  - [x] `EnsureCustomMonosaccharideFileExists_UnwritableTarget_ThrowsMetaMorpheusException`
  - [x] `EnsureCustomMonosaccharideFileExists_WritesTheEmbeddedTemplateVerbatim`
  - [x] `EnsureCustomMonosaccharideFileExists_CopiesLegacyFileWhenPresent` — covers the one-time legacy migration

  All new/changed tests isolate against a temp directory; none depend on real global state. The legacy-file lookup is derived from the target `path`'s own directory (equivalent to
  `DataDir` in production) rather than reading the global `DataDir` directly, so the migration test stays fully isolated.

  ## Review notes

  Addresses all feedback from review:
  - Startup failures now throw `MetaMorpheusException` instead of `Console.WriteLine`, which went nowhere in the windowed GUI app
  - Removed an unused `using CsvHelper;`
  - Fixed filename/doc accuracy and a stale directory-creation comment
  - Added the one-time legacy-file migration
  - Renamed the template-guard test off a nonexistent member reference


trishorts (commented):
Reviewed the full diff against `master`, with every sibling quoted from the checkout rather than from memory. This is a real fix for a real problem, and deleting the `Product.wxs` component is the right move — that component is the whole cause of #2752.

**Verified locally, not just read** (worktree at `5e206b362`):

* `dotnet build EngineLayer/EngineLayer.csproj` — **0 errors**. The 26 warnings are all pre-existing (CS8632/CS8625 nullable noise in DIA/PEP/DatabaseLoading, plus the long-standing `obo` CS8981); none come from the changed files.
* `dotnet test --filter "FullyQualifiedName~TestNGlyco"` — **36/36 pass**.
* **The embedded resource name is correct.** This is the part I most expected to be wrong, so I checked it directly rather than trusting the test: I loaded the built `EngineLayer.dll` and enumerated `GetManifestResourceNames()`, which returns exactly `EngineLayer.Glycan_Mods.MonosaccharidesCustom.tsv` — matching the literal at `GlycanDatabase.cs:257`. `EngineLayer.csproj` sets no `<RootNamespace>` so it defaults to the project name, and your `<None Remove>` + `<EmbeddedResource Include>` pair mirrors the existing `Mods\RnaMods.txt` block at `:13-19` exactly.
* **No stale path references survive.** I grepped the whole tree for the old `Glycan_Mods\MonosaccharidesCustom.tsv` layout: `GlobalVariables.cs`, `GlycanDatabase.cs`, `EngineLayer.csproj`, `Product.wxs` and both `TestNGlyco.cs` probe paths are all converted to `GlobalVariables.CustomMonosaccharidePath`. Nothing points at the old location any more. The remaining hits are comments naming the file without a path (`GlycanDatabase.cs:149,339,369`, `Glycan.cs:189`) and are still accurate.
* **The installer CI check is not affected.** `.github/workflows/InstallRunAndArtifact.yml`'s `validate_installer_contents` job is scoped to `*.dll` only (`Get-ChildItem -Filter *.dll`), and no workflow file mentions this `.tsv` at all — so dropping it from build output will not trip the whitelist. Worth stating explicitly, because removing a hand-listed installer file is normally exactly the thing that reddens CI.
* **Nothing here is duplicated.** There is no generic "write embedded default if missing" helper anywhere in MetaMorpheus (`EnsureFile`, `WriteDefault`, `CreateIfMissing`, `EnsureExists` — all zero hits), and mzLib has none either: `ProteaseDictionary`/`RnaseDictionary` only load their embedded resources into memory and expose path-based merge APIs, with no file-writing code in the library at all. EngineLayer is the correct home. One detail in your favour that is worth calling out — `LoadDigestionAgents` has to reach into *mzLib's* assembly using a private resource name it hard-codes and comments as such (`GlobalVariables.cs:607`); your method reads its own assembly's resource, so it does not inherit that brittleness.

**One note on the description.** It says the new layout matches `proteases_custom.tsv`/`rnase_custom.tsv` "exactly". That is true of the *path* — I confirmed both siblings are `Path.Combine(DataDir, "...")` with no subfolder (`:52-53`) — but not of the *ownership model*. Proteases are two files merged at load: the shipped list stays an embedded resource never written into user space, the user file is seeded with only its header row (`:617-628`), and `LoadAndMergeCustomProteases` combines them. This PR deliberately collapses both roles into one file so the instruction banner lives where the user actually edits, which is a defensible reason to diverge — it just is not the same convention, and the description reads as though it were. Detail in the first comment; a design note, not a defect.

**Seven comments below. None of them is a merge blocker.** The cheapest two are the stray `using CsvHelper;` and the failure contract on the new `catch`, both one-liners. The first comment is the one worth an actual decision.

<sub>Reviewed with Claude Code assistance; the build, test, resource-name and grep results above were run locally against `5e206b362` before posting.</sub>

MetaMorpheus/EngineLayer/GlobalVariables.cs
```diff
@@ -473,8 +474,8 @@ private static void LoadGlycans()
             // Custom monosaccharides must be registered FIRST so any custom tokens are recognized
             // by the glycan-database parsers below. The file is optional; if it does not exist,
             // LoadCustomMonosaccharides is a no-op.
-            string customMonosaccharidePath = Path.Combine(DataDir, @"Glycan_Mods", "MonosaccharidesCustom.tsv");
-            GlycanDatabase.LoadCustomMonosaccharides(customMonosaccharidePath);
+            GlycanDatabase.EnsureCustomMonosaccharideFileExists(CustomMonosaccharidePath);
```
trishorts, line 478: **One consequence worth recording explicitly: the instruction banner is now frozen at whatever version the user first launched.**

`EnsureCustomMonosaccharideFileExists` writes the embedded template only when the file is missing, which is exactly right for protecting user rows. It also means the 85-line documented banner is written once and never refreshed, so any later improvement to it reaches new installs only.

That is not hypothetical: #2665 edits `EngineLayer/Glycan_Mods/MonosaccharidesCustom.tsv` to document the diagnostic-ion column — it is one of the 11 files in that PR. Under this design no existing user ever sees those docs, and nothing signals that their copy is stale.

I am flagging this rather than asking you to change it, because the obvious alternative is worse than it first looks and I convinced myself of that while checking. #2752 proposed a two-file split, and that is genuinely what the protease registry does — `LoadDigestionAgents` (`GlobalVariables.cs:587-681`) keeps `proteases.tsv` as an embedded resource that is never written into user space, seeds `proteases_custom.tsv` with **only its header row** (the writer walks the embedded file and `break`s at the `Name\t` line, `:617-628`), and merges the two via `LoadAndMergeCustomProteases`.

Two reasons that does not transfer here:

1. **The banner is the feature.** A user who opens this file to hand-add a row finds the column spec, the reserved name/code table and worked examples right there. Splitting the files would move the documentation somewhere the person editing is not looking.
2. **A split invites exactly the thing this file is careful to avoid — a user-editable copy of the built-ins.** The template gets this right today: all 11 built-ins appear only as `#` comments, as a reserved-names table rather than as rows, which is why `EnsureCustomMonosaccharideFileExists_ThenLoadCustomMonosaccharides_DoesNotThrow` can assert `KindCapacity == 11` on the freshly-seeded file. The defaults stay owned by `Glycan`; the file only ever carries customs. Worth keeping that property whatever else changes.

So the single-file design looks right to me. What I would ask for is only bookkeeping:

1. **Say in the description why the single file was chosen over #2752's split** — that keeping the instructions inside the user's own file is the point — so the issue's proposal reads as considered and rejected rather than bypassed. The stated justification right now is that this matches `proteases_custom.tsv`, and on the ownership axis it deliberately does not.
2. **Coordinate with #2665**, since whoever merges second should know its template edit now only reaches new installs.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
```diff
@@ -231,6 +234,37 @@ public static string PersistCustomMonosaccharide(string name, string codeText, s
...
+                    // same pattern as RnaMods.txt (GlobalVariables.LoadRnaModifications) and the
+                    // mzLib-embedded default protease/rnase templates (GlobalVariables.LoadDigestionAgents).
+                    var assembly = typeof(GlycanDatabase).Assembly;
+                    using var stream = assembly.GetManifestResourceStream("EngineLayer.Glycan_Mods.MonosaccharidesCustom.tsv");
+                    using var reader = new StreamReader(stream);
+                    File.WriteAllText(path, reader.ReadToEnd());
+                }
+                catch (Exception ex)
+                {
+                    Console.WriteLine($"Could not create custom monosaccharide file '{path}': {ex.Message}");
```
trishorts: **This swallows every failure and writes it to a console nobody is reading.** It is the only `Console.WriteLine` in all of `EngineLayer` — I grepped the whole project, one hit, this line. The GUI is a windowed app with no attached console on any platform, so on exactly the path this method exists to serve, the message goes nowhere.

**Please do not "fix" it by throwing**, even though `LoadDigestionAgents` does. I went looking for a handler and there isn't one: `GUI/App.xaml.cs` is an empty partial class, and there is no `DispatcherUnhandledException` or `AppDomain.UnhandledException` registration anywhere in `GUI` or `CMD`. `GlobalVariables.SetUpGlobalVariables()` is called at `MainWindow.xaml.cs:45`, inside the constructor and *before* `InitializeComponent()` at `:46` — so anything that escapes `LoadGlycans()` takes the whole application down at launch with a raw unhandled exception. For a file `LoadCustomMonosaccharides` itself documents as optional (`GlycanDatabase.cs:70-73`, "file is optional"), that is much too blunt. (The same hazard already exists on the protease path; not this PR's problem, but it is why I am not asking you to copy it.)

**The event-based channels are also unavailable here**, which is worth knowing before reaching for one. Every `WarnHandler` — `MetaMorpheusTask` `:80`, `MetaMorpheusEngine` `:83`, `EverythingRunnerEngine` `:62`, `MyFileManager` `:85` — is wired to `NotificationHandler` in that same constructor *after* the `SetUpGlobalVariables()` call, so none is attached yet. And `EngineLayer.csproj` declares no `<ProjectReference>` at all, so `GlobalVariables` cannot reach `TaskLayer` or the GUI regardless of timing.

What *is* live at that moment is the plain static list, which is precisely why it exists in that form. `GlobalVariables.ErrorsReadingMods` (`:47`) is initialized inside `LoadModifications()` at `:407`, and `SetUpGlobalVariables` calls `LoadModifications()` at `:92` before `LoadGlycans()` at `:94` — so the list is guaranteed to exist by the time you need it. The GUI drains it unconditionally at `MyWindow_Loaded` (`MainWindow.xaml.cs:88` → `PrintErrorsReadingMods()` at `:1628`), which pushes each entry through `NotificationHandler` into the notifications pane and auto-expands it (`:117-130`). No WPF coupling on the producing side, so it survives the move to a different GUI toolkit.

Two ways to use that, either of which beats the current line:

- **Minimal, one line here:** `GlobalVariables.ErrorsReadingMods.Add($"Could not create custom monosaccharide file '{path}': {ex.Message}");`. Works today in the GUI with no other edits. The honest caveat is that all five existing uses are modification-related (`:107,124,149` plus the two drains), so the name would be carrying something it was not built for — and `CMD/Program.cs` only drains it inside the `if (Path.GetExtension(db).Equals(".xml"))` branch at `:193-206`, so a CLI run with no XML database would never print it.
- **Cleaner, four small edits:** add a sibling `public static List<string> StartupWarnings` on `GlobalVariables`, initialize it at the top of `SetUpGlobalVariables()`, drain it next to `PrintErrorsReadingMods()` in the GUI, and drain it unconditionally right after `SetUpGlobalVariables()` at `CMD/Program.cs:90` rather than inside the XML-only loop. Correctly named, and it covers the CLI.

I would take either; the second is what I would write, but I would not hold the PR over the choice. The thing worth changing is the `Console.WriteLine`.

One more reason this matters more than it looks: `GetManifestResourceStream` returns `null` rather than throwing when a resource name is wrong or an `<EmbeddedResource>` entry is dropped. Line 257 hands that straight to `new StreamReader(stream)`, which throws `ArgumentNullException`, which this `catch` eats — so a build misconfiguration would present as "the feature quietly does not exist". `MonoSaccharidesInstruction_EveryLineIsCommentOrBlank` guards the name at test time, which is good and I am glad it is there, but the runtime path should not be silent about it.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
```diff
@@ -1,4 +1,5 @@
-﻿using EngineLayer.GlycoSearch;
+﻿using CsvHelper;
```
trishorts: **Unused `using`.** Nothing in this file references any `CsvHelper` type — `Csv` appears exactly once in the whole file, on this line. All the parsing here is hand-rolled `StreamReader` / `string.Split` on tabs, and the only `CsvHelper` usage anywhere in the repo is `GuiFunctions/MetaDraw/BioPolymerCoverage/BioPolymerCoverageMapViewModel.cs`.

Same class of thing as the four stray usings on #2751 — reads like an IDE autocomplete accident. Safe to delete.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
```diff
@@ -231,6 +234,37 @@ public static string PersistCustomMonosaccharide(string name, string codeText, s
             }
         }
 
+        /// <summary>
+        /// Ensure the CustomMonoSaccharide.tsv existed in the directory. If the file is missing,
```
trishorts: Small doc-comment accuracy point, now that this is public `EngineLayer` API:

1. **Wrong filename.** `CustomMonoSaccharide.tsv` is not a file that exists — it is `MonosaccharidesCustom.tsv`. Anyone grepping the name out of the doc finds nothing.
2. **"header-only template" undersells what this actually writes**, and undersells it in the direction that matters. The embedded resource is the full 85-line documented banner — instructions, column spec, the built-in name/code table, worked examples — with the header row as its single non-comment line. That banner is the whole reason for embedding a file rather than writing a header string from code, and it is what `MonoSaccharidesInstruction_EveryLineIsCommentOrBlank` is guarding. A future reader who trusts this comment will think a one-line `File.WriteAllText(path, MonoSaccharidesHeader)` would be an equivalent simplification, and it would not be.

Also `<param name="path"></param>` is empty; either say what it is — the destination path, normally `GlobalVariables.CustomMonosaccharidePath` — or drop the tag.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
```diff
@@ -231,6 +234,37 @@ public static string PersistCustomMonosaccharide(string name, string codeText, s
...
+        /// write the header-only template; do nothing if it already exists.
+        /// </summary>
+        /// <param name="path"></param>
+        public static void EnsureCustomMonosaccharideFileExists(string path) 
+        {
+            if (!File.Exists(path)) 
+            { 
+                try
+                {
+                    // Make sure the directory exists before writing the file, however, it should be created while the installing
```
trishorts: Nit: this comment is now stale in a way that inverts the PR's own point. "it should be created while the installing process" was true of `Glycan_Mods\`, but `path` is `DataDir\MonosaccharidesCustom.tsv` now, and the entire premise of this change is that the installer no longer manages this file.

`DataDir` itself is guaranteed by `SetUpDataDirectory()` (`GlobalVariables.cs:362-381`), which runs at `SetUpGlobalVariables():86`, well before `LoadGlycans()`. So the `CreateDirectory` call is genuine belt-and-braces rather than installer-dependent — `LoadDigestionAgents` writes to the same `DataDir` root and does not call it at all. Either drop the call or reword to something like "DataDir is created by SetUpDataDirectory before this runs; this is defensive only."

MetaMorpheus/Test/TestNGlyco.cs
```diff
@@ -784,5 +784,101 @@ public static void PersistCustomMonosaccharide_RoundTripsThroughLoadCustomMonosa
                 }
             }
         }
+
+        [Test]
+        public static void MonoSaccharidesInstruction_EveryLineIsCommentOrBlank()
+        {
+            // GlycanDatabase.MonoSaccharidesInstruction is hand-written directly into the source file
```
trishorts: Nit: `GlycanDatabase.MonoSaccharidesInstruction` does not exist — there is no such member, and the content this test reads is not "hand-written directly into the source file". It is the embedded `Glycan_Mods/MonosaccharidesCustom.tsv` resource, which is exactly what makes the test worth having: a `.tsv` under `Glycan_Mods/` looks like inert data nobody needs to think about, right up until a stray non-`#` line breaks startup.

The test name inherits the same phantom. Worth renaming to something like `EmbeddedMonosaccharideTemplate_EveryLineIsCommentBlankOrHeader` and pointing the comment at the resource, so a future reader can find the thing being guarded.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
```diff
@@ -231,6 +234,37 @@ public static string PersistCustomMonosaccharide(string name, string codeText, s
             }
         }
 
+        /// <summary>
+        /// Ensure the CustomMonoSaccharide.tsv existed in the directory. If the file is missing, 
+        /// write the header-only template; do nothing if it already exists.
+        /// </summary>
+        /// <param name="path"></param>
+        public static void EnsureCustomMonosaccharideFileExists(string path)
```
trishorts, line 244: **The old file is left on disk and silently ignored — worth a short one-time migration.**

Your description is right that installer upgrades cannot be intercepted from here: the old product's uninstall clears `Glycan_Mods\` before any new-version code runs, so those rows are gone regardless. But that is not the only way people get here, and the other paths are recoverable:

- **Source / portable runs**, where `SetUpDataDirectory` leaves `DataDir = AppDomain.CurrentDomain.BaseDirectory` (`GlobalVariables.cs:372-375`) — no installer involved, so `DataDir\Glycan_Mods\MonosaccharidesCustom.tsv` is still sitting there with the user's rows in it.
- **`UserSpecifiedDataDir`** (`:377-380`), same.

In both cases the file survives, this PR stops reading it, and the user gets no warning — their monosaccharides just stop existing. That is arguably a worse failure than the installer case, because nothing about it is visible.

Since this method already runs at exactly the right moment — new file missing, before the load — the migration is nearly free:

    if (!File.Exists(path))
    {
        string legacyPath = Path.Combine(GlobalVariables.DataDir, "Glycan_Mods", "MonosaccharidesCustom.tsv");
        if (File.Exists(legacyPath))
        {
            File.Copy(legacyPath, path);   // one-time carry-over; leave the old file alone
            return;
        }
        // ... existing embedded-template seed
    }

Fair warning that MetaMorpheus has no precedent for this — I searched, and there is no user-data migration anywhere in the repo; the only `File.Move`/`Directory.Move` calls in the tree are test backup/restore helpers in `Test/GuiTests/LoadDigestionAgentTest.cs:48,61`. So this is a judgment call, not a convention you would be breaking.

But the entire point of #2752 is "don't silently lose user data", and shipping a fix that silently loses it on the non-installer paths is an odd note to end on. If you would rather not, that is reasonable — but then please carry the migration note from your description into the release notes, since the PR body is not something a user will ever read.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
trishorts: Follow-up, because I found a better argument for the second option than the one I gave above — this shape is already the house idiom, not something new.

The duplicate-protein-accession warning does exactly this. `DatabaseLoadingEngine.SanitizeBioPolymerDatabase(..., out List<string> errors)` collects messages deep in the call (`DatabaseLoadingEngine.cs:210`, "The protein '...' has multiple entries ... was renamed."), hands them back as a plain `List<string>`, and the engine drains them near the surface with `Warn(error)` (`:53`, `:60`), which raises `MetaMorpheusEngine.WarnHandler` into the same notifications pane. Producer collects into a list because it cannot reach the UI; something closer to the front end pushes the list out.

The important caveat is the one that makes it *not* directly reusable here: that path only runs after the user hits Start. `MetaMorpheusEngine.WarnHandler` is wired at `MainWindow.xaml.cs:83`, well after the `SetUpGlobalVariables()` call at `:45`, and `Warn` is an instance member of an engine that does not exist yet at startup. So the channel is unavailable, but the *pattern* is established — which is why a `GlobalVariables.StartupWarnings` list drained at `MyWindow_Loaded` is the startup-time analogue of what database loading already does after Start, rather than a new mechanism.

That is the version I would write. Still not a merge blocker — replacing the `Console.WriteLine` is the part that matters.

trishorts:
Heads up — I pushed one test-only commit to this branch (`90c0461b5`), since patch coverage was sitting at 81.5% and the bar is ~92%. No production code touched. Shout if you'd rather I hadn't and I'll drop it.

The entire uncovered patch was three lines: the `catch` in `EnsureCustomMonosaccharideFileExists` (`GlycanDatabase.cs:261,263,264`). `GlobalVariables.cs` had none. Two tests:

- **`EnsureCustomMonosaccharideFileExists_UnwritableTarget_SwallowsAndDoesNotThrow`** — stands a directory where the file should go. `File.Exists` is false for a directory, so the seed is attempted and `File.WriteAllText` throws `UnauthorizedAccessException` on Windows and Unix alike. That reaches the `catch` with no permissions manipulation and nothing to restore afterwards.
- **`EnsureCustomMonosaccharideFileExists_WritesTheEmbeddedTemplateVerbatim`** — asserts the seeded file matches the embedded resource byte for byte, and that the header is its only data line. That pins two properties worth keeping: the instruction banner is the reason a file is embedded rather than a header string written from code, and the eleven built-ins stay `#` documentation rather than editable rows a user edit could shadow.

Verified with `dotnet-coverage` rather than assumed: lines 261, 263 and 264 go from 0 hits to 1, which is the whole uncovered set — so patch should come back at 100%. `TestNGlyco` is 38/38 locally.

One caveat, tied to the failure-contract thread above: the first test asserts the seed **does not throw**. If you take the suggestion there and move from `Console.WriteLine` to collecting the message for the notifications pane, it keeps passing unchanged. If you instead decide to throw, that test is the one to update — deliberately, since it is encoding the contract.

MetaMorpheus/EngineLayer/GlobalVariables.cs
RayMSMS: Thanks for the thorough analysis — the ownership distinction and the #2665 flag are both well taken, and I'll add a design note to the description covering why the single file was chosen over the split.

One correction on the "frozen banner" point, though. That concern assumes the existing user's file is already present, so EnsureCustomMonosaccharideFileExists sees it and never re-seeds. But this PR moves the file to the DataDir root, while existing users' files live in the old Glycan_Mods location.

On upgrade:

The old installer's uninstall step clears Glycan_Mods (the pre-existing RemoveFolder On="both"), so the old file is gone.
The DataDir root has no such file, so EnsureCustomMonosaccharideFileExists re-seeds it from the embedded template — which ships in the current DLL.

So an existing user who upgrades gets the current banner, not a frozen one. The banner isn't pinned to first-launch; it's re-seeded whenever the root path is empty, which is exactly the case on the first upgrade to this version.

By the same token, #2665's template edits do reach upgrading users, not just new installs — they arrive via the updated DLL and the root re-seed.

There is one coordination requirement this creates: #2665 needs to be merged before the next published release. The re-seed only happens once per user (on their first upgrade to this version, when the root path is empty); after that the file exists and is never overwritten again. So whatever the embedded template contains at the moment a user first upgrades is what they get — permanently. If #2665 lands in a release after this one, existing users will already have re-seeded from the pre-#2665 template and won't pick up its diagnostic-ion docs. Merging both into the same release closes that window: every upgrading user re-seeds once, from a template that already includes #2665's edits.

(The one real cost here is unrelated to the banner: existing users lose their custom rows once on this upgrade, because step 1 clears them before any new-version code runs. That's unavoidable from the new version's side and is documented as a migration note in the description.)

trishorts (dismissed):

trishorts:
Pushed `fc2296298` to this branch — patch coverage was at 80.8% against a ~92% bar. Shout if you'd rather I hadn't and I'll drop it.

The uncovered set was exactly five lines, in two places:

**`Program.cs` (2 missing, 2 partial).** The warning loop sits inline in `Run`, which no test can reach without running a whole task, so it's extracted verbatim into `Program.FlushStartupWarnings(verbosity)` and `Run` now calls that. Behaviour is unchanged; the one edit to the logic is hoisting the verbosity check out of the loop, where it was invariant — every iteration was re-evaluating a condition that can't change. Say the word if you'd rather I left it inside.

Three tests in `TestCmd.cs`, redirecting `Console.Out`: `minimal` and `normal` both write every queued warning; `none` writes nothing **but still clears**, so a later caller can't report the same warning twice; and the empty-list case, which is what actually happens on a healthy startup.

**`GlycanDatabase.cs:275` (1 partial).** The `?.` on `StartupWarnings`. The existing unwritable-target test only ever exercised the non-null side. Two tests in `TestNGlyco.cs`:

- `..._RecordsStartupWarning` — asserts the swallowed failure actually lands on `StartupWarnings` and that the message names the path. That's the property worth pinning: the seed failing quietly is only acceptable *because* the message resurfaces in the notifications pane / CLI. Without this the swallow and the report are tested separately and nothing checks they're connected.
- `..._NoWarningSink_DoesNotThrow` — the null side. `EnsureCustomMonosaccharideFileExists` is public and its sink is a static field that `SetUpGlobalVariables` assigns, so a caller that hasn't been through setup must not get an NRE out of the failure path itself — that would defeat the point of swallowing.

The three tests that touch the process-wide `StartupWarnings` are `[NonParallelizable]` and restore the original list in a `finally`, matching the existing pattern in `PsmTsvWriterTests`.

Verified with `dotnet-coverage` rather than assumed: every changed line is hit, the verbosity check reports 4/4 branches, the loop 2/2, the `?.` 2/2 — so patch should come back at 100%. 151 tests across the glyco, CMD and GlobalVariables fixtures pass locally.

trishorts (dismissed):

zhuoxinshi (dismissed):

MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
```diff
@@ -231,6 +233,50 @@ public static string PersistCustomMonosaccharide(string name, string codeText, s
...
+                    // The default template—instructions is embedded in the DLL so it survives
+                    // install/repair/upgrade regardless of what the installer does to Glycan_Mods --
+                    // same pattern as RnaMods.txt (GlobalVariables.LoadRnaModifications) and the
+                    // mzLib-embedded default protease/rnase templates (GlobalVariables.LoadDigestionAgents).
+                    var assembly = typeof(GlycanDatabase).Assembly;
+                    using var stream = assembly.GetManifestResourceStream("EngineLayer.Glycan_Mods.MonosaccharidesCustom.tsv");
+                    using var reader = new StreamReader(stream);
+                    File.WriteAllText(path, reader.ReadToEnd());
+                }
+                catch (Exception ex)
```
nbollis, line 273: remove try catch. Allow crash to happen if the path is wrong

Another option would be to throw a new MetaMorpheus exceptions.

MetaMorpheus/EngineLayer/GlobalVariables.cs
```diff
@@ -45,12 +45,14 @@ public static class GlobalVariables
         public static AnalyteType AnalyteType;
 
         public static List<string> ErrorsReadingMods;
+        public static List<string> StartupWarnings;
```
nbollis: remove

MetaMorpheus/EngineLayer/GlobalVariables.cs
```diff
@@ -85,6 +87,7 @@ public static void SetUpGlobalVariables()
             ExperimentalDesignFileName = "ExperimentalDesign.tsv";
             SeparationTypes = new List<string> { { "HPLC" }, { "CZE" } };
 
+            StartupWarnings = new List<string>();
```
nbollis: remove

MetaMorpheus/GUI/MainWindow.xaml.cs
```diff
@@ -1633,6 +1633,13 @@ private void PrintErrorsReadingMods()
                 NotificationHandler(null, new StringEventArgs(error, null));
             }
             GlobalVariables.ErrorsReadingMods.Clear();
+
+            // print any startup warnings (e.g. failure to create the custom monosaccharide file)
+            foreach (var warning in GlobalVariables.StartupWarnings)
```
nbollis: remove

MetaMorpheus/CMD/Program.cs
```diff
@@ -89,6 +109,10 @@ private static int Run(CommandLineSettings settings)
 
             GlobalVariables.SetUpGlobalVariables();
 
+            // Unconditionally surface startup warnings (e.g. custom monosaccharide file creation failure).
+            // Not gated on database type -- unlike ErrorsReadingMods below, these must reach every CLI run.
+            FlushStartupWarnings(settings.Verbosity);
```
nbollis: remove

MetaMorpheus/CMD/Program.cs
```diff
@@ -39,6 +39,26 @@ public static int Main(string[] args)
             return errorCode;
         }
 
+        /// <summary>
+        /// Writes the warnings collected by <see cref="GlobalVariables.SetUpGlobalVariables"/> (e.g. a failure to
+        /// create the custom monosaccharide file) to the console, then clears them so they are reported once.
+        /// The GUI does the same thing through its notifications pane; this is the command-line equivalent.
+        /// </summary>
+        /// <param name="verbosity">Nothing is written at <see cref="CommandLineSettings.VerbosityType.none"/>.</param>
+        public static void FlushStartupWarnings(CommandLineSettings.VerbosityType verbosity)
```
nbollis: remove

nbollis (approved):

trishorts (approved):

zhuoxinshi (approved):

PR #2758, opened 2026-08-26
Store the fragment index as flat arrays and build it in parallel (index build 5.9x, 1.8x less memory)

trishorts:
Rewrites the fragment/precursor index as flat arrays, fragments peptides in parallel while building it, and takes two allocations out of the per-scan scoring loop.

Results are unchanged: **1833 tests pass, 0 fail**. The full suite was run after each commit, not just at the end.

## Why

The fragment index was a `List<int>[]` with one bin per 0.001 Da. At the default 30000 Da ceiling that is 30,000,001 bins, so the reference array alone cost 240 MB before a single peptide id was stored, and every occupied bin added a `List` object plus its own doubling `int[]`. Millions of small objects for the GC to trace, and a pointer chase on every bin lookup in the search inner loop. It is also what the `OutOfMemoryException` handler telling users to fall back to Classic Search was apologizing for.

Separately, the fragment/bin pass ran single-threaded even though the digestion immediately above it was already parallel.

## Measured

Index build only, on a 64-core machine, comparing the old layout against the new one from the same sorted peptide list.

| workload | entries | occupied bins | entries/bin | old to new (1 thread) | new (64 threads) | index heap |
|---|---|---|---|---|---|---|
| human, **top-down** ETD | 48.9M | 18.0M (60.1%) | **2.7** | 14.0 s to 5.4 s | **2.1 s** (6.68x) | 1527 to **301 MB** (5.08x) |
| yeast, bottom-up | 243.7M | n/a | n/a | 43.3 s to 14.6 s | **7.3 s** (5.91x) | 1912 to 1044 MB (1.83x) |
| human, bottom-up | 316.2M | 7.1M (23.8%) | 44 | 46.3 s to 17.6 s | **9.0 s** (5.16x) | 2355 to 1321 MB (1.78x) |
| human XML + 2 var mods, one of 4 partitions | 1.08B | 7.2M (24.0%) | 150 | 100.1 s to 58.6 s | **31.2 s** (3.21x) | 6532 to 4236 MB (1.54x) |

Entry counts identical in every run. Databases: `uniprot-yeast-filtered-reviewed_yes.fasta.gz` and UniProt human (53,722 and 104,303 proteins with decoys).

**The win tracks bin sparsity, not database size.** When bins hold ~150 entries a `List` amortizes its header over 150 ints and the old layout is merely wasteful. When bins hold 2.7 entries it spends more on object headers than on data. Top-down spreads proteoform fragments across 60% of the mass range at 2.7 entries per bin, which is why memory drops 5x there and 1.8x for tryptic peptides -- the change helps most exactly where the index scales worst.

Worth noting `Test/TopDownTestData/TopDownSearchToml.toml` sets `SearchType = "Classic"`, so top-down does not build a fragment index at all today. A 1.5 GB index for a 107k-proteoform search is a plausible reason Modern was never the default there; 301 MB may change that calculus.

This is a win on index construction and memory, not a 5x faster search end to end -- how much it shows up depends on what share of a run goes into indexing, which varies with file count and index-cache reuse.

## What had to be preserved

Three invariants govern this change, and each has a test:

- **Peptide ids ascend within a bin.** `BinarySearchBinForPrecursorIndex` searches a bin *by peptide mass*, which only works because peptides are mass-sorted, so id order is mass order. The build records emissions in peptide order and counting-sorts them stably.
- **The parallel build must equal the sequential one.** Blocks are contiguous and their emissions concatenated in block order, so the index does not depend on thread count. `TestIndexIsIdenticalRegardlessOfThreadCount` builds the same database at 1, 2, 3 and 8 threads and compares every bin.
- **`OrderByDescending` is a stable sort.** Ties decide which peptide becomes the PSM, so the counting sort replacing it is stable on purpose. An unstable sort would have changed results silently.

The precursor index keeps its two waves -- base masses, then interior terminal mods -- in that order, because it is iterated rather than searched, so its within-bin order also feeds tie-breaks.

## Known regression: a 2.1B entry ceiling

Holding every entry in one `int[]` caps an index at `int.MaxValue` entries. `List<int>[]` had no such cap -- each bin carried its own list, so only free memory set the limit.

The cap is reachable, not theoretical. Human UniProt XML with oxidation and acetylation as variable mods, unpartitioned:

| | result |
|---|---|
| old `List<int>[]` | succeeds: 367.6 s, **22.6 GB** heap, 3,948,838,078 entries |
| new CSR | refused |

So on a large-memory machine there is a class of search the old code built and this one will not.

Mitigating it: partitioning is how MetaMorpheus already bounds index size, and the same database at 4 partitions builds fine at 1.08B entries each. **Partitioned new code beats unpartitioned old code on both axes** -- 4 x 31.2 s = 125 s of index building against 367.6 s, and 4.2 GB peak against 22.6 GB. The workaround is faster and lighter than the behaviour it replaces.

`Build` therefore throws a `MetaMorpheusException` naming the entry count, the limit, and the partition multiplier required:

> This database needs 3,948,838,078 fragment index entries, more than the 2,147,483,647 a single index can hold. Multiply the number of database partitions by at least 2, or reduce the maximum fragment mass, the number of variable modifications, or the missed cleavages. Classic Search does not build a fragment index at all and has no such limit.

Lifting the cap would mean segmenting the entry array on bin boundaries so no bin straddles a segment, at the cost of an extra indirection on every bin lookup. Left as a follow-up.

## Scoring loop

`Array.Clear` over the whole peptide index between scans is O(peptides) of bookkeeping per scan whatever the scan touched. `ScanScoringTable` instead stamps each cell with the scan it belongs to and reads a stale stamp as zero, making scan setup O(1) apart from a clear every 255 scans.

That is **not** unconditionally better, so it is selected rather than assumed. A stamped cell is two bytes wide, so a scan touching most of the index pays double the memory traffic to skip a memset it was already amortizing. On a synthetic version of the loop:

| touched fraction | stamping vs clearing |
|---|---|
| 0.1% (normal ppm tolerance) | 8-13x faster |
| 1% | 1.8x faster |
| 10% | 0.8x -- slower |
| 100% (open search) | 0.6x -- slower |

Open searches therefore keep clearing. Crosslink and glyco search in `OpenSearchMode` and are behaviorally untouched.

`OrderByDescending` on the coarse scores ran once per scan in three engines, allocating a LINQ ordering and a key array each time. Coarse scores are bytes, so `DescendingScoreSorter` counting-sorts them in two linear passes with no per-scan allocation.

## Index cache format

The cache is now two flat arrays instead of NetSerializer over millions of `List` objects. Writing the offset table verbatim would have cost 4 bytes per bin regardless of density -- 114 MB for a small database where the old format wrote ~36 MB -- so only occupied bins are stored, as a varint gap plus a varint count. That same small database now writes well under a megabyte.

Reads validate magic, version, bin ids and that counts add up, so a truncated or foreign file raises `InvalidDataException`. `MetaMorpheusTask` already treats a throw there as "rebuild the index", so **existing cached indexes are regenerated on first run** rather than misread.

## Other limits

- `MassBinIndex.Build` is now the bottleneck in index construction -- count and scatter are sequential, roughly 7 s of the 7.3 s parallel yeast stage, and the reason the parallel speedup fades as bins get denser. Parallelizing the scatter needs per-block-per-bin offsets, a ~2 GB table at 16 threads, so it is not a straightforward follow-up.
- Small databases build ~70 ms slower, because the offset table is sized by bin count (30M) rather than entry count, so it pays a fixed prefix sum the old layout did not. Irrelevant against a real search.

## Scope

MetaMorpheus only -- no mzLib change, no version bump, no repackaging. mzLib carries the same array-of-lists shape in `MassSpectrometry/PeakIndexing/IndexingEngine.cs`, but at ~200k bins rather than 30M and with reference-typed payloads, so the same treatment would return far less there; the real win in that one would be struct peaks, which is a public-API change and a separate discussion.

### DIFF
--- a/MetaMorpheus/EngineLayer/CrosslinkSearch/CrosslinkSearchEngine.cs
+++ b/MetaMorpheus/EngineLayer/CrosslinkSearch/CrosslinkSearchEngine.cs
@@ -1,4 +1,5 @@
 ﻿using EngineLayer.ModernSearch;
+using EngineLayer.Indexing;
 using MassSpectrometry;
 using MzLibUtil;
 using Omics.Fragmentation;
@@ -34,12 +35,12 @@ namespace EngineLayer.CrosslinkSearch
         private Modification NH2DeadEnd;
         private Modification Loop;
         private readonly char[] AllCrosslinkerSites;
-        private readonly List<int>[] SecondFragmentIndex;
+        private readonly MassBinIndex SecondFragmentIndex;
         private readonly double[] PrecursorMassTable;
         private readonly double[] NextPrecursorMassTable;
 
         public CrosslinkSearchEngine(List<CrosslinkSpectralMatch>[] globalCsms, Ms2ScanWithSpecificMass[] listOfSortedms2Scans, List<PeptideWithSetModifications> peptideIndex,
-            List<int>[] fragmentIndex, List<int>[] secondFragmentIndex, int currentPartition, CommonParameters commonParameters, 
+            MassBinIndex fragmentIndex, MassBinIndex secondFragmentIndex, int currentPartition, CommonParameters commonParameters, 
             List<(string fileName, CommonParameters fileSpecificParameters)> fileSpecificParameters,
             Crosslinker crosslinker, int CrosslinkSearchTopNum, bool CleaveAtCrosslinkSite, bool quench_H2O, bool quench_NH2, bool quench_Tris, List<string> nestedIds, 
             List<(int, int, int)>[] candidates, int nextPartition, 
--- a/MetaMorpheus/EngineLayer/CrosslinkSearch/CrosslinkSearchEngine.cs
+++ b/MetaMorpheus/EngineLayer/CrosslinkSearch/CrosslinkSearchEngine.cs
@@ -1,5 +1,6 @@
 ﻿using EngineLayer.ModernSearch;
 using EngineLayer.Indexing;
+using EngineLayer.Util;
 using MassSpectrometry;
 using MzLibUtil;
 using Omics.Fragmentation;
@@ -116,10 +117,11 @@ namespace EngineLayer.CrosslinkSearch
             int[] threads = Enumerable.Range(0, maxThreadsPerFile).ToArray();
             Parallel.ForEach(threads, (scanIndex) =>
             {
-                byte[] scoringTable = new byte[PeptideIndex.Count];
+                var scoringTable = new ScanScoringTable(PeptideIndex.Count, UseStampedScoringTable);
                 List<int> idsOfPeptidesPossiblyObserved = new List<int>();
-                byte[] secondScoringTable = new byte[PeptideIndex.Count];
+                var secondScoringTable = new ScanScoringTable(PeptideIndex.Count, UseStampedScoringTable);
                 List<int> childIdsOfPeptidesPossiblyObserved = new List<int>();
+                var scoreSorter = new DescendingScoreSorter();
 
                 byte scoreAtTopN = 0;
                 int peptideCount = 0;
@@ -130,7 +132,7 @@ namespace EngineLayer.CrosslinkSearch
                     if (GlobalVariables.StopLoops) { return; }
 
                     // empty the scoring table to score the new scan (conserves memory compared to allocating a new array)
-                    Array.Clear(scoringTable, 0, scoringTable.Length);
+                    scoringTable.BeginScan();
                     idsOfPeptidesPossiblyObserved.Clear();      
 
                     var scan = ListOfSortedMs2Scans[scanIndex];
@@ -147,7 +149,7 @@ namespace EngineLayer.CrosslinkSearch
                     //child scan first - pass scoring
                     if (scan.ChildScans != null && CommonParameters.MS2ChildScanDissociationType != DissociationType.Unknown && CommonParameters.MS2ChildScanDissociationType != DissociationType.LowCID)
                     {
-                        Array.Clear(secondScoringTable, 0, secondScoringTable.Length);
+                        secondScoringTable.BeginScan();
                         childIdsOfPeptidesPossiblyObserved.Clear();
 
                         List<int> childBinsToSearch = new List<int>();
@@ -166,7 +168,7 @@ namespace EngineLayer.CrosslinkSearch
                             {
                                 idsOfPeptidesPossiblyObserved.Add(childId);
                             }
-                            scoringTable[childId] = (byte)(scoringTable[childId] + secondScoringTable[childId]);
+                            scoringTable.Set(childId, (byte)(scoringTable[childId] + secondScoringTable[childId]));
                         }
                     }
 
@@ -176,7 +178,7 @@ namespace EngineLayer.CrosslinkSearch
                         scoreAtTopN = 0;
                         peptideCount = 0;
 
-                        foreach (int id in idsOfPeptidesPossiblyObserved.OrderByDescending(p => scoringTable[p]))
+                        foreach (int id in scoreSorter.Sort(idsOfPeptidesPossiblyObserved, scoringTable))
                         {
                             peptideCount++;
                             // Whenever the count exceeds the TopN that we want to keep, we removed everything with a score lower than the score of the TopN-th peptide in the ids list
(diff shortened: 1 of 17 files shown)


trishorts:
CI is green on `13da65b4` — all 13 checks pass, including both halves of the Windows suite and codecov patch/project.

That covers the two paths the scoring-table change splits on: `Test on Windows (glyco)` exercises `OpenSearchMode`, which keeps the existing `Array.Clear`, and `Test on Windows (core)` covers the stamped path used by narrow-tolerance modern and non-specific search. Both agree with master.

Reviewer notes, in the order I'd want them if I were reviewing this:

1. **The three invariants in "What had to be preserved" are the whole risk surface.** If any of them is wrong the tests still pass on small databases and results drift on large ones. Worth a second pair of eyes on the stable counting sort in `DescendingScoreSorter` and on the two-wave precursor ordering in `IndexingEngine.CreatePrecursorBaseEmissions` / `AddInteriorTerminalModsToPrecursorIndex`.

2. **`span != null` compiles and is always true** for a non-empty backing array, via the implicit `T[]` conversion. `Test/SearchEngineTests.cs` had exactly that pattern against the old `List<int>[]`; it now uses `CountInBin(bin) > 0`. I swept for other instances and found none, but it is an easy thing to reintroduce.

3. **Cached indexes regenerate on first run** after this merges, since the on-disk format changed. No user action needed — `MetaMorpheusTask` already treats a failed index read as "rebuild" — but the first search against an existing index folder will be slower than usual.

4. **The 2.1B entry ceiling is a real regression**, documented in the PR body with the measurement showing the old layout building the index this one refuses. Accepting it was a deliberate call, on the grounds that partitioning is the existing mechanism for exactly this and that partitioned new code is faster and lighter than unpartitioned old code. Happy to do the segmented-array fix instead if reviewers would rather not carry the limit.

The benchmark harness used for the numbers in the description is not part of this PR — it lives outside the repo. I can add it under `Test` as an `[Explicit]` fixture if that would be useful for future index work.

nbollis (approved):

trishorts:
This overlaps [#2742](https://github.com/smith-chem-wisc/MetaMorpheus/pull/2742) almost entirely, and I opened this without noticing that. Flagging it here so nobody reviews this PR without knowing.

**These two need to become one thing before either merges — one gets picked and the other closed, or they get folded into a single PR.** Both staying open means whichever merges second rewrites its half against 14 files the other one already changed.

Verified:

- Not stacked. `git merge-base` puts this and #2742 both on master `99711097`.
- `MassBinIndex` here and `FragmentIndex` there are the same design down to the member names: CSR layout, `Length => binStart.Length - 1`, `EntryCount`, `ReadOnlySpan<int> this[int bin]`, empty-not-null bins, `MetaMorpheusException` past `int.MaxValue`.
- The layout win is common to both. The difference is the build: this one fragments each peptide once into per-thread emission runs, `Parallel.For` over blocks, then a stable counting sort; #2742 counts then fills, serial, fragmenting every peptide twice (`IndexingEngine.cs:179` and `:236`).
- #2742 also carries four partition-dependence bug fixes that this PR does not, and those are the more valuable half of the pair.

So the honest summary is that #2742 is the more complete change and this one has the better build. One PR with both beats either. Same comment on #2742; the decision belongs on that thread as much as this one.

trishorts:
Closing this in favour of two smaller PRs, which resolves the "pick one" I raised above. Nothing is being thrown away except the duplicated half.

This PR was three changes in one branch, and only one of them collided with #2742:

| | where it went |
|---|---|
| CSR fragment index (`MassBinIndex`) | **dropped** — #2742's `FragmentIndex` is the same design and got there first |
| parallel index build | [acesnik/MetaMorpheus#8](https://github.com/acesnik/MetaMorpheus/pull/8), against #2742's branch |
| per-scan scoring table + coarse-score sort | #2761, against master |

**#2761** is independent of the layout question entirely — it never touches the index. Full suite 1830 passed, 0 failed.

**acesnik/MetaMorpheus#8** applies the build strategy to `FragmentIndex` rather than to a class of my own, so it's a diff against his code with `FragmentIndex.cs` untouched. On the full reviewed human proteome it takes the index build from 47.1 s to 23.4 s and drops 17 GB of allocation, because count-then-fill fragmented the whole database twice and this fragments it once. Verified byte-identical to his serial build by hashing every occupied bin, including the non-specific case with a precursor index and interior terminal mods. Full suite 1892 passed, 0 failed.

The one thing genuinely lost is the index cache file format from `93e92433d`, which stored only occupied bins. That was specific to `MassBinIndex`; if it's still wanted it can go on top of `FragmentIndex` separately.

Reopen if any of that turns out to be the wrong split.

PR #2759, opened 2026-08-26
Let modern search index nucleic acids (RNA/oligos)

trishorts:
Modern (index-based) search could not run against a nucleic acid database. This makes it work, and adds tests proving it reaches the same answer classic search does.

**1833 tests pass, 0 fail.** 1830 were already there; the fourth new test replaces one that asserted the old refusal.

## What failed, exactly

Two blockers, in the order a user hits them. Both reproduced before changing anything.

**First**, `SearchTask`'s modern branch cast the biopolymer list to `Protein`:

```
InvalidCastException: Unable to cast object of type 'Transcriptomics.RNA'
to type 'Proteomics.Protein'
    at System.Linq.Enumerable.CastICollectionIterator`1.ToList()
    SearchTask.cs:382 — bioPolymerList.Cast<Protein>().ToList()
```

The line above it read `// Assume modern search is for proteins.` The message names two types and explains neither, which is a bad first thing to hit.

**Second**, behind it, `IndexingEngine` refused non-protein digestion outright:

```
MetaMorpheusException: Digestion parameters must be of type DigestionParams.
Not yet implemented for Rna Digestion
    IndexingEngine.cs:98
```

`IndexingEngine` was typed `List<Protein>` in and `List<PeptideWithSetModifications>` out.

**A third thing worth knowing**, which is not a bug but is confusing: `RnaSearchParameters`'s constructor sets `SearchType = SearchType.Classic`. Anyone who sets `SearchType` *before* constructing it has the value silently overwritten. That behaviour is unchanged here — flagged because it can look like the fix did not take.

## This was a decision, not an oversight

`Test/Transcriptomics/RnaSpecificTypeSwitches.cs` contained:

```csharp
[Test]
public static void IndexingEngine_ThrowsOnRNA()
{
    var indexEngine = new IndexingEngine(..., new RnaDigestionParams(), ...);
    Assert.Throws<MetaMorpheusException>(() => indexEngine.Run());
}
```

Someone deliberately pinned the refusal. **This PR reverses that decision**, so the test becomes `IndexingEngine_IndexesRNA`. That inversion is the single line in this diff most deserving of a second opinion — if the refusal was protecting something I have not found, this is where it gets discussed.

## Why the change is small: the seams already existed

Almost nothing had to be invented. mzLib and the match layer were already generic:

| seam | state before this PR |
|---|---|
| `IBioPolymer.Digest(IDigestionParams, fixed, variable, silacLabels, turnoverLabels, topDown)` | already generic — **and already carries the SILAC parameters**, so one call serves proteins and nucleic acids |
| `IBioPolymerWithSetMods` | already exposes `Fragment(...)` and, via `IHasChemicalFormula`, `MonoisotopicMass` — everything the index reads |
| `SpectralMatch`, `PeptideSpectralMatch`, `OligoSpectralMatch` | **all three constructors already take `IBioPolymerWithSetMods`** |
| `ClassicSearchEngine` | already takes `List<IBioPolymer>`, already switches on `GlobalVariables.AnalyteType` to pick the match type |

So the work is type-widening plus copying a switch that classic search already makes. No new abstraction is introduced.

## The parameters are `IEnumerable`, and that is deliberate

The obvious move is a `List<Protein>` convenience overload next to a `List<IBioPolymer>` one, mirroring `ClassicSearchEngine`'s pair of constructors. **I tried that first and it is wrong here.** `List<T>` is invariant, so the overload is genuinely needed for the ~45 existing protein call sites — but it makes an empty collection expression ambiguous, and three existing tests broke with:

```
CS0121: The call is ambiguous between IndexingEngine(List<Protein>, ...)
        and IndexingEngine(List<IBioPolymer>, ...)
```

from `new IndexingEngine([], [], [], ...)`. `null` would have been ambiguous too. That is a wart every future caller would meet.

Taking `IEnumerable<IBioPolymer>` instead solves it properly: `IEnumerable<T>` is **covariant**, so `List<Protein>` binds implicitly, there is no overload, and nothing is ambiguous. The constructor materializes with `as List<...> ?? .ToList()`, so passing a `List` costs no copy.

The same reasoning applies to the three derived engines. Their constructors now take `IEnumerable<IBioPolymerWithSetMods>`, which is why **~20 protein call sites in `SearchEngineTests` and `XLTest` needed zero edits** — `List<PeptideWithSetModifications>` binds by covariance too.

The one place an overload survives is `GenerateIndexes`, because a `ref` parameter cannot be covariant. Overloads on a `ref` parameter type cannot be ambiguous (a `ref` argument must be a typed variable, never `null` or `[]`), so the wart above cannot recur there.

## Crosslink, glyco and non-specific stay proteomics-only

These are built around proteases and glycans — terminal mod placement, the "single" digestion agents, glycan boxes, the FDR categories. None has a nucleic acid counterpart, and `CrosslinkSearchEngine` already threw on non-protein digestion of its own accord. Generifying them speculatively would be a much larger change with nothing exercising it.

Each therefore keeps a peptide-typed view of the index that `ModernSearchEngine` now holds generically:

```csharp
protected new readonly List<PeptideWithSetModifications> PeptideIndex;
```

Two things to know about that:

- **It shadows the base field with `new`.** Same objects, narrower type; the derived bodies bind to the derived field and needed no other edits. Shadowing is a smell in general, so it is called out here rather than hidden — the alternative was rewriting several hundred lines of crosslink and glyco internals for no behavioural gain.
- **Where those engines call *base* methods** (`IndexedScoring`) they pass `base.PeptideIndex` explicitly, since the base signature wants the generic list. Three call sites, in crosslink and glyco.

Non-specific search now fails with a sentence instead of a cast exception:

> Non-specific search is only implemented for proteins. Use Classic or Modern search for nucleic acid databases.

That is the one behavioural change for proteins-adjacent paths, and it only fires on input that previously crashed anyway.

## The index cache is skipped for oligos

This is the only place the fix is a workaround rather than a fix, so it deserves the detail.

The on-disk peptide index round-trips through NetSerializer and is then relinked:

```csharp
peptide.SetNonSerializedPeptideInfo(GlobalVariables.AllModsKnownDictionary, proteinDictionary, storedDigestParams);
```

In mzLib, `PeptideWithSetModifications` is `[Serializable]` and has that method. **`OligoWithSetMods` has neither.** So the cache cannot hold oligos without an mzLib change — which means a version bump and a repackaging cycle before RNA modern search could work at all.

Instead, `GenerateIndexes` checks `GlobalVariables.AnalyteType != AnalyteType.Oligo` and, for oligos, runs the engine straight into memory without consulting or writing the cache. Nucleic acid databases are small enough that rebuilding each run costs nothing measurable.

Consequences, stated plainly:

- **No mzLib change is needed for this PR.** mzLib stays at 1.0.585, untouched.
- Oligo searches rebuild the index every run. For the databases people actually use, this is not noticeable.
- Protein caching is completely unchanged — same code path, same files, same reuse.
- If someone later wants oligo index caching, that is an mzLib PR adding `[Serializable]` and a parent-relink to `OligoWithSetMods`, and a small follow-up here to drop the check.

## Match type

`ModernSearchEngine` now makes the same choice `ClassicSearchEngine` makes at `ClassicSearchEngine.cs:241`:

```csharp
PeptideSpectralMatches[scanIndex] = GlobalVariables.AnalyteType == AnalyteType.Oligo
    ? new OligoSpectralMatch(...)
    : new PeptideSpectralMatch(...);
```

`AnalyteType` is set inside `MetaMorpheusEngine.Run()` from the engine's own `CommonParameters`, so this is correct for an engine constructed directly in a test as well as for one driven by a task.

## Tests

New file `Test/Transcriptomics/TestRnaModernSearch.cs`:

- `IndexingEngine_IndexesOligos` — the index digests RNA and bins its fragments; entries are `OligoWithSetMods`.
- `OligoIndex_BinsAreSortedByOligoId` — ids ascend within every fragment bin, and the peptide index is ordered by mass. **This one matters**: the search binary-searches a bin *by mass*, which is only valid because id order equals mass order. If that ever stopped holding for oligos, results would go quietly wrong rather than crash. The test also asserts it found multi-entry bins, so it cannot pass vacuously.
- `ModernSearch_FindsSimpleSixmer` — the payoff. Modern search finds the same GUACUG sixmer `TestRnaSearchEngine.FindsSimpleSixmer` finds with classic search, returns it as an `OligoSpectralMatch`, and clears the same score threshold (>22).

Changed: `IndexingEngine_ThrowsOnRNA` → `IndexingEngine_IndexesRNA`, as described above.

## Not covered

Stated so nobody assumes otherwise:

- **Non-specific search on RNA** — refused with a clear message, not implemented.
- **Crosslink and glyco on RNA** — unchanged, proteomics-only.
- **Oligo index caching** — needs the mzLib change described above.
- **No performance claim.** This is about making the path work. Whether modern search is *faster* than classic for realistic nucleic acid databases is unmeasured, and on small databases classic may well win.

## Relationship to #2758

#2758 rewrites `IndexingEngine` and `ModernSearchEngine` heavily (flat CSR fragment index, parallel build). **This PR is deliberately branched off `master`, not off #2758, so it can be reviewed on its own.** The two will conflict — whichever merges second needs a rebase. The conflicts are mechanical (both touch the same declarations, for different reasons) and I am happy to do that rebase in whichever order you prefer.

### DIFF
--- a/MetaMorpheus/EngineLayer/Indexing/IndexingEngine.cs
+++ b/MetaMorpheus/EngineLayer/Indexing/IndexingEngine.cs
@@ -192,9 +192,23 @@ namespace EngineLayer.Indexing
                 }
 
                 //Add terminal mods if needed (do it here rather than earlier so that we don't have to fragment twice)
-                // Only reachable for the "single" proteases, which are protein-only, so the cast is safe.
-                if (addInteriorTerminalModsToPrecursorIndex && peptides[peptideId] is PeptideWithSetModifications peptideForTerminalMods)
+                // The "single" agents are NOT protein-only: mzLib's rnases.tsv ships RNases named
+                // singleN and singleC, which the Name.Contains("single") test above matches. What makes
+                // this reachable only for peptides is the refusal in SearchTask.RunSpecific -- a
+                // non-specific search rejects a nucleic acid database before it gets here. That guard
+                // lives in another file, so throw rather than skip if it is ever relaxed: silently
+                // omitting interior terminal mods would cost identifications with nothing reporting why.
+                if (addInteriorTerminalModsToPrecursorIndex)
                 {
+                    if (peptides[peptideId] is not PeptideWithSetModifications peptideForTerminalMods)
+                    {
+                        throw new MetaMorpheusException(
+                            "Interior terminal modifications are only implemented for peptides, but a " +
+                            $"{peptides[peptideId].GetType().Name} reached the precursor index under digestion agent " +
+                            $"'{CommonParameters.DigestionParams.DigestionAgent.Name}'. Non-specific search is " +
+                            "supposed to refuse a nucleic acid database before indexing.");
+                    }
+
                     AddInteriorTerminalModsToPrecursorIndex(precursorIndex, fragments, peptideForTerminalMods, peptideId, terminalModifications);
                 }
 
--- a/MetaMorpheus/EngineLayer/Indexing/IndexingEngine.cs
+++ b/MetaMorpheus/EngineLayer/Indexing/IndexingEngine.cs
@@ -21,7 +21,7 @@ namespace EngineLayer.Indexing
         private static readonly double WaterMonoisotopicMass = PeriodicTable.GetElement("H").PrincipalIsotope.AtomicMass * 2 + PeriodicTable.GetElement("O").PrincipalIsotope.AtomicMass;
 
         private const int FragmentBinsPerDalton = 1000;
-        private readonly List<Protein> ProteinList;
+        private readonly List<IBioPolymer> BioPolymerList;
         private readonly List<Modification> FixedModifications;
         private readonly List<Modification> VariableModifications;
         private readonly List<SilacLabel> SilacLabels;
@@ -33,13 +33,18 @@ namespace EngineLayer.Indexing
         public readonly List<FileInfo> ProteinDatabases;
         public readonly TargetContaminantAmbiguity TcAmbiguity;
 
-        public IndexingEngine(List<Protein> proteinList, List<Modification> variableModifications, List<Modification> fixedModifications,
+        /// <summary>
+        /// Takes IEnumerable rather than List so that a List&lt;Protein&gt; still binds here by
+        /// covariance -- the protein callers need no change, and there is no overload to be ambiguous
+        /// about when someone passes an empty collection expression.
+        /// </summary>
+        public IndexingEngine(IEnumerable<IBioPolymer> bioPolymerList, List<Modification> variableModifications, List<Modification> fixedModifications,
             List<SilacLabel> silacLabels, SilacLabel startLabel, SilacLabel endLabel, int currentPartition, DecoyType decoyType,
-            CommonParameters commonParams, List<(string fileName, CommonParameters fileSpecificParameters)> fileSpecificParameters, 
+            CommonParameters commonParams, List<(string fileName, CommonParameters fileSpecificParameters)> fileSpecificParameters,
             double maxFragmentSize, bool generatePrecursorIndex, List<FileInfo> proteinDatabases, TargetContaminantAmbiguity tcAmbiguity, List<string> nestedIds)
             : base(commonParams, fileSpecificParameters, nestedIds)
         {
-            ProteinList = proteinList;
+            BioPolymerList = bioPolymerList as List<IBioPolymer> ?? bioPolymerList.ToList();
             VariableModifications = variableModifications;
             FixedModifications = fixedModifications;
             SilacLabels = silacLabels;
@@ -63,7 +68,7 @@ namespace EngineLayer.Indexing
             sb.AppendLine("Partitions: " + CurrentPartition + "/" + CommonParameters.TotalPartitions);
             sb.AppendLine("Precursor Index: " + GeneratePrecursorIndex);
             sb.AppendLine("Search Decoys: " + DecoyType);
-            sb.AppendLine("Number of proteins: " + ProteinList.Count);
+            sb.AppendLine("Number of proteins: " + BioPolymerList.Count);
             sb.AppendLine("Number of fixed mods: " + FixedModifications.Count);
             sb.AppendLine("Number of variable mods: " + VariableModifications.Count);
             sb.AppendLine("Dissociation Type: " + CommonParameters.DissociationType);
@@ -83,7 +88,7 @@ namespace EngineLayer.Indexing
                 sb.AppendLine("specificProtease: " + digestionParam.SpecificProtease);
             sb.AppendLine("maximumFragmentSize" + (int)Math.Round(MaxFragmentSize));
 
-            sb.Append("Localizeable mods: " + ProteinList.Select(b => b.OneBasedPossibleLocalizedModifications.Count).Sum());
+            sb.Append("Localizeable mods: " + BioPolymerList.Select(b => b.OneBasedPossibleLocalizedModifications.Count).Sum());
             return sb.ToString();
         }
 
@@ -93,26 +98,25 @@ namespace EngineLayer.Indexing
             int oldPercentProgress = 0;
 
             // digest database
-            List<PeptideWithSetModifications> peptides = new List<PeptideWithSetModifications>();
+            // IBioPolymer.Digest covers proteins and nucleic acids alike, and already carries the SILAC
+            // parameters, so the index no longer needs to know which kind of polymer it is holding.
+            List<IBioPolymerWithSetMods> peptides = new List<IBioPolymerWithSetMods>();
 
-            if (CommonParameters.DigestionParams is not DigestionParams digestionParams)
-                throw new MetaMorpheusException("Digestion parameters must be of type DigestionParams. Not yet implemented for Rna Digestion");
-            
             int maxThreadsPerFile = CommonParameters.MaxThreadsToUsePerFile;
             int[] threads = Enumerable.Range(0, maxThreadsPerFile).ToArray();
             Parallel.ForEach(threads, (i) =>
             {
-                List<PeptideWithSetModifications> localPeptides = new List<PeptideWithSetModifications>();
+                List<IBioPolymerWithSetMods> localPeptides = new List<IBioPolymerWithSetMods>();
 
-                for (; i < ProteinList.Count; i += maxThreadsPerFile)
+                for (; i < BioPolymerList.Count; i += maxThreadsPerFile)
                 {
                     // Stop loop if canceled
                     if (GlobalVariables.StopLoops) { return; }
 
-                    localPeptides.AddRange(ProteinList[i].Digest(digestionParams, FixedModifications, VariableModifications, SilacLabels, TurnoverLabels));
+                    localPeptides.AddRange(BioPolymerList[i].Digest(CommonParameters.DigestionParams, FixedModifications, VariableModifications, SilacLabels, TurnoverLabels));
 
                     progress++;
-                    var percentProgress = (int)((progress / ProteinList.Count) * 100);
+                    var percentProgress = (int)((progress / BioPolymerList.Count) * 100);
 
                     if (percentProgress > oldPercentProgress)
                     {
@@ -188,9 +192,10 @@ namespace EngineLayer.Indexing
                 }
 
                 //Add terminal mods if needed (do it here rather than earlier so that we don't have to fragment twice)
-                if (addInteriorTerminalModsToPrecursorIndex)
+                // Only reachable for the "single" proteases, which are protein-only, so the cast is safe.
+                if (addInteriorTerminalModsToPrecursorIndex && peptides[peptideId] is PeptideWithSetModifications peptideForTerminalMods)
                 {
-                    AddInteriorTerminalModsToPrecursorIndex(precursorIndex, fragments, peptides[peptideId], peptideId, terminalModifications);
+                    AddInteriorTerminalModsToPrecursorIndex(precursorIndex, fragments, peptideForTerminalMods, peptideId, terminalModifications);
                 }
 
                 progress++;
@@ -206,7 +211,7 @@ namespace EngineLayer.Indexing
             return new IndexingResults(peptides, fragmentIndex, precursorIndex, this);
         }
 
-        private List<int>[] CreateNewPrecursorIndex(List<PeptideWithSetModifications> peptidesSortedByMass)
+        private List<int>[] CreateNewPrecursorIndex(List<IBioPolymerWithSetMods> peptidesSortedByMass)
         {
             // create precursor index
             List<int>[] precursorIndex = null;
--- a/MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
+++ b/MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
@@ -1,4 +1,4 @@
-using Chemistry;
+﻿using Chemistry;
 using EngineLayer;
 using EngineLayer.Indexing;
 using MassSpectrometry;
@@ -1271,8 +1271,15 @@ namespace TaskLayer
             }
         }
 
+        /// <summary>
+        /// Cast rather than OfType, to match the write side one method up. Both respond to the same
+        /// violated precondition -- a cached index is only reachable when indexIsCacheable, which is
+        /// AnalyteType != Oligo -- and they must fail the same way. OfType here would drop the oligos
+        /// and carry on with whatever proteins remained, returning a plausible index silently built
+        /// from a subset of the database; the write side already throws on the first oligo.
+        /// </summary>
         private static List<IBioPolymerWithSetMods> ReadPeptideIndex(string peptideIndexFileName, IEnumerable<IBioPolymer> allKnownBioPolymers)
-            => ReadPeptideIndex(peptideIndexFileName, allKnownBioPolymers.OfType<Protein>().ToList())
+            => ReadPeptideIndex(peptideIndexFileName, allKnownBioPolymers.Cast<Protein>().ToList())
                 .Cast<IBioPolymerWithSetMods>().ToList();
 
         private static List<PeptideWithSetModifications> ReadPeptideIndex(string peptideIndexFileName, List<Protein> allKnownProteins)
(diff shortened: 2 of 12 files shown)


nbollis (dismissed):

trishorts:
@nbollis — your approval here was dismissed by two commits of mine, not by a change in what the PR does. Flagging it rather than letting you rediscover it.

Both were tests, closing the `codecov/patch` gate:

| | |
|---|---|
| `5e4a074cf` | `GenerateIndexes` with `AnalyteType.Oligo` — asserts the index is built in memory and **nothing is cached**, since an oligo index cannot be read back |
| `70b66b70b` | non-specific search refuses a nucleic acid database as a `MetaMorpheusException` rather than a bare `InvalidCastException` |

Patch coverage went **74.00% → 86.27% → 92.15%**, and every check is now green. No production code changed in either commit.

Two things worth knowing if you re-review:

**The first test found a real gap, not just a coverage one.** Nothing in the suite reached `GenerateIndexes` at all — the three existing tests drive `IndexingEngine` and `ModernSearchEngine` directly. So `indexIsCacheable`, which is the decision this PR exists to make, had no test rather than merely no coverage. Verified by forcing it the other way: the `DatabaseIndex` folder appears and the test fails.

**My first version of that test was vacuous** and I want that on the record rather than buried. It watched an output folder `GenerateIndexes` is never given — the cache is written beside the *database* — so it would have passed against either branch. The database is now copied somewhere of its own and the assertion is that no `DatabaseIndex` folder appears next to it.

Four lines remain uncovered and I left them deliberately: three in `MetaMorpheusTask.cs`, and `NextPrecursorMassTable` in `CrosslinkSearchEngine.cs`, which needs a crosslink search with two or more index partitions. Happy to build that fixture if you would rather it were covered than argued for.

Also brought the branch up to date with master just now (`5003042e7`), via `update-branch` so nothing further is dismissed.

acesnik (commented):
🤖 The decision I most wanted to see in this PR is the one that is already made and documented: oligo indexes are not written to disk, because `OligoWithSetMods` is neither `[Serializable]` nor able to restore its parent the way `SetNonSerializedPeptideInfo` does, and nucleic acid databases are small enough to rebuild. That is the right call, and it quietly disposes of the thing I went looking for first — whether an RNA index could collide with a cached protein one. It cannot, because oligo searches never touch the cache at all, which is a stronger guarantee than a distinguishing key would be.

For protein searches I checked the key text is unchanged: `"Number of proteins: "` and `"Localizeable mods: "` still render identically off `BioPolymerList`, so existing cached indexes still hit rather than silently rebuilding. And `"protease: " + DigestionParams.DigestionAgent` would separate an `Rnase` from a `Protease` anyway if the caching decision were ever revisited.

`NonSpecificSearch_RefusesANucleicAcidDatabase` is a good test — asserting the message contains "Classic or Modern" and not just that it failed is the part most people skip. Taking `IEnumerable<IBioPolymer>` on the constructor so `List<Protein>` still binds by covariance is a nice touch too; it is why the protein call sites needed no change.

Three things, one of which is your own argument turned around.

**The refusal reaches a GUI user as a crash report.** Inline at `SearchTask.cs:459`. This is the point you made to me on #2721, traced end to end, and it applies here unchanged: the message is good and the user never sees it as a message.

**`Cast` and `OfType` disagree about the same violated precondition.** Inline at `MetaMorpheusTask.cs:1275`.

**The comment about the `"single"` agents is wrong, though its conclusion holds.** Inline at `IndexingEngine.cs:196`.

Not verified: I did not run an RNA search end to end, and `Test/Test.csproj` cannot execute on macOS, so the 277 new test lines are CI's to confirm. The solution builds on `-c UbuntuMac` with 0 errors and `GUI.csproj` compiles with `EnableWindowsTargeting`. The PR is `BEHIND` master, so a rebase is due regardless.

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
```diff
@@ -452,6 +450,16 @@ protected override MyTaskResults RunSpecific(string OutputFolder, List<DbForTask
                     //foreach terminus we're going to look at
                     foreach (CommonParameters paramToUse in paramsToUse)
                     {
+                        // Non-specific search is built around proteases -- terminal mod placement, the
+                        // "single" agents, the FDR categories -- none of which have a nucleic acid
+                        // counterpart yet. Say so, rather than letting the cast below throw a bare
+                        // InvalidCastException that names Protein and RNA and explains neither.
+                        if (bioPolymerList.Any(p => p is not Protein))
+                        {
+                            throw new MetaMorpheusException(
```
acesnik, line 459: This is the argument you made to me on #2721, and it lands the same way here. A `MetaMorpheusException` thrown from `RunSpecific` gets caught by `RunTask`, dumped into `results.txt` with system info and a stack trace, and rethrown; `EverythingRunnerEngine.Run` has no try/catch; and `MainWindow` routes the faulted task to `EverythingRunnerExceptionHandler`, which does not special-case `MetaMorpheusException` and shows "Would you like to report this crash?" with a mailto to mm_support.

So a user who picked Non-Specific with an RNA database is told MetaMorpheus crashed and invited to file a bug — and the message here, which is a good one and says what to do instead, is the thing they do not get shown.

Your own test acknowledges the shape of it: *"RunTask writes the failure into results.txt and rethrows, so the exception is observable here."* That is exactly what makes it observable to a test and invisible to a user.

The house channel is the one you pointed me at: `EverythingRunnerEngine.Run`, inside the per-task loop next to the existing gates, with `Warn(...)` + `FinishedAllTasks(OutputFolder)` + `return`. Both things the gate needs are in scope there — the search type off the task and the analyte type off `GlobalVariables` — and `engine.Warnings` gives the test something better to assert on than a thrown exception. Keeping this throw as a backstop for a direct `RunTask` caller is the split you and I settled on for #2721, and it works here for the same reason.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -1268,6 +1271,10 @@ private static void WritePeptideIndex(List<PeptideWithSetModifications> peptideI
             }
         }
 
+        private static List<IBioPolymerWithSetMods> ReadPeptideIndex(string peptideIndexFileName, IEnumerable<IBioPolymer> allKnownBioPolymers)
+            => ReadPeptideIndex(peptideIndexFileName, allKnownBioPolymers.OfType<Protein>().ToList())
```
acesnik: These two helpers respond to the same violated precondition in opposite ways. `WritePeptideIndex` at 1261 uses `Cast<PeptideWithSetModifications>()`, which throws `InvalidCastException` on the first oligo. `ReadPeptideIndex` here uses `OfType<Protein>()`, which silently drops them and carries on with whatever proteins remain.

Both are unreachable today — `indexIsCacheable` is `AnalyteType != Oligo`, and `GenerateIndexes_ForOligos_BuildsInMemoryAndCachesNothing` pins that. So this is not a live defect. But the guard is one boolean in one method, and if it is ever wrong the write path fails loudly while the read path returns a plausible, quietly wrong index. Given the alternative is an index built from a subset of the database, I would rather both were loud: `Cast` on the read side too, or a check on `indexIsCacheable` inside these helpers rather than only above them.

MetaMorpheus/EngineLayer/Indexing/IndexingEngine.cs
```diff
@@ -188,9 +192,10 @@ protected override MetaMorpheusEngineResults RunSpecific()
                 }
 
                 //Add terminal mods if needed (do it here rather than earlier so that we don't have to fragment twice)
-                if (addInteriorTerminalModsToPrecursorIndex)
+                // Only reachable for the "single" proteases, which are protein-only, so the cast is safe.
+                if (addInteriorTerminalModsToPrecursorIndex && peptides[peptideId] is PeptideWithSetModifications peptideForTerminalMods)
```
acesnik: The conclusion is right but the stated reason is not, and the reason is what a future reader will rely on. The comment says the `"single"` agents are protein-only. They are not — mzLib's `rnases.tsv` ships RNases named exactly `singleN` and `singleC`, the last two entries in the file, and `addInteriorTerminalModsToPrecursorIndex` keys on `DigestionAgent.Name.Contains("single")`, which those satisfy.

What actually makes this safe is somewhere else: non-specific search refuses a nucleic acid database at `SearchTask.cs:459`, so an RNA search never reaches a `"single"` agent here. That is a guard in a different file, and it is the kind of guard someone lifts.

Worth caring about because of how this fails if that happens. Before this PR the whole method threw `"Not yet implemented for Rna Digestion"` on an RNA digestion; now the `is PeptideWithSetModifications` pattern simply does not match, so interior terminal modifications are skipped for oligos with no error and no warning — a precursor index missing entries, and fewer identifications for a reason nothing reports. Citing the real reason would help; an `else` that throws would mean it cannot degrade quietly if the refusal is ever relaxed.

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
trishorts: Fixed in `33b9cf4d`. The refusal now runs in `EverythingRunnerEngine.Run` next to the existing database gates -- `Warn(...)` + `FinishedAllTasks(OutputFolder)` + `return` -- with the throw kept in `SearchTask` as the backstop for a direct `RunTask` caller, the same split as #2721.

New test `NonSpecificSearch_OverANucleicAcidDatabase_IsRefusedByTheRunnerNotThrown` asserts on `runner.Warnings` and on `Assert.DoesNotThrow`, plus that the task never got an output folder. I ran it against a build with the gate forced off: it fails there and passes with it, so it discriminates refusal from crash rather than just re-asserting the message.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
trishorts: Fixed in `33b9cf4d` -- `ReadPeptideIndex` now uses `Cast<Protein>()`, so both helpers fail the same way on the same violated precondition. Agreed on the reasoning: an index quietly built from a subset of the database is the worse of the two failures, and `indexIsCacheable` being one boolean in one method is thin protection for the difference.

MetaMorpheus/EngineLayer/Indexing/IndexingEngine.cs
trishorts: You are right and my comment was wrong. I checked `rnases.tsv` and initially did not see `singleN`/`singleC` -- my local mzLib clone's master was stale at 08-24. On the real master they are lines 95-96, the last two entries, added by mzLib #1217 which merged 08-27. My own PR, four days before I wrote that comment.

So `Name.Contains("single")` does match RNases, and the comment's stated reason was wrong. Fixed in `33b9cf4d`: the comment now cites the real guard -- the refusal in `SearchTask.RunSpecific` -- and says it lives in another file, and the `is not PeptideWithSetModifications` case throws a `MetaMorpheusException` naming the type and the digestion agent rather than skipping. If the refusal is ever relaxed it fails loudly instead of dropping interior terminal mods with nothing reporting it.

trishorts:
@acesnik all three comments addressed and resolved — ready for another look. `33b9cf4d`.

The refusal moved to `EverythingRunnerEngine.Run` with the throw kept as a backstop, the same split as #2721; the new test asserts on `Warnings` and `DoesNotThrow`, and I checked it fails against a build with the gate forced off. `ReadPeptideIndex` now uses `Cast<Protein>()` to match the write side.

On the `"single"` agents — you were right and I was wrong. My local mzLib clone's master was stale at 08-24, which is why I did not see them; on real master `singleN`/`singleC` are lines 95-96 of `rnases.tsv`, added by mzLib #1217, which merged 08-27. My own PR, four days before I wrote that comment. The comment now cites the real guard and the mismatch throws instead of silently skipping interior terminal mods. 38/38 green.

acesnik (approved):
🤖 All three fixed, and I checked each against the tree rather than taking the replies on trust.

**The runner gate.** Placement is right: inside the per-task loop, after `var ok = TaskList[i]`, alongside the existing database and spectra gates, and — the part that makes your output-folder assertion mean something — before `Path.Combine`/`CreateDirectory` for the task. `Warn` + `FinishedAllTasks` + `return`, with the throw kept below as the backstop.

One thing I checked that you did not claim, because it is where this gate could have been incomplete. The runner tests `GlobalVariables.AnalyteType == AnalyteType.Oligo`, while `SearchTask`'s throw tests `bioPolymerList.Any(p => p is not Protein)` — different predicates, and if they diverged the gate would let something through to the crash path. They do not: `DatabaseLoadingEngine.cs:93` dispatches `LoadOligoDb` versus `LoadProteinDb` on that same analyte type, so a non-`Protein` polymer can only appear when the type is `Oligo`. The gate covers exactly what the throw covered. Where they differ the runner is stricter — an Oligo run whose database yields nothing is refused earlier — which is the harmless direction.

**The test discriminates on three independent axes**, not just the message: no throw out of `runner.Run()`, the warning present with both the diagnosis and the remedy, and no per-task output folder. `[NonParallelizable]` is there too, consistent with the other two tests in the file that mutate `GlobalVariables.AnalyteType`, and the type is restored in a `finally` — which matters more than it looks, since that global is exactly what the gate reads.

**`ReadPeptideIndex` now uses `Cast<Protein>()`.** Both helpers fail the same way on the same violated precondition, so an index built from a silent subset of the database is no longer a possible outcome.

**The `IndexingEngine` comment now cites the real guard and throws instead of skipping.** Throwing is the right call *here* specifically because the user-facing path is now gated in the runner: this is an invariant violation rather than a configuration mistake, so the crash-report route is the correct destination for it, and the message names the type and the digestion agent so whoever relaxes the refusal upstream finds out immediately. Same split as #2721 — warn where a user can act, throw where only a caller can.

The stale-clone diagnosis is worth recording rather than waving past: `singleN`/`singleC` landed in `rnases.tsv` via mzLib #1217 on 08-27, your local master was at 08-24, and the comment was written against what you could see. That is a failure mode with no code smell to catch it, and it will recur for anyone reviewing mzLib behaviour from a MetaMorpheus checkout — the pin makes the two repositories drift by construction.

Approving. Not verified by me: the suite itself, since `Test/Test.csproj` is `net*.0-windows` and cannot execute here, so CI is the first real run of the new test.

nbollis (approved):

RayMSMS (approved):

Issue #1277, opened 2018-07-10
File names in command line


ncarrut:
Hi,
I'm using MetaMorpheus in command line mode and then parsing the results.  Can I set file names from the command line?  For example make the result folder "myResult" instead of the date stamp.  Also set the base result file name to get "nyName_Peptides.psmtsv" instead of the current result which is either aggregate_Peptides.psmtsv or the raw-file name _Peptides.psmtsv depending on how many files were searched.  
Nick

rmillikin:
Hi Nick,

You can specify the output folder with -o. For example:

`.\CMD.exe -t "C:\Data\SlicedYeastRawFiles\2018-07-02-14-07-16\Task2-SearchTask\SearchTaskconfig.toml" -o "C:\Data\SlicedYeastRawFiles\MyOutputFolder" -s "C:\Data\SlicedYeastRawFiles\SmallCalibratibleYeast.mzml" -d "C:\Data\SlicedYeastRawFiles\smalldb.fasta"`

Currently there's no way to rename the tsv output files; we organize the output such that each file has a list of PSMs/proteins, and then the aggregated one with all the PSMs/proteins from all the files. I'm not quite sure how to organize the output with a custom name unless we append the spectra file name on top of the user-defined name. How do you want this to work? Do you just want the aggregate file with a custom name and the individual file results remain the same as current?

As a side note, you can give each task a name, if that helps you organize your results. The task's name is defined in its .toml.

Rob

ncarrut:
Thanks for the response Rob.  I'm trying to make an automated search for QC
runs.  Sometimes it runs with just one .raw file and sometimes more.  One
file gives just the individual results named according to the raw file (eg
"myRawfile_PSMs.psmtsv").  Multiple files gives the aggregate file (eg.
"aggregate_PSMs.psmtsv").  I'm making my R script look for different file
names depending on whether I searched one .raw file or multiple and it
would make my life easier if there was just one file name to expect.  I see
your point about the awkwardness of a user-defined name.  Maybe you could
create the aggregate file regardless of the number of files searched.  Or
maybe I could just stick with searching for the raw-file based name when
appropriate.
Nick

rmillikin:
I think that solution makes sense; always creating the aggregate output file (even with 1 spectra file searched) and giving the option of a user-defined name. And if >1 files are searched, the individual file results are additionally written with their respective spectra file names.

[...]

Issue #1248, opened 2018-06-29
Require Localization for Some Discovered PTMs


trishorts:
Some ptms have masses that could be confussed with other ptms of the same or similar mass (+/- 1 Da). Rare mods may occur at a level difficult to discern from the noise. Both types would benefit from extra scrutiny. We could add a feature to the modification that requires an increase metamorpheus score for certain mods. Something like:

LR   True/False

LR = localization required

Mods like phospho and glyco which are prevalent but cant be easily localized would be left as false. I think acetyl is fine to keep as false. Oxidations and hydroxylations may require localization..

rmillikin:
@trishorts want to add the LR flag to mzLib? seems useful...

trishorts:
I like that

rmillikin:
I think this is worth investigating:
in the first-pass search, what if we searched with targets only, got rid of the q-value stuff, and only annotated mods where the score improved over unmodified? that way there's at least one fragment that helps localize the mod, and the search would be faster (probably). there would be a sensitivity decrease but the confidence in PTM localization would be better. less junk to search through in the second-pass search, too. 

we could evaluate this strategy with an open-mass search post-GPTMD and see what the mass-diff histogram looks like. if there are still large peaks then maybe that strategy wouldn't work well (i.e., it misses too many mods).

rmillikin:
I think this would create a bias towards targets in the second-pass search though, yeah? i.e., the PTM-containing targets are guaranteed to have a higher score, but the decoys aren't. Not quite sure, maybe @zrolfs can chime in. I can't quite wrap my head around it.

Maybe we can put the PTM on the best-localized position for the peptide in the first-pass search, even if the localized score is worse than the unmodified. That way we cut down on the search space in the first pass (no decoys) and second pass (only the best PTM positions are annotated) while still maintaining TD requirements?

zrolfs:
Rob is correct. When we add target PTMs that matched well to spectra (but not decoy PTMs), we are biasing our second pass search to over-report. We can (somewhat) get away with it when we limit the additions to just PTMs found above a 1% FDR, because GPTMD has a relatively high score threshold with all the notches such that a peptide needs to have a pretty good score to be identified at a 1% FDR in GPTMD. If we add every target PTM (and no decoys), then we're now making it ~N times more likely for a random sequence to match to a target than a decoy, where N is the number of notches used.

PR #2760, opened 2026-08-26
Protease Guru in MetaMorpheus

nbollis:
just click the button and digest away. Product.wxs has been updated with all dependencies of ProteaseGuru. 

image

image

### DIFF
--- a/MetaMorpheus/GUI/GUI.csproj
+++ b/MetaMorpheus/GUI/GUI.csproj
@@ -67,7 +67,7 @@
     <PackageReference Include="Newtonsoft.Json" Version="13.0.4" />
     <PackageReference Include="OxyPlot.Core" Version="2.2.0" />
     <PackageReference Include="OxyPlot.Wpf" Version="2.2.0" />
-    <PackageReference Include="ProteaseGuru" Version="0.2.3" />
+    <PackageReference Include="ProteaseGuru" Version="0.2.4" />
     <PackageReference Include="SharpLearning.CrossValidation" Version="0.31.8" />
     <PackageReference Include="SharpLearning.GradientBoost" Version="0.31.8" />
     <PackageReference Include="SharpLearning.Metrics" Version="0.31.8" />
--- a/MetaMorpheus/GUI/GUI.csproj
+++ b/MetaMorpheus/GUI/GUI.csproj
@@ -67,6 +67,7 @@
     <PackageReference Include="Newtonsoft.Json" Version="13.0.4" />
     <PackageReference Include="OxyPlot.Core" Version="2.2.0" />
     <PackageReference Include="OxyPlot.Wpf" Version="2.2.0" />
+    <PackageReference Include="ProteaseGuru" Version="0.2.3" />
     <PackageReference Include="SharpLearning.CrossValidation" Version="0.31.8" />
     <PackageReference Include="SharpLearning.GradientBoost" Version="0.31.8" />
     <PackageReference Include="SharpLearning.Metrics" Version="0.31.8" />
--- a/MetaMorpheus/GUI/GUI.csproj
+++ b/MetaMorpheus/GUI/GUI.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>WinExe</OutputType>
@@ -65,8 +65,8 @@
     <PackageReference Include="mzLib" Version="1.0.585" />
     <PackageReference Include="Nett" Version="0.15.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.4" />
-    <PackageReference Include="OxyPlot.Core" Version="2.0.0" />
-    <PackageReference Include="OxyPlot.Wpf" Version="2.0.0" />
+    <PackageReference Include="OxyPlot.Core" Version="2.2.0" />
+    <PackageReference Include="OxyPlot.Wpf" Version="2.2.0" />
     <PackageReference Include="SharpLearning.CrossValidation" Version="0.31.8" />
     <PackageReference Include="SharpLearning.GradientBoost" Version="0.31.8" />
     <PackageReference Include="SharpLearning.Metrics" Version="0.31.8" />
--- a/MetaMorpheus/GUI/MainWindow.xaml.cs
+++ b/MetaMorpheus/GUI/MainWindow.xaml.cs
@@ -1226,9 +1226,16 @@ namespace MetaMorpheusGUI
                 var selectedItem = (TabItem)MainWindowTabControl.SelectedItem;
                 var selectedItemHeader = selectedItem.Header.ToString();
 
-                if (selectedItemHeader == "Visualize")
+                switch (selectedItemHeader)
                 {
-                    MenuItem_MetaDraw_Click(sender, e);
+                    case "Visualize":
+                        MenuItem_MetaDraw_Click(sender, e);
+                        break;
+                    case "Protease Guru":
+                        {
+                            OpenProteaseGuru_Click(sender, e);
+                            break;
+                        }
                 }
             }
         }
@@ -1324,6 +1331,15 @@ namespace MetaMorpheusGUI
                 MetaDraw metaDrawGui = new MetaDraw(filesToLoad);
                 metaDrawGui.Show();
             }
+        }
+
+        private void OpenProteaseGuru_Click(object sender, RoutedEventArgs e)
+        {
+            var pgWindow = new ProteaseGuruGui.MainWindow
+            {
+                Owner = this
+            };
+            pgWindow.Show();
         }
 
         private void MenuItem_ResetDefaults_Click(object sender, RoutedEventArgs e)
--- a/MetaMorpheus/GUI/MainWindow.xaml.cs
+++ b/MetaMorpheus/GUI/MainWindow.xaml.cs
@@ -38,6 +38,7 @@ namespace MetaMorpheusGUI
         private readonly ObservableCollection<RawDataForDataGrid> SelectedSpectraFiles = new ObservableCollection<RawDataForDataGrid>();
         private readonly ObservableCollection<ProteinDbForDataGrid> SelectedProteinDatabaseFiles = new ObservableCollection<ProteinDbForDataGrid>();
         private ObservableCollection<InRunTask> InProgressTasks;
+        private ProteaseGuruGui.MainWindow? _proteaseGuruWindow;
         public static string NewestKnownMetaMorpheusVersion { get; private set; }
 
         public MainWindow()
@@ -1335,11 +1336,20 @@ namespace MetaMorpheusGUI
 
         private void OpenProteaseGuru_Click(object sender, RoutedEventArgs e)
         {
-            var pgWindow = new ProteaseGuruGui.MainWindow
+            // ProteaseGuru relies on process-wide static state (static events, GlobalVariables,
+            // NotificationService singleton, shared DispatcherTimer), so only one window is safe.
+            if (_proteaseGuruWindow is { IsLoaded: true })
             {
-                Owner = this
-            };
-            pgWindow.Show();
+                if (_proteaseGuruWindow.WindowState == WindowState.Minimized)
+                    _proteaseGuruWindow.WindowState = WindowState.Normal;
+                _proteaseGuruWindow.Activate();
+                _proteaseGuruWindow.Focus();
+                return;
+            }
+
+            _proteaseGuruWindow = new ProteaseGuruGui.MainWindow();
+            _proteaseGuruWindow.Closed += (_, _) => _proteaseGuruWindow = null;
+            _proteaseGuruWindow.Show();
         }
 
         private void MenuItem_ResetDefaults_Click(object sender, RoutedEventArgs e)
--- a/MetaMorpheus/GUI/MainWindow.xaml.cs
+++ b/MetaMorpheus/GUI/MainWindow.xaml.cs
@@ -38,7 +38,7 @@ namespace MetaMorpheusGUI
         private readonly ObservableCollection<RawDataForDataGrid> SelectedSpectraFiles = new ObservableCollection<RawDataForDataGrid>();
         private readonly ObservableCollection<ProteinDbForDataGrid> SelectedProteinDatabaseFiles = new ObservableCollection<ProteinDbForDataGrid>();
         private ObservableCollection<InRunTask> InProgressTasks;
-        private ProteaseGuruGui.MainWindow? _proteaseGuruWindow;
+        private ProteaseGuru.Gui.MainWindow? _proteaseGuruWindow;
         public static string NewestKnownMetaMorpheusVersion { get; private set; }
 
         public MainWindow()
@@ -1347,7 +1347,7 @@ namespace MetaMorpheusGUI
                 return;
             }
 
-            _proteaseGuruWindow = new ProteaseGuruGui.MainWindow();
+            _proteaseGuruWindow = new ProteaseGuru.Gui.MainWindow();
             _proteaseGuruWindow.Closed += (_, _) => _proteaseGuruWindow = null;
             _proteaseGuruWindow.Show();
         }
--- a/MetaMorpheus/GUI/MainWindow.xaml.cs
+++ b/MetaMorpheus/GUI/MainWindow.xaml.cs
@@ -1232,7 +1232,7 @@ namespace MetaMorpheusGUI
                     case "Visualize":
                         MenuItem_MetaDraw_Click(sender, e);
                         break;
-                    case "Protease Guru":
+                    case "ProteaseGuru":
                         {
                             OpenProteaseGuru_Click(sender, e);
                             break;
--- a/MetaMorpheus/GuiFunctions/GuiFunctions.csproj
+++ b/MetaMorpheus/GuiFunctions/GuiFunctions.csproj
@@ -16,10 +16,8 @@
   <ItemGroup>
     <PackageReference Include="itext7" Version="9.7.0" />
     <PackageReference Include="itext7.bouncy-castle-adapter" Version="9.7.0" />
-    <PackageReference Include="mzLib" Version="1.0.587" />
     <PackageReference Include="OxyPlot.Wpf" Version="2.2.0" />
     <PackageReference Include="mzLib" Version="1.0.588" />
-    <PackageReference Include="OxyPlot.Wpf" Version="2.2.0" />
     <PackageReference Include="Svg" Version="3.4.7" />
   </ItemGroup>
 
--- a/MetaMorpheus/GuiFunctions/GuiFunctions.csproj
+++ b/MetaMorpheus/GuiFunctions/GuiFunctions.csproj
@@ -17,7 +17,7 @@
     <PackageReference Include="itext7" Version="9.7.0" />
     <PackageReference Include="itext7.bouncy-castle-adapter" Version="9.7.0" />
     <PackageReference Include="mzLib" Version="1.0.585" />
-    <PackageReference Include="OxyPlot.Wpf" Version="2.0.0" />
+    <PackageReference Include="OxyPlot.Wpf" Version="2.2.0" />
     <PackageReference Include="Svg" Version="3.4.7" />
   </ItemGroup>
 
--- a/MetaMorpheus/GuiFunctions/MetaDraw/PlotModelStat.cs
+++ b/MetaMorpheus/GuiFunctions/MetaDraw/PlotModelStat.cs
@@ -452,6 +452,9 @@ namespace GuiFunctions
 
                 foreach (string sourceFile in psmsBySourceFile.Keys)
                 {
+                    // XAxisKey/YAxisKey are load-bearing: with a value axis Key "Primary" at Left and a
+                    // CategoryAxis Key "Category" at Bottom, omitting either key makes OxyPlot 2.2.0's
+                    // BarSeries render nothing at all, silently, with no exception. Do not remove them.
                     BarSeries column = new BarSeries
                     {
                         BarWidth = 200,
--- a/MetaMorpheus/GuiFunctions/MetaDraw/PlotModelStat.cs
+++ b/MetaMorpheus/GuiFunctions/MetaDraw/PlotModelStat.cs
@@ -137,7 +137,7 @@ namespace GuiFunctions
 
         private void histogramPlot(int plotType)
         {
-            privateModel.LegendTitle = "Source file(s)";
+            privateModel.Legends.Add(new OxyPlot.Legends.Legend { LegendTitle = "Source file(s)" });
 
             bool isGroupingEnabled = parameters.GroupingProperty != "None";
             Dictionary<string, Dictionary<string, ObservableCollection<SpectrumMatchFromTsv>>> groupedPsmsBySourceFile = null;
@@ -452,13 +452,15 @@ namespace GuiFunctions
 
                 foreach (string sourceFile in psmsBySourceFile.Keys)
                 {
-                    ColumnSeries column = new ColumnSeries
+                    BarSeries column = new BarSeries
                     {
-                        ColumnWidth = 200,
+                        BarWidth = 200,
                         IsStacked = true,
                         Title = sourceFile,
                         TrackerFormatString = "Category: {bin}\n{0}: {2}\nGroup: {group}\nTotal: {total}",
-                        BaseValue = parameters.UseLogScaleYAxis ? 0.1 : 0
+                        BaseValue = parameters.UseLogScaleYAxis ? 0.1 : 0,
+                        XAxisKey = "Primary",
+                        YAxisKey = "Category"
                     };
 
                     foreach (string groupKey in allGroupKeys)
@@ -507,13 +509,15 @@ namespace GuiFunctions
             {
                 foreach (string key in dictsBySourceFile.Keys)
                 {
-                    ColumnSeries column = new ColumnSeries
+                    BarSeries column = new BarSeries
                     {
-                        ColumnWidth = 200,
+                        BarWidth = 200,
                         IsStacked = true,
                         Title = key,
                         TrackerFormatString = "Bin: {bin}\n{0}: {2}\nTotal: {total}",
-                        BaseValue = parameters.UseLogScaleYAxis ? 0.1 : 0
+                        BaseValue = parameters.UseLogScaleYAxis ? 0.1 : 0,
+                        XAxisKey = "Primary",
+                        YAxisKey = "Category"
                     };
 
                     foreach (var d in dictsBySourceFile[key])
@@ -653,13 +657,15 @@ namespace GuiFunctions
 
                 foreach (string sourceFile in psmsBySourceFile.Keys)
                 {
-                    var column = new ColumnSeries
+                    var column = new BarSeries
                     {
-                        ColumnWidth = 200,
+                        BarWidth = 200,
                         IsStacked = true,
                         Title = sourceFile,
                         TrackerFormatString = "Bin: {bin}\n{0}: {2}\nGroup: {group}\nTotal: {total}",
-                        BaseValue = parameters.UseLogScaleYAxis ? 0.1 : 0
+                        BaseValue = parameters.UseLogScaleYAxis ? 0.1 : 0,
+                        XAxisKey = "Primary",
+                        YAxisKey = "Category"
                     };
 
                     foreach (string groupKey in allGroupKeys)
@@ -713,13 +719,15 @@ namespace GuiFunctions
             {
                 foreach (string key in dictsBySourceFile.Keys)
                 {
-                    var column = new ColumnSeries
+                    var column = new BarSeries
                     {
-                        ColumnWidth = 200,
+                        BarWidth = 200,
                         IsStacked = true,
                         Title = key,
                         TrackerFormatString = "Bin: {bin}\n{0}: {2}\nTotal: {total}",
-                        BaseValue = parameters.UseLogScaleYAxis ? 0.1 : 0
+                        BaseValue = parameters.UseLogScaleYAxis ? 0.1 : 0,
+                        XAxisKey = "Primary",
+                        YAxisKey = "Category"
                     };
 
                     foreach (var d in dictsBySourceFile[key])
@@ -768,6 +776,7 @@ namespace GuiFunctions
             var mainAxis = new CategoryAxis
             {
                 Position = AxisPosition.Bottom,
+                Key = "Category",
                 ItemsSource = category,
                 Title = isGroupingEnabled ? null : xAxisTitle,
                 GapWidth = 0.3,
@@ -1072,7 +1081,7 @@ namespace GuiFunctions
             return values;
         }
 
-        private class HistItem : ColumnItem
+        private class HistItem : BarItem
         {
             public int total { get; set; }
             public string bin { get; set; }
--- a/MetaMorpheus/GuiFunctions/MetaDraw/SpectrumMatch/PeptideSpectrumMatchPlot.cs
+++ b/MetaMorpheus/GuiFunctions/MetaDraw/SpectrumMatch/PeptideSpectrumMatchPlot.cs
@@ -50,7 +50,6 @@ namespace GuiFunctions
             width = width > 0 ? width : 700;
             height = height > 0 ? height : 300;
             // Use unique temp file names to ensure thread safety
-            string tempModelPath = System.IO.Path.Combine(System.IO.Path.GetDirectoryName(path), Path.GetRandomFileName() + "." + MetaDrawSettings.ExportType);
             string tempStationarySequencePngPath = System.IO.Path.Combine(System.IO.Path.GetDirectoryName(path), Path.GetRandomFileName() + "_annotation.png");
             string tempPtmLegendPngPath = System.IO.Path.Combine(System.IO.Path.GetDirectoryName(path), Path.GetRandomFileName() + "_legend.png");
             List<System.Drawing.Bitmap> bitmaps = new();
@@ -92,9 +91,7 @@ namespace GuiFunctions
                 encoder.Save(file);
             }
 
-            // export model as png and load both stationary and model as bitmap
-            ExportToPng(tempModelPath, (int)width, (int)height);
-            bitmaps.Add(new System.Drawing.Bitmap(tempModelPath));
+            bitmaps.Add(MetaDrawLogic.ConvertUIElementToBitmap(PlotView, System.IO.Path.GetDirectoryName(path)));
             points.Add(new Point(0, 0));
 
             var tempStatSequenceBitmap = new System.Drawing.Bitmap(tempStationarySequencePngPath);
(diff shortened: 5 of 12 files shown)


MetaMorpheus/GUI/GUI.csproj
```diff
@@ -65,8 +65,9 @@
     <PackageReference Include="mzLib" Version="1.0.585" />
     <PackageReference Include="Nett" Version="0.15.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.4" />
-    <PackageReference Include="OxyPlot.Core" Version="2.0.0" />
-    <PackageReference Include="OxyPlot.Wpf" Version="2.0.0" />
+    <PackageReference Include="OxyPlot.Core" Version="2.2.0" />
+    <PackageReference Include="OxyPlot.Wpf" Version="2.2.0" />
```
trishorts, line 69: **The OxyPlot bump leaves the two projects on different versions, and the one that isn't bumped holds most of the code.**

This moves `GUI` to OxyPlot 2.2.0, but `GuiFunctions.csproj:20` still pins:

```xml
<PackageReference Include="OxyPlot.Wpf" Version="2.0.0" />
```

OxyPlot is used in 33 files, and they are not distributed the way the diff suggests — **20 are in GuiFunctions**, 10 in Test, only 3 in GUI. So the majority of MetaMorpheus's plotting code compiles against 2.0.0 while the app resolves 2.2.0 at runtime, since NuGet unifies to the highest version in the graph.

That usually works, and it may well be fine here. But it means the MetaDraw plotting code will be running against a version it was never compiled against, and this PR reads as a ProteaseGuru integration — nobody reviewing it for that is going to be looking for a plotting-library upgrade. If ProteaseGuru 0.2.3 requires 2.2.0, I would bump `GuiFunctions` and `Test` in the same commit so the pins agree, and say so in the description. If it does not require it, the bump probably belongs in its own PR where it can be exercised against MetaDraw deliberately.

MetaMorpheus/MetaMorpheusSetup/Product.wxs
```diff
@@ -433,8 +433,32 @@
 		<Component Id="src_PredictionClients.dll" Guid="C6398A43-F845-47C2-9F35-0BB2435AFE16">
 			<File Id="src_PredictionClients.dll" Name="PredictionClients.dll" Source="$(var.GUI_TargetDir)PredictionClients.dll" />
 		</Component>
-
-      <!--Commandline Components-->
+		<Component Id="src_ProteaseGuru_Engine.dll" Guid="c8f3e773-9adb-4285-b364-aba840db1534">
+			<File Id="src_ProteaseGuru_Engine.dll" Name="Engine.dll" Source="$(var.GUI_TargetDir)Engine.dll" />
```
trishorts: **`Engine.dll` and `Tasks.dll` are very generic names to drop into the shared install directory.**

These two, plus the `ProteaseGuru*` ones below, all land flat in `$(var.GUI_TargetDir)`. `ProteaseGuruGui.dll` and `ProteaseGuruGuiFunctions.dll` carry their origin in the name; `Engine.dll` and `Tasks.dll` do not.

MetaMorpheus's own equivalents are `EngineLayer.dll` and `TaskLayer.dll`, so there is no collision today. The risk is the next dependency that ships an `Engine.dll` — the build would silently resolve one of them, and because MSI component GUIDs are keyed to the installed file path, the installer would too, in a way that is unpleasant to unpick on upgrade.

Not this PR's fault — the names come from ProteaseGuru's project layout. Worth knowing it is now MetaMorpheus's problem, and worth a comment here recording where those two files come from, since `Engine.dll` in a component list is otherwise unattributable.

trishorts (commented):
Nice to have ProteaseGuru reachable from the main window, and the tab-as-launcher matches how Visualize already works. Two things I would want settled before this goes in, one of them beyond the stated scope of the PR.

I checked the launch pattern against `MenuItem_MetaDraw_Click` rather than assuming: MetaDraw also constructs a fresh window and calls `Show()` with no single-instance guard, so re-selecting the tab spawning another window is the existing convention here, not something this PR introduces. The one difference is `Owner = this`, which MetaDraw does not set — an owned window always floats above the main window and closes with it. Probably what you want for a launcher; flagging only because it is a deliberate divergence from the sibling case.

`ProteaseGuru 0.2.3` is the newest published version on nuget.org, so the pin is current.

MetaMorpheus/GUI/GUI.csproj
nbollis: Good catch. Fixed this. Required making some changes to our plotting code.

MetaMorpheus/MetaMorpheusSetup/Product.wxs
nbollis: Agreed that it is not ideal. In the future, it would be nice to change those to proteaseguru.engine and proteaseguru.task

nbollis:
> Nice to have ProteaseGuru reachable from the main window, and the tab-as-launcher matches how Visualize already works. Two things I would want settled before this goes in, one of them beyond the stated scope of the PR.
> 
> I checked the launch pattern against `MenuItem_MetaDraw_Click` rather than assuming: MetaDraw also constructs a fresh window and calls `Show()` with no single-instance guard, so re-selecting the tab spawning another window is the existing convention here, not something this PR introduces. The one difference is `Owner = this`, which MetaDraw does not set — an owned window always floats above the main window and closes with it. Probably what you want for a launcher; flagging only because it is a deliberate divergence from the sibling case.
> 
> `ProteaseGuru 0.2.3` is the newest published version on nuget.org, so the pin is current.

Good points. I made it so that there is only one ProteaseGuru window and it can lose focus to MetaMorpehus

trishorts:
Following up on my two comments above, both of which you have resolved. I re-checked them against the head of the branch rather than taking the resolution on faith, and both are genuinely closed.

**The OxyPlot pin.** `GUI.csproj:68-69` and `GuiFunctions.csproj:20` are all on 2.2.0 now, so the split is gone and the 20 GuiFunctions files that hold most of the plotting code compile against the version the app actually resolves.

**The `ColumnSeries` to `BarSeries` migration this forced.** I went looking for a regression here and did not find one, so I am recording the check rather than the suspicion. `OxyPlot.Series.ColumnSeries` does not exist in 2.2.0 at all — reflecting over the packaged `OxyPlot.dll`, the only bar types are `BarSeries`, `IntervalBarSeries`, `TornadoBarSeries`, `LinearBarSeries`, `RectangleBarSeries` and `ErrorBarSeries` — so the swap was forced rather than stylistic. And `BarSeries` being the horizontal one is handled correctly by the `XAxisKey = "Primary"` / `YAxisKey = "Category"` binding: with the value axis at `Left` and the `CategoryAxis` at `Bottom`, OxyPlot classifies the series as transposed, which is what keeps the histogram bars vertical. I rebuilt the exact axis configuration from `ConfigureHistogramAxes` against the 2.2.0 assembly and `PlotElementExtensions.IsTransposed` returns `True` for it. So this is the documented replacement, not a plot that happens to still pass its assertions.

Worth knowing that the updated `MetaDrawTest` assertions would not have distinguished the two cases either way — they read `Items[i].Value` off the series, which is populated during `UpdateData` and is orientation-independent. I am not asking for a rendering test; just flagging that the green suite is not the evidence here, and the check above is.

On `Engine.dll` and `Tasks.dll`: agreed, and `proteaseguru.engine` / `proteaseguru.task` is the right eventual shape. Since that has to happen in the ProteaseGuru repo and cannot happen here, an issue over there would keep it from evaporating — and note it is not purely their problem, because MSI component GUIDs are keyed to the installed file path, so renaming those assemblies later is a change to this installer too.

MetaMorpheus/GuiFunctions/MetaDraw/PlotModelStat.cs
```diff
@@ -452,13 +452,15 @@ private HistogramRawData GetHistogramData(int plotType)
...
+                    BarSeries column = new BarSeries
                     {
-                        ColumnWidth = 200,
+                        BarWidth = 200,
                         IsStacked = true,
                         Title = sourceFile,
                         TrackerFormatString = "Category: {bin}\n{0}: {2}\nGroup: {group}\nTotal: {total}",
-                        BaseValue = parameters.UseLogScaleYAxis ? 0.1 : 0
+                        BaseValue = parameters.UseLogScaleYAxis ? 0.1 : 0,
+                        XAxisKey = "Primary",
```
trishorts, line 465: These two keys are load-bearing, and worth a comment saying so — they read like decoration and they are not.

`ColumnSeries` really is gone in OxyPlot.Core 2.2.0 (I reflected the assembly: `BarSeries`, `BarItem`, `BarSeriesBase\`1`, `LinearBarSeries`, `RectangleBarSeries` — no `ColumnSeries`, no `ColumnItem`), so the rewrite was forced rather than optional. I built a standalone OxyPlot 2.2.0 probe replicating `ConfigureHistogramAxes` exactly — `CategoryAxis` Key `Category` at `Bottom`, value axis Key `Primary` at `Left` — and exported to SVG:

```
=== as written: XAxisKey=Primary, YAxisKey=Category ===
  resolved XAxis: key=Primary  pos=Left   type=LinearAxis
  resolved YAxis: key=Category pos=Bottom type=CategoryAxis
    rect x=45.8  y=304.2 w=145 h=66.7   -> vertical column
    rect x=234.3 y=37.4  w=145 h=333.5  -> vertical column
    rect x=422.7 y=170.8 w=145 h=200.1  -> vertical column

=== with the keys omitted (the naive ColumnSeries -> BarSeries rename) ===
  resolved XAxis: key=Category pos=Bottom type=CategoryAxis
  resolved YAxis: key=Primary  pos=Left   type=LinearAxis
    no bar rects emitted
```

Heights 66.7 / 333.5 / 200.1 for values 1 / 5 / 3, so as written this is correct, and all four `BarSeries` sites got both keys — I checked, 4 constructions, 4 `XAxisKey`, 4 `YAxisKey`.

The reason it deserves a comment: dropping either key produces **no bars and no exception** — a silently blank histogram. And the tests can't see that. `MetaDrawTest` pulls `series.Items` by reflection and asserts on counts and values, which are populated whether or not anything renders, so the suite cannot distinguish a correct histogram from an empty one. A one-line note here is the cheapest guard against someone "tidying" these away later.

trishorts (commented):
Both of my earlier threads are properly settled — GuiFunctions is on 2.2.0 now, and I'm content to leave the `Engine.dll`/`Tasks.dll` naming to a future ProteaseGuru rename given Validate Installer Contents is green. The single-window guard in `OpenProteaseGuru_Click`, and the comment explaining that ProteaseGuru's process-wide static state is what requires it, is exactly right.

Three things from this round, none of which I'd call a blocker on their own, though the second and third are related and I'd want the third written down somewhere before merge:

- The `ColumnSeries` → `BarSeries` migration is **correct**, and I verified it rather than assuming: a standalone OxyPlot 2.2.0 probe of your exact axis configuration renders vertical columns with heights proportional to the values. Dropping the `XAxisKey`/`YAxisKey` pair — which is what a naive rename would have done — emits no bars at all, silently, with no exception. Worth a comment in the source, because no test can see the difference.
- The four assertions `5d8b880b` deleted from `ExportPlot_RefragmentationWithAdditionalFragmentIons_WritesExpectedImage` were not bad assertions. `ConvertUIElementToBitmap` falls back to a 200x100 bitmap when `RenderSize` is zero, which is every headless test run, so the exported PNG no longer contains the spectrum. No behaviour change in the app — the callers already pass `plotView.ActualWidth`/`ActualHeight` — but it costs the only pixel-level check of MetaDraw's image export, in the same PR that rewrote the plotting stack.
- The version skew isn't fully closed. mzLib 1.0.587 declares `OxyPlot.Wpf 2.0.0` and `mzPlot.dll` references `ColumnSeries`/`ColumnItem`, which 2.2.0 does not have. Nothing in MetaMorpheus reaches those paths today, so it's latent rather than broken — but it's also the reason `ExportToPng` had to be abandoned, and it wants an mzLib issue.

MetaMorpheus/Test/MetaDraw/MetaDrawTest.cs
```diff
@@ -2341,45 +2341,6 @@ public void ExportPlot_RefragmentationWithAdditionalFragmentIons_WritesExpectedI
...
-                refragmentHasC = HasColor(bmpRefragment, cColor);
-                refragmentHasZDot = HasColor(bmpRefragment, zDotColor);
-            }
-
-            Assert.That(noRefragmentHasC || noRefragmentHasZDot, Is.False,
-                "No c or zDot ion colors should be present in the PNG without refragmentation.");
-            Assert.That(refragmentHasC || refragmentHasZDot, Is.True,
-                "c or zDot ion colors should be present in the PNG with refragmentation.");
-
             Assert.That(errorsNoRefragment, Is.Null, "No errors should be reported for no refragmentation.");
```
trishorts, line 2350: The four assertions `5d8b880b` removed above here weren't bad — the export path moved underneath them, and this is the one test in the file that looked at what was actually drawn.

They checked that the refragmented PNG is larger than the non-refragmented one, and that the c and zDot ion colours are present in the refragmented image and absent from the other. That is the only assertion anywhere that refragmentation reaches the exported picture.

Why they stopped holding: `ExportToPng` rendered the `PlotModel` headlessly at the requested size. `MetaDrawLogic.ConvertUIElementToBitmap` renders the live visual instead, and opens with

```csharp
int width  = (int)visual.RenderSize.Width  == 0 ? 200 : (int)visual.RenderSize.Width;
int height = (int)visual.RenderSize.Height == 0 ? 100 : (int)visual.RenderSize.Height;
```

In a headless NUnit run the `PlotView` was never realised, so `RenderSize` is 0 and every export collapses to a 200x100 bitmap of an unlaid-out control. No ion colours, no size difference — hence the failures.

In the app this is a no-op, and I want to be clear about that: `MetaDrawLogic.cs:558,563` already pass `plotView.ActualWidth`/`ActualHeight`, so the rendered size is the same either way. The loss is entirely in the test, and it lands in the same PR that rewrote the plotting stack, which is the worst possible moment to give up the only pixel-level check.

What's left — the file exists, no errors were reported — would pass against a blank 200x100 bitmap.

`OxyPlot.Wpf.PngExporter` still ships in 2.2.0; the only breaking change is that `Export(model, path, width, height, OxyColor background, double resolution)` lost the `OxyColor` parameter. So MetaMorpheus can call it directly with the 2.2 signature to keep a headless model-render path for tests, and these assertions can come back as they were.

MetaMorpheus/GuiFunctions/GuiFunctions.csproj
```diff
@@ -17,7 +17,7 @@
     <PackageReference Include="itext7" Version="9.7.0" />
     <PackageReference Include="itext7.bouncy-castle-adapter" Version="9.7.0" />
     <PackageReference Include="mzLib" Version="1.0.587" />
-    <PackageReference Include="OxyPlot.Wpf" Version="2.0.0" />
+    <PackageReference Include="OxyPlot.Wpf" Version="2.2.0" />
```
trishorts, line 19: This closes the GUI/GuiFunctions split I raised, thanks — but there is a third consumer of OxyPlot in this graph that neither of us named, and it can't be fixed from inside this PR.

`mzLib` 1.0.587's nuspec, one line above, declares:

```xml
<dependency id="OxyPlot.Wpf" version="2.0.0" />
```

and `mzPlot.dll` in that package carries type references to `OxyPlot.Series.ColumnSeries` and `OxyPlot.Series.ColumnItem` — neither of which exists in OxyPlot.Core 2.2.0. NuGet unifies the graph up to 2.2.0 without a warning, so mzPlot now runs against an OxyPlot it was neither compiled nor tested against.

Latent today rather than broken: those two types almost certainly live behind mzPlot's `AddBarPlot` / `HistogramPlot`, and nothing in MetaMorpheus calls either (I grepped `AddBarPlot`, `HistogramPlot`, `AddLinePlot`, `AddScatterPlot`, `AddSpectrumPlot` across the branch — no hits outside a test method's own name). The first caller of those gets a `TypeLoadException`, and there is nothing in the tree recording why they can't be used.

It is also the reason `ExportToPng` had to go in this PR. That method is mzPlot's, not OxyPlot's, and it is compiled against 2.0's `PngExporter.Export(model, path, width, height, OxyColor, double)`; 2.2 dropped the `OxyColor` parameter, so calling it would throw `MissingMethodException`. That single signature change is what cascaded into the export rewrite and the deleted assertions on `MetaDrawTest.cs`.

For clarity, the rest of the graph is consistent — `ProteaseGuru` 0.2.3 pins `OxyPlot.Core`/`OxyPlot.Wpf`/`OxyPlot.SkiaSharp` 2.2.0, which is what forces this bump in the first place, since two OxyPlot versions can't coexist in one process.

The real fix is an mzLib PR moving mzPlot to OxyPlot 2.2.0, which has to migrate `AddBarPlot`/`HistogramPlot` off `ColumnSeries` the same way this PR just migrated `PlotModelStat`. I don't think that needs to block this, but it should exist as an issue before this merges, otherwise the constraint lives only in this thread.

trishorts: Filed as smith-chem-wisc/mzLib#1248, so the constraint no longer lives only in this thread.

I bumped `mzPlot.csproj` to 2.2.0 on mzLib master and built it, so the issue carries the exact list rather than a guess: five errors in one file, from three breaking changes — `ColumnSeries`/`ColumnItem` removed, `PngExporter.Background` removed, and `PlotModel.LegendBorderThickness`/`LegendBorder` moved onto `Model.Legends`. It also records the `XAxisKey`/`YAxisKey` trap you already hit here, since `AddBarPlot` has no test and a silently empty chart is easy to miss.

I don't think it blocks this PR.

MetaMorpheus/GuiFunctions/MetaDraw/PlotModelStat.cs
nbollis: Added the load-bearing note above the first `BarSeries` construction in `PlotModelStat.cs` (`GetHistogramData`, around line 455): explicitly documents that omitting either `XAxisKey`/`YAxisKey` makes OxyPlot 2.2.0 render no bars at all, silently, with no exception. All four `BarSeries` sites keep both keys.

MetaMorpheus/Test/MetaDraw/MetaDrawTest.cs
nbollis: Restored all four pixel-level assertions in `ExportPlot_RefragmentationWithAdditionalFragmentIons_WritesExpectedIons`, exactly as you suggested: after each export the test now renders `plotView.Model` headlessly via `OxyPlot.Wpf.PngExporter.Export(model, path, width, height)` (the 2.2 signature — the `OxyColor` overload no longer exists), and asserts the refragmented render is larger and contains c/zDot ion colours while the non-refragmented one does not. The size/count checks and `HasColor` pixel scan are back verbatim. Green: `ExportPlot_...WritesExpectedIons` passes, and the full MetaDraw suite is 57/57.

MetaMorpheus/GUI/GUI.csproj
nbollis: Closed — `GuiFunctions.csproj:20` and `GUI.csproj:68-69` all pin `OxyPlot.Core`/`OxyPlot.Wpf` 2.2.0 now, so the split is gone and there is no version skew between the projects.

MetaMorpheus/GuiFunctions/GuiFunctions.csproj
nbollis: Tracked — this now lives as mzLib issue #1248 ("mzPlot is pinned to OxyPlot 2.0.0..."), which documents the five compile sites, the `ColumnSeries`→`BarSeries` axis-key trap, and the suggested migration scope including a test for `AddBarPlot`. Nothing in MetaMorpheus calls those paths today, so it stays latent until that issue is picked up.

MetaMorpheus/MetaMorpheusSetup/Product.wxs
nbollis: Resolved properly — ProteaseGuru 0.2.4 has now renamed these assemblies: `Engine.dll` → `ProteaseGuru.Engine.dll` and `Tasks.dll` → `ProteaseGuru.Tasks.dll` (with `ProteaseGuruGui`/`ProteaseGuruGuiFunctions` → `ProteaseGuru.Gui`/`ProteaseGuru.GuiFunctions` following the same scheme), and this PR's `Product.wxs` now installs the renamed files, with fresh component GUIDs since the installed file paths changed. No generic `Engine.dll`/`Tasks.dll` land in the shared install directory anymore, and the MSI was rebuilt and verified to contain only the `ProteaseGuru.*` names.

MetaMorpheus/GuiFunctions/MetaDraw/SpectrumMatch/PeptideSpectrumMatchPlot.cs
```diff
@@ -92,9 +91,7 @@ public PeptideSpectrumMatchPlot(OxyPlot.Wpf.PlotView plotView, SpectrumMatchFrom
                 encoder.Save(file);
             }
 
-            // export model as png and load both stationary and model as bitmap
-            ExportToPng(tempModelPath, (int)width, (int)height);
-            bitmaps.Add(new System.Drawing.Bitmap(tempModelPath));
+            bitmaps.Add(MetaDrawLogic.ConvertUIElementToBitmap(PlotView, System.IO.Path.GetDirectoryName(path)));
```
pcruzparri, line 94: **High** — Child-scan exports lose the spectrum

**Problem:**
`ExportToPng` couldn't stay — it sets `PngExporter.Background`, which 2.2.0 removed, so under the new pin it would throw `MissingMethodException` on every spectrum-match export. Replacing it was right. The replacement covers realised views but not detached ones.

`ConvertUIElementToBitmap` sizes from `visual.RenderSize` and falls back to 200x100 when that's zero (`MetaDrawLogic.cs:928-929`). Fine for the main spectrum, whose `PlotView` is MetaDraw's real control. Child scans are built on a placeholder that's deliberately detached (`MetaDrawLogic.cs:195-209`):

```csharp
PlotView childPlotView = new PlotView(); // placeholder
var childPlot = new PeptideSpectrumMatchPlot(childPlotView, sm, childScan, matchedIons, annotateProperties: false);
...
// remove model from placeholder (the model can only be referenced by 1 plotview at a time)
childPlotView.Model = null;
CurrentlyDisplayedPlots.Add(childPlot);
```

That placeholder is never parented, because the Parent/Child Scan View builds its own control and binds only the model — `MetaDraw.xaml:400`, `<oxy:PlotView x:Name="ChildPlotView" Model="{Binding Plot.Model}" />`. So `childPlot.PlotView` stays unmeasured with a null `Model`, while `childPlot.Model` — what's actually on screen — lives on the plot object.

`ExportPlot` iterates `CurrentlyDisplayedPlots` (`MetaDrawLogic.cs:539`) and dispatches child plots through this method (`:557`); `skipPlotRegeneration` needs `Count == 1` (`:488-491`), which is never true when children exist.

**Impact:**
Every child scan exports without its spectrum. Not a blank file — the stationary sequence is still composited (`PeptideSpectrumMatchPlot.cs:97-103`) and `CombineBitmap` sizes the composite as the max over its inputs (`MetaDrawLogic.cs:706-710`), saved unrescaled by the Png branch (`MassSpectrumPlot.cs:219-220`). So the file keeps the sequence annotation, the spectrum region is the empty 200x100 fallback, and the output dimensions come from the sequence bitmap.

Ordinary data, not an edge case: child plots are built whenever `sm.ChildScanMatchedIons != null` (`MetaDrawLogic.cs:135`) — the normal shape of crosslink and glyco results. It fails silently, with files appearing under the right names and nothing reported.

**Suggested fix:**
Render from `Model` when the view isn't realised and from the visual otherwise. `PngExporter` is still an instance class in 2.2.0 with `Width`/`Height` — only `Background` is gone — and the 4-arg `Export(model, path, width, height)` is what `MetaDrawTest.cs:2340` already calls.

Same call shape at `CrosslinkSpectrumMatchPlot.cs:93` and `ChimeraSpectrumMatchPlot.cs:67`. Only the peptide path is reached by parent/child export today, but all three assume a realised `PlotView`.

**Affected test scenarios:**
Export a PSM with child scans — the crosslink PSM at `Test/XLTest.cs:1508-1510`, or the glyco PSM at `MetaDrawTest.cs:1166` — and assert the child image contains the spectrum; a dimension check against the 200x100 fallback would do it. Nothing covers child-scan export output today. Worth knowing while you're in there: the refragmentation test's pixel assertions were re-pointed at `PngExporter.Export(plotView.Model, ...)` (`MetaDrawTest.cs:2340`, `:2348`), which nothing in GuiFunctions or GUI calls, so they no longer check that the export pipeline draws anything. Once this is fixed they can point back at the real output.

MetaMorpheus/GuiFunctions/MetaDraw/PlotModelStat.cs
```diff
@@ -137,7 +137,7 @@ private void createPlot(string plotType)
 
         private void histogramPlot(int plotType)
         {
-            privateModel.LegendTitle = "Source file(s)";
+            privateModel.Legends.Add(new OxyPlot.Legends.Legend { LegendTitle = "Source file(s)" });
```
pcruzparri, line 140: **Medium** — The two scatter plots lose their legend

**Problem:**
2.2.0 moved legend configuration off `PlotModel` and into `Model.Legends`, which is why this line had to change. That fix landed in `histogramPlot` only.

`linePlot` (`PlotModelStat.cs:881`) never adds a `Legend`, and it drives the other two plots in this class — **Precursor PPM Error vs. RT** (`linePlot(1)`) and **Predicted RT vs. Observed RT** (`linePlot(3)`). Under 2.0.0 a legend was drawn from series titles with no legend object present; with an empty `Legends` collection under 2.2.0, nothing is drawn.

The tell is that `linePlot` adds two `ScatterSeries` carrying no points at all, whose only purpose is to produce legend entries:

```csharp
privateModel.Series.Add(new ScatterSeries { Title = "non-variant PSMs", MarkerFill = OxyColors.Blue });   // :961
privateModel.Series.Add(new ScatterSeries { Title = "variant PSMs", MarkerFill = OxyColors.DarkRed });    // :972
```

Those two now render nothing.

**Impact:**
Both scatter plots lose the key that tells a user which colour is variant and which is non-variant. The plots still draw, so the loss is silent. Tests don't catch it — the `PlotModelStat` tests read `Series` items by reflection, which are populated either way.

**Suggested fix:**
Add a bare `new Legend()` in `linePlot`. If you'd rather hoist it into the constructor next to `privateModel = new PlotModel { ... }` (`:57`) so every plot type gets one, hoist a titleless `Legend` and set `LegendTitle` inside `histogramPlot` — hoisting this line as written would put "Source file(s)" on the scatter plots, where the entries are variant/non-variant PSMs.

**Affected test scenarios:**
Construct `PlotModelStat("Precursor PPM Error vs. RT", ...)` and assert `Model.Legends` is non-empty. The same assertion would have caught the `histogramPlot` break.

pcruzparri (commented):
Code review — 2 findings, 1 at severity >= High, posted as inline comments.

acesnik:
This is cool. Should the places where we cite papers be added if people use this? Or somehow mention what to cite for ProteaseGuru?

MetaMorpheus/GUI/MainWindow.xaml.cs
```diff
@@ -1338,6 +1346,24 @@ private void MenuItem_MetaDraw_Click(object sender, RoutedEventArgs e)
             }
         }
 
+        private void OpenProteaseGuru_Click(object sender, RoutedEventArgs e)
```
trishorts, line 1349: **Note, not a blocker — this is the only way to open ProteaseGuru, and it can't be re-triggered from the tab you land on.**

`OpenProteaseGuru_Click` has exactly one caller: the `TabControl_SelectionChanged` switch at line 1249. There's no button or menu item bound to it anywhere in `MainWindow.xaml` (I grepped), despite the `_Click` name and the PR description's "just click the button".

`SelectionChanged` only fires when the selection actually changes, so:

1. Click the ProteaseGuru tab → window opens, tab stays selected.
2. Close the ProteaseGuru window → the `Closed` handler nulls `_proteaseGuruWindow`.
3. Click the ProteaseGuru tab again → already selected, no event, nothing happens.

The user is left on the description page with no way back in until they select a different tab first.

I checked before raising it: the Visualize tab has exactly the same shape — `MenuItem_MetaDraw_Click` is also reachable only from this switch, with no menu entry — so this is a pre-existing pattern you're extending faithfully, not something you introduced. That's why it's a note rather than a finding. But since this PR adds a fresh instance of the pattern, and the tab body is already a mostly-empty description page, a small "Open ProteaseGuru" `Button` in that grid would make the entry point idempotent and cost about three lines.

One smaller thing in the same method: the window is `Show()`n without `Owner = this`, so it won't minimize or restore with MetaMorpheus and can end up buried behind it. Setting `Owner` is most of what the `WindowState`/`Activate`/`Focus` block above is compensating for. (MetaDraw doesn't set it either, so again — pre-existing.)

MetaMorpheus/MetaMorpheus.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 17
-VisualStudioVersion = 17.1.32228.430
+# Visual Studio Version 18
```
trishorts, line 3: **Minor — unrelated VS-version churn, and it pins an *insiders* build string.**

```
-# Visual Studio Version 17
-VisualStudioVersion = 17.1.32228.430
+# Visual Studio Version 18
+VisualStudioVersion = 18.8.11925.187 insiders
```

Nothing in this PR needs it, and `18.8.11925.187 insiders` records a preview VS build as the solution's version for everyone who opens it afterwards.

Worth flagging because you already made the opposite call on the sibling PR — mzLib #1277's description says *"Exclude the unrelated Visual Studio 17->18 solution-file change that was already on the branch."* Same change, same author, opposite treatment. Whichever way you want it, it'd be good for the two repos to agree.

MetaMorpheus/MetaMorpheusSetup/Product.wxs
```diff
@@ -528,10 +552,13 @@
         <File Id="src_libiompstubs5md.dll" Name="libiompstubs5md.dll" Source="$(var.GUI_TargetDir)runtimes\win-x64\native\libiompstubs5md.dll" />
       </Component>
       <!--Native backing library for the managed SkiaSharp.dll, which ships but has no native pair.-->
-      <Component Id="src_libSkiaSharp.dll" Guid="08B9E210-A08C-457F-9618-7D7DBEFE233A">
-        <File Id="src_libSkiaSharp.dll" Name="libSkiaSharp.dll" Source="$(var.GUI_TargetDir)runtimes\win-x64\native\libSkiaSharp.dll" />
-      </Component>
-    </ComponentGroup>
+       <Component Id="src_libSkiaSharp.dll" Guid="08B9E210-A08C-457F-9618-7D7DBEFE233A">
```
trishorts, line 555: **Minor, non-blocking — whitespace-only reindent of an existing component.**

`src_libSkiaSharp.dll` is byte-identical apart from indentation (same `Guid`, same `Source`), but it shows up as a delete + add. In `Product.wxs` specifically that costs something: this is the file where a reviewer has to eyeball every component and GUID, and a reindented block reads at a glance like a component that changed.

The new blocks just above (`src_HarfBuzzSharp.dll` through `src_SkiaSharp.HarfBuzz.dll`, ~line 454) are also indented with a leading space before their tabs, which is what pulled the surrounding lines out of alignment.

Not worth another round on its own — just worth reverting the untouched block's indentation next time you push.

For what it's worth, the substance here is right: adding `libHarfBuzzSharp.dll` as the native pair for the managed `HarfBuzzSharp.dll` is the correct companion to the OxyPlot 2.2.0 SkiaSharp dependency, and it mirrors the existing `libSkiaSharp.dll` comment above it.

trishorts (commented):
Re-reviewed at `3533243`. My five earlier threads are all properly addressed and resolved — the OxyPlot pins now agree across `GUI`/`GuiFunctions`, the ProteaseGuru assemblies are renamed out of the generic `Engine.dll`/`Tasks.dll` namespace with fresh component GUIDs, the load-bearing axis-key comment is in, and the pixel assertions are restored.

**No blocking findings from me this round.** I chased two hypotheses and both were wrong, so recording them so nobody re-treads: (1) I thought swapping `ExportToPng` for `ConvertUIElementToBitmap` would make raster exports ignore the requested width/height — it doesn't for the main plot, since `MetaDrawLogic.cs:557` passes `plotView.ActualWidth/ActualHeight` and that's the same control the bitmap is sized from; (2) I thought the `CanvasPdfExportDpi` scaling would leave the spectrum bitmap 6.25x the sequence bitmap — it doesn't, `ConvertUIElementToBitmap` downscales back to logical size at `MetaDrawLogic.cs:947`, matching what the sequence path does.

What I'm leaving below is three minors. Separately: **@pcruzparri's two open threads are both still unaddressed at this commit** — `linePlot` (`PlotModelStat.cs:881`) still adds no `Legend` while `histogramPlot:140` does, and `PeptideSpectrumMatchPlot.cs:94` still renders through `ConvertUIElementToBitmap(PlotView, ...)`. I verified both and left them open.

MetaMorpheus/GuiFunctions/GuiFunctions.csproj
```diff
@@ -16,8 +16,10 @@
   <ItemGroup>
     <PackageReference Include="itext7" Version="9.7.0" />
     <PackageReference Include="itext7.bouncy-castle-adapter" Version="9.7.0" />
+    <PackageReference Include="OxyPlot.Wpf" Version="2.2.0" />
+    <PackageReference Include="mzLib" Version="1.0.588" />
```
trishorts, line 20: **This is what all nine red checks are, and it is a bad merge resolution rather than a stale pin.**

`GuiFunctions.csproj` now declares both packages twice:

```xml
19:  <PackageReference Include="OxyPlot.Wpf" Version="2.2.0" />
20:  <PackageReference Include="mzLib" Version="1.0.588" />
21:  <PackageReference Include="mzLib" Version="1.0.589" />
22:  <PackageReference Include="OxyPlot.Wpf" Version="2.2.0" />
```

NuGet resolves the duplicate `mzLib` to the lower version and `EngineLayer` still asks for 1.0.589, so every build fails identically:

```
GuiFunctions.csproj : error NU1605: Warning As Error: Detected package downgrade: mzLib from 1.0.589 to 1.0.588.
  GuiFunctions -> EngineLayer -> mzLib (>= 1.0.589)
  GuiFunctions -> mzLib (>= 1.0.588)
```

That is a build break, not a test failure, which is why `Test on Windows` fails in 4 seconds and why the installer, `Search with Built`, all three OS jobs and even the non-blocking external-service job go red together. `Test on Windows` is the only required check on master.

Deleting lines 20 and 22 should clear all nine. `GUI.csproj` has no duplicates, so this is the only file affected.

The six master merges since 2026-09-03 have not fixed it, so it is worth checking how the resolution is being done -- a merge that keeps both sides of a `PackageReference` line will keep reappearing on the next one.

trishorts (commented):
Nothing new to review since my 2026-09-06 pass -- the six commits since are all master merges, and none of my three threads or @pcruzparri's two have been addressed. Flagging the CI state, since it has a single cause and it is a one-line fix.

PR #2761, opened 2026-08-26
Stamp the per-scan scoring table and counting-sort the coarse scores

trishorts:
> **Project of origin:** `MetaMorpheus_index_perf`

🤖 Split out of #2758, which is being retired — see the discussion there and on #2742. This half is independent of the fragment-index layout question those two PRs are about, and does not touch the index at all.

## What

Two things the per-scan scoring loop was paying for, in `ModernSearchEngine` and the three engines that inherit from it.

**`Array.Clear` over the whole peptide index between scans.** O(peptides) of pure bookkeeping per scan, whatever the scan actually touched. `ScanScoringTable` instead stamps each cell with the scan it belongs to and reads a stale stamp as zero, so starting a scan is O(1) apart from a clear every 255 scans.

That is not unconditionally better, so it is selected rather than assumed. A stamped cell is two bytes wide, so a scan touching most of the index pays double the memory traffic to skip a memset it was already amortizing. On a synthetic version of the loop, stamping runs ~8-13x faster than clearing when a scan touches ~0.1% of the index — a normal ppm-tolerance search — and ~0.6-0.8x as fast when it touches most of it. So `UseStampedScoringTable` is `massDiffAcceptor is not OpenSearchMode`: open searches keep clearing, which is what crosslink and glyco get.

**`OrderByDescending` on the coarse scores, once per scan, in three engines** — allocating a LINQ ordering and a key array every time. Coarse scores are bytes, so `DescendingScoreSorter` counting-sorts them in two linear passes with no per-scan allocation.

It is stable on purpose. LINQ's ordering is stable, and ties here decide which peptide becomes the PSM, so an unstable sort would change reported results.

## Verification

Full suite on this branch: **1830 passed, 1 skipped, 0 failed**.

Results are unchanged by construction — both changes are representation, not scoring — and the suite includes the search tests that compare PSM counts and scores.

## Note for reviewers

This will conflict with #2742, which edits the same four engine files. The conflicts are narrow (the `scoringTable` parameter type on three signatures) and mechanical either way round.

### DIFF
--- a/MetaMorpheus/EngineLayer/CrosslinkSearch/CrosslinkSearchEngine.cs
+++ b/MetaMorpheus/EngineLayer/CrosslinkSearch/CrosslinkSearchEngine.cs
@@ -1,4 +1,5 @@
 ﻿using EngineLayer.ModernSearch;
+using EngineLayer.Util;
 using MassSpectrometry;
 using MzLibUtil;
 using Omics.Fragmentation;
@@ -115,10 +116,11 @@ namespace EngineLayer.CrosslinkSearch
             int[] threads = Enumerable.Range(0, maxThreadsPerFile).ToArray();
             Parallel.ForEach(threads, (scanIndex) =>
             {
-                byte[] scoringTable = new byte[PeptideIndex.Count];
+                var scoringTable = new ScanScoringTable(PeptideIndex.Count, UseStampedScoringTable);
                 List<int> idsOfPeptidesPossiblyObserved = new List<int>();
-                byte[] secondScoringTable = new byte[PeptideIndex.Count];
+                var secondScoringTable = new ScanScoringTable(PeptideIndex.Count, UseStampedScoringTable);
                 List<int> childIdsOfPeptidesPossiblyObserved = new List<int>();
+                var scoreSorter = new DescendingScoreSorter();
 
                 byte scoreAtTopN = 0;
                 int peptideCount = 0;
@@ -129,7 +131,7 @@ namespace EngineLayer.CrosslinkSearch
                     if (GlobalVariables.StopLoops) { return; }
 
                     // empty the scoring table to score the new scan (conserves memory compared to allocating a new array)
-                    Array.Clear(scoringTable, 0, scoringTable.Length);
+                    scoringTable.BeginScan();
                     idsOfPeptidesPossiblyObserved.Clear();      
 
                     var scan = ListOfSortedMs2Scans[scanIndex];
@@ -146,7 +148,7 @@ namespace EngineLayer.CrosslinkSearch
                     //child scan first - pass scoring
                     if (scan.ChildScans != null && CommonParameters.MS2ChildScanDissociationType != DissociationType.Unknown && CommonParameters.MS2ChildScanDissociationType != DissociationType.LowCID)
                     {
-                        Array.Clear(secondScoringTable, 0, secondScoringTable.Length);
+                        secondScoringTable.BeginScan();
                         childIdsOfPeptidesPossiblyObserved.Clear();
 
                         List<int> childBinsToSearch = new List<int>();
@@ -165,7 +167,7 @@ namespace EngineLayer.CrosslinkSearch
                             {
                                 idsOfPeptidesPossiblyObserved.Add(childId);
                             }
-                            scoringTable[childId] = (byte)(scoringTable[childId] + secondScoringTable[childId]);
+                            scoringTable.Set(childId, (byte)(scoringTable[childId] + secondScoringTable[childId]));
                         }
                     }
 
@@ -175,7 +177,7 @@ namespace EngineLayer.CrosslinkSearch
                         scoreAtTopN = 0;
                         peptideCount = 0;
 
-                        foreach (int id in idsOfPeptidesPossiblyObserved.OrderByDescending(p => scoringTable[p]))
+                        foreach (int id in scoreSorter.Sort(idsOfPeptidesPossiblyObserved, scoringTable))
                         {
                             peptideCount++;
                             // Whenever the count exceeds the TopN that we want to keep, we removed everything with a score lower than the score of the TopN-th peptide in the ids list
--- a/MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
+++ b/MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using EngineLayer.Util;
 using MassSpectrometry;
 using EngineLayer.SpectrumMatch;
 
@@ -120,11 +121,12 @@ namespace EngineLayer.GlycoSearch
             int[] threads = Enumerable.Range(0, maxThreadsPerFile).ToArray(); // We can do the parallel search on different threads
             Parallel.ForEach(threads, (scanIndex) =>
             {
-                byte[] scoringTable = new byte[PeptideIndex.Count];
+                var scoringTable = new ScanScoringTable(PeptideIndex.Count, UseStampedScoringTable);
                 List<int> idsOfPeptidesPossiblyObserved = new List<int>();
 
-                byte[] secondScoringTable = new byte[PeptideIndex.Count]; // We didn't use that right now.
+                var secondScoringTable = new ScanScoringTable(PeptideIndex.Count, UseStampedScoringTable); // We didn't use that right now.
                 List<int> childIdsOfPeptidesPossiblyObserved = new List<int>();
+                var scoreSorter = new DescendingScoreSorter();
 
                 List<int> idsOfPeptidesTopN = new List<int>();
                 byte scoreAtTopN = 0;
@@ -136,7 +138,7 @@ namespace EngineLayer.GlycoSearch
                     if (GlobalVariables.StopLoops) { return; }
 
                     // empty the scoring table to score the new scan (conserves memory compared to allocating a new array)
-                    Array.Clear(scoringTable, 0, scoringTable.Length);
+                    scoringTable.BeginScan();
                     idsOfPeptidesPossiblyObserved.Clear();
                     idsOfPeptidesTopN.Clear();
 
@@ -183,7 +185,7 @@ namespace EngineLayer.GlycoSearch
                     {
                         scoreAtTopN = 0;
                         peptideCount = 0;
-                        foreach (int id in idsOfPeptidesPossiblyObserved.OrderByDescending(p => scoringTable[p])) //from the higest score to the lowest score
+                        foreach (int id in scoreSorter.Sort(idsOfPeptidesPossiblyObserved, scoringTable)) //from the higest score to the lowest score
                         {
                             if (scoringTable[id] < (int)byteScoreCutoff) //if the score is lower than the cutoff, we can skip this peptide.
                             {
--- a/MetaMorpheus/EngineLayer/ModernSearch/ModernSearchEngine.cs
+++ b/MetaMorpheus/EngineLayer/ModernSearch/ModernSearchEngine.cs
@@ -1,4 +1,5 @@
 ﻿using Chemistry;
+using EngineLayer.Util;
 using MassSpectrometry;
 using Omics.Fragmentation;
 using Proteomics.ProteolyticDigestion;
@@ -21,6 +22,13 @@ namespace EngineLayer.ModernSearch
         protected readonly DissociationType DissociationType;
         protected readonly double MaxMassThatFragmentIonScoreIsDoubled;
 
+        /// <summary>
+        /// Whether the per-scan score table stamps cells instead of clearing them. Worth it when a scan
+        /// touches a small slice of the peptide index; a losing trade for an open search, which touches
+        /// most of it. See ScanScoringTable.
+        /// </summary>
+        protected readonly bool UseStampedScoringTable;
+
         public ModernSearchEngine(SpectralMatch[] globalPsms, Ms2ScanWithSpecificMass[] listOfSortedms2Scans, List<PeptideWithSetModifications> peptideIndex,
             List<int>[] fragmentIndex, int currentPartition, CommonParameters commonParameters, List<(string fileName, CommonParameters fileSpecificParameters)> fileSpecificParameters, MassDiffAcceptor massDiffAcceptor, double maximumMassThatFragmentIonScoreIsDoubled,
             List<string> nestedIds) : base(commonParameters, fileSpecificParameters, nestedIds)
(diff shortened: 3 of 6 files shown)


acesnik:
🤖 This is a satisfying optimisation to review. Replacing a full-array clear per scan with a generation stamp, and a comparison sort with a counting sort over a byte domain, are both the kind of change where the idea is obviously right and the risk is entirely in the details — so I went at the details rather than the idea. The behaviour-preserving claim holds, and I couldn't break it.

## Results are unchanged (measured)

Modern search, HCD, `OneMM` (`DotMassDiffAcceptor`, so the stamped path), trypsin, 1 partition, 8 threads. 4,000 mouse reviewed entries plus cRAP (8,230 with decoys), peptide index **1,918,766**, **26,355 MS2 scans**. Four runs of `master` and four of this branch, all against one shared cached peptide index:

- `AllPSMs.psmtsv` **byte-identical across all eight runs** — 12,285 rows, 9,559 scans.
- `AllPeptides.psmtsv`, `AllPSMs_FormattedForPercolator.tab`, all three `AllQuantified*` files and the `.mzID` byte-identical too.
- Column-wise on shared scans: Score, Delta Score, Full Sequence, Base Sequence, Accession, Notch, Spectrum Match Count, Matched Ion Counts, QValue, PEP, PEP_QValue — all identical.
- Only `results.txt`, `allResults.txt` and `AutoGeneratedManuscriptProse.txt` differ, all timings.

**If anyone else reproduces this, pin the index.** My first attempt gave each arm its own copy of the database, so each built its own index — and the two `peptideIndex.ind` files differ. That alone moved 4 reported identifications (`DLSTIEPLK` ↔ `DISTLEPLK`, isobaric I/L plus S/T permutations), 20 Delta Scores, 5,044 q-values and 8,747 PEPs, and I nearly wrote it up as your bug. Control: two `master` runs sharing one index are byte-identical in every column including PEP, so search plus PEP is deterministic once the index is fixed. (`.mzID` `<Modification>` element order also varies per process — 68 lines between two `master` runs — so compare that one sorted.)

## The stability claim, checked directly

Tie order was my main worry: a counting sort and a comparison sort can order equal scores differently, and if anything downstream takes a top-N of equally scored candidates, that changes which peptide gets reported at identical score. So I pulled the two new types into a standalone harness and diffed them against exactly the code they replace.

- `Sort` vs `peptideIds.OrderByDescending(p => scoringTable[p])`: 40,000 randomised scans, **4,196,914 ids**, exact sequence equality. Score alphabets were kept deliberately small so ties were dense — **38,119 of 40,000** scans contained at least one tie, so the test has teeth.
- `ScanScoringTable` stamped and unstamped vs a `byte[]` reference model: 5,000 `BeginScan` calls, crossing the 255-stamp wrap about 19 times, 185,000 cell reads, including forced 255→0 score wraps. Identical.
- A value written under stamp 1 stayed invisible across 2,999 later scans, so a stale stamp never aliases a live one. Worth recording why the zero-stamp case is safe: a fresh table has `_stamp == 0` and zero cells, so every cell reads live-with-score-0 — which is the right answer — and the first `BeginScan` moves to 1 and retires them all.

The 257-entry prefix array covers the whole byte domain, and since scores are `byte` negatives can't arise. Happy to hand the harness over as two unit tests.

## Two questions about `UseStampedScoringTable`

Both are performance rather than correctness, and the first is why I'm holding off approving.

**1. `massDiffAcceptor is not OpenSearchMode` misses `ModOpen`.** `MassDiffAcceptorType.ModOpen` is `IntervalMassDiffAcceptor("-187andUp", [-187, +∞))`. `GetAllowedPrecursorMassIntervalsFromObservedMass` maps that `+∞` maximum to a lower bound of `-∞`, so in `IndexScoreScan` `lowestMassPeptideToLookFor` is `-∞` and `GetFirstAndLastIndexesInBinToIncrement` returns `start = 0` for **every** bin — it increments every peptide in the bin up to precursor + 187 Da. That's the same index-touching regime as an open search, but it isn't `OpenSearchMode`, so it takes the stamped path: the 0.6–0.8x case in your own numbers.

What makes it worth settling rather than noting: `MassDifferenceAcceptorSelectionViewModel` calls ModOpen an "open-mass search" and its tooltip says *"Please use Modern Search mode when using this search type"*. So the configuration the GUI actively steers toward Modern Search is the one that would land on the regressing side. A `Custom` `interval` acceptor with a wide range has the same shape. Keying on the width of the allowed precursor interval, or on `MassDiffAcceptorType`, would cover all of them. I did not run a ModOpen search — this is read from the code, so if the reading is wrong I'd rather hear it than have you work around me.

**2. The non-specific engine's touch fraction has nothing to do with its acceptor.** `SnesIndexedScoring` increments every peptide in each matching bin — no `GetFirstAndLastIndexesInBinToIncrement`, no precursor filtering, at all four sites. So how much of the index a non-specific scan touches isn't a function of the mass-diff acceptor, yet that engine inherits the stamping decision from it and stamps for any non-open acceptor. It's also the one engine here whose innermost loop gets only `Increment` calls and no sorter benefit. Not covered by the synthetic benchmark — worth measuring before assuming stamping wins there.

## Speed: I'd like a real search number in the body

Warm-index `ModernSearchEngine`, 4 reps per arm, alternating, same machine, same shared index:

| | run 1 | run 2 | run 3 | run 4 | mean |
|---|---|---|---|---|---|
| `master` | 18.2 s | 14.5 s | 15.6 s | 11.1 s | **14.9 s** |
| this branch | 11.9 s | 11.6 s | 11.4 s | 12.8 s | **11.9 s** |

Mean ratio 1.24x, and the branch is far more consistent — spread 1.13x against master's 1.65x — which is what you'd expect from deleting a large variable memset. For scale, `master` does 26,355 × 1,918,766 = **50.6 GB** of `Array.Clear` traffic on this one file. But master's own spread is wider than the gap between the means, and its fastest run beats this branch's median. My machine was under heavy external memory pressure, so treat the absolutes as soft.

The point stands regardless: the body quotes only a synthetic microbenchmark, with no repetition count and no end-to-end timing. Given a 1.65x spread on unchanged code, a single number wouldn't mean anything — worth saying how many reps the 8–13x came from, and adding one whole-search figure.

## Smaller things

- **The stamped path is unreachable in crosslink and glyco.** Both pass `new OpenSearchMode()` to the `ModernSearchEngine` base constructor unconditionally, so `UseStampedScoringTable` is always false there and the only change those two get is the sorter plus a branch per table access. Correct as written — the body just reads as though it were a per-run choice for them.
- **`Sort` returns a span over a buffer the next `Sort` overwrites.** All three call sites are safe today (none sorts again inside the loop) but nothing enforces it, and a future nested sort would silently corrupt an in-flight iteration rather than throwing. A line on the method, or a cheap version guard, would stop that being rediscovered the hard way.
- **Memory** for the table doubles: 3.8 MB against 1.9 MB per thread here, roughly 31 MB against 15 MB at 8 threads. Fine, but it scales with `PeptideIndex.Count` × threads and this engine already has an OOM story.
- **No tests added.** Coverage comes from existing search tests, and nothing in the suite pins tie order — so an unstable sort wouldn't have failed anything, which means "1830 passed" isn't evidence for the stability claim specifically. The two harnesses above would make it evidence.
- The ~15-line rationale blocks with benchmark ratios inline will go stale in the file; the body already carries all of it.
- "a clear every 255 scans" is accurate — stamps run 1..255 and the clear lands on the 256th `BeginScan`.

Not verified: the test suite (Windows-only, so relying on your CI); end-to-end crosslink, glyco, non-specific and top-down — only Modern/HCD/OneMM was run; ModOpen or Open end to end; multi-partition; the LowCID branch; and your synthetic microbenchmark, which I didn't reproduce. My timings came from a machine with 28.8 of 29.7 GB of swap in use by unrelated processes, on a 4,000-protein subset with `MaxFragmentSize` reduced to 5000 to make the index build tractable — direction and variance, not a production multiplier.

acesnik (commented):
🤖 This is a satisfying optimisation to review. Replacing a full-array clear per scan with a generation stamp, and a comparison sort with a counting sort over a byte domain, are both the kind of change where the idea is obviously right and the risk is entirely in the details — so I went at the details rather than the idea. The behaviour-preserving claim holds, and I couldn't break it.

## Results are unchanged (measured)

Modern search, HCD, `OneMM` (`DotMassDiffAcceptor`, so the stamped path), trypsin, 1 partition, 8 threads. 4,000 mouse reviewed entries plus cRAP (8,230 with decoys), peptide index **1,918,766**, **26,355 MS2 scans**. Four runs of `master` and four of this branch, all against one shared cached peptide index:

- `AllPSMs.psmtsv` **byte-identical across all eight runs** — 12,285 rows, 9,559 scans.
- `AllPeptides.psmtsv`, `AllPSMs_FormattedForPercolator.tab`, all three `AllQuantified*` files and the `.mzID` byte-identical too.
- Column-wise on shared scans: Score, Delta Score, Full Sequence, Base Sequence, Accession, Notch, Spectrum Match Count, Matched Ion Counts, QValue, PEP, PEP_QValue — all identical.
- Only `results.txt`, `allResults.txt` and `AutoGeneratedManuscriptProse.txt` differ, all timings.

**If anyone else reproduces this, pin the index.** My first attempt gave each arm its own copy of the database, so each built its own index — and the two `peptideIndex.ind` files differ. That alone moved 4 reported identifications (`DLSTIEPLK` ↔ `DISTLEPLK`, isobaric I/L plus S/T permutations), 20 Delta Scores, 5,044 q-values and 8,747 PEPs, and I nearly wrote it up as your bug. Control: two `master` runs sharing one index are byte-identical in every column including PEP, so search plus PEP is deterministic once the index is fixed. (`.mzID` `<Modification>` element order also varies per process — 68 lines between two `master` runs — so compare that one sorted.)

## The stability claim, checked directly

Tie order was my main worry: a counting sort and a comparison sort can order equal scores differently, and if anything downstream takes a top-N of equally scored candidates, that changes which peptide gets reported at identical score. So I pulled the two new types into a standalone harness and diffed them against exactly the code they replace.

- `Sort` vs `peptideIds.OrderByDescending(p => scoringTable[p])`: 40,000 randomised scans, **4,196,914 ids**, exact sequence equality. Score alphabets were kept deliberately small so ties were dense — **38,119 of 40,000** scans contained at least one tie, so the test has teeth.
- `ScanScoringTable` stamped and unstamped vs a `byte[]` reference model: 5,000 `BeginScan` calls, crossing the 255-stamp wrap about 19 times, 185,000 cell reads, including forced 255→0 score wraps. Identical.
- A value written under stamp 1 stayed invisible across 2,999 later scans, so a stale stamp never aliases a live one. Worth recording why the zero-stamp case is safe: a fresh table has `_stamp == 0` and zero cells, so every cell reads live-with-score-0 — which is the right answer — and the first `BeginScan` moves to 1 and retires them all.

The 257-entry prefix array covers the whole byte domain, and since scores are `byte` negatives can't arise. Happy to hand the harness over as two unit tests.

## Two questions about `UseStampedScoringTable`

Both are performance rather than correctness, and the first is why I'm holding off approving.

**1. `massDiffAcceptor is not OpenSearchMode` misses `ModOpen`.** `MassDiffAcceptorType.ModOpen` is `IntervalMassDiffAcceptor("-187andUp", [-187, +∞))`. `GetAllowedPrecursorMassIntervalsFromObservedMass` maps that `+∞` maximum to a lower bound of `-∞`, so in `IndexScoreScan` `lowestMassPeptideToLookFor` is `-∞` and `GetFirstAndLastIndexesInBinToIncrement` returns `start = 0` for **every** bin — it increments every peptide in the bin up to precursor + 187 Da. That's the same index-touching regime as an open search, but it isn't `OpenSearchMode`, so it takes the stamped path: the 0.6–0.8x case in your own numbers.

What makes it worth settling rather than noting: `MassDifferenceAcceptorSelectionViewModel` calls ModOpen an "open-mass search" and its tooltip says *"Please use Modern Search mode when using this search type"*. So the configuration the GUI actively steers toward Modern Search is the one that would land on the regressing side. A `Custom` `interval` acceptor with a wide range has the same shape. Keying on the width of the allowed precursor interval, or on `MassDiffAcceptorType`, would cover all of them. I did not run a ModOpen search — this is read from the code, so if the reading is wrong I'd rather hear it than have you work around me.

**2. The non-specific engine's touch fraction has nothing to do with its acceptor.** `SnesIndexedScoring` increments every peptide in each matching bin — no `GetFirstAndLastIndexesInBinToIncrement`, no precursor filtering, at all four sites. So how much of the index a non-specific scan touches isn't a function of the mass-diff acceptor, yet that engine inherits the stamping decision from it and stamps for any non-open acceptor. It's also the one engine here whose innermost loop gets only `Increment` calls and no sorter benefit. Not covered by the synthetic benchmark — worth measuring before assuming stamping wins there.

## Speed: I'd like a real search number in the body

Warm-index `ModernSearchEngine`, 4 reps per arm, alternating, same machine, same shared index:

| | run 1 | run 2 | run 3 | run 4 | mean |
|---|---|---|---|---|---|
| `master` | 18.2 s | 14.5 s | 15.6 s | 11.1 s | **14.9 s** |
| this branch | 11.9 s | 11.6 s | 11.4 s | 12.8 s | **11.9 s** |

Mean ratio 1.24x, and the branch is far more consistent — spread 1.13x against master's 1.65x — which is what you'd expect from deleting a large variable memset. For scale, `master` does 26,355 × 1,918,766 = **50.6 GB** of `Array.Clear` traffic on this one file. But master's own spread is wider than the gap between the means, and its fastest run beats this branch's median. My machine was under heavy external memory pressure, so treat the absolutes as soft.

The point stands regardless: the body quotes only a synthetic microbenchmark, with no repetition count and no end-to-end timing. Given a 1.65x spread on unchanged code, a single number wouldn't mean anything — worth saying how many reps the 8–13x came from, and adding one whole-search figure.

## Smaller things

- **The stamped path is unreachable in crosslink and glyco.** Both pass `new OpenSearchMode()` to the `ModernSearchEngine` base constructor unconditionally, so `UseStampedScoringTable` is always false there and the only change those two get is the sorter plus a branch per table access. Correct as written — the body just reads as though it were a per-run choice for them.
- **`Sort` returns a span over a buffer the next `Sort` overwrites.** All three call sites are safe today (none sorts again inside the loop) but nothing enforces it, and a future nested sort would silently corrupt an in-flight iteration rather than throwing. A line on the method, or a cheap version guard, would stop that being rediscovered the hard way.
- **Memory** for the table doubles: 3.8 MB against 1.9 MB per thread here, roughly 31 MB against 15 MB at 8 threads. Fine, but it scales with `PeptideIndex.Count` × threads and this engine already has an OOM story.
- **No tests added.** Coverage comes from existing search tests, and nothing in the suite pins tie order — so an unstable sort wouldn't have failed anything, which means "1830 passed" isn't evidence for the stability claim specifically. The two harnesses above would make it evidence.
- The ~15-line rationale blocks with benchmark ratios inline will go stale in the file; the body already carries all of it.
- "a clear every 255 scans" is accurate — stamps run 1..255 and the clear lands on the 256th `BeginScan`.

Not verified: the test suite (Windows-only, so relying on your CI); end-to-end crosslink, glyco, non-specific and top-down — only Modern/HCD/OneMM was run; ModOpen or Open end to end; multi-partition; the LowCID branch; and your synthetic microbenchmark, which I didn't reproduce. My timings came from a machine with 28.8 of 29.7 GB of swap in use by unrelated processes, on a 4,000-protein subset with `MaxFragmentSize` reduced to 5000 to make the index build tractable — direction and variance, not a production multiplier.

PR #2763, opened 2026-08-27
Name the LoadRnaXML arguments so the thread count stops binding to a variant limit

### LINKED ISSUE #2762: RNA XML loading passes the thread count as the heterozygous-variant limit
🤖 `DatabaseLoadingEngine` passes the per-file thread count into mzLib's heterozygous-variant limit when loading an RNA XML database, because `LoadRnaXML` and `LoadProteinXML` order their parameters differently and the RNA call was written as though they matched.

## The call

`MetaMorpheus/EngineLayer/DatabaseLoading/DatabaseLoadingEngine.cs:151`

```csharp
rnaList = RnaDbLoader.LoadRnaXML(fileName, generateTargets, decoyType, isContaminant,
    GlobalVariables.AllRnaModsKnown, modTypesToExclude, out unknownMods,
    commonParameters.MaxThreadsToUsePerFile, decoyIdentifier: decoyIdentifier);
```

Eight positional arguments. mzLib's signature (single overload, `RnaDbLoader.cs:318-323`):

```
 1 rnaDbLocation      6 modTypesToExclude          11 fivePrimeTerm
 2 generateTargets    7 unknownModifications (out) 12 threePrimeTerm
 3 decoyType          8 maxHeterozygousVariants    13 decoyIdentifier
 4 isContaminant      9 minAlleleDepth             14 entrapmentIdentifier
 5 allKnownModifications  10 maxThreads             15 isEntrapment
```

So `commonParameters.MaxThreadsToUsePerFile` binds to **`maxHeterozygousVariants`**, and `maxThreads` keeps its default of `1`.

## Why it was easy to miss

The protein loader in the same method has the opposite order — `maxThreads` comes *before* the variant parameters — and its call at `:180` binds correctly:

```csharp
proteinList = ProteinDbLoader.LoadProteinXML(fileName, generateTargets, decoyType, GlobalVariables.AllModsKnown,
    isContaminant, modTypesToExclude, out um, commonParameters.MaxThreadsToUsePerFile, 0,
    commonParameters.MinVariantDepth, addTruncations: commonParameters.AddTruncations, decoyIdentifier: decoyIdentifier);
```

`ProteinDbLoader.LoadProteinXML` is `(..., out unknownModifications, int maxThreads = -1, int maxHeterozygousVariants = 4, int minAlleleDepth = 1, ...)`. Same three concepts, different order, and every argument is a defaulted `int`, so nothing warns.

## Two consequences

**RNA XML decoy generation is single-threaded.** `maxThreads` reaches `RnaDecoyGenerator.GenerateDecoys(targets, decoyType, maxThreads, decoyIdentifier)` as `1` no matter what the user configured.

**Variant expansion depth is set by the thread count.** How many heterozygous variant combinations get applied depends on `MaxThreadsToUsePerFile` — so the same database and the same task produce different variant proteins on a machine configured with 4 threads than with 16. That is machine-dependent output rather than data-dependent, the same class of problem as smith-chem-wisc/mzLib#1155.

Note the protein path deliberately passes `0` for this parameter — the previous value is still visible commented out on the line above — so the two paths also disagree about what the limit should be, independent of the binding.

## Not related to the open variant PRs

Found while reviewing smith-chem-wisc/mzLib#972, but this is present on master today and independent of that series. The `// TODO: Add in variant params when fixed in MzLib.` comment directly above the call suggests the variant parameters were known to be unwired; the part that was not noticed is that the thread count is landing in one of them.

Fix is to name the arguments. Any fix changes RNA variant output relative to today's accidental value, since today's value is the thread count — worth deciding explicitly whether RNA should match the protein path's `0` or take mzLib's default of `4`.

acesnik:
🤖 Closes #2762. One call site, arguments named instead of positional.

## The bug

`DatabaseLoadingEngine.cs:151` passed `commonParameters.MaxThreadsToUsePerFile` as the eighth positional argument to `RnaDbLoader.LoadRnaXML`. Parameter eight is `maxHeterozygousVariants`; `maxThreads` is parameter ten.

```
LoadRnaXML     ... out unknownModifications, maxHeterozygousVariants, minAlleleDepth, maxThreads, ...
LoadProteinXML ... out unknownModifications, maxThreads, maxHeterozygousVariants, minAlleleDepth, ...
```

Same three concepts, opposite order. The protein call five lines below binds correctly; the RNA call was written as though the two matched. Every parameter involved is a defaulted `int`, so nothing warned and nothing failed.

Consequences on master:

- **RNA XML decoy generation ran single-threaded.** `maxThreads` kept its default of `1`, so `RnaDecoyGenerator.GenerateDecoys` never saw the configured thread count.
- **Variant expansion depth was set by the thread count.** The same database and the same task produced different variant RNA on a 4-thread configuration than on a 16-thread one. That is machine-dependent output rather than data-dependent — the same class of problem as smith-chem-wisc/mzLib#1155.

## This changes RNA variant output

Unavoidably. Today's `maxHeterozygousVariants` is whatever `MaxThreadsToUsePerFile` happened to be, so any fix has to replace it with something.

I used `maxHeterozygousVariants: 0`, matching the choice the protein path already makes explicitly on the line below, rather than letting it fall back to mzLib's default of `4`. That is a judgement call and a one-word change in review if you would rather have `4` — but 0 is the value this file has already chosen for the equivalent protein parameter, with the original still visible commented out above it.

Nothing else about RNA loading changes.

## Deliberately not fixed here

The protein path wires `minAlleleDepth` from `commonParameters.MinVariantDepth`; the RNA path takes mzLib's default of `1`. That inconsistency is real and survives this PR. It is a separate decision about what RNA searches should do, not part of the binding bug, so it stays out — and the `// TODO: Add in variant params when fixed in MzLib.` comment above the call is left in place because it is still true.

## Verification

`dotnet build MetaMorpheus.sln -c UbuntuMac -p:EnableWindowsTargeting=true` — 0 errors.

Not verified: the RNA search tests, which are Windows-targeted, so CI is the check. Worth watching whether any RNA variant expectation moves — if one does, that is this fix taking effect rather than breaking something, since the previous value depended on the thread count of whatever machine recorded the expectation.

### DIFF
--- a/MetaMorpheus/EngineLayer/DatabaseLoading/DatabaseLoadingEngine.cs
+++ b/MetaMorpheus/EngineLayer/DatabaseLoading/DatabaseLoadingEngine.cs
@@ -148,7 +148,20 @@ public class DatabaseLoadingEngine(
             GlobalVariables.AddMods(ProteinDbLoader.GetPtmListFromProteinXml(fileName), true, true);
             // TODO: Add in variant params when fixed in MzLib. 
             List<string> modTypesToExclude = GlobalVariables.AllRnaModTypesKnown.Where(b => !localizeableModificationTypes.Contains(b)).ToList();
-            rnaList = RnaDbLoader.LoadRnaXML(fileName, generateTargets, decoyType, isContaminant, GlobalVariables.AllRnaModsKnown, modTypesToExclude, out unknownMods, commonParameters.MaxThreadsToUsePerFile, decoyIdentifier: decoyIdentifier);
+            // Named, not positional. LoadRnaXML orders these as (maxHeterozygousVariants, minAlleleDepth,
+            // maxThreads); LoadProteinXML below orders them as (maxThreads, maxHeterozygousVariants,
+            // minAlleleDepth). Passing the thread count positionally here bound it to
+            // maxHeterozygousVariants, so decoy generation ran at maxThreads = 1 and the variant
+            // combinatorics limit was whatever thread count the run happened to use. Every parameter
+            // involved is a defaulted int, so nothing warned. See #2762.
+            //
+            // maxHeterozygousVariants: 0 matches the protein path below rather than taking mzLib's
+            // default of 4 -- the value has to become something, and 0 is the choice already made here.
+            rnaList = RnaDbLoader.LoadRnaXML(fileName, generateTargets, decoyType, isContaminant,
+                GlobalVariables.AllRnaModsKnown, modTypesToExclude, out unknownMods,
+                maxHeterozygousVariants: 0,
+                maxThreads: commonParameters.MaxThreadsToUsePerFile,
+                decoyIdentifier: decoyIdentifier);
         }
 
         emptyEntriesCount = rnaList.Count(p => p.BaseSequence.Length == 0);


trishorts (approved):
Approving. I checked the binding rather than trusting the diff, and it is exactly as described.

`RnaDbLoader.LoadRnaXML` (`UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs:318-323`) orders the tail as:

```csharp
out Dictionary<string, Modification> unknownModifications,
int maxHeterozygousVariants = 4, int minAlleleDepth = 1,
int maxThreads = 1, ...
```

so the eighth argument really was landing on `maxHeterozygousVariants` and `maxThreads` really was staying at its default of 1. Both consequences follow: single-threaded RNA decoy generation, and a variant-combinatorics limit set by the machine's thread count. The mzLib#1155 comparison is apt, and machine-dependent rather than data-dependent output is the part that makes this worth fixing loudly rather than quietly.

`maxHeterozygousVariants: 0` is the right call, and I verified the precedent you cite instead of taking it on trust: the protein path passes a literal `0` in that position at `DatabaseLoadingEngine.cs:180`, with `commonParameters.MaxHeterozygousVariants` still visible in the commented-out line above it. So this is consistency with a choice already made in this file, not a new one, and it is the only value that does not make the fix a silent behaviour change of its own.

Naming the arguments rather than merely reordering them is the durable part. Two loaders carrying the same three concepts in opposite orders, every one of them a defaulted `int`, is a trap that re-arms itself the moment either signature grows; named arguments are the only thing that disarms it permanently.

Agreed on leaving `minAlleleDepth` alone. The protein path wiring `commonParameters.MinVariantDepth` while RNA takes mzLib's default of 1 is a real inconsistency, but it is a decision about what RNA searches should do rather than part of the binding bug, and folding it in would make this one-line fix un-revertable.

MetaMorpheus/EngineLayer/DatabaseLoading/DatabaseLoadingEngine.cs
```diff
@@ -148,7 +148,20 @@ public static IEnumerable<RNA> LoadOligoDb(string fileName, bool generateTargets
...
+            // minAlleleDepth). Passing the thread count positionally here bound it to
+            // maxHeterozygousVariants, so decoy generation ran at maxThreads = 1 and the variant
+            // combinatorics limit was whatever thread count the run happened to use. Every parameter
+            // involved is a defaulted int, so nothing warned. See #2762.
+            //
+            // maxHeterozygousVariants: 0 matches the protein path below rather than taking mzLib's
+            // default of 4 -- the value has to become something, and 0 is the choice already made here.
+            rnaList = RnaDbLoader.LoadRnaXML(fileName, generateTargets, decoyType, isContaminant,
+                GlobalVariables.AllRnaModsKnown, modTypesToExclude, out unknownMods,
+                maxHeterozygousVariants: 0,
```
trishorts, line 162: **This is what is failing CI.** `Test/Transcriptomics/RnaDatabaseLoadingTests.cs:129` does

```csharp
IBioPolymer variantTarget = targets.First(p => p.AppliedSequenceVariations.Count >= 1);
```

and now throws `InvalidOperationException : Sequence contains no matching element`. With `maxHeterozygousVariants: 0` no variant is applied, so no such element exists. `TwoTruncationsAndSequenceVariant_DbLoading` exists specifically to assert that RNA sequence variants *are* applied, and it passed before only because the old bug happened to pass a positive number (the thread count) into that slot.

So the rationale in the comment above doesn't carry over. "0 matches the protein path" is accurate as a statement about `LoadProteinDb` (`DatabaseLoadingEngine.cs:180` does pass a literal `0`), but the protein path has no test asserting variants get applied and the RNA path does. Matching it here silently turns RNA variant support off.

The value has to become *something*, agreed — but 0 is the one choice that changes behaviour beyond fixing the bug. Options, in the order I'd prefer them:

1. **Don't pass it at all.** Name only `maxThreads:` and let `maxHeterozygousVariants` take mzLib's default of 4. That fixes exactly the reported bug and nothing else, and the existing test stays green.
2. **Pass `commonParameters.MaxHeterozygousVariants`**, which is what the commented-out protein call at `DatabaseLoadingEngine.cs:179` was reaching for before someone hardcoded 0. This is the most defensible long-term, but it changes RNA behaviour on purpose, so it wants to be its own PR with the test updated deliberately.

Separately: `minAlleleDepth` is left to take mzLib's default of 1 here, while the protein path passes `commonParameters.MinVariantDepth`. If option 1 or 2 restores variant application, that divergence becomes live and is worth naming explicitly too rather than inheriting a default.

trishorts (commented):
The diagnosis is right and worth having written down: `RnaDbLoader.LoadRnaXML` orders its optionals `(maxHeterozygousVariants = 4, minAlleleDepth = 1, maxThreads = 1)` while `ProteinDbLoader.LoadProteinXML` orders them `(maxThreads, maxHeterozygousVariants, minAlleleDepth)` — I confirmed both signatures at `RnaDbLoader.cs:318-323` and `ProteinDbLoader.cs:87`. Passing the thread count positionally did bind it to `maxHeterozygousVariants`, and every parameter being a defaulted `int` is why nothing warned. Naming the arguments is the correct fix.

The value chosen for `maxHeterozygousVariants` is the problem, and CI is already saying so — `Test on Windows` and `Test on Windows (core)` are both red on `TwoTruncationsAndSequenceVariant_DbLoading`. That is not unrelated flake; it is this change. See the inline note.

Once that is settled this is a clean fix, and the explanatory comment is the right amount of detail for a bug that was invisible by construction.

PR #2764, opened 2026-08-27
Make IntensitiesByFile drop isobaric samples, as its companion view already does

trishorts:
`ProteinGroup` exposes two `SpectraFileInfo`-shaped views over the base class's sample-keyed quantification store. They disagree about what to do with a sample they cannot represent:

```csharp
// :108
public IReadOnlyList<SpectraFileInfo> FilesForQuantification
    => SamplesForQuantification?.OfType<SpectraFileInfo>().ToList();

// :120
public IReadOnlyDictionary<SpectraFileInfo, double> IntensitiesByFile
    => IntensitiesBySample?.ToDictionary(kvp => (SpectraFileInfo)kvp.Key, kvp => kvp.Value);
```

The first filters. The second casts, and throws `InvalidCastException` on an `IsobaricQuantSampleInfo`.

**Both behaviours are documented**, twelve lines apart — *"isobaric samples are dropped by the getter"* on one and *"the getter throws if an isobaric sample is present"* on the other. So this was a deliberate choice, not an oversight, and I want to be clear that I am arguing against a decision rather than reporting a slip.

### Why the choice is now the wrong one

It is unreachable on master today, because nothing puts isobaric samples into that store. It becomes reachable the moment mzLib's `QuantificationEngine` writes per-channel intensities onto a protein group — which is what #2755 does.

The failure is not graceful. The read happens on the ordinary output path:

```
System.InvalidCastException : Unable to cast object of type
'MassSpectrometry.IsobaricQuantSampleInfo' to type 'MassSpectrometry.SpectraFileInfo'
   at TaskLayer.PostSearchAnalysisTask.Run()
```

So a TMT search **crashes**, rather than simply not populating a file-keyed view that could never have held channels anyway. "Fail loudly" is a reasonable instinct, but the thing it fails on is a caller doing something correct.

### Why filtering

It makes the pair consistent, which is the part that matters more than either behaviour on its own. A caller reading both views on the same data currently gets one success and one exception; afterwards it gets the same set of files from both.

Dropping is also the honest answer: `IntensitiesByFile` is keyed by `SpectraFileInfo` and one file carries many channels, so there is no correct value to return for them. `IntensitiesBySample` is where that data lives, and it is unaffected.

### Scope

One property getter, one comment. `IntensitiesBySample`, `SamplesForQuantification` and label-free quantification are all untouched.

Three tests: the isobaric-only case that used to throw, a mixed case asserting the two views agree on what they drop, and a label-free case pinning that ordinary quantification is unchanged.

### Why this is its own PR

It was found by #2755's end-to-end TMT tests, and it would have been easy to fix in passing there. But it is a bug in master's own code, on the FlashLFQ output path, and nobody reviewing a quantification-engine PR would look for it. It should be reviewable on its own terms — and #2755 is blocked until it lands.

### DIFF
--- a/MetaMorpheus/EngineLayer/ProteinParsimony/ProteinGroup.cs
+++ b/MetaMorpheus/EngineLayer/ProteinParsimony/ProteinGroup.cs
@@ -113,11 +113,16 @@ namespace EngineLayer
         /// Snapshot view of <see cref="BioPolymerGroup.IntensitiesBySample"/> keyed by SpectraFileInfo.
         /// The getter returns a fresh read-only copy on every call, so mutating the returned value has no
         /// effect on the protein group; assign a new dictionary through the setter to change it.
-        /// SpectraFileInfo only: the getter throws if an isobaric sample is present.
+        /// SpectraFileInfo only: isobaric samples are dropped by the getter, so a round-trip loses them.
+        /// This matches <see cref="FilesForQuantification"/>, which is the companion view over the same
+        /// backing store -- the two describe the same subset and should agree on what happens to a
+        /// sample they cannot represent.
         /// </summary>
         public IReadOnlyDictionary<SpectraFileInfo, double> IntensitiesByFile
         {
-            get => IntensitiesBySample?.ToDictionary(kvp => (SpectraFileInfo)kvp.Key, kvp => kvp.Value);
+            get => IntensitiesBySample?
+                .Where(kvp => kvp.Key is SpectraFileInfo)
+                .ToDictionary(kvp => (SpectraFileInfo)kvp.Key, kvp => kvp.Value);
             set => IntensitiesBySample = value?.ToDictionary(kvp => (ISampleInfo)kvp.Key, kvp => kvp.Value);
         }
 
--- a/MetaMorpheus/Test/ProteinGroupIsobaricViewTests.cs
+++ b/MetaMorpheus/Test/ProteinGroupIsobaricViewTests.cs
@@ -1,0 +1,88 @@
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using EngineLayer;
+using MassSpectrometry;
+using NUnit.Framework;
+using Omics;
+using Proteomics;
+
+namespace Test
+{
+    /// <summary>
+    /// ProteinGroup exposes two SpectraFileInfo-shaped views over the base class's sample-keyed
+    /// quantification store: FilesForQuantification and IntensitiesByFile. Isobaric quantification
+    /// puts IsobaricQuantSampleInfo into that store -- one entry per channel rather than per file --
+    /// so both views have to answer for a sample they cannot represent, and they have to answer the
+    /// same way.
+    /// </summary>
+    [TestFixture]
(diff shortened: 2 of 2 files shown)


nbollis (approved):

trishorts:
@Alexander-Sol — 95 lines, 1/3 with nbollis, CI green (13/13). This one gates #2755.

`ProteinGroup` has two `SpectraFileInfo`-shaped views over the same sample-keyed store, and they disagree about a sample they cannot represent — one filters, one casts and throws.

**Both behaviours were documented, twelve lines apart**, so I want to be explicit that I am arguing against a deliberate choice rather than reporting a slip. "Fail loudly" is defensible; it stops being defensible once the thing it fails on is a caller doing something correct.

It is unreachable on master today. It becomes reachable the moment mzLib's `QuantificationEngine` writes per-channel intensities onto a protein group — and the read is on the ordinary output path, so a TMT search **crashes**:

```
System.InvalidCastException : Unable to cast object of type
'MassSpectrometry.IsobaricQuantSampleInfo' to type 'MassSpectrometry.SpectraFileInfo'
   at TaskLayer.PostSearchAnalysisTask.Run()
```

I found it by running #2755's end-to-end TMT tests against master, which is also the evidence that filtering fixes it.

Dropping is the honest answer as well as the consistent one: `IntensitiesByFile` is keyed by `SpectraFileInfo` and one file carries many channels, so there is no correct value to return for them. `IntensitiesBySample` holds that data and is untouched.

Three tests: the isobaric case that used to throw, a mixed case asserting the two views now agree on what they drop, and a label-free case pinning that ordinary quantification is unaffected.

acesnik:
🤖 Correct and worth landing. The getter now matches `FilesForQuantification` exactly, and the `is` filter genuinely excludes channels — `IsobaricQuantSampleInfo` implements `ISampleInfo` rather than deriving from `SpectraFileInfo`, which is also why the old hard cast threw rather than quietly succeeding.

One thing to carry into #2755 rather than fix here.

The setter is unchanged and replaces the whole backing store (`ProteinGroup.cs:126`):

```csharp
set => IntensitiesBySample = value?.ToDictionary(kvp => (ISampleInfo)kvp.Key, kvp => kvp.Value);
```

So the asymmetry you documented as "a round-trip loses them" has five production callers, each assigning a `SpectraFileInfo`-keyed dictionary built from FlashLFQ output, or null:

- `PostSearchAnalysisTask.cs:608`, and `:674` which assigns `null`
- `PostGlycoSearchAnalysisTask.cs:619`
- `SilacConversions.cs:503`
- `ProteinGroup.cs:699` and `:703` — inherent there, since a subset group is per-file by definition

Today that is invisible for the same reason the bug you are fixing is: nothing puts isobaric samples in the store. Once #2755 does, whichever of those runs afterwards erases them — and this fix is what makes it silent, since before, the read would have thrown first.

Whether it matters is a question of ordering in #2755. If channels are written after `PostSearchAnalysisTask` has assigned, nothing is lost. Flagging so that gets checked there rather than found from a TMT run with empty channels.

acesnik (approved):

zhuoxinshi (approved):

PR #2765, opened 2026-08-27
Bump mzLib to 1.0.587

trishorts:
Nothing but the version, in all six csprojs. Reviewable on its own and merges without waiting on anything else in the quant batch.

Master has been on `1.0.585` across two mzLib releases.

## What 1.0.587 carries that an existing caller would notice

Three things, and MetaMorpheus is exposed to none of them — worth stating explicitly rather than asserting the bump is safe:

- **`QuantifiedPeaks.tsv` gains `Peak FWHM` and `Peak FWHM Status`** ([mzLib #1180](https://github.com/smith-chem-wisc/mzLib/pull/1180)), inserted mid-row. Anything parsing that file *by column index* reads the wrong field. MetaMorpheus writes the file through FlashLFQ and does not parse it back.
- **`QuantificationEngine` now excludes ambiguous spectral matches** ([mzLib #1210](https://github.com/smith-chem-wisc/mzLib/pull/1210)). `PostSearchAnalysisTask` already passes `includeAmbiguous: false` at its own call site, so it was never taking that path.
- **`SumCollapse`, `MeanCollapse` and `SumRollUp.ArrayPool` were removed** ([mzLib #1194](https://github.com/smith-chem-wisc/mzLib/pull/1194), [#1211](https://github.com/smith-chem-wisc/mzLib/pull/1211)). Master references none of them — `grep` for all three comes back empty.

The full release notes are on [mzLib 1.0.587](https://github.com/smith-chem-wisc/mzLib/releases/tag/1.0.587).

## Why separately

Four quant PRs behind this one each needed the same bump, and were each carrying it. Landing it once here means their diffs are about what they actually do, and none of them has to be reviewed with a version change stapled to the front.

## Verification

CI is the verification that counts. A local build proves nothing on this machine — `E:\NugetLocal` is a registered source holding development builds with deliberately high version numbers, so a csproj asking for a version that is not on nuget.org does not fail: NuGet emits `NU1603` and silently floats *up* to the local package. `1.0.587` is confirmed live on nuget.org (`api.nuget.org/v3-flatcontainer/mzlib/1.0.587/mzlib.1.0.587.nupkg` returns 200).

### DIFF
--- a/MetaMorpheus/CMD/CMD.csproj
+++ b/MetaMorpheus/CMD/CMD.csproj
@@ -24,7 +24,7 @@
     <PackageReference Include="Microsoft.ML" Version="3.0.1" />
     <PackageReference Include="Microsoft.ML.CpuMath" Version="3.0.1" />
     <PackageReference Include="Microsoft.ML.FastTree" Version="3.0.1" />
-    <PackageReference Include="mzLib" Version="1.0.585" />
+    <PackageReference Include="mzLib" Version="1.0.587" />
     <PackageReference Include="Nett" Version="0.15.0" />
     <PackageReference Include="SQLite.Interop.dll" Version="1.0.103" />
     <PackageReference Include="System.Data.SQLite.Core" Version="1.0.118" />
--- a/MetaMorpheus/EngineLayer/EngineLayer.csproj
+++ b/MetaMorpheus/EngineLayer/EngineLayer.csproj
@@ -30,7 +30,7 @@
     <PackageReference Include="Microsoft.ML" Version="3.0.1" />
     <PackageReference Include="Microsoft.ML.CpuMath" Version="3.0.1" />
     <PackageReference Include="Microsoft.ML.FastTree" Version="3.0.1" />
-    <PackageReference Include="mzLib" Version="1.0.585" />
+    <PackageReference Include="mzLib" Version="1.0.587" />
     <PackageReference Include="Nett" Version="0.15.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.4" />
     <PackageReference Include="SQLite.Interop.dll" Version="1.0.103" />
--- a/MetaMorpheus/GUI/GUI.csproj
+++ b/MetaMorpheus/GUI/GUI.csproj
@@ -62,7 +62,7 @@
     <PackageReference Include="Microsoft.ML" Version="3.0.1" />
     <PackageReference Include="Microsoft.ML.CpuMath" Version="3.0.1" />
     <PackageReference Include="Microsoft.ML.FastTree" Version="3.0.1" />
-    <PackageReference Include="mzLib" Version="1.0.585" />
+    <PackageReference Include="mzLib" Version="1.0.587" />
     <PackageReference Include="Nett" Version="0.15.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.4" />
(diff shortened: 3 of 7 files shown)


trishorts:
`Validate Installer Contents` failed on the first run, and the cause is worth recording because it is not what it looks like:

```
❌ Missing DLLs from installer (not whitelisted):
- ZstdSharp.dll (from GUI\bin\Release\net10.0-windows)
```

**The bump did not add the dependency.** mzLib declared `ZstdSharp.Port` in its nuspec from **1.0.586**, in [mzLib #1179 — "declare the four dependencies missing from mzLib.nuspec"](https://github.com/smith-chem-wisc/mzLib/pull/1179). The nuspec is byte-identical between 1.0.586 and 1.0.587.

MetaMorpheus was still pinned to `1.0.585`, which never declared it — so the DLL was never restored into the output, and `Product.wxs` never had to list it. This PR is simply the first build that takes mzLib's corrected dependency set, so a gap that has been latent for two releases surfaces here.

Fixed in `7c57212` by adding the component to `ExecutablesAndLibrariesComponentGroup`.

### Worth noting for next time

This is the **paired-PR rule running in the direction our release procedure does not document**. The documented case is mzLib *dropping* a file from the nuspec, which needs its MetaMorpheus WiX removal merged **first** or the installer build fails with `WIX0103`. The inverse — mzLib *adding* a nuspec dependency — needs a component **added**, and there is nothing to merge first, because the consumer only discovers it when it bumps.

Two consequences:

1. A consumer that lags several mzLib releases accumulates these silently. The failure appears on the bump PR and looks like the bump broke something, when the change landed releases earlier.
2. `Validate Installer Contents` earned its keep here. Without it the installer would have shipped without a DLL that is present in every developer build, and failed at runtime with a load error rather than at build time with a clear message.

nbollis (approved):

trishorts:
@Alexander-Sol — small one, and it unblocks four others. 1/3 with nbollis, CI fully green (13/13).

Twelve lines: the mzLib pin in all six csprojs, plus one WiX component.

Two things worth your eye rather than a rubber stamp:

**The WiX component is the interesting part.** `Validate Installer Contents` failed on the first run for a missing `ZstdSharp.dll`, and the cause is not this PR. mzLib declared `ZstdSharp.Port` in its nuspec from **1.0.586** (mzLib #1179, "declare the four dependencies missing from mzLib.nuspec") — the nuspec is byte-identical 1.0.586 → 1.0.587. MetaMorpheus was pinned at 1.0.585, which never declared it, so the DLL was never restored and `Product.wxs` never had to list it. This is simply the first build taking mzLib's corrected dependency set.

That is the paired-PR rule running the direction our release procedure does not document — the known case is mzLib *dropping* a nuspec file needing a WiX removal merged first; this is mzLib *adding* one, which needs a component added and has nothing to merge first.

**Nothing in 1.0.587 changes MetaMorpheus's behaviour**, and I checked each rather than assuming: the new `QuantifiedPeaks.tsv` columns are written through FlashLFQ and never parsed back; `QuantificationEngine`'s new ambiguous-match exclusion was already being avoided by `PostSearchAnalysisTask`'s own `includeAmbiguous: false`; and `SumCollapse`, `MeanCollapse` and `SumRollUp.ArrayPool` — removed in 1.0.587 — are referenced nowhere on master.

Merging this fixes the identical installer failure currently red on #2718 and #2755.

acesnik (approved):

pcruzparri (approved):

PR #2766, opened 2026-08-27
Write calibrated spectra as mgf (#1673)

### LINKED ISSUE #1673: mgf as output option
User requested that we write .mgf as an output option.

acesnik:
🤖 Closes #1673. Three commits: the output option, a search fix the option exposed, and a portable mgf variant.

**Blocked on smith-chem-wisc/mzLib#1227 and a release.** The branch pins mzLib 1.0.586 and will not compile until the pin moves to whatever version carries that PR — `ExportAsMgf(path, includeMs1Scans:)` and the extension headers both come from it. I have left the pin wrong rather than guess a version number.

## What you get

`CalibrationParameters.OutputFormat`, defaulting to `MzML` so nothing changes unless asked. Exposed as `OutputFormat = "MzML"` in the task toml and as a dropdown in the Calibrate task window. The calibrated file's extension, the experimental design entry and the write all follow it, so a Calibrate → Search chain hands the next task a file whose extension matches what was written.

| | PSMs | peptides | protein groups | precursors |
|---|---|---|---|---|
| `MzML` | 80 | 55 | 46 | 482 |
| `Mgf` | 80 | 55 | 46 | 482 |
| `MgfMs2Only` | 63 | 44 | 36 | 118 |

`Mgf` and `MzML` agree byte for byte across all 57 columns of `AllPSMs.psmtsv`.

## The search fix, which is the substantive part

`SearchTask` switched precursor deconvolution off for anything typed as `Mgf` or `Ms2Align`. The premise — an mgf has no MS1 scans, only header precursor information — holds for every third-party mgf and not for one mzLib writes. The symptom was **exactly one precursor per MS2 scan**, 118 against mzML's 482, which is what made it findable.

Now gated on whether the file can support deconvolution: every MS2 scan must name an MS1 scan present in the file, and must have a resolvable `IsolationRange`. Both are required — without a precursor scan number `_GetMs2Scans` falls through to the scan header, and a null `IsolationRange` makes `GetIsolatedMassesAndCharges` return nothing. A partially annotated file goes header-only for the whole file rather than deconvoluting some scans and not others.

msalign keeps the unconditional type check; it is pre-deconvoluted by construction.

**This changes search output for mgf files carrying MS1 scans, and only those.** All 17 checked-in mgf fixtures have zero MS1 scans and every one still takes the header-only path. I also checked three synthetic degraded files — our headers stripped, one block missing `PRECURSORSCAN`, one missing `ISOLATIONWIDTH` — and all three fall back.

One case I could not make fall back: dropping only `ISOLATIONMZ` still deconvolutes, because the reader defaults the isolation centre to `PEPMASS` so `IsolationRange` stays non-null. Reaching it needs MS1 blocks *and* `PRECURSORSCAN` *and* `ISOLATIONWIDTH` but no `ISOLATIONMZ` — an mzLib-written mgf someone has edited. I left it rather than add a check that only appears to cover it.

## Why `MgfMs2Only` is a third enum value

The mgf `Mgf` writes includes MS1 scans, which is what preserves the 80 PSMs. Those blocks have no `PEPMASS`, which the spec requires, and MSToolkit — Comet's reader — answers with `Error in MGF file: no PEPMASS found.` then `exit(-12)`. ProteoWizard and OpenMS read it fine. `MgfMs2Only` writes MS2 and above only, so every block carries the mandatory `PEPMASS`.

A third enum value rather than a checkbox because the choice is meaningless when the format is mzML, and a control that does nothing two thirds of the time is worse. The tooltip explains per value which readers need which.

## Tests

11 across `MgfPrecursorGateTests.cs` and `MgfOutputTests.cs`, all passing: six gate cases (three negative), an all-or-nothing case, a write/read round trip, toml round trip of all three enum values, the `MzML` default, an end-to-end calibration per value, and an end-to-end search proving deconvolution happens only with MS1 scans present. That last one was confirmed to fail against the old type check with `Expected: greater than 118, But was: 118`.

`HasDeconvolutablePrecursorScans` is `internal` with `InternalsVisibleTo("Test")`, following `EngineLayer/Util/PrecursorSet.cs`.

Not verified: `Test.csproj` is Windows-targeted, so these ran through a throwaway net10.0 harness. It compiles under `-p:EnableWindowsTargeting=true` and the full solution builds clean under `UbuntuMac`; CI is the check for the rest of the suite.

### DIFF
--- a/MetaMorpheus/TaskLayer/CalibrationTask/CalibrationTask.cs
+++ b/MetaMorpheus/TaskLayer/CalibrationTask/CalibrationTask.cs
@@ -49,6 +49,15 @@ namespace TaskLayer
 
         public const string CalibSuffix = "-calib";
 
+        /// <summary>
+        /// Extension for the calibrated file, following <see cref="CalibrationParameters.OutputFormat"/>.
+        /// The next task in the chain is handed this file, so the extension has to match what was written
+        /// or the reader picked for it will not be the one that wrote it.
+        /// </summary>
+        private bool WritingMgf => CalibrationParameters.OutputFormat == SpectraFileOutputFormat.Mgf;
+
+        private string CalibratedFileExtension => WritingMgf ? ".mgf" : ".mzML";
+
 
         private List<string> _unsuccessfullyCalibratedFilePaths;
         private string _taskId;
@@ -85,8 +94,9 @@ namespace TaskLayer
                 string originalUncalibratedFilePath = currentRawFileList[spectraFileIndex];
                 string uncalibratedNewFullFilePath = Path.Combine(outputFolder, Path.GetFileName(currentRawFileList[spectraFileIndex]));
                 string originalUncalibratedFilenameWithoutExtension = Path.GetFileNameWithoutExtension(originalUncalibratedFilePath);
-                string calibratedNewFullFilePath = Path.Combine(OutputFolder, originalUncalibratedFilenameWithoutExtension + CalibSuffix + ".mzML")
-                    .ToSafeOutputPath(CalibSuffix + ".mzML");
+                string calibratedExtension = CalibratedFileExtension;
+                string calibratedNewFullFilePath = Path.Combine(OutputFolder, originalUncalibratedFilenameWithoutExtension + CalibSuffix + calibratedExtension)
+                    .ToSafeOutputPath(CalibSuffix + calibratedExtension);
 
                 // mark the file as in-progress
                 StartingDataFile(originalUncalibratedFilePath, new List<string> { _taskId, "Individual Spectra Files", originalUncalibratedFilePath });
@@ -187,7 +197,7 @@ namespace TaskLayer
             assumedPathToExperDesign = Path.Combine(assumedPathToExperDesign, GlobalVariables.ExperimentalDesignFileName);
             if (File.Exists(assumedPathToExperDesign))
             {
-                WriteNewExperimentalDesignFile(assumedPathToExperDesign, outputFolder, currentRawFileList, _unsuccessfullyCalibratedFilePaths);
+                WriteNewExperimentalDesignFile(assumedPathToExperDesign, outputFolder, currentRawFileList, _unsuccessfullyCalibratedFilePaths, CalibratedFileExtension);
             }
         }
 
@@ -523,7 +533,21 @@ namespace TaskLayer
 
         private void CalibrationOutput(MsDataFile msDataFile, string mzFilePath, FileSpecificParameters fileParams, string tomlName, string taskId, string mzFilenameNoExtension)
         {
-            msDataFile.ExportAsMzML(mzFilePath, CalibrationParameters.WriteIndexedMzml);
+            if (WritingMgf)
+            {
+                // Temporary, and should be deleted once mgf carries these: dissociation type, precursor scan
+                // number and isolation width have no header in the format today, so a task run after this one
+                // sees fewer fields. Measured on the bundled yeast file, a Calibrate -> Search chain finds 85
+                // PSMs through mzML and 61 through mgf.
+                Warn("Writing calibrated spectra as mgf. Dissociation type, precursor scan number and "
+                     + "isolation width are not carried in mgf today, so a task run after this one will "
+                     + "identify fewer spectra than it would from mzML.");
+                msDataFile.ExportAsMgf(mzFilePath);
+            }
+            else
+            {
+                msDataFile.ExportAsMzML(mzFilePath, CalibrationParameters.WriteIndexedMzml);
+            }
             MyTaskResults.NewSpectra.Add(mzFilePath);
             Toml.WriteFile(fileParams, tomlName, tomlConfig);
             FinishedWritingFile(tomlName, new List<string> { taskId, "Individual Spectra Files", mzFilenameNoExtension });
@@ -532,7 +556,7 @@ namespace TaskLayer
         }
 
         private static void WriteNewExperimentalDesignFile(string pathToOldExperDesign, string outputFolder, List<string> originalUncalibratedFileNamesWithExtension,
-            List<string> unsuccessfullyCalibratedFilePaths)
+            List<string> unsuccessfullyCalibratedFilePaths, string calibratedExtension)
         {
             List<SpectraFileInfo> oldExperDesign = ExperimentalDesign.ReadExperimentalDesign(pathToOldExperDesign, originalUncalibratedFileNamesWithExtension, out var errors);
 
@@ -558,7 +582,7 @@ namespace TaskLayer
                 }
                 else
                 {
-                    SpectraFileInfo calibratedSpectraFile = new(Path.Combine(outputFolder, originalFileName + CalibSuffix + ".mzML"),
+                    SpectraFileInfo calibratedSpectraFile = new(Path.Combine(outputFolder, originalFileName + CalibSuffix + calibratedExtension),
                     originalSpectraFile.Condition, originalSpectraFile.BiologicalReplicate, originalSpectraFile.TechnicalReplicate, originalSpectraFile.Fraction);
                     newExperDesign.Add(calibratedSpectraFile);
                 }
--- a/MetaMorpheus/TaskLayer/CalibrationTask/CalibrationTask.cs
+++ b/MetaMorpheus/TaskLayer/CalibrationTask/CalibrationTask.cs
@@ -54,7 +54,8 @@ namespace TaskLayer
         /// The next task in the chain is handed this file, so the extension has to match what was written
         /// or the reader picked for it will not be the one that wrote it.
         /// </summary>
-        private bool WritingMgf => CalibrationParameters.OutputFormat == SpectraFileOutputFormat.Mgf;
+        private bool WritingMgf => CalibrationParameters.OutputFormat is SpectraFileOutputFormat.Mgf
+                                                                       or SpectraFileOutputFormat.MgfMs2Only;
 
         private string CalibratedFileExtension => WritingMgf ? ".mgf" : ".mzML";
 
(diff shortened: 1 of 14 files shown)


trishorts (commented):
Read this as a draft, so the comments are about shape rather than blockers. The substance is good and the search fix is the valuable half — "the premise holds for every third-party mgf and not for one mzLib writes" is exactly the right diagnosis, and 482 precursors against 118 is the kind of number that makes a bug findable instead of arguable.

Three comments inline. The one that matters is where `HasDeconvolutablePrecursorScans` lives. `SearchTask` is the only one of `GetMs2Scans`'s four production callers that gets the gate, and `MetaMorpheusTask.cs:283-287` dereferences `ms2scan.IsolationRange` with no null check, so the partially-annotated mgf your own doc comment describes is an NRE in GPTMD, GlycoSearch and Calibrate. Moving the predicate onto `MetaMorpheusTask` fixes that and drops the need for the first TaskLayer `InternalsVisibleTo`; pushing it into mzLib#1227 as an `MsDataFileExtensions` member would be better still, since every fact it reads is mzLib's and it is the round-trip assertion for the feature #1227 adds.

Verified rather than assumed, since several of these turn on release ordering:

- Outside `Test`, `DoPrecursorDeconvolution` is assigned in TaskLayer exactly once, at `SearchTask.cs:341`. `GPTMDTask.cs:137`, `GlycoSearchTask.cs:109` and `CalibrationTask.cs:219` all call `GetMs2Scans` ungated.
- On mzLib master the mgf reader passes `null` for both `isolationWidth` and `oneBasedPrecursorScanNumber` (`Readers/Mgf/Mgf.cs:271-275`), and `IsolationRange` requires both `IsolationWidth` and `IsolationMz` (`MsDataScan.cs:135-145`). So against the pinned 1.0.586 the new predicate is always false and the new gate is behaviourally identical to the old type check.
- `CanFileTypeBeCalibrated` still refuses `.mgf`, so the new output is not re-calibratable.

The enum I would leave exactly where it is. mzLib has no write-side format vocabulary to extend — `SupportedFileType` is a read-side extension-to-reader map whose contract is a one-to-one extension round trip, which `Mgf` and `MgfMs2Only` would break by colliding on `.mgf` — and `MgfMs2Only` is not a container anyway, it is `Mgf` plus `includeMs1Scans: false`. A task-settings enum in TaskLayer beside `SearchType` is the right home, and giving it its own file mirrors `SearchTask/SearchType.cs`. The toml round trip needs no `ConfigureType`, which `TestToml.cs:34-46` already demonstrates for `MassDiffAcceptorType`; worth adding the equivalent assertion for `OutputFormat` to the Calibrate case at `TestToml.cs:64-65`, which currently writes and reads back but asserts nothing about the result.

The GUI wiring is clean, incidentally, and I went looking for the opposite: `Enum.GetValues` into the ComboBox with a cast back on save avoids the radio-button-plus-MessageBox pattern the other task windows use, which is the pattern that breaks whenever an enum gains a member. `UpdateFieldsFromTask` runs on both construction paths, so the cast on save cannot see a null selection.

One small thing not worth its own comment: `WriteIndexedMzml` stays enabled and meaningless when an mgf format is selected. The XML doc says it is ignored; the checkbox does not.

MetaMorpheus/TaskLayer/SearchTask/SearchTask.cs
```diff
@@ -155,6 +158,30 @@ private void MigrateLegacyMostAbundantRequest()
...
 
+        /// <summary>
+        /// True only when every MS2 scan names an MS1 scan that is present in the file and has a resolvable
+        /// isolation range. Both are required to deconvolute a precursor: without a precursor scan number
+        /// <see cref="MetaMorpheusTask._GetMs2Scans"/> falls through to the scan header, and a null IsolationRange
+        /// makes GetIsolatedMassesAndCharges return nothing. IsolationRange needs an isolation width as well as a
+        /// centre, and the MGF reader defaults only the centre. A partially annotated file is treated as
+        /// header-only rather than deconvoluted for some scans and not others.
+        /// </summary>
+        internal static bool HasDeconvolutablePrecursorScans(MsDataFile file)
```
trishorts, line 169: **The right idea in the wrong class — and the file this makes producible reaches three other tasks that have no gate at all.**

The gate you are replacing is the only one of its kind: outside `Test`, `DoPrecursorDeconvolution` is assigned in TaskLayer exactly once, at line 341 here. But `GetMs2Scans` has four production callers:

- `SearchTask.cs:338` — gated, by this PR
- `GPTMDTask.cs:137` — no gate
- `GlycoSearchTask.cs:109` — no gate
- `CalibrationTask.cs:219` — no gate

Today that costs nothing, and it is worth being precise about why, because it is exactly what this PR changes. mzLib's mgf reader on master passes literal `null` for both `isolationWidth` and `oneBasedPrecursorScanNumber` (`Readers/Mgf/Mgf.cs:271-275`, the two nulls straddling `DissociationType.Unknown`). So for every mgf, `OneBasedPrecursorScanNumber` is null, the whole deconvolution branch in `_GetMs2Scans` is skipped, and everything falls through to `PrecursorFromScanHeader`. The existing gate is about setting `UseProvidedPrecursorInfo`, not about avoiding a crash.

mzLib#1227 changes that, and this PR is what turns such files from a curiosity into ordinary output. Once an mgf carries `PRECURSORSCAN`, the branch is entered, and `MetaMorpheusTask.cs:283-287` dereferences `ms2scan.IsolationRange.Minimum` and `.Maximum` inside the `DoPrecursorDeconvolution` block with no null check. `IsolationRange` is non-null only when both `IsolationWidth` and `IsolationMz` have values (`MsDataScan.cs:135-145`). So an mgf with `PRECURSORSCAN` but one block missing `ISOLATIONWIDTH` — precisely the partially-annotated file your doc comment names — is a NullReferenceException in GPTMD, GlycoSearch and Calibrate, and only Search is protected. The all-or-nothing choice is the right one; it just needs to cover all four callers rather than one.

Which is the structural point: this belongs on `MetaMorpheusTask`, beside `_GetMs2Scans`, not on `SearchTask`. That is the class all four callers already inherit from, it is where every other file-shape question in this stack lives, and it removes the need for line 28 — the first `[assembly: InternalsVisibleTo("Test")]` on the TaskLayer assembly. The placement idiom itself is fine and has two precedents (`EngineLayer/Util/PrecursorSet.cs:8`, `GuiFunctions/MetaDraw/MetaDrawLogic.cs:29`), but granting Test access to every TaskLayer internal is a real widening, and not something to acquire as a side effect of one predicate.

The stronger version of the argument, worth weighing while you still own both sides: every fact this reads — `MsnOrder`, `OneBasedPrecursorScanNumber`, `IsolationRange` — is mzLib's, and mzLib already encodes half the rule silently and per-scan at `MsDataScan.cs:187`, where a null `IsolationRange` returns no envelopes. The file-level generalization is the natural round-trip assertion for the feature #1227 adds, and only mzLib can state it: an mgf written with `includeMs1Scans: true` is re-deconvolutable, one written `false` or by a third party is not. As an `MsDataFileExtensions` member in #1227 it would ship with the tests that already construct these scan shapes, and FlashLFQ would get it too. As `internal` on `SearchTask` it is stranded.

One last thing specific to this method: against the pinned mzLib 1.0.586 it cannot return `true` for any input, for the reader reason above. That makes the new gate behaviourally identical to the old type check until the pin moves — which you have flagged — but it also means `MgfPrecursorGateTests` exercises a path the shipped dependency cannot reach, so those greens are not yet evidence about the real behaviour.

MetaMorpheus/TaskLayer/CalibrationTask/CalibrationTask.cs
```diff
@@ -49,6 +49,16 @@ public CalibrationTask() : base(MyTask.Calibrate)
...
 
+        /// <summary>
+        /// Extension for the calibrated file, following <see cref="CalibrationParameters.OutputFormat"/>.
+        /// The next task in the chain is handed this file, so the extension has to match what was written
+        /// or the reader picked for it will not be the one that wrote it.
+        /// </summary>
+        private bool WritingMgf => CalibrationParameters.OutputFormat is SpectraFileOutputFormat.Mgf
+                                                                       or SpectraFileOutputFormat.MgfMs2Only;
+
+        private string CalibratedFileExtension => WritingMgf ? ".mgf" : ".mzML";
```
trishorts, line 60: **Calibrate refuses `.mgf` as input, so this produces an output its own task type cannot re-consume.**

`CanFileTypeBeCalibrated` (line 423, unchanged here) rejects `.mgf` outright, and `RunSpecific` skips such files at line 85 with the warning at 416. So a Calibrate then Calibrate chain, or any re-run of calibration over a previous output folder, now meets a file the task declines. And because `WriteNewExperimentalDesignFile` writes the `.mgf` path into the experimental design (line 201), the refusal travels with the design file rather than staying local to one run.

Defensible for an opt-in option whose default is mzML, but it undercuts the line in the description that the extension follows the format "so a Calibrate then Search chain hands the next task a file whose extension matches what was written". True for Search; not true for Calibrate. One sentence naming which downstream tasks accept the output would settle it.

The related one, since both are silent: FlashLFQ quantification needs MS1 scans. `MgfMs2Only` produces a file that cannot be quantified at all, and nothing in the chain says so, so a Calibrate(MgfMs2Only) then Search-with-quantification run simply quantifies nothing. The tooltip explains the identification cost well; the quantification cost is the one a user is least likely to predict from "MS2 and above only".

MetaMorpheus/TaskLayer/CalibrationTask/CalibrationTask.cs
```diff
@@ -523,7 +534,28 @@ public void CalibrationWarnMessage(DataPointAquisitionResults acquisitionResults
 
         private void CalibrationOutput(MsDataFile msDataFile, string mzFilePath, FileSpecificParameters fileParams, string tomlName, string taskId, string mzFilenameNoExtension)
         {
-            msDataFile.ExportAsMzML(mzFilePath, CalibrationParameters.WriteIndexedMzml);
+            if (WritingMgf)
```
trishorts, line 537: **Both branches warn on every file written, and neither matches how `Warn` is used elsewhere in TaskLayer.**

Every `Warn` in TaskLayer today falls into one of three buckets: a refusal (`CalibrationTask.cs:416`, "Calibration for .mgf files is not supported"), a report of a failure or of an otherwise-silent fallback (`CalibrationTask.cs:502-520`; `SearchTask.cs:342`, whose comment reads "That is intended, but silent — report how often it happens so it is noticed"), or a legacy-parameter migration (`SearchTask.cs:150`). There is no precedent for warning on every exercise of a valid, deliberately-chosen option — and this sits inside `CalibrationOutput`, so a 20-file run emits 20 identical warnings for a dropdown the user set once, on purpose.

I would keep the `MgfMs2Only` one and drop the other. Dropping MS1 really is information loss a user may not have connected to a fifth of the identifications, which puts it squarely in the `SearchTask.cs:342` bucket. The `Mgf` warning restates the tooltip they read while choosing it. If both stay, hoist them to once per task rather than once per file.

trishorts:
Status note rather than new review feedback -- my earlier comments still stand as written.

**This does not compile, and the cause is upstream rather than anything in this diff.** All three platform legs plus the four Windows test jobs fail with the same error:

```
CalibrationTask.cs(553,28): error CS1501: No overload for method 'ExportAsMgf' takes 2 arguments
```

The two-argument form comes from smith-chem-wisc/mzLib#1227 (*Carry precursor metadata through mgf, and trim mgf peaks by MS level*), which changes the signature to `ExportAsMgf(this MsDataFile file, string destinationPath, bool includeMs1Scans = true)`. That PR is open and unreleased, so no published mzLib carries the overload this branch calls.

So the ordering is: mzLib#1227 merges, a release goes out, this pins to it. Worth naming #1227 in the description as the counterpart -- without that line the red CI reads as a defect in this PR rather than a dependency, and a reviewer has to go and find it. The branch is also `CONFLICTING` with master now.

Nothing to do about it here beyond the sequencing; flagging it so it is not mistaken for something reviewable.

PR #2767, opened 2026-08-27
Bugfix: RNA Combos was not added to the installer. It is now

nbollis:
This causes RNA GPTMD to crash once installed, now we can run fine and dandy


MetaMorpheus/MetaMorpheusSetup/Product.wxs
```diff
@@ -668,6 +668,11 @@
         <RegistryValue Root="HKCU" Key="Software\SmithGroup\MetaMorpheus" Name="installed" Type="integer" Value="1" KeyPath="yes" />
         <File ReadOnly="no" Id="Appsrc_Data_combos.txt" Name="combos.txt" Source="$(var.GUI_TargetDir)Data\combos.txt" />
       </Component>
+      <Component Id="Appsrc_Data_rnaCombos.txt" Guid="6f1a9a8d-5f0d-4cbe-a1eb-f1f7fa7e8f5b" Condition="OVERWRITEAPPDATA">
+
+        <RegistryValue Root="HKCU" Key="Software\SmithGroup\MetaMorpheus" Name="installed" Type="integer" Value="1" KeyPath="yes" />
+        <File ReadOnly="no" Id="Appsrc_Data_rnaCombos.txt" Name="rnaCombos.txt" Source="$(var.GUI_TargetDir)Data\rnaCombos.txt" />
+      </Component>
```
trishorts, line 675: Worth knowing while you are in here: two sibling folders have the identical omission, and unlike `rnaCombos.txt` they fail silently.

`Mods\ProteaseMods.txt` (`EngineLayer.csproj:105`) and `Glycan_Mods\OGlycan\OGlycan_withIsobaric.gdb` (`EngineLayer.csproj:78`) are both copied to the build output and both absent from `Product.wxs`. Their folders are read by enumeration —

```csharp
foreach (var modFile in Directory.GetFiles(Path.Combine(DataDir, @"Mods")))                        // GlobalVariables.cs:415
foreach (var glycanFile in Directory.GetFiles(Path.Combine(DataDir, @"Glycan_Mods", @"OGlycan")))  // GlobalVariables.cs:484
```

— so an installed build just has fewer mods and one fewer glycan database, with no error. `rnaCombos.txt` threw only because `LoadCombos` names its file.

Fine to leave them out of this PR; they are in #2769 with the CI gap that let all three through. Flagging here in case you would rather add the two components now while the file is open.

trishorts (approved):
🤖 Approving. The fix is right, and I checked the two things a hand-listed WiX component can quietly get wrong.

**It will actually install.** The new `<Component>` sits inside the `Appsrc_Data_files` `ComponentGroup`, which `ProductFeature` references at `Product.wxs:32` — a component in no feature is compiled and never installed, and that is not what happened here. The GUID is unique across all 148 file components, and `Condition="OVERWRITEAPPDATA"` matches every sibling in the group. The `Component`/`File` id collision (`Appsrc_Data_rnaCombos.txt` used for both) is legal and is what `combos.txt` does two lines above.

**It fixes the reported crash, and the diagnosis holds.** `GptmdTask.LoadCombos` is an iterator, so its `StreamReader` on `Path.Combine(GlobalVariables.DataDir, "Data", specificPath)` does not open until `.ToList()` at `GPTMDTask.cs:64` forces it — which is why the failure lands inside GPTMD rather than at startup. And it only bites installed users: `SetUpDataDirectory` (`GlobalVariables.cs:358-381`) points `DataDir` at `%LOCALAPPDATA%\MetaMorpheus` under Program Files, populated only by the `OVERWRITEAPPDATA` components here, and at the build output everywhere else, populated by `EngineLayer.csproj`.

**`rnaCombos.txt` was the only `Data\` omission.** Diffed all nine tracked files in `EngineLayer/Data` against the `Source=` values in the group; the other eight are all present.

Two things beyond this diff, neither blocking, both now written up as #2769 rather than left in a comment thread:

- The same omission is live for `Mods\ProteaseMods.txt` and `Glycan_Mods\OGlycan\OGlycan_withIsobaric.gdb`. Both folders are read by `Directory.GetFiles` enumeration rather than by name, so those degrade silently — a missing `Protease` mod type and a missing entry in the O-glycan database dropdown, no exception — which is presumably why they have outlived this one.
- `Validate Installer Contents` would be the natural guard and cannot be: it is `-Filter *.dll` on both sides (`InstallRunAndArtifact.yml:320`, `:355`), and the content folders install to `%LOCALAPPDATA%`, not `$installDir`. The issue has the shape of the fix.

**Coordination:** #2749 also edits `Product.wxs` and deletes four `Data\` files, so whichever of the two lands second will need a rebase there. It leaves `rnaCombos.txt` alone, so it does not overlap in substance — only in the file.

acesnik (approved):

RayMSMS (approved):

PR #2771, opened 2026-08-28
Name iTRAQ 8-plex's eighth channel 121, which is the one that exists

trishorts:
`GetReporterIonLabels` ended iTRAQ 8-plex `"119", "120"`. There is no 120 channel in the 8-plex kit — it is skipped because the phenylalanine immonium ion sits at 120.081 — and the eighth reagent is **121**.

**The m/z was always right.** `H12 C{13}6 N{15}2` plus a proton is 121.1215 against a published 121.122, so MetaMorpheus has been extracting the correct ion and reporting it under a name no kit sells. The `# 120` comment on that DI line in `Mods/tmt.txt` is where it came from.

**This reaches users.** `GetReporterIonLabels` *is* the reporter-ion column header in the `.psmtsv` (`PostSearchAnalysisTask.WritePsmPlusMultiplexIons`), so every iTRAQ 8-plex result file already written names a channel that does not exist. The values under it are correct; only the header is wrong. Anyone joining on that column name is joining on a channel no instrument produced.

Extracted from #2598 to stand alone, because it is wrong data in shipped output and should not wait on a 3,300-line feature review. It is also why #2598 and #2755 had a merge-order constraint — both touch this file and only #2598 carried the fix — so landing it here removes that.

## The test is the point of the PR

A hardcoded label list can only ever pin whatever was there when it was written, which is how this survived. `EveryTagsLabelsNameTheChannelAtTheirOwnIndex` asserts the invariant that actually matters instead: every consumer pairs `channelLabels[i]` with `ReporterIonMzs[i]` **positionally**, so label `i` must name the channel whose reporter ion sits at `i`. Every label begins with its nominal mass, so it is cheap, and it covers all eight tag types rather than the one that was wrong.

Verified it catches the original bug rather than merely passing — against the old label:

```
iTRAQ8 label 120 at index 7 names channel 120,
but the reporter ion at that index is 121.1215 (channel 121)
```

**It also checks the N/C suffix,** which the nominal mass cannot see. `127N` and `127C` are both `"127"`, so a swapped suffix would pass a nominal-mass check while pairing each channel with the others ion — and that is the same family as the `131`/`131N` disagreement between mzLibs two hardcoded label lists. They are separable by mass rather than convention: at one nominal mass the N form is lighter, and `ReporterIonMzs` is sorted ascending, so N must precede C. Verified against a mutant that swaps TMT10s pair:

```
TMT10 orders 127C before 127N at index 1, but the N form is the lighter of the
pair and the reporter ions ascend (127.1248 then 127.1311)
```

The existing `TestGetReporterIonLabelsWithTagType` pinned `"120"`, so it **moved** to `"121"` rather than being loosened, with a pointer to the index test beside it.

97 multiplex, isobaric and TMT tests green.

### DIFF
--- a/MetaMorpheus/EngineLayer/Util/IsobaricMassTag.cs
+++ b/MetaMorpheus/EngineLayer/Util/IsobaricMassTag.cs
@@ -194,7 +194,13 @@ namespace EngineLayer
                 IsobaricMassTagType.TMT11 => new List<string> { "126", "127N", "127C", "128N", "128C", "129N", "129C", "130N", "130C", "131N", "131C" },
                 IsobaricMassTagType.TMT18 => new List<string> { "126", "127N", "127C", "128N", "128C", "129N", "129C", "130N", "130C", "131N", "131C", "132N", "132C", "133N", "133C", "134N", "134C", "135N" },
                 IsobaricMassTagType.iTRAQ4 => new List<string> { "114", "115", "116", "117" },
-                IsobaricMassTagType.iTRAQ8 => new List<string> { "113", "114", "115", "116", "117", "118", "119", "120" },
+                // iTRAQ 8-plex has no 120 channel: it is skipped because the phenylalanine immonium
+                // ion sits at 120.081, and the eighth reagent is 121. The m/z extracted here has
+                // always been the right one -- H12 C{13}6 N{15}2 plus a proton is 121.1215 against a
+                // published 121.122 -- so only the NAME was wrong, and it reached users as a column
+                // header in the .psmtsv. The "# 120" comment on that DI line in Mods/tmt.txt is where
+                // it came from.
+                IsobaricMassTagType.iTRAQ8 => new List<string> { "113", "114", "115", "116", "117", "118", "119", "121" },
                 IsobaricMassTagType.diLeu4 => new List<string> { "115", "116", "117", "118" },
                 IsobaricMassTagType.diLeu12 => new List<string> { "115a", "115b", "116a", "116b", "116c", "117a", "117b", "117c", "118a", "118b", "118c", "118d" },
                 _ => null
--- a/MetaMorpheus/Test/Multiplex_Labeling_TMT_iTRAQ.cs
+++ b/MetaMorpheus/Test/Multiplex_Labeling_TMT_iTRAQ.cs
@@ -870,7 +870,10 @@ namespace Test
             Assert.That(itraq4Labels, Is.EqualTo(new List<string> { "114", "115", "116", "117" }));
 
             var itraq8Labels = IsobaricMassTag.GetReporterIonLabels(IsobaricMassTagType.iTRAQ8);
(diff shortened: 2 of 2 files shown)


acesnik:
🤖 Nice find. A reporter-ion label that names a channel the kit doesn't have is the sort of thing that survives for years because the extracted intensities are correct and only the header lies — so anyone checking the numbers finds nothing wrong.

I verified the chemistry independently rather than taking it on faith. The eighth iTRAQ-8plex diagnostic-ion line at `EngineLayer/Mods/tmt.txt:279` (and `:301` for the K motif) is `HCD: H12 C{13}6 N{15}2# 120`, which is 120.1142 neutral and **121.1215** as a singly-protonated m/z — the 121 reagent. 113–119 plus 121 is the AB Sciex 8-plex set; 120 is skipped precisely because the Phe immonium ion sits at 120.081. So the ion was always right and only the name was wrong, exactly as you describe.

**Output change, for the record:** one column header in an iTRAQ 8-plex `.psmtsv` goes `120` → `121`. `GetReporterIonLabels` has exactly one non-test caller, `PostSearchAnalysisTask.cs:1311`, and it only `string.Join`s the list into the header — the values come from `psm.IsobaricMassTagReporterIonIntensities` positionally. So no value moves, and nothing inside MetaMorpheus keys on the label. External scripts joining on the header name are the only thing affected.

**No merge-order constraint with mzLib**, which was the thing I most expected to bite. mzLib's psmtsv reader would not stop matching the renamed column: `SpectrumMatchFromTsvHeader.TmtChannelNames` — the list both `SpectrumMatchFromTsv.ParseReporterIonColumns` and `LightWeightSpectralMatch` iterate — is 126, 127N/C … 135N only, with no iTRAQ or DiLeu channel at all. Confirmed against mzLib `upstream/master` source *and* against the packaged artifact's string constants. The reader never saw `120` either, so this can land in any order.

**The index test does what you say it does.** I checked the label/ion pairing by hand for all eight types rather than trusting the count assertion. DiLeu-12 is the interesting case: `tmt.txt` lists its DI ions as 115a, 116a, 116b, 116c, 115b, 117a, …, but sorting by computed m/z reproduces the code's `115a, 115b, 116a, …` order exactly — so the positional assertion is meaningful there rather than accidentally passing.

Two small things, neither blocking. `if (tag == null) continue;` makes the whole loop a silent no-op if the mods fail to load — `TestGetIsobaricMassTagWithTagType` already asserts all eight non-null so it can't pass vacuously in practice, but a checked count at the end is one line and removes the shape. And of the 7 added lines in `IsobaricMassTag.cs`, one is code; the 6-line comment restates the PR body, as does the 3-line comment above the moved assertion — the body is the better home for it.

Not verified: I didn't run the tests (`Test/Test.csproj` is `net*.0-windows` and won't execute here), so the test claims rest on reading plus your green Windows CI. I also didn't run an actual iTRAQ 8-plex search, so "the values under the column are unchanged" is read off the writer rather than measured against real spectra, and I didn't check the wiki for a hardcoded `120`.

acesnik (approved):

RayMSMS (approved):

nbollis (approved):

PR #2772, opened 2026-08-28
Make the protein table rectangular when intensities are omitted, not zeroed

trishorts:
`AllProteinGroups.tsv` writes one header — from `proteinGroups.First()` — then a row per group. The `Intensity_` columns were gated on `SampleGroupResult.HasIntensityData`, which is **per-group**, so a protein the quantifier found nothing for emits two fewer columns than its neighbour and **every value after the gap lands under the wrong column name**. No exception, no warning.

Measured on two groups sharing a sample list where one had nothing observed: **header 28 fields, row 26**.

## It has never fired, and that is an accident about to end

`QuantificationAnalysis` assigns `IntensitiesByFile` for every protein *including unmeasured ones*, so the zeros are present and `HasIntensityData` is uniformly true. `TestMultipleProteinGroupsHeaderAndRowsHaveSameColumnCount` pins exactly that, and its own comment says so:

> mirrors how QuantificationAnalysis always assigns IntensitiesByFile, even for unmeasured proteins

mzLibs `QuantificationEngine` **omits** zero-valued cells instead, leaving an assigned-but-empty dictionary. That is what the isobaric path produces once it starts calling that engine — so this lands before #2755 rather than after it.

## The fix

The gate is now `HasAssignedSampleIntensities` — a non-empty sample list **and** an assigned dictionary. Whether a column *exists* has to be a property of the group as a whole, or a row cannot line up with a header written from a different group.

This is the same reasoning the class already applies elsewhere: `HasPeptideLevelQuantification` is documented as *"assigned uniformly so every group carries the same schema."* That was applied to `IntensityOccupancy_` and not to `Intensity_`.

`WasSearched` is untouched — it answers a different question (whether a sample groups files were searched, which SILAC needs because quantification invents one file per label). The SILAC tests confirm it still behaves.

## Three tests, and two exist because the first cannot see them

Each was verified by mutating the gate and watching the right one fail while the others passed:

| test | kills |
|---|---|
| rectangularity, one group with nothing observed | the original per-group gate (28 vs 26 before the fix) |
| samples assigned, no dictionary | `\|\|` in place of `&&` |
| no samples at all, but PSMs present | the non-empty check weakened to a null check |

The third needs the PSM to be observable: with none there are no sample groups and the row is empty either way. That is the shape a subset group takes when no sample matches the requested file — `ConstructSubsetProteinGroup` assigns an empty sample list beside an empty-but-non-null dictionary, which is the ordinary case for a protein not seen in that file.

Also noted on `TestTmtProteinGroupsHaveCountColumnsButNoIntensityColumns` that the absence it asserts describes **master**, where the multiplex branch returns before quantifying anything — not a claim that TMT output should have no intensity columns. Its `AllSame()` assertion is the half that stays true either way, and the design-file-present case still has no coverage in either repo.

## Counterpart

**mzLib #1240** makes the same correction in `BioPolymerGroup` and exposes the predicate as `protected`. The two are independent — this one fixes MetaMorpheuss own copy and needs no mzLib release. When the pin moves past #1240, this private copy goes away and `ProteinGroup` overrides the header rather than hiding it with `new`, which also closes a latent dispatch hole: `IBioPolymerGroup` declares `GetTabSeparatedHeader`, so an interface-typed call today would pair mzLibs header with MetaMorpheuss row.

133 tests green across the ProteinGroup, multiplex, SILAC and parsimony suites; full suite was green before merging the current master.

### DIFF
--- a/MetaMorpheus/EngineLayer/ProteinParsimony/ProteinGroup.cs
+++ b/MetaMorpheus/EngineLayer/ProteinParsimony/ProteinGroup.cs
@@ -152,6 +152,30 @@ namespace EngineLayer
         /// </summary>
         public HashSet<string> SearchedSpectraFilePaths { get; set; }
 
+        /// <summary>
+        /// Whether sample intensities were assigned to this group at all. Gates the Intensity_ columns
+        /// in both the header and the row.
+        /// </summary>
+        /// <remarks>
+        /// Deliberately not the per-group SampleGroupResult.HasIntensityData. Whether a column EXISTS
+        /// has to be a property of the group as a whole, or a row cannot line up with a header written
+        /// from a different group -- and AllProteinGroups.tsv writes one header, from
+        /// proteinGroups.First(), then a row per group.
+        ///
+        /// It did line up until now, but by accident: QuantificationAnalysis assigns IntensitiesByFile
+        /// for every protein including unmeasured ones, so the zeros were present and HasIntensityData
+        /// was uniformly true. mzLib's QuantificationEngine omits zero-valued cells instead, leaving an
+        /// assigned-but-empty dictionary, and that is what the isobaric path produces.
+        ///
+        /// Both fields are required: a subset group built for a file none of the samples match gets an
+        /// empty sample list beside an empty dictionary, and gating on the dictionary alone would
+        /// advertise columns nothing can fill. Mirrors BioPolymerGroup.HasAssignedSampleIntensities from
+        /// the mzLib release after 1.0.587; this copy goes away when the pin moves and ProteinGroup
+        /// overrides the header rather than hiding it.
+        /// </remarks>
+        private bool HasAssignedSampleIntensities =>
+            SamplesForQuantification is { Count: > 0 } && IntensitiesBySample is not null;
+
         // Fails open: an unset path list or a group with no file info means we cannot tell, and dropping
         // columns on a guess is worse than keeping them.
         private bool WasSearched(SampleGroupResult group) =>
@@ -261,7 +285,7 @@ namespace EngineLayer
 
                     if (searched)
                         sb.Append($"SpectralCount_{group.Label}\t");
-                    if (group.HasIntensityData)
+                    if (HasAssignedSampleIntensities)
                         sb.Append($"Intensity_{group.Label}\t");
                     if (searched)
                         sb.Append($"CountOccupancy_{group.Label}\t");
@@ -393,9 +417,11 @@ namespace EngineLayer
                         sb.Append("\t");
                     }
 
-                    if (group.HasIntensityData)
+                    if (HasAssignedSampleIntensities)
                     {
-                        if (group.Intensity > 0)
+                        // Empty rather than 0 when this group has nothing: the cell must not
+                        // assert a measurement that was not made.
+                        if (group.HasIntensityData && group.Intensity > 0)
                             sb.Append(group.Intensity);
                         sb.Append("\t");
                     }
--- a/MetaMorpheus/Test/Multiplex_Labeling_TMT_iTRAQ.cs
+++ b/MetaMorpheus/Test/Multiplex_Labeling_TMT_iTRAQ.cs
@@ -986,6 +986,15 @@ namespace Test
         }
 
         [Test]
+        /// <remarks>
+        /// The absence asserted here describes master, where QuantificationAnalysis returns before
+        /// quantifying anything for a multiplex search, so no group is ever assigned intensities. It is
+        /// not a statement that TMT output should have no intensity columns -- when the isobaric path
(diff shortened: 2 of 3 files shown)


acesnik:
🤖 Same class of problem as mzLib #1240 and worth fixing in both places: a row that omits cells its own header declares puts every subsequent value under the wrong column name, and nothing complains. Deciding the schema once for the run rather than per group is the right shape, and I like that the new test asserts `pgB.GetTabSeparatedHeader() == pgA.GetTabSeparatedHeader()` — that's the invariant that actually matters.

The headline for me is that it's **inert for every path that exists on master today**, which I traced rather than assumed. Old and new gates agree everywhere:

| path | old (`group.HasIntensityData`) | new (`HasAssignedSampleIntensities`) |
|---|---|---|
| `QuantificationAnalysis`, `PostSearchAnalysisTask.cs:627` — full zero-filled dictionary for every group | true | true |
| `PopulateCountBasedOccupancy`, `:693` — `IntensitiesByFile = null`, non-empty file list | false | false |
| `ConstructSubsetProteinGroup` matched, `ProteinGroup.cs:693` | true | true |
| same, unmatched (`:697` nulls the dictionary) | false | false |

The unmatched-subset row is uniform across groups too, since the file list every group is matched against is the same object. The empirical half is the SILAC suite, which asserts exact header substrings (`SpectralCount_silac\tIntensity_silac\tCountOccupancy_silac\t…`) and is green.

I reproduced the 28-vs-26 arithmetic: 14 static columns before the dynamic block and 8 after = 22, plus 2 sample groups × (`SpectralCount_`, `Intensity_`, `CountOccupancy_`) = 28; pgB drops its two `Intensity_` cells → 26. Matches.

The case that *does* change is the isobaric one — `PopulateSampleGroupResults` gates per channel on `IntensitiesBySample.TryGetValue(sample, …)`, so an omitted channel drops its column mid-row and a single header can't describe both. That's the shape #2755 produces, so landing this first is the right order.

I also went looking for index-based readers, since inserting a column is the classic way to break one, and there are none: nothing in MetaMorpheus reads `AllProteinGroups.tsv`/`AllQuantifiedProteinGroups.tsv` back, the tests that inspect it resolve columns by `Array.IndexOf(header, …)`, and mzLib's readers cover psmtsv rather than protein groups. Writing an empty cell instead of `0` also matches what the file already did for a quantified-but-zero protein, so consumers get no new convention.

One residual worth naming: the gate is a property of one group, but the invariant it protects belongs to the collection — the header is `proteinGroups.First()`'s. Uniform on every path above, so this is a note rather than a request; just flagging that the guarantee lives in the caller, not in the gate.

Three things I'd like fixed before this merges - all documentation, none of it behaviour, but one of them means a comment silently doesn't work:

1. `Multiplex_Labeling_TMT_iTRAQ.cs` — the `<remarks>` block sits between `[Test]` and the signature (lines 988–989), so it's orphaned trivia rather than the method's doc comment. `Test.csproj` sets no `GenerateDocumentationFile` so nothing warns, but compiling the same shape with `-p:GenerateDocumentationFile=true` gives `warning CS1587: XML comment is not placed on a valid language element` and the text never reaches the doc file. It's also the only `///`-after-`[Test]` in that file.
2. `ProteinGroupWithNoSamplesButPsms_HasNoIntensityColumns` opens `<summary>` twice (diff lines 174–175).
3. That test's docstring attributes "an empty sample list beside an empty-but-non-null dictionary" to `ConstructSubsetProteinGroup`, but MetaMorpheus's override doesn't produce that shape — with no matching sample it leaves `FilesForQuantification` null and sets `IntensitiesByFile = null` (`ProteinGroup.cs:693-697`). That shape comes from mzLib's `BioPolymerGroup.ConstructSubsetBioPolymerGroup`, which assigns `SamplesForQuantification = matchingSamples` even when the filter is empty. The test still guards the `Count > 0` half of the gate, so only the comment is wrong.

Holding the second checkmark only because these need a push to fix and a comment on a merged PR tends to stay a comment. None of it is a disagreement with the change.

Last, gently: 22 of the 29 added lines in `ProteinGroup.cs` are comment, and the `<remarks>` on a one-expression private property runs longer than the PR section explaining the same thing.

Not verified: I didn't run any test (Windows-only project, no harness built), so the three new tests' results rest on your green CI — I verified their logic and their non-vacuity by reading, and the third test's `SampleGroupResults, Is.Not.Empty` pre-assertion does avoid the empty-collection trap. I also didn't run a real search, so "no produced file changes" is traced through the code plus the SILAC header assertions rather than measured on an output diff, and the mzLib-side behaviour was read from the source checkout rather than decompiled from the pinned 1.0.587 package.

RayMSMS (approved):

nbollis (approved):

zhuoxinshi (approved):

PR #2773, opened 2026-08-29
Build calibration test fixture paths with Path.Combine so they run off Windows

acesnik:
🤖 Test-only. No production code changes, and no behaviour changes on Windows.

`Test/CalibrationTests.cs` built its fixture paths as `Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\SmallCalibratible_Yeast.mzML")`. That literal is a path only on Windows; on macOS and Linux the backslash is an ordinary filename character, so the file resolves to a single entry named `TestData\SmallCalibratible_Yeast.mzML` that does not exist. This replaces all 27 such literals with separate `Path.Combine` segments.

On Windows the two forms produce the identical string, because `Path.Combine` inserts a `DirectorySeparatorChar` between non-rooted segments and does not rewrite separators inside a segment. CI runs the test project on `windows-latest` only, so this class of bug has never been visible there — the same reason #2721 existed on the production side.

Measured off Windows, in a throwaway net10.0 harness that compiles this file against `EngineLayer` and `TaskLayer`: **15 of the 29 cases failed before, 0 fail after.** Fourteen failed as `FileNotFoundException` naming the backslash path; the fifteenth, `CalibrationRunsInMostAbundantMode`, wraps its run in `Assert.DoesNotThrow` and so failed as `Expected: No Exception to be thrown`. Nothing is left failing for any other reason, and every fixture these tests want is already in the repo.

Only path strings were changed. The 14 regex literals (`@"\d+\.\d*"`) are untouched, as are the single-segment directory and filename literals, which were already correct as `Path.Combine` arguments. No assertion in this file compares a path, and nothing derived from these strings is written into an experimental-design file or a TOML, so there is no string-identity risk on the Windows side.

One deliberate exception: `Path.Combine(experimentalDesignPath, rawFile)` in `TestExperimentalDesignErrors` combines a file path with a filename, producing a nonsense path. That is what the test wants — it only asserts on the resulting error message, and it passes on both platforms. Left alone, and flagged here because a mechanical sweep would be tempted to change it.

This is a slice of a wider pattern rather than the whole of it: roughly 550 more backslash path literals remain across about 45 other files under `Test/`, with `TestOGlyco.cs` (87) and `MetaDraw/MetaDrawTest.cs` (51) the largest. A nastier variant also exists in the same family — hardcoded drive literals like `@"C:\work\out\good_name.mzML"` — which `Path.Combine` cannot rescue. Deliberately not touched here.

### DIFF
--- a/MetaMorpheus/Test/CalibrationTests.cs
+++ b/MetaMorpheus/Test/CalibrationTests.cs
@@ -40,10 +40,10 @@ namespace Test
 
             // set up original spectra file (input to calibration)
             string nonCalibratedFilePath = Path.Combine(unitTestFolder, nonCalibratedFile);
-            File.Copy(Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\SmallCalibratible_Yeast.mzML"), nonCalibratedFilePath, true);
+            File.Copy(Path.Combine(TestContext.CurrentContext.TestDirectory, "TestData", "SmallCalibratible_Yeast.mzML"), nonCalibratedFilePath, true);
 
             // protein db
-            string myDatabase = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\smalldb.fasta");
+            string myDatabase = Path.Combine(TestContext.CurrentContext.TestDirectory, "TestData", "smalldb.fasta");
 
             // set up original experimental design (input to calibration)
             SpectraFileInfo fileInfo = new(nonCalibratedFilePath, "condition", 0, 0, 0);
@@ -107,10 +107,10 @@ namespace Test
 
                 // set up original spectra file (input to calibration)
                 string nonCalibratedFilePath = Path.Combine(unitTestFolder, "filename1.mzML");
-                File.Copy(Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\SmallCalibratible_Yeast.mzML"), nonCalibratedFilePath, true);
+                File.Copy(Path.Combine(TestContext.CurrentContext.TestDirectory, "TestData", "SmallCalibratible_Yeast.mzML"), nonCalibratedFilePath, true);
 
                 // protein db
-                string myDatabase = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestData\smalldb.fasta");
+                string myDatabase = Path.Combine(TestContext.CurrentContext.TestDirectory, "TestData", "smalldb.fasta");
 
                 // run calibration
                 CalibrationTask calibrationTask = new();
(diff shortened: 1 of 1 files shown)


trishorts (approved):
Approving. Test-only, and the transformation is provably identity on Windows: `Path.Combine` inserts a `DirectorySeparatorChar` between non-rooted segments and does not rewrite separators inside a segment, so the two forms produce the same string there. Nothing in this file compares a path or feeds one into an experimental-design file or a toml, so there is no string-identity risk on the side CI actually exercises.

The `Path.Combine(experimentalDesignPath, rawFile)` exception in `TestExperimentalDesignErrors` is the right call, and I am glad it is called out. A nonsense path is the input that test wants, and a mechanical sweep is exactly what would have "fixed" it into passing for the wrong reason.

Being explicit about what is left is the most useful part of this PR: roughly 550 more backslash literals across about 45 files, and the `@"C:\work\out\..."` drive-literal variant that `Path.Combine` cannot rescue at all. Worth an issue so the remainder does not depend on someone rediscovering it, because this slice being green will otherwise read as the problem being solved.

zhuoxinshi (approved):

trishorts (approved):
Approving. Test-only, and I checked the sweep is complete rather than partial: at the head of this branch `CalibrationTests.cs` has **27** `Path.Combine(..., "TestData", "...")` call sites and **zero** remaining backslash path literals. The only backslashes left in the file are the `@"\d+\.\d*"` regexes, which are correctly untouched.

Two judgement calls I agree with:

- `Path.Combine(experimentalDesignPath, rawFile)` in `TestExperimentalDesignErrors` is left alone. It does produce a nonsense path, and that is exactly what the test asserts on, so a mechanical sweep "fixing" it would have broken the test. Good call flagging it in the description.
- The single-segment literals (`@"TestCalibrationLow"`, `@"TestCalibrationMostAbundant"`) are already correct as `Path.Combine` arguments and did not need touching.

On Windows the two forms produce identical strings, so this cannot regress the Windows run. The ~550 remaining literals across the other ~45 test files are worth a follow-up issue so the rest of the file set does not have to be rediscovered later.

trishorts (commented):
Read the diff directly (one file, 27 mechanical replacements). Nothing inline to attach a finding to, so this is the verification plus one correction to the body.

**The sweep is complete, and I checked that rather than assuming it.** In the head revision of `CalibrationTests.cs` the only remaining backslashes inside string literals are exactly the 14 `@"\d+\.\d*"` regex matches, at lines 71/72, 132/133, 266/267, 293/294, 506/507 and 565-571. No drive letters, and no paths built by concatenation or interpolation, so there is no `+ "\\"` variant hiding behind the grep for literals. `Path.Combine(experimentalDesignPath, rawFile)` at line 426 is intact, as you said it would be.

**The residual risk `Path.Combine` cannot fix is case, and it is clean here.** Separators are only half of what breaks off Windows; a fixture whose literal disagrees with the on-disk name in case resolves fine on Windows and fails on Linux, and no amount of `Path.Combine` helps. All seven fixtures this file names match `Test/TestData/` byte for byte including case -- `SmallCalibratible_Yeast.mzML`, `smalldb.fasta`, `gapdh.fa`, `mouseOne.mzML`, `mouseOne.xml`, `TaGe_SA_A549_3_snip.mzML`, `TaGe_SA_A549_3_snip.fasta` -- and `TestData/TestExperimentalDesign/WrongNumberOfCells` exists exactly as spelled in the two lines you left alone. Worth knowing for the wider sweep: the remaining ~45 files will need that check too, and it is not mechanical.

**One correction.** The body says "the same reason #2721 existed on the production side". #2721 is "Refuse to update a spectral library when none was given" and has nothing to do with path separators. The production-side counterpart is **#2747** -- "Three small cleanups: glyco output paths, ..." -- whose first commit fixes `Path.Combine(OutputFolder + "\\oglyco" + ".psmtsv")` and describes the identical mechanism. Worth fixing so the reference lands somewhere useful.

**On the remaining ~550.** I searched issues for this and found nothing tracking it. The scoping in your last paragraph -- the count, the two worst files, and the drive-literal variant that `Path.Combine` cannot rescue -- is already the content of a good issue, and without one it lives only in the body of a merged PR. Happy to open it if you would rather not.

Nothing blocking. All thirteen checks pass, including `ubuntu-latest` and `macos-latest`, though as you note those build rather than run this project, so they are not what validates the change.

nbollis (approved):

PR #2774, opened 2026-08-29
Add the TMT experimental design model and its parser

trishorts:
**First of three PRs splitting #2598.** That PR is 3,300 lines carrying two new WPF windows and has sat at 0/3. #2596 sat 77 days and #2602 sat 233, both smaller — so this split is about getting the work reviewed, not about disagreeing with any of it.

This one is **headless**: `TmtExperimentalDesign`, its parser and writer, and the tests. No CMD wiring, no GUI, nothing that needs a WPF reviewer.

`TmtDesign.txt` describes which data file and reporter channel each sample sits in, so a TMT search can attribute per-channel intensities to conditions. `GlobalVariables` gains the default filename alongside the existing `ExperimentalDesign.tsv`.

**58 tests**: the parser against well-formed and malformed files, sample-type round-tripping, and projection onto mzLib's `IExperimentalDesign` / `IsobaricQuantSampleInfo` — the contract `QuantificationEngine` consumes.

### Two things deliberately not here

- `ResolveExperimentalDesignTests` exercises `CMD.Program.ResolveExperimentalDesign`, so it travels with the CMD wiring. The compiler drew that boundary, not me — it was the one file that would not build without `Program`.
- A comment-only edit to `Multiplex_Labeling_TMT_iTRAQ.cs` that #2771 has since made redundant. It was churn by the time this was split; carrying it would have put a cosmetic diff in front of a reviewer for nothing.

### The other two

| | |
|---|---|
| **CMD wiring** | stacked on this branch |
| **GUI** | stacked on this branch, independently of CMD |

Both show this PR's commits until it merges, then rebase to their own files. **Merge this one first.**

### DIFF
--- a/MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
+++ b/MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
@@ -127,6 +127,15 @@ namespace EngineLayer
             // How many cells a row needs to carry every required column.
             int minimumCells = new[] { idxFile, idxPlex, idxSample, idxChannel, idxCond, idxBio, idxFrac, idxTech }.Max() + 1;
 
+            // Normalize the provided paths ONCE, the same way the "did not contain the file(s)" check
+            // below normalizes them. The row-level in-this-run test used to compare a resolved path
+            // against these strings raw, so a provided path carrying a '..' segment matched there but
+            // not here: the row was skipped and the file was then reported as missing from a design
+            // that names it. Both file checks now read the same set.
+            var providedFullPaths = new HashSet<string>(
+                fullFilePathsWithExtension.Select(NormalizePath),
+                StringComparer.OrdinalIgnoreCase);
+
             for (int i = 1; i < lines.Length; i++)
             {
                 var line = lines[i];
@@ -162,8 +171,7 @@ namespace EngineLayer
                 string full = ResolveAgainstDesignFile(file, designDirectory);
 
                 // Only consider lines for files actually in this run (mirrors ExperimentalDesign behavior)
-                bool inThisRun = !fullFilePathsWithExtension.Any() ||
-                    fullFilePathsWithExtension.Contains(full, StringComparer.OrdinalIgnoreCase);
+                bool inThisRun = providedFullPaths.Count == 0 || providedFullPaths.Contains(full);
 
                 dataRowsSeen++;
                 if (inThisRun)
@@ -312,15 +320,11 @@ namespace EngineLayer
             if (fullFilePathsWithExtension != null && fullFilePathsWithExtension.Count > 0)
             {
                 // normalize and compare case-insensitively using full paths
-                var provided = new HashSet<string>(
-                    fullFilePathsWithExtension.Select(p => { try { return Path.GetFullPath(p); } catch { return p; } }),
-                    StringComparer.OrdinalIgnoreCase);
-
                 var defined = new HashSet<string>(
-                    fileState.Keys.Select(p => { try { return Path.GetFullPath(p); } catch { return p; } }),
+                    fileState.Keys.Select(NormalizePath),
                     StringComparer.OrdinalIgnoreCase);
 
-                var notDefined = provided.Where(p => !defined.Contains(p)).ToList();
+                var notDefined = providedFullPaths.Where(p => !defined.Contains(p)).ToList();
                 if (notDefined.Any())
                 {
                     errors.Add("Error: The TMT design did not contain the file(s): " + string.Join(", ", notDefined));
@@ -336,6 +340,16 @@ namespace EngineLayer
         /// TmtDesign.txt written alongside its raw files expects to be read from, and only then
         /// against the process working directory.
         /// </summary>
+        /// <summary>
+        /// Full path when the runtime can produce one, the original string when it cannot. Used on both
+        /// sides of every file comparison so the two file checks can never disagree.
+        /// </summary>
+        private static string NormalizePath(string path)
+        {
+            try { return Path.GetFullPath(path); }
+            catch { return path; }
+        }
+
         private static string ResolveAgainstDesignFile(string file, string designDirectory)
         {
             try
--- a/MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
+++ b/MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
@@ -29,8 +29,20 @@ namespace EngineLayer
         /// <summary>The column name carrying <see cref="TmtSampleType"/>.</summary>
         public const string SampleTypeColumn = "Sample Type";
 
-        // New validation helper: every (SampleName, BioRep, Fraction, TechRep) must be unique
-        private static string ValidateUniqueSampleBioFracTech(IEnumerable<(string Sample, int Bio, int Fraction, int Tech)> tuples)
+        /// <summary>
+        /// Every (Sample, BioRep, Fraction, TechRep) must be unique WITHIN A PLEX.
+        /// </summary>
+        /// <remarks>
+        /// Per plex rather than per file, because a bridge channel is by definition the same material
+        /// carried in more than one plex under one name -- which is what <see cref="TmtSampleType.Bridge"/>
+        /// exists to describe. Collecting these across the whole file rejected exactly that design, and
+        /// the only way through was to rename the bridges apart, which throws away the fact that they
+        /// are the same material and so defeats the point of having them.
+        ///
+        /// Keyed per plex still catches the case worth catching: one plex describing the same sample
+        /// twice.
+        /// </remarks>
+        private static string ValidateUniqueSampleBioFracTech(IEnumerable<(string Plex, string Sample, int Bio, int Fraction, int Tech)> tuples)
         {
             var duplicates = tuples
                 .GroupBy(t => t)
@@ -42,14 +54,21 @@ namespace EngineLayer
                 return null;
 
             var msgs = duplicates.Select(d =>
-                $"Duplicate combination detected: Sample \"{d.Sample}\" Biorep {d.Bio} Fraction {d.Fraction} Techrep {d.Tech}");
+                $"Duplicate combination detected in plex \"{d.Plex}\": Sample \"{d.Sample}\" " +
+                $"Biorep {d.Bio} Fraction {d.Fraction} Techrep {d.Tech}");
             return string.Join(Environment.NewLine, msgs);
         }
 
         // RETURN: files where each file carries its plex annotations
+        /// <param name="fullFilePathsWithExtension">
+        /// The files this run actually searched, used to ignore design rows naming anything else.
+        /// Null or empty means "do not filter" -- the same meaning the guard in ToMzLibDesign gives it,
+        /// and the reason this no longer throws on null.
+        /// </param>
         public static List<TmtFileInfo> Read(string tmtDesignPath, List<string> fullFilePathsWithExtension, out List<string> errors)
         {
             errors = new List<string>();
+            fullFilePathsWithExtension ??= new List<string>();
             var files = new List<TmtFileInfo>();
 
             // How many data rows the file had, and how many named a file in this run. A design whose
@@ -64,7 +83,7 @@ namespace EngineLayer
             var fileStateConflicts = new Dictionary<string, HashSet<string>>(StringComparer.OrdinalIgnoreCase);
 
             // Collect tuples for uniqueness validation
-            var uniqueTuples = new List<(string Sample, int Bio, int Fraction, int Tech)>();
+            var uniqueTuples = new List<(string Plex, string Sample, int Bio, int Fraction, int Tech)>();
 
             if (!File.Exists(tmtDesignPath))
             {
@@ -127,6 +146,14 @@ namespace EngineLayer
                 var file = cols[idxFile].Trim();
                 if (string.IsNullOrEmpty(file)) continue;
 
+                // A row naming a file and no channel is the placeholder Write emits for a file with no
+                // annotations yet, so that a part-authored design does not lose the file. Read used to
+                // reject it -- the blank Biological Replicate failed int.TryParse, the row was skipped,
+                // and the file was then reported as missing from the design. The placeholder therefore
+                // lost exactly what it existed to preserve, and a part-authored design is the NORMAL
+                // state while someone is filling one in through the GUI.
+                bool channelless = string.IsNullOrWhiteSpace(cols[idxChannel]);
+
                 // Resolve a relative name against the design file's own directory, not the process
                 // working directory. A TmtDesign.txt written with bare file names sits beside the raw
                 // files it names, and Path.GetFullPath alone would only match when the process happened
@@ -142,6 +169,22 @@ namespace EngineLayer
                 if (inThisRun)
                 {
                     dataRowsMatched++;
+
+                    if (channelless)
+                    {
+                        // Register the file with no annotations and move on. Fraction and technical
+                        // replicate still parse when present, so a placeholder carrying them keeps them.
+                        int.TryParse(cols[idxFrac].Trim(), out int placeholderFrac);
+                        int.TryParse(cols[idxTech].Trim(), out int placeholderTech);
+                        if (!fileState.ContainsKey(full))
+                        {
+                            fileState[full] = (cols[idxPlex].Trim(),
+                                placeholderFrac < 1 ? 1 : placeholderFrac,
+                                placeholderTech < 1 ? 1 : placeholderTech);
+                        }
+                        continue;
+                    }
+
                     var plex    = cols[idxPlex].Trim();
                     var sample  = cols[idxSample].Trim();
                     var tag     = cols[idxChannel].Trim();
@@ -155,9 +198,13 @@ namespace EngineLayer
                         continue;
                     }
 
-                    if (!int.TryParse(cols[idxBio].Trim(), out var bio))
+                    // >= 1, matching Fraction and Technical Replicate. A bare TryParse accepted 0 and
+                    // -3 silently, and ToMzLibDesign passes the value straight through to
+                    // ISampleInfo.BiologicalReplicate. If these are ever meant to be 0-based, all three
+                    // columns have to change together rather than one drifting from the other two.
+                    if (!int.TryParse(cols[idxBio].Trim(), out var bio) || bio < 1)
                     {
-                        errors.Add($"Line {i + 1}: invalid Biological Replicate.");
+                        errors.Add($"Line {i + 1}: Biological Replicate must be >= 1.");
                         continue;
                     }
                     if (!int.TryParse(cols[idxFrac].Trim(), out var frac) || frac < 1)
@@ -173,7 +220,7 @@ namespace EngineLayer
 
                     // Record tuple for uniqueness validation (ignore completely blank sample name)
                     if (!string.IsNullOrWhiteSpace(sample))
-                        uniqueTuples.Add((sample, bio, frac, tech));
+                        uniqueTuples.Add((plex, sample, bio, frac, tech));
 
                     // Per-file consistency
                     if (fileState.TryGetValue(full, out var state))
@@ -201,7 +248,7 @@ namespace EngineLayer
                         byTag = new Dictionary<string, TmtPlexAnnotation>(StringComparer.OrdinalIgnoreCase);
                         plexToAnnotations[plex] = byTag;
                     }
-                    if (!byTag.ContainsKey(tag))
+                    if (!byTag.TryGetValue(tag, out var existing))
                     {
                         byTag[tag] = new TmtPlexAnnotation
                         {
@@ -211,6 +258,22 @@ namespace EngineLayer
                             BiologicalReplicate = bio,
                             SampleType = sampleType
                         };
+                    }
+                    // Fractions of one plex legitimately repeat the same channel-to-sample map, which is
+                    // why this is keyed per plex and why a repeat is collapsed rather than refused. But
+                    // collapsing rows that DISAGREE silently discards one of them: two rows naming
+                    // channel 126 as different samples kept the first and said nothing. The uniqueness
+                    // rule above cannot catch it either -- two different sample names are two different
+                    // tuples, so it only fires when the duplicates agree, which is the harmless case.
+                    else if (!string.Equals(existing.SampleName, sample, StringComparison.OrdinalIgnoreCase)
+                             || !string.Equals(existing.Condition, cond, StringComparison.OrdinalIgnoreCase)
+                             || existing.BiologicalReplicate != bio
+                             || existing.SampleType != sampleType)
+                    {
+                        errors.Add(
+                            $"Line {i + 1}: plex \"{plex}\" channel \"{tag}\" is described twice and the " +
+                            $"descriptions disagree — \"{existing.SampleName}\"/\"{existing.Condition}\" " +
+                            $"versus \"{sample}\"/\"{cond}\".");
                     }
                 }
             }
--- a/MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
+++ b/MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
@@ -1,0 +1,639 @@
+﻿using FlashLFQ;
+using MassSpectrometry;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace EngineLayer
+{
+    public static class TmtExperimentalDesign
+    {
+        /// <summary>
+        /// The header this class writes. "Sample Type" is the newest column and is OPTIONAL on read
+        /// (see <see cref="RequiredColumns"/>), so a design file written before it existed still loads.
+        /// </summary>
+        public const string Header = "File\tPlex\tSample Name\tTMT Channel\tCondition\tBiological Replicate\tFraction\tTechnical Replicate\tSample Type";
+
+        /// <summary>
+        /// The columns a design file must carry to be readable. "Sample Type" is deliberately absent:
+        /// requiring it would invalidate every TmtDesign.txt written before it was added, and a file
+        /// without it is unambiguous - every channel is a study sample.
+        /// </summary>
+        public static readonly IReadOnlyList<string> RequiredColumns = new[]
+        {
+            "File", "Plex", "Sample Name", "TMT Channel", "Condition",
+            "Biological Replicate", "Fraction", "Technical Replicate"
+        };
+
+        /// <summary>The column name carrying <see cref="TmtSampleType"/>.</summary>
+        public const string SampleTypeColumn = "Sample Type";
+
+        // New validation helper: every (SampleName, BioRep, Fraction, TechRep) must be unique
+        private static string ValidateUniqueSampleBioFracTech(IEnumerable<(string Sample, int Bio, int Fraction, int Tech)> tuples)
+        {
+            var duplicates = tuples
+                .GroupBy(t => t)
+                .Where(g => g.Count() > 1)
+                .Select(g => g.Key)
+                .ToList();
+
+            if (duplicates.Count == 0)
+                return null;
+
+            var msgs = duplicates.Select(d =>
+                $"Duplicate combination detected: Sample \"{d.Sample}\" Biorep {d.Bio} Fraction {d.Fraction} Techrep {d.Tech}");
+            return string.Join(Environment.NewLine, msgs);
+        }
+
+        // RETURN: files where each file carries its plex annotations
+        public static List<TmtFileInfo> Read(string tmtDesignPath, List<string> fullFilePathsWithExtension, out List<string> errors)
+        {
+            errors = new List<string>();
+            var files = new List<TmtFileInfo>();
+
+            // How many data rows the file had, and how many named a file in this run. A design whose
+            // rows all fail to resolve is not an empty design -- it is a design pointed at the wrong
+            // place -- and it used to return no errors at all.
+            int dataRowsSeen = 0;
+            int dataRowsMatched = 0;
+
+            // Collect per-plex annotations (unique by tag) and per-file (plex, frac, tech)
+            var plexToAnnotations = new Dictionary<string, Dictionary<string, TmtPlexAnnotation>>(StringComparer.OrdinalIgnoreCase);
+            var fileState = new Dictionary<string, (string Plex, int Fraction, int TechRep)>(StringComparer.OrdinalIgnoreCase);
+            var fileStateConflicts = new Dictionary<string, HashSet<string>>(StringComparer.OrdinalIgnoreCase);
+
+            // Collect tuples for uniqueness validation
+            var uniqueTuples = new List<(string Sample, int Bio, int Fraction, int Tech)>();
+
+            if (!File.Exists(tmtDesignPath))
+            {
+                errors.Add("TMT design file not found!");
+                return files;
+            }
+
+            string designDirectory;
+            try { designDirectory = Path.GetDirectoryName(Path.GetFullPath(tmtDesignPath)); }
+            catch { designDirectory = null; }
+
+            string[] lines;
+            try
+            {
+                lines = File.ReadAllLines(tmtDesignPath);
+            }
+            catch (Exception ex)
+            {
+                errors.Add("Could not read TMT design file: " + ex.Message);
+                return files;
+            }
+
+            if (lines.Length == 0 || !IsHeaderValid(lines[0]))
+            {
+                errors.Add("TMT design file header is invalid.");
+                return files;
+            }
+
+            var headers = lines[0].Split('\t');
+            int idxFile   = Array.FindIndex(headers, h => string.Equals(h.Trim(), "File", StringComparison.OrdinalIgnoreCase));
+            int idxPlex   = Array.FindIndex(headers, h => string.Equals(h.Trim(), "Plex", StringComparison.OrdinalIgnoreCase));
+            int idxSample = Array.FindIndex(headers, h => string.Equals(h.Trim(), "Sample Name", StringComparison.OrdinalIgnoreCase));
+            int idxChannel= Array.FindIndex(headers, h => string.Equals(h.Trim(), "TMT Channel", StringComparison.OrdinalIgnoreCase));
+            int idxCond   = Array.FindIndex(headers, h => string.Equals(h.Trim(), "Condition", StringComparison.OrdinalIgnoreCase));
+            int idxBio    = Array.FindIndex(headers, h => string.Equals(h.Trim(), "Biological Replicate", StringComparison.OrdinalIgnoreCase));
+            int idxFrac   = Array.FindIndex(headers, h => string.Equals(h.Trim(), "Fraction", StringComparison.OrdinalIgnoreCase));
+            int idxTech   = Array.FindIndex(headers, h => string.Equals(h.Trim(), "Technical Replicate", StringComparison.OrdinalIgnoreCase));
+            // -1 when the file predates the column; every channel then reads as a study sample.
+            int idxType   = Array.FindIndex(headers, h => string.Equals(h.Trim(), SampleTypeColumn, StringComparison.OrdinalIgnoreCase));
+
+            // How many cells a row needs to carry every required column.
+            int minimumCells = new[] { idxFile, idxPlex, idxSample, idxChannel, idxCond, idxBio, idxFrac, idxTech }.Max() + 1;
+
+            for (int i = 1; i < lines.Length; i++)
+            {
+                var line = lines[i];
+                if (string.IsNullOrWhiteSpace(line)) continue;
+
+                var cols = line.Split('\t');
+                // A row must reach every REQUIRED column, not every column in the header. Comparing
+                // against the full header width would reject a row whose only omission is the
+                // optional trailing Sample Type cell — which is exactly what a design file written
+                // before that column existed looks like.
+                if (cols.Length < minimumCells)
+                {
+                    errors.Add($"Line {i + 1} has fewer columns than expected.");
+                    continue;
+                }
+
+                var file = cols[idxFile].Trim();
+                if (string.IsNullOrEmpty(file)) continue;
+
+                // Resolve a relative name against the design file's own directory, not the process
+                // working directory. A TmtDesign.txt written with bare file names sits beside the raw
+                // files it names, and Path.GetFullPath alone would only match when the process happened
+                // to be running in that folder -- so the same design file worked or silently matched
+                // nothing depending on where MetaMorpheus was launched from.
+                string full = ResolveAgainstDesignFile(file, designDirectory);
+
+                // Only consider lines for files actually in this run (mirrors ExperimentalDesign behavior)
+                bool inThisRun = !fullFilePathsWithExtension.Any() ||
+                    fullFilePathsWithExtension.Contains(full, StringComparer.OrdinalIgnoreCase);
+
+                dataRowsSeen++;
+                if (inThisRun)
+                {
+                    dataRowsMatched++;
+                    var plex    = cols[idxPlex].Trim();
+                    var sample  = cols[idxSample].Trim();
+                    var tag     = cols[idxChannel].Trim();
+                    var cond    = cols[idxCond].Trim();
+
+                    var rawType = idxType >= 0 && idxType < cols.Length ? cols[idxType].Trim() : string.Empty;
+                    if (!TryParseSampleType(rawType, out var sampleType))
+                    {
+                        errors.Add($"Line {i + 1}: '{rawType}' is not a recognised Sample Type. " +
+                                   $"Use one of: {string.Join(", ", SampleTypeNames)}.");
+                        continue;
+                    }
+
+                    if (!int.TryParse(cols[idxBio].Trim(), out var bio))
+                    {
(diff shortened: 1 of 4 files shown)


MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
```diff
@@ -0,0 +1,639 @@
...
+                        fileState[full] = (plex, frac, tech);
+                    }
+
+                    // Plex annotations
+                    if (!plexToAnnotations.TryGetValue(plex, out var byTag))
+                    {
+                        byTag = new Dictionary<string, TmtPlexAnnotation>(StringComparer.OrdinalIgnoreCase);
+                        plexToAnnotations[plex] = byTag;
+                    }
+                    if (!byTag.ContainsKey(tag))
```
acesnik: The dedup itself is right — fractions of one plex legitimately repeat the same channel→sample map, which is exactly why it is keyed per plex rather than per file. What is missing is a check that the rows being collapsed agree. Measured against a two-row design:

```
sample.raw  Plex1  Tumour  126  Disease  1  1  1  study sample
sample.raw  Plex1  Normal  126  Control  2  1  1  study sample
```
```
ERRORS: 0
KEPT channel 126 -> sample 'Tumour' condition 'Disease' bio 1
```

`Normal`/`Control` is gone and nothing says so. The `(Sample, Bio, Fraction, Tech)` rule cannot catch it, because two *different* sample names are two different tuples — it only fires when the duplicate rows agree in those four cells, which is the harmless case.

`fileStateConflicts` a few lines up already does exactly the right thing for Plex/Fraction/TechRep: accumulate the disagreement and report it. Reporting when an existing `byTag[tag]` differs from the incoming row would reuse it. Worth noting SDRF's one hard row constraint is this same key — `SdrfValidator` refuses a duplicate `source name + assay name + comment[label]`.

MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
```diff
@@ -0,0 +1,639 @@
...
+        {
+            "File", "Plex", "Sample Name", "TMT Channel", "Condition",
+            "Biological Replicate", "Fraction", "Technical Replicate"
+        };
+
+        /// <summary>The column name carrying <see cref="TmtSampleType"/>.</summary>
+        public const string SampleTypeColumn = "Sample Type";
+
+        // New validation helper: every (SampleName, BioRep, Fraction, TechRep) must be unique
+        private static string ValidateUniqueSampleBioFracTech(IEnumerable<(string Sample, int Bio, int Fraction, int Tech)> tuples)
```
acesnik: This collects tuples across the **whole file**, which rejects a design the PR's own new column exists to describe. `TmtSampleType.Bridge` is documented as "a bridge channel, shared across plexes to make them comparable" — and a bridge shared across two unfractionated plexes under one name produces the same tuple twice. Measured, calling `Read` directly:

```
File   Plex   Sample   Channel  Cond     Bio Frac Tech  Type
p1.raw Plex1  Tumour   126      Disease  1   1    1     study sample
p1.raw Plex1  Bridge   127N     Bridge   1   1    1     bridge
p2.raw Plex2  Normal   126      Control  2   1    1     study sample
p2.raw Plex2  Bridge   127N     Bridge   1   1    1     bridge
```
```
files=2  errors=1
ERR: Duplicate combination detected: Sample "Bridge" Biorep 1 Fraction 1 Techrep 1
```

Renaming to `Bridge_P1`/`Bridge_P2` clears it — 0 errors — but that is the workaround that throws away the information, since nothing downstream can then tell the two channels are the same material, which is the whole point of a bridge.

Keying this per plex, the way `plexToAnnotations` already is, would still catch two rows describing the same sample twice within one plex — the case actually worth catching — and would pair naturally with the agreement check on the dedup above.

MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
```diff
@@ -0,0 +1,639 @@
...
+
+                    var rawType = idxType >= 0 && idxType < cols.Length ? cols[idxType].Trim() : string.Empty;
+                    if (!TryParseSampleType(rawType, out var sampleType))
+                    {
+                        errors.Add($"Line {i + 1}: '{rawType}' is not a recognised Sample Type. " +
+                                   $"Use one of: {string.Join(", ", SampleTypeNames)}.");
+                        continue;
+                    }
+
+                    if (!int.TryParse(cols[idxBio].Trim(), out var bio))
```
acesnik: `Fraction` and `Technical Replicate` both enforce `>= 1` on the next two branches, and there is a test named for it; `bio` is a bare `int.TryParse`. A design with `Bio` of `0` and `-3` reads back with 0 errors — measured. Given `ToMzLibDesign` passes bio straight through, and the 0-based/1-based question in my main comment, worth closing while the file is open.

MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
```diff
@@ -0,0 +1,639 @@
...
+
+                // Resolve a relative name against the design file's own directory, not the process
+                // working directory. A TmtDesign.txt written with bare file names sits beside the raw
+                // files it names, and Path.GetFullPath alone would only match when the process happened
+                // to be running in that folder -- so the same design file worked or silently matched
+                // nothing depending on where MetaMorpheus was launched from.
+                string full = ResolveAgainstDesignFile(file, designDirectory);
+
+                // Only consider lines for files actually in this run (mirrors ExperimentalDesign behavior)
+                bool inThisRun = !fullFilePathsWithExtension.Any() ||
```
acesnik: No null check here, while line 249 guards `fullFilePathsWithExtension != null` for the same parameter. `Read(path, null, out errors)` throws `ArgumentNullException: Value cannot be null. (Parameter 'source')` — measured. One of the two is wrong, and the guard downstream suggests null was meant to be legal.

MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
```diff
@@ -0,0 +1,639 @@
...
+                if (anns.Count > 0)
+                {
+                    foreach (var a in anns)
+                    {
+                        sw.WriteLine($"{file.FullFilePathWithExtension}\t{plex}\t{a.SampleName}\t{a.Tag}\t{a.Condition}\t{a.BiologicalReplicate}\t{file.Fraction}\t{file.TechnicalReplicate}\t{ToDesignFileValue(a.SampleType)}");
+                    }
+                }
+                else
+                {
+                    sw.WriteLine($"{file.FullFilePathWithExtension}\t{plex}\t\t\t\t\t{file.Fraction}\t{file.TechnicalReplicate}\t");
```
acesnik, line 404: This placeholder row does not survive a round trip. `WriteEmitsAPlaceholderRowForAFileWithNoAnnotations` asserts the exact text but never reads it back; reading it back:

```
FILES BACK: 0
ERR: Line 2: invalid Biological Replicate.
ERR: Error: The TMT design did not contain the file(s): /…/file.1.raw
```

The blank Biological Replicate cell fails `int.TryParse`, the row is skipped, and the file is then also reported as undefined. That test's own comment says the placeholder exists "so a design the user is part-way through authoring does not silently lose the file" — reading it back does lose it. Either `Read` should treat a row with no channel as a file-only row, or `Write` should not emit one. The choice matters for #2776, where a part-authored design is the normal state rather than an edge case.

Same family, lower priority: `Write` neither escapes nor rejects a tab inside `Plex`, `Sample Name` or `Condition`. A condition of `Dose<TAB>10uM` writes a ten-cell row, and `Read` then reports `Line 2: '1' is not a recognised Sample Type` — caught, but naming a column the user never touched.

MetaMorpheus/Test/TmtSampleTypeAndMzLibDesignTests.cs
```diff
@@ -0,0 +1,560 @@
...
+        /// positionally. A design file listing channels in any other order must not change it.
+        /// </summary>
+        [Test]
+        public static void ToMzLibDesign_OrdersChannelsByReporterMz_NotByDesignFileOrder()
+        {
+            var tag = Tmt10();
+            Assert.IsNotNull(tag, "TMT10 must resolve from the loaded modifications");
+
+            // Deliberately scrambled relative to the tag's own channel order.
+            var file = FileWith(@"C:\data\run1.raw", "PlexA", 1, 1,
```
acesnik: This literal is why six of the 58 tests only pass on Windows: `Path.GetFileName` keeps the whole string on Unix, so the dictionary key is never `run1.raw`. 52 pass off Windows, 6 fail — the five `ToMzLibDesign_*` tests and `TwoEntriesWithNoFileNameAreReportedAsMissingNamesNotACollision`.

The production code is platform-clean, so this is fixture paths only — and this same file already has the right idiom: `TwoEntriesSharingAFileNameAreRefusedWithoutLosingTheRest` and `EntryWithNoFileNameIsReported` cover the same branches with `Path.Combine` and `Path.DirectorySeparatorChar`, and both pass. Same direction as #2721 and #2773. CI is `windows-latest` only, so these cannot fail there.

MetaMorpheus/Test/TmtSampleTypeAndMzLibDesignTests.cs
```diff
@@ -0,0 +1,560 @@
...
+        /// The stronger half of the same invariant, and the one that actually protects the projection:
+        /// ToMzLibDesign pairs channelLabels[i] with ReporterIonMzs[i], so label i has to name the
+        /// channel whose m/z sits at i. A count cannot catch a mislabelled channel -- iTRAQ 8-plex
+        /// carried the name "120" for the 121 reagent for exactly that reason, extracting the right ion
+        /// under a name no kit has.
+        ///
+        /// Every label begins with its nominal mass, so the assertion is available cheaply.
+        /// </summary>
+        [Test]
+        public static void EveryTagsLabelsNameTheChannelAtTheirOwnIndex()
```
acesnik: #2771 already put this test at `Multiplex_Labeling_TMT_iTRAQ.cs:1049`, and its version is the stronger one — it also checks the N/C suffix ordering that a nominal-mass check cannot see. Dropping this and `EveryTagsLabelsMatchItsReporterIonCount` loses nothing. Your description is right that the `Multiplex_Labeling_TMT_iTRAQ.cs` edit became redundant; the test just came back in through this fixture.

acesnik (commented):
🤖 This lands the headless half of the TMT work: a design format saying which file, plex and reporter channel each sample sits in, a parser that resolves it against the files a run actually searched, a writer, the `Sample Type` column that finally lets a pooled reference or bridge channel be marked as one, and `ToMzLibDesign` to project it all onto the `IExperimentalDesign` that quantification consumes. Splitting it out of #2598 was the right call and I would say so even if nothing else came of it — 3,300 lines sitting at 0/3 for weeks is not a review problem, it is a reviewability problem, and the parser is genuinely readable on its own.

The shape is right, and the four findings from the #2598 review are addressed. `ResolveAgainstDesignFile` plus the `dataRowsSeen == 0` error close the hole where a design whose paths did not resolve read as success, which was the one I would have worried about most: silently annotating nothing is worse than failing. Resolving columns by name so their order is free, and making `Sample Type` optional so a design written before it existed still loads, are both the kind of decision that costs nothing now and saves a support thread later. And `ToMzLibDesign` documenting its order and length contracts *at the point where getting them wrong would be silent* is the detail I appreciated most — that projection is exactly where an off-by-one becomes a mislabelled quantification column rather than an exception.

I built this off Windows and ran both new fixtures in a throwaway net10.0 harness: 58 tests, matching your count. I have left seven notes on lines. The first two are the ones I would fix before #2775 and #2776 rebase onto a merged version of this; the rest are cheap. The first two are also the same rule seen from opposite sides — one accepts two different samples on one channel, the other rejects the same sample on two plexes — so a single uniqueness rule keyed per plex, with an agreement check, would close both.

## The question I would most like your answer to

`TmtDesign.txt` and `ExperimentalDesign.tsv` describe the same five concepts under different spellings — `File` against `FileName`, `Biological Replicate` against `Biorep` — and that divergence generates most of the work downstream. It is why #2775 needs a mutual-exclusivity rule and a second `GlobalVariables` filename, and it is why the two files disagree on a detail that will eventually surface as an output column that is off by one: `ExperimentalDesign` stores `biorep - 1, techrep - 1, fraction - 1` while `ToMzLibDesign` passes them through, so for the identical design row `1  1  1` the two report `bio=0 frac=0 tech=0` and `bio=1 frac=1 tech=1` through the same `ISampleInfo` interface. `ISampleInfo.Fraction` is documented as "0 if not applicable" and `Read` enforces `Fraction >= 1`, so an unfractionated TMT run says fraction 1 where an unfractionated LFQ run says 0. Whichever convention is right, it should be one convention.

Which raises whether a third design format is warranted at all. The alternatives are optional `Plex`/`Channel`/`Sample Type` columns on the existing file, or SDRF as the carrier with `ToMzLibDesign` as the only new code — and mzLib 1.0.587, the version already pinned here, ships `Readers.Sdrf` with a reader, writer and validator keyed on file + assay + channel, which is the same key the per-plex annotation dictionary needs. The `Sample Type` spellings came from SDRF already. I am not asking you to change direction in this PR; I am asking whether the answer is written down before two front ends encode `TmtDesign.txt` as a user-visible format, because that is the point after which it stops being a code change.

The naming carries the same shape of risk. Every type here is `Tmt*` — `TmtExperimentalDesign`, `TmtFileInfo`, `TmtPlexAnnotation`, `TmtSampleType`, and the column `TMT Channel` — while the model accepts every `IsobaricMassTagType` and mzLib spells them `Isobaric*`. That rename is cheap today and much less so once #2776's XAML binds to it. Relatedly, `ToMzLibDesign` takes `ChannelLabel` from `GetReporterIonLabels`, which is right for #2771 but disagrees with mzLib's `SpectrumMatchFromTsvHeader.TmtChannelNames` on five of eight tag types: TMT10/11/18 are covered, TMT6 is missing 127-130, iTRAQ4 all four, iTRAQ8 all eight, diLeu4 all four, diLeu12 all twelve. That list also still carries `131` where `GetReporterIonLabels` now spells `131N`. Pre-existing in mzLib and not yours to fix, but the model now depends on it, so `ToMzLibDesign` for iTRAQ or DiLeu produces labels nothing on the psmtsv read path will find.

Two notes on the stack rather than the code. **#2755 carries a copy of `TmtExperimentalDesign.cs` that is 64 lines behind this one** — no `ResolveAgainstDesignFile`, no no-match error, i.e. exactly the fix #2598 was asked for — so that PR is worth a comment lest the merge quietly take the older side. And on where a bad design surfaces: the model never throws `MetaMorpheusException`, which is the right call, but of the two consumers of its error list, #2775 uses it for a CMD exit code and discards the parsed design while #2776 calls `Read(designPath, filesInDir, out var _)` and discards the errors. So once all three merge, a malformed design is refused on the command line and ignored in the GUI. Reasonable as a staging order, but worth stating so nobody reads the merge of #2776 as "the GUI validates the design". On #2598 itself: #2775 and #2776 together reproduce its file list exactly, minus the edit #2771 made redundant, so I would close it once this merges rather than rebase it.

Checked and clean: no culture exposure (every parse is `int.TryParse` on integer fields, confirmed under `de-DE`), column order is free, a missing required column is refused at the header, channel case is canonicalised so `127n` lands on `127N`, and the order and length contracts in `ToMzLibDesign` do what the remark says. Not verified: nothing here reaches quantification, so there is no output to compare and the projection is only proven against hand-built inputs; `QuantificationEngine`'s consumption of it I read from the `IExperimentalDesign` doc comment rather than exercised; and the GUI half is untestable here by construction.

MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
trishorts: Fixed in `b1a269efc`. Silent data loss, and your analysis of why the uniqueness rule cannot cover it is the part that made it obvious — it only fires when the duplicates *agree*, which is precisely the harmless case.

Reporting now when an existing `byTag[tag]` disagrees with the incoming row, on sample name, condition, biological replicate or sample type. Took your suggestion of reusing the `fileStateConflicts` shape.

`ChannelDescribedTwiceWithDifferentSamplesIsReported` uses your exact two-row fixture. It fails against the previous parser.

Good catch on the SDRF parallel — `source name + assay name + comment[label]` being SDRF's one hard row constraint is a useful sanity check that the key is the right one.

MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
trishorts: Fixed in `b1a269efc` — keyed per plex, as you suggested.

This was the worst of the seven: the design being rejected is the one `TmtSampleType.Bridge` was added *in this PR* to describe. And the workaround makes it worse rather than better — renaming to `Bridge_P1`/`Bridge_P2` clears the error by destroying the only thing that made them bridges, namely that downstream can tell they are the same material.

Two tests, both from your fixture: `ABridgeSharedAcrossPlexesIsNotADuplicate` (0 errors, 2 files) and `TheSameSampleTwiceInOnePlexIsStillADuplicate`, so the rule still fires where it should. The first fails against the previous parser.

MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
trishorts: Fixed in `b1a269efc` — `>= 1`, matching Fraction and Technical Replicate.

I took consistency with the two sibling columns rather than trying to settle your 0-based/1-based question, and left a note in place saying that if these are ever meant to be 0-based then all three have to change together rather than one drifting from the other two. Happy to go the other way if you would rather settle it now.

`BiologicalReplicateBelowOneIsRejected` covers `0` and `-3`, your two measured values.

MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
trishorts: Fixed in `b1a269efc` — null now means "do not filter".

You are right that the guard downstream settles which of the two is wrong: `ToMzLibDesign` already treats null that way, so `Read` throwing on it made the two ends of the same parameter disagree. Documented on the parameter so the meaning is stated rather than inferred.

`ANullFileListMeansDoNotFilterRatherThanThrowing` pins it, and fails against the previous parser with the `ArgumentNullException` you measured.

MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
trishorts: Fixed in `b1a269efc`. I took your first option — `Read` treats a channel-less row as a file-only row.

The deciding argument is the one you make at the end: for #2776 a part-authored design is the normal state, not an edge case. Given that, `Write` emitting the placeholder is right and `Read` refusing it was the bug — the row lost exactly what its own comment says it exists to preserve.

`APlaceholderRowForAnUnannotatedFileSurvivesARoundTrip` reads back the row that test only ever asserted the text of, and checks both halves: the file comes back, and it carries no annotations. Fails against the previous parser with your two errors.

**The tab case I have not done**, and I would rather say so than let it look handled. It is real — `Write` neither escapes nor rejects a tab in `Plex`, `Sample Name` or `Condition`, and the resulting error names a column the user never touched. Left as a follow-up on the grounds you flagged it lower priority; rejecting on write is about five lines if you would rather it went in here.

MetaMorpheus/Test/TmtSampleTypeAndMzLibDesignTests.cs
trishorts: Fixed in `b1a269efc`. Six tests, exactly the ones you named.

Routed every literal through one `FixturePath` helper rather than sprinkling `Path.Combine` inline, so the intent — a rooted path ending in this file name — is stated once and the next fixture added here inherits it. The remark on the helper records that CI is `windows-latest` only, which is why nothing caught this.

Confirmed your read that the production code is platform-clean; this was fixture paths alone.

MetaMorpheus/Test/TmtSampleTypeAndMzLibDesignTests.cs
trishorts: Dropped both in `b1a269efc`.

You are right on both counts, and the second is the interesting one: the description claiming the `Multiplex_Labeling_TMT_iTRAQ.cs` edit was redundant was true, but the test had come back in through this fixture — so the redundancy I described removing, I had reintroduced one file over. Exactly the kind of thing a split is supposed to surface and I did not check for.

Keeping #2771's version, which also checks the N/C suffix ordering that a nominal-mass check cannot see.

trishorts:
@acesnik — all seven threads addressed and resolved in `b1a269efc`. Two were silent data loss and one rejected a design this PR’s own new column exists to describe.

| your finding | fix |
|---|---|
| a channel described twice with different samples was collapsed silently | reported, reusing the `fileStateConflicts` shape you pointed at |
| the uniqueness tuple was file-wide, rejecting a bridge shared across plexes | keyed per plex; still fires within one plex |
| the placeholder row did not survive a round trip | `Read` treats a channel-less row as a file-only row |
| `Biological Replicate` was a bare `TryParse` | `>= 1`, matching its two siblings |
| `Read(path, null, …)` threw while `ToMzLibDesign` guarded null | null means “do not filter”, documented on the parameter |
| six tests only passed on Windows | one `FixturePath` helper; production code was already clean |
| a test #2771 already owns, better | dropped both |

Six new tests, one per parser finding. **All six fail against the previous parser** — verified by reverting it wholesale rather than trusting the diff. 156 green across the design, sample-type and multiplex fixtures.

Three things worth flagging back:

**The bridge one was the worst, and the workaround is worse than the bug.** Renaming to `Bridge_P1`/`Bridge_P2` clears the error by destroying the only property that made them bridges — that downstream can tell they are the same material.

**Your last finding caught me reintroducing something I had described removing.** The PR description says the `Multiplex_Labeling_TMT_iTRAQ.cs` edit became redundant, which was true — and the test had come straight back in through this fixture. Exactly what a split is meant to surface, and I did not check for it.

**I have not done the tab case**, and would rather say so than let it look handled. Real, and about five lines to reject on write — say the word and it goes in here rather than a follow-up.

Ready for another look.

MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
```diff
@@ -0,0 +1,702 @@
...
+                // Resolve a relative name against the design file's own directory, not the process
+                // working directory. A TmtDesign.txt written with bare file names sits beside the raw
+                // files it names, and Path.GetFullPath alone would only match when the process happened
+                // to be running in that folder -- so the same design file worked or silently matched
+                // nothing depending on where MetaMorpheus was launched from.
+                string full = ResolveAgainstDesignFile(file, designDirectory);
+
+                // Only consider lines for files actually in this run (mirrors ExperimentalDesign behavior)
+                bool inThisRun = !fullFilePathsWithExtension.Any() ||
+                    fullFilePathsWithExtension.Contains(full, StringComparer.OrdinalIgnoreCase);
```
Alexander-Sol: Non-blocking: this `inThisRun` check compares the *resolved* path `full` (run through `Path.GetFullPath` in `ResolveAgainstDesignFile`) against the *raw* strings in `fullFilePathsWithExtension` via `Contains`. But the later "did not contain the file(s)" check (~L312) normalizes *both* sides with `Path.GetFullPath` before comparing. So a provided path that isn't already normalized (a `..` segment, or a differently-cased directory component) can fail this `Contains`, get the row skipped, and then be reported as missing from a design that actually names it. Suggest normalizing both sides the same way here as well so the two file checks can't disagree. Realistically GUI/CMD callers pass full paths, hence non-blocking.

MetaMorpheus/Test/TmtExperimentalDesignTests.cs
```diff
@@ -0,0 +1,777 @@
...
+            Assert.Throws<InvalidOperationException>(() => TmtExperimentalDesign.Write(null));
+            Assert.Throws<InvalidOperationException>(() => TmtExperimentalDesign.Write(new List<TmtFileInfo>()));
+        }
+
+        /// <summary>
+        /// A file with no channel annotations still gets a row, so a design the user is part-way
+        /// through authoring does not silently lose the file.
+        /// </summary>
+        [Test]
+        /// <summary>
```
Alexander-Sol: Nit: two `<summary>` blocks straddle the `[Test]` attribute here — one above it (L509-512) and one below (this one). The first is orphaned from the method for XML-doc purposes. Drop one so a single summary sits directly above the method.

Alexander-Sol (dismissed):
Approving. Self-contained, headless, and thoroughly tested — the load-bearing `ToMzLibDesign` contracts (channel order = ascending reporter m/z, one entry per tag channel) are documented and directly covered. Two non-blocking comments inline; both are minor hardening, neither needs to hold up the merge.

MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
trishorts: Fixed in 873407d1. The provided paths are now normalized once, before the row loop, into a single `providedFullPaths` set that both file checks read — so the row-level test and the "did not contain the file(s)" check can no longer disagree. Both sides go through a shared `NormalizePath` helper (`Path.GetFullPath`, original string on throw). Added `AnUnnormalizedProvidedPathStillMatchesItsDesignRow`, which routes the provided path through a `..` segment; it fails on the old comparison and passes on the new one.

MetaMorpheus/Test/TmtExperimentalDesignTests.cs
trishorts: Fixed in 873407d1 — the two blocks are merged into one summary directly above the method.

MetaMorpheus/Test/TmtSampleTypeAndMzLibDesignTests.cs
```diff
@@ -0,0 +1,521 @@
...
+        /// after the first report is what keeps them separate: without it the first is keyed under
+        /// the empty string, and the second is then reported as sharing the file name '' -- naming a
+        /// fault the design does not have, and sending the user looking for a duplicate.
+        /// </summary>
+        [Test]
+        public static void TwoEntriesWithNoFileNameAreReportedAsMissingNamesNotACollision()
+        {
+            var tag = Tmt10();
+            var first = FileWith(FixturePath("data") + Path.DirectorySeparatorChar, "PlexA", 1, 1, ("126", "A", 1, TmtSampleType.StudySample));
+            var second = FileWith(@"D:\other\", "PlexA", 1, 1, ("126", "B", 1, TmtSampleType.StudySample));
```
acesnik: This one did not get converted, and it is in the test most sensitive to it. `first` now goes through `FixturePath`, `second` is still `@"D:\other\"`, so the two entries are no longer the same kind of path.

Measured on macOS:

```
GetFileName(/data/     ) = ''           -> NO file name
GetFileName(D:\other\  ) = 'D:\other\'  -> HAS a file name
```

There is no separator in `D:\other\` off Windows, so `Path.GetFileName` keeps the whole string and that entry reports *no* missing-name error. One instead of two:

```
Failed TwoEntriesWithNoFileNameAreReportedAsMissingNamesNotACollision
  Assert.That(errors.Count(e => e.Contains("no file name")), Is.EqualTo(2))
  Expected: 2  But was: 1
```

On Windows both are rooted with no file name, so it passes — which means CI stays green and this is red only for anyone working off Windows. Same blind spot as the original finding, one test later.

`FixturePath("other") + Path.DirectorySeparatorChar`, matching the line above, fixes it. It is the last one: `grep -n '@"[A-Z]:\\'` over both test files returns nothing else.

acesnik (commented):
🤖 All seven verified fixed, and I checked them by re-running the probes that found them rather than reading the diff. The measured before/after, same fixtures as last time:

| finding | before | now |
|---|---|---|
| bridge shared across two plexes | 1 error, design refused | **0 errors, 2 files, all four annotations kept** |
| same sample twice in *one* plex | — | still errors, and now names the plex |
| channel 126 given two different samples | 0 errors, `Normal`/`Control` silently gone | **1 error naming line 3, both descriptions quoted** |
| `Biological Replicate` 0 and −3 | 0 errors | 2 errors |
| `Read(path, null, out errors)` | `ArgumentNullException` | 1 file, 0 errors |

The disagreement message is better than what I suggested — quoting both descriptions rather than just flagging the collision is what turns it from "something is wrong here" into something a user can act on.

I also checked one thing you did not claim, because it is the way this fix could have gone wrong. The per-plex dedup exists so that fractions of one plex may repeat the same channel→sample map, and a disagreement check is exactly the sort of thing that starts rejecting that. One plex across three fractions with identical maps: **0 errors**. Same design with fraction 3 disagreeing on channel 126: caught, naming line 4. So the new check discriminates rather than just being stricter.

One straggler from the cross-platform fix, on line 505 — details inline. It is one literal, and it is the last one: `grep -n '@"[A-Z]:\\'` over both test files returns nothing else.

**On the tab case — follow-up, don't hold this for it.** It produces an error rather than corrupting anything, and the error naming an untouched column is a confusing message, not a wrong result. Against two silent-data-loss fixes and a design this PR's own column exists to describe, it is not worth another round on the base of a three-PR stack.

**On 0-based versus 1-based, since you offered to settle it.** Taking consistency with the two sibling columns was the right call for this PR, and I would not settle it here. But it does want settling before #2775 and #2776 encode it: `ExperimentalDesign` stores `biorep - 1, techrep - 1, fraction - 1` while `ToMzLibDesign` passes them through, so the identical design row `1 1 1` reports `bio=0 frac=0 tech=0` from one and `bio=1 frac=1 tech=1` from the other through the same `ISampleInfo`. Whichever way it goes, one convention.

Commenting rather than approving only because line 505 needs a push. Everything else is verified; I will approve on the next one.

Scope of what I ran: 64 tests across `TmtExperimentalDesignTests` and `TmtSampleTypeAndMzLibDesignTests` in a throwaway net10.0 harness — 63 pass, the one below fails. I did not compile `Multiplex_Labeling_TMT_iTRAQ.cs`, so your 156 is a superset of my count, not a disagreement with it.

MetaMorpheus/Test/TmtSampleTypeAndMzLibDesignTests.cs
trishorts: Fixed in `7b1a5a1c` -- `second` now uses `FixturePath("other") + Path.DirectorySeparatorChar`, matching the line above.

One correction to the sweep, in your favour on the conclusion: `grep -nE '"[A-Z]:'` does find two more, `TmtExperimentalDesignTests.cs:601` and `:603` in `TmtFileInfoIdentityIsTheFilePath`. They are safe -- that test only compares the strings for case-insensitive equality and hashing, and never sends them through `Path.GetFileName` -- so 505 was indeed the last behaviourally platform-sensitive one. 68/68 green.

trishorts:

`7b1a5a1c` converts the last fixture path (`TmtSampleTypeAndMzLibDesignTests.cs:505`) to `FixturePath("other") + Path.DirectorySeparatorChar`. Also `873407d1` earlier, for Alexander-Sol's two: the provided paths are normalized once into a set both file checks read, with a regression test for the `..` case.

One correction to your sweep, in your favour on the conclusion: `grep -nE '"[A-Z]:'` does find two more at `TmtExperimentalDesignTests.cs:601` and `:603`, but that test only compares the strings for equality and hashing and never calls `Path.GetFileName`, so 505 really was the last platform-sensitive one. 68/68 green.

zhuoxinshi (approved):

acesnik (approved):
🤖 Approving, as promised last round. Line 505 now reads `FixturePath("other") + Path.DirectorySeparatorChar`, matching 504 exactly, so the two entries are the same kind of path and both report a missing file name off Windows as well as on it.

**Your correction to my sweep was right, and I was wrong about more than you said.** You framed it as your broader pattern catching two that mine missed — in fact `@"[A-Z]:\`, the pattern I quoted, matches `TmtExperimentalDesignTests.cs:601` and `:603` too. So "returns nothing else" was simply wrong rather than narrow, and you were right not to take it on trust. Your conclusion holds: `TmtFileInfoIdentityIsTheFilePath` only exercises `Equals`, `GetHashCode` and `HashSet` membership, never `Path.GetFileName`, so a Windows-shaped literal there is testing string identity rather than filesystem behaviour and is safe. I confirmed there is no remaining `Path.GetFileName` reached from a literal drive path in either fixture.

One thing worth acting on: `7b1a5a1c` dismissed @Alexander-Sol's review, and that review was an **approval** with two notes he called non-blocking. So the push that fixed my finding cost you an approval you already had. That puts this at 2 of 3 with @zhuoxinshi and me, and re-requesting him is the cheapest way to the third.

The larger questions from my first read are unchanged and I am not reopening them here — they were never this PR's to answer. Restating only so they do not get lost between merges: the `Tmt*` naming against a model that accepts every `IsobaricMassTagType`, 0-based versus 1-based replicate indices where `ExperimentalDesign` subtracts one and `ToMzLibDesign` does not, whether a third design format is warranted at all when the pinned mzLib already ships `Readers.Sdrf`, and mzLib's `TmtChannelNames` disagreeing with `GetReporterIonLabels` on five of eight tag types. All four get more expensive the moment #2775 and #2776 bind to them, which is the argument for settling them between this merging and those rebasing rather than after.

And since this is the one that lands first: `TmtExperimentalDesign.cs` is 715 lines here against 638 on #2755, #2775, #2776 and #2598. This is the only branch carrying the two silent-data-loss fixes, so every one of those wants rebasing onto a merged version of this rather than merging its own copy — resolved the wrong way, any of them reverts the fix that #2598's review asked for.

Verification unchanged in kind: 68 tests in a throwaway net10.0 harness off Windows, all green at this head. `Test/Test.csproj` is `net10.0-windows`, so your 156 remains a superset of what I can run rather than a figure I have confirmed.

MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
```diff
@@ -0,0 +1,716 @@
...
+        /// carried in more than one plex under one name -- which is what <see cref="TmtSampleType.Bridge"/>
+        /// exists to describe. Collecting these across the whole file rejected exactly that design, and
+        /// the only way through was to rename the bridges apart, which throws away the fact that they
+        /// are the same material and so defeats the point of having them.
+        ///
+        /// Keyed per plex still catches the case worth catching: one plex describing the same sample
+        /// twice.
+        /// </remarks>
+        private static string ValidateUniqueSampleBioFracTech(IEnumerable<(string Plex, string Sample, int Bio, int Fraction, int Tech)> tuples)
+        {
```
Alexander-Sol, line 46: Having this return bool with a string out would match more closely with patterns seen elsewhere in the codebase. I won't hold up the review on it though

MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
```diff
@@ -0,0 +1,716 @@
...
+
+        public override string ToString()
+        {
+            return Path.GetFileName(FullFilePathWithExtension);
+        }
+    }
+
+    public sealed class TmtPlexAnnotation
+    {
+        public string Tag { get; set; } = "";
```
Alexander-Sol, line 677: Should this implement ISampleInfo?

Alexander-Sol (approved):

MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
trishorts: Looked into this properly rather than picking on preference, and I'd like to leave it as-is — reasoning below, and happy to be overruled.

The closest sibling is `ExperimentalDesign.cs`, which this file is deliberately modelled on (`TmtExperimentalDesign.cs:319` — "mirrors ExperimentalDesign behavior"). It uses this exact pair:

- `ReadExperimentalDesign(..., out List<string> errors)` — `ExperimentalDesign.cs:12`
- `GetErrorsInExperimentalDesign(...)` returning `string`, doc-commented "Will return null if there are no errors" — `ExperimentalDesign.cs:116`
- caller does the same `!= null` → `errors.Add` — `ExperimentalDesign.cs:78-85`

`ValidateUniqueSampleBioFracTech` is `GetErrorsInExperimentalDesign` with a different predicate.

Searching upstream master for the shape you're describing, I find one `bool` + `out string` in the repo: `BrukerDataDirectory.TryGetParentDotDFolder` (`EngineLayer/Util/BrukerDataDirectory.cs:100`). Its `out string` carries a *result value* the caller consumes (`Spectra[i] = dotDFolder;`), not an error message — the `TryGet` value-retrieval idiom, and it's public and multi-caller, which is the scope the Framework Design Guidelines put `Try*` at. `GUI/Util/TaskValidator.cs` has ~20 `bool Check*` validators but none carry an `out string`; they raise a `MessageBox` themselves. The error-reporting convention in EngineLayer is the accumulator: `ReadExperimentalDesign:12`, `DatabaseLoadingEngine.LoadBioPolymers:83`, `SanitizeBioPolymerDatabase:190`.

The reason the distinction bites here is that my caller doesn't branch on success — it appends to a list (`:304-306`), alongside four other `errors.Add` sites. `bool` + `out` would hand it a return value to discard and an out-parameter to say what the one string already says.

Where I think you're onto something: the helper is the only validation in `Read` that doesn't just call `errors.Add` itself. Passing it the list and dropping the return entirely would match the four sites around it and delete the null check. That's a genuine simplification — but it's a code change, and with `dismiss_stale_reviews` on and the bar at 3, pushing it here costs all three approvals on the PR that #2775, #2776 and #2755 are all waiting on. I'll take it into the follow-up that already owes the tab-escaping fix.

MetaMorpheus/EngineLayer/TmtExperimentalDesign.cs
trishorts: Fair question, and it was on my own open list rather than something I'd settled — I had it written down as "the design types don't implement `ISampleInfo`/`IExperimentalDesign` and carry no `ReporterIonMz` or `IsReferenceChannel`", from back when this was still #2598. So it deserves a real answer rather than a deflection.

Half of it is closed by this PR, just not the way that note assumed: `ToMzLibDesign` produces the `IExperimentalDesign`, and the `IsobaricQuantSampleInfo`s it emits carry both `ReporterIonMz` and `IsReferenceChannel`. The remaining half — `TmtPlexAnnotation : ISampleInfo` — I now think should stay closed deliberately, and the reason is cardinality rather than taste.

`ISampleInfo` (`MassSpectrometry/ExperimentalDesign/ISampleInfo.cs:9`) requires `FullFilePathWithExtension`, `Condition`, `BiologicalReplicate`, `TechnicalReplicate` and `Fraction`. `TmtPlexAnnotation` has two of those. The other three are file-level and live on the parent `TmtFileInfo`, so implementing the interface means either a back-pointer from every annotation to its file, or duplicating file state onto every row.

That's a symptom of the real difference: **`TmtPlexAnnotation` is per (plex, channel); `ISampleInfo` is per (file, channel).** A plex spans its fractions, and `Read` deliberately shares one annotation instance across every file of that plex (`:312-314`); `ToMzLibDesign` then fans it out to one `IsobaricQuantSampleInfo` per file (`:527`, taking `file.FullFilePathWithExtension`). It's 1:N, so a type whose identity is coarser than the interface's can't implement it honestly. Public SDRF agrees on that grain, for what it's worth — `PXD004683` is a 2-plex experiment with 48 assay names.

Two things the projection does that implementing the interface couldn't:

**Ordering.** mzLib consumes the array positionally against `ISpectralMatch.Intensities` — `QuantificationEngine.cs:495` copies by index and nothing ever matches on channel label. `ToMzLibDesign` emits one entry per *tag* channel in ascending reporter-m/z order and materializes unannotated channels as `Empty`. A collection of annotations-as-`ISampleInfo` would be in design-file row order and would carry only the annotated channels — the exact two shifts the ordering contract exists to prevent. The annotation also has no source for `ReporterIonMz` or `PlexId`.

**`ISampleInfo` isn't the extension seam — `IExperimentalDesign` is.** mzLib ships two `ISampleInfo` implementations and type-tests them downstream: `CollapseStrategyBase.cs:116`, `ReferenceChannelNormalization.cs:38`, `QuantificationWriter.cs:167`, `BioPolymerGroup.cs:571`. A third implementation from EngineLayer compiles and flows through `PivotByFile`, then loses its channel identity at collapse and is skipped by reference normalization — wrong numbers, no exception. `CollapseStrategyBase.cs:102-107` says exactly this about returning the wrong concrete type.

On SDRF, which is where it matters most: keeping the carrier model separate from the quant model is what makes an SDRF reader a second `ToMzLibDesign()` rather than a rewrite. And `ISampleInfo` is going to move — it carries no `source name` and no plex *name* (`PlexId` is an int ordinal, excluded from equality), which are the two fields an SDRF round trip needs, and adding a sample name is also what settles whether it participates in `Equals`. Binding an EngineLayer DTO to the interface now means inheriting each of those across a release boundary. It's also why `TmtSampleType` is a five-value enum rather than mzLib's `bool IsReferenceChannel` — the design file is worth keeping more faithful than the quant object, and the projection is where the two are reconciled.

What's fair in the comment is that none of this is written on the type. A remark saying `TmtPlexAnnotation` is a design-file row projected to `IsobaricQuantSampleInfo` by `ToMzLibDesign`, and why it isn't an `ISampleInfo`, belongs there — I'll add it in the follow-up rather than push here, since with `dismiss_stale_reviews` on and the bar at 3, a doc commit costs all three approvals on the PR the rest of the stack is waiting on.

PR #2775, opened 2026-08-29
Wire the TMT experimental design into the CMD front end

trishorts:
**Second of three PRs splitting #2598.** #2774 (the model and its parser) **merged 2026-09-02** as `d233c17b`, so this branch has been updated from master and now shows **only its own layer**: `CMD/Program.cs` and `ResolveExperimentalDesignTests.cs`.

`Program.ResolveExperimentalDesign` locates the design file for a run, parses it, and **gates the run on the result** — one design file type or the other but not both, a design required once any search task asks for normalization, and a clean parse required before normalizing against it. A headless TMT run now fails fast on a design it could not read, instead of normalizing against one and silently producing different numbers.

### What this PR does *not* do: consume the design

`TmtExperimentalDesign.Read` at `Program.cs:373` — and `ExperimentalDesign.ReadExperimentalDesign` beside it — is called for its `out errors`. **The returned design is dropped**, and `ResolveExperimentalDesign` hands back only an exit code (`0` continue, `5` stop). So the CLI validates a design it does not yet consume.

That is deliberate staging rather than an oversight: **#2755** is the PR that carries the parsed design into the search. Saying it here because the title reads as though the design reaches the run, and it does not — this is the resolution-and-gating layer only.

### The design file is no longer deleted

The path this replaces deleted the user's design file when it failed to parse — silently at `-v none`, and otherwise on a `y/n` prompt whose question ("Delete and continue empty?") was about continuing while the answer was taken as consent to delete. Porting it forward would also have widened it to a second file type, `TmtDesign.txt`.

`File.Delete` is gone from this path entirely. At `-v none` an unreadable design is skipped and the file left alone; the prompt now reads **"Continue without an experimental design? y/n"** and continues only on `y`/`yes`, leaving the file alone either way. That is issue **#2256**, and the wording matches **#2780** — which removes the equivalent delete on the GUI side — word for word, so the CLI and the GUI cannot drift on whether a malformed design is the user's file or ours to discard.

The three tests that pinned the old behaviour are inverted rather than deleted: `AgreeingContinuesWithoutTheDesignAndKeepsTheFile`, `DecliningStopsTheRunAndKeepsTheFile` (covering `"n"`, `""` and `null`, so a closed stdin declines rather than falls through), and `AtVerbosityNoneAnUnreadableDesignIsSkippedNotDeleted`, each asserting `File.Exists(designPath)` with the reason on the assertion.

### Why the tests live here

`ResolveExperimentalDesignTests` travels with this PR rather than with the model because it exercises `CMD.Program` directly — that is where the compiler put the boundary when the split was built. Pulling the decision out of `Run` into one method with `Action<string> write` and `Func<string> readLine` injected is what makes it testable at all; previously this path could only be exercised by running the CLI and typing at it.

### On the merge from master

This branch had carried its own copy of `TmtExperimentalDesign.cs` since the split, at 639 lines against the 715 that merged. That copy and the two test fixtures came back as add/add conflicts and were resolved **wholesale to master's side**, which is the only correct resolution — every line unique to this branch was code #2774 replaced, including the two silent-data-loss fixes its review had asked for (the per-plex uniqueness key, which had been rejecting a bridge channel shared across plexes, and the report when one channel is described twice with different samples). All four shared files are byte-identical to `upstream/master` after the merge; verified rather than assumed.

The mzLib pin comes with the merge: **1.0.587 → 1.0.588**, untouched by either side.

**191 tests green** across the TMT design, sample-type, multiplex and experimental-design fixtures at `d7d1694f3`; solution builds clean.

### DIFF
--- a/MetaMorpheus/CMD/Program.cs
+++ b/MetaMorpheus/CMD/Program.cs
@@ -260,75 +260,21 @@ namespace MetaMorpheusCommandLine
             List<string> startingRawFilenameList = settings.Spectra.Select(b => Path.GetFullPath(b)).ToList();
             List<DbForTask> startingXmlDbFilenameList = settings.Databases.Select(b => new DbForTask(Path.GetFullPath(b), IsContaminant(b))).ToList();
 
-            // check that experimental design is defined if normalization is enabled
+            // check that an experimental design is defined if normalization is enabled
             var searchTasks = taskList
                 .Where(p => p.Item2.TaskType == MyTask.Search)
                 .Select(p => (SearchTask)p.Item2);
 
-            string pathToExperDesign = Directory.GetParent(startingRawFilenameList.First()).FullName;
-            pathToExperDesign = Path.Combine(pathToExperDesign, GlobalVariables.ExperimentalDesignFileName);
+            int designExitCode = ResolveExperimentalDesign(
+                Directory.GetParent(startingRawFilenameList.First()).FullName,
+                startingRawFilenameList,
+                searchTasks.Any(p => p.SearchParameters.Normalize),
+                settings.Verbosity == CommandLineSettings.VerbosityType.minimal
+                    || settings.Verbosity == CommandLineSettings.VerbosityType.normal);
 
-            if (!File.Exists(pathToExperDesign))
+            if (designExitCode != 0)
             {
-                if (searchTasks.Any(p => p.SearchParameters.Normalize))
-                {
-                    if (settings.Verbosity == CommandLineSettings.VerbosityType.minimal || settings.Verbosity == CommandLineSettings.VerbosityType.normal)
-                    {
-                        Console.WriteLine("Experimental design file was missing! This must be defined to do normalization. Download a template from https://github.com/smith-chem-wisc/MetaMorpheus/wiki/Experimental-Design");
-                    }
-                    return 5;
-                }
-            }
-            else
-            {
-                ExperimentalDesign.ReadExperimentalDesign(pathToExperDesign, startingRawFilenameList, out var errors);
-
-                if (errors.Any())
-                {
-                    if (searchTasks.Any(p => p.SearchParameters.Normalize))
-                    {
-                        if (settings.Verbosity == CommandLineSettings.VerbosityType.minimal || settings.Verbosity == CommandLineSettings.VerbosityType.normal)
-                        {
-                            foreach (var error in errors)
-                            {
-                                Console.WriteLine(error);
-                            }
-                        }
-                        return 5;
-                    }
-                    else
-                    {
-                        if (settings.Verbosity == CommandLineSettings.VerbosityType.minimal || settings.Verbosity == CommandLineSettings.VerbosityType.normal)
-                        {
-                            Console.WriteLine("An experimental design file was found, but an error " +
-                            "occurred reading it. Do you wish to continue with an empty experimental design? (This will delete your experimental design file) y/n" +
-                            "\nThe error was: " + errors.First());
-
-                            var result = Console.ReadLine();
-
-                            if (result.ToLowerInvariant() == "y" || result.ToLowerInvariant() == "yes")
-                            {
-                                File.Delete(pathToExperDesign);
-                            }
-                            else
-                            {
-                                return 5;
-                            }
-                        }
-                        else
-                        {
-                            // just continue on if verbosity is on "none"
-                            File.Delete(pathToExperDesign);
-                        }
-                    }
-                }
-                else
-                {
-                    if (settings.Verbosity == CommandLineSettings.VerbosityType.minimal || settings.Verbosity == CommandLineSettings.VerbosityType.normal)
-                    {
-                        Console.WriteLine("Read ExperimentalDesign.tsv successfully");
-                    }
-                }
+                return designExitCode;
             }
 
             EverythingRunnerEngine a = new EverythingRunnerEngine(taskList, startingRawFilenameList, startingXmlDbFilenameList, settings.OutputFolder);
@@ -354,6 +300,124 @@ namespace MetaMorpheusCommandLine
             }
 
             return errorCode;
+        }
+
+        /// <summary>
+        /// Works out which experimental design this run will use, reports whatever is wrong with it,
+        /// and returns the process exit code: 0 to carry on, 5 to stop.
+        ///
+        /// Static, and parameterised over its console I/O, so the decision can be tested without
+        /// driving a whole command line run. These branches decide whether a search starts at all,
+        /// and before this they were reachable only by launching the CLI against real spectra.
+        /// </summary>
+        /// <param name="designDirectory">The folder beside the spectra files, where a design lives.</param>
+        /// <param name="startingRawFilenameList">Full paths of the spectra files this run will search.</param>
+        /// <param name="normalizationRequested">True when any search task asks for normalization, which is what makes a design mandatory rather than optional.</param>
+        /// <param name="reportToConsole">True at minimal or normal verbosity. At "none" there is nobody to ask, so a recoverable problem is recovered from silently.</param>
+        /// <param name="write">Console writer; defaults to <see cref="Console.WriteLine(string)"/>.</param>
+        /// <param name="readLine">Console reader; defaults to <see cref="Console.ReadLine"/>.</param>
+        public static int ResolveExperimentalDesign(
+            string designDirectory,
+            List<string> startingRawFilenameList,
+            bool normalizationRequested,
+            bool reportToConsole,
+            Action<string> write = null,
+            Func<string> readLine = null)
+        {
+            write ??= Console.WriteLine;
+            readLine ??= Console.ReadLine;
+
+            string pathToExperDesign = Path.Combine(designDirectory, GlobalVariables.ExperimentalDesignFileName);
+            string pathToTmtDesign = Path.Combine(designDirectory, GlobalVariables.TmtExperimentalDesignFileName);
+
+            bool hasClassicDesign = File.Exists(pathToExperDesign);
+            bool hasTmtDesign = File.Exists(pathToTmtDesign);
+
+            // Only one may be present. Two files carrying replicate structure can drift apart, and
+            // nothing here could say which of them the user meant.
+            if (hasClassicDesign && hasTmtDesign)
+            {
+                if (reportToConsole)
+                {
+                    write("Both ExperimentalDesign and TmtDesign.txt were found. Only one design file type may be present.");
+                }
+
+                return 5;
+            }
+
+            if (!hasClassicDesign && !hasTmtDesign)
+            {
+                // A design is optional until something needs it. Normalization is that something.
+                if (normalizationRequested)
+                {
+                    if (reportToConsole)
+                    {
+                        write("No experimental design file present. Normalization requires a design (ExperimentalDesign.tsv or TmtDesign.txt).");
+                    }
+
+                    return 5;
+                }
+
+                return 0;
+            }
+
+            string designPath = hasClassicDesign ? pathToExperDesign : pathToTmtDesign;
+            List<string> errors;
+
+            if (hasClassicDesign)
+            {
+                ExperimentalDesign.ReadExperimentalDesign(designPath, startingRawFilenameList, out errors);
+            }
+            else
+            {
+                TmtExperimentalDesign.Read(designPath, startingRawFilenameList, out errors);
+            }
+
+            if (!errors.Any())
+            {
+                if (reportToConsole)
+                {
+                    write("Read " + Path.GetFileName(designPath) + " successfully");
+                }
+
+                return 0;
+            }
+
+            // Normalizing against a design that could not be read would silently produce different
+            // numbers, so this is the one case that cannot be recovered from.
+            if (normalizationRequested)
+            {
+                if (reportToConsole)
+                {
+                    foreach (string error in errors)
+                    {
+                        write(error);
+                    }
+                }
+
+                return 5;
+            }
+
+            // Otherwise the run can continue without a design -- but only by discarding the one that
+            // is there, which is the user's call whenever there is a user to ask.
+            if (!reportToConsole)
+            {
+                File.Delete(designPath);
+                return 0;
+            }
+
+            write((hasClassicDesign ? "An experimental design file" : "A TMT design file")
+                + " was found, but an error occurred reading it. Delete and continue empty? y/n");
+            write("First error: " + errors.First());
+
+            string answer = readLine();
+            if (answer?.ToLowerInvariant() == "y" || answer?.ToLowerInvariant() == "yes")
+            {
+                File.Delete(designPath);
+                return 0;
+            }
+
+            return 5;
         }
 
         private static void WriteMultiLineIndented(string toWrite)
--- a/MetaMorpheus/CMD/Program.cs
+++ b/MetaMorpheus/CMD/Program.cs
@@ -398,22 +398,24 @@ namespace MetaMorpheusCommandLine
                 return 5;
             }
 
-            // Otherwise the run can continue without a design -- but only by discarding the one that
-            // is there, which is the user's call whenever there is a user to ask.
+            // Otherwise the run continues without reading the design, leaving the file alone. It used
+            // to be deleted -- silently at verbosity none, and on a y/n whose question was about
+            // continuing rather than about deletion. Deleting an input the user cannot be asked about
+            // is a lot to infer from a verbosity flag, and #2256 is a user who lost their design that
+            // way. The GUI side stops deleting in #2780; the CLI and the GUI have to agree about
+            // whether a design that fails to parse is the user's file or ours to discard.
             if (!reportToConsole)
             {
-                File.Delete(designPath);
                 return 0;
             }
 
             write((hasClassicDesign ? "An experimental design file" : "A TMT design file")
-                + " was found, but an error occurred reading it. Delete and continue empty? y/n");
+                + " was found, but an error occurred reading it. Continue without an experimental design? y/n");
             write("First error: " + errors.First());
 
             string answer = readLine();
             if (answer?.ToLowerInvariant() == "y" || answer?.ToLowerInvariant() == "yes")
             {
-                File.Delete(designPath);
                 return 0;
             }
 
--- a/MetaMorpheus/Test/ResolveExperimentalDesignTests.cs
+++ b/MetaMorpheus/Test/ResolveExperimentalDesignTests.cs
@@ -187,14 +187,16 @@ namespace Test
         }
 
         /// <summary>
-        /// Not normalizing, so an unreadable design is recoverable -- by discarding it. That is
-        /// destructive, so it only happens on an explicit yes.
+        /// Not normalizing, so an unreadable design is recoverable -- by not reading it. Agreeing
+        /// continues the run and LEAVES THE FILE ALONE. The prompt only ever asked about continuing,
+        /// so deleting was an answer to a question nobody was asked; see #2256, and #2780 for the
+        /// same change on the GUI side.
         /// </summary>
         [TestCase("y")]
         [TestCase("YES")]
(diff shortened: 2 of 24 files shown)


MetaMorpheus/CMD/Program.cs
```diff
@@ -356,6 +302,124 @@ private static int Run(CommandLineSettings settings)
...
+                }
+
+                return 5;
+            }
+
+            // Otherwise the run can continue without a design -- but only by discarding the one that
+            // is there, which is the user's call whenever there is a user to ask.
+            if (!reportToConsole)
+            {
+                File.Delete(designPath);
```
acesnik: This is the branch taken when verbosity is `none`, and it deletes the user's design file without asking. The comment two lines above is honest that there is nobody to ask — but the outcome is that `-v none` silently destroys input.

Two things make it worth stopping on rather than porting.

It now applies to `TmtDesign.txt` as well as `ExperimentalDesign.tsv`, since `designPath` is whichever was found. So a destructive path gains a second file type.

And it is the behaviour of issue #2256, where a user reported losing their experimental design. **#2780 is open and removes the equivalent delete on the GUI side** — its comment reads "Proceed without reading the file rather than deleting it." As things stand the CLI and the GUI would disagree about whether a design that fails to parse is the user's file or ours to discard.

I should be straight that this is a decision rather than an oversight: `AtVerbosityNoneAnUnreadableDesignIsDeletedWithoutAsking` pins it, asserting both that nothing was asked and that the file is gone. So changing it means inverting a test you wrote on purpose, and it is a disagreement to settle rather than a slip to fix.

Mine is that `-v none` means "do not print", not "act differently", and that deleting an input the user cannot be asked about is a lot to infer from a verbosity flag. But if the reasoning is that an unattended run has to make progress, saying so in the description would settle it and I will stop raising it. Either way it wants deciding against #2780 rather than twice.

acesnik (commented):
🤖 `ResolveExperimentalDesign` is a real improvement on what it replaces, independent of TMT. The old code inlined the whole design-resolution decision into `Run`, and pulling it into one method with `Action<string> write` and `Func<string> readLine` injected is what makes `ResolveExperimentalDesignTests` possible at all — 320 lines of tests over a path that previously could only be exercised by running the CLI and typing at it. The mutual-exclusivity rule is right too, and for the right stated reason: two files carrying replicate structure can drift apart and nothing there could say which one the user meant.

One housekeeping note rather than a finding: the copy of `TmtExperimentalDesign.cs` here is 638 lines against #2774's current 715, so it predates the seven fixes that went in yesterday. Presumably the rebase handles it — flagging only because resolved the wrong way it would revert two silent-data-loss fixes, and #2755 has the same shape of drift.

## The silent delete, which now reaches TmtDesign.txt

Details inline at line 405. Short version: under `-v none` a design file that fails to read is deleted without asking, and this PR extends that from `ExperimentalDesign.tsv` to `TmtDesign.txt`. That is the behaviour issue #2256 was reported for, and #2780 — open, `Fixes #2256` — removes the equivalent delete on the GUI side. So the two PRs currently move in opposite directions on the same question, and it would be better to settle it once than to have the CLI and the GUI disagree about whether a malformed design is the user's file or ours to discard.

## Smaller

The parsed design is discarded — `TmtExperimentalDesign.Read` is called for its `out errors` and the returned `List<TmtFileInfo>` is dropped (line 373), so the CLI validates a design it then does not consume. Reasonable as staging if #2755 is what consumes it, but worth a line in the description saying so, because "wire the TMT experimental design into the CMD front end" reads as though the design reaches the run.

Checked and clean: the solution builds on `-c UbuntuMac` with 0 errors. I did not run the CLI end to end, and `Test/Test.csproj` cannot execute on macOS, so the 320 new tests are CI's to confirm.

acesnik:
🤖 One thing I skipped and should have led with: the 320 test lines are the strongest part of this PR.

They cover the decision table rather than sampling it — both designs present with and without normalization, no design stopping only when normalizing, both design types accepted through one parameterised test, `y`/`YES` deleting and `n`/`""`/`null` not, and `MainStopsANormalizingSearchWhoseDesignFileCannotBeRead` going through `Main` end to end and asserting the output directory is never created. `AnUnreadableDesignStopsANormalizingRunAndIsKept` carries the assertion I would otherwise have asked for.

That is a path that previously could only be exercised by running the CLI and typing at it, and the injected `write`/`readLine` are what made it reachable. Worth saying because a reader skimming +2,294 lines would not guess how much of the value is there.

MetaMorpheus/CMD/Program.cs
trishorts: You have convinced me, and I have changed it rather than defended it. `441ab73c`.

The argument that settled it is the one about the prompt: it asked *"Delete and continue empty? y/n"*, but what the user is agreeing to is continuing -- deletion was an answer to a question nobody was asked. At `-v none` there was not even that. You are right that a verbosity flag does not carry the authority to change what happens to an input; it means do not print.

The CLI now leaves the file alone on both paths, and the prompt reads "Continue without an experimental design?", matching #2780 word for word so the two cannot drift.

On the pinned test -- fair to call it out, and it deserved inverting rather than deleting. All three are now the other way round: `AgreeingContinuesWithoutTheDesignAndKeepsTheFile`, `DecliningStopsTheRunAndKeepsTheFile`, and `AtVerbosityNoneAnUnreadableDesignIsSkippedNotDeleted` asserts `File.Exists(designPath)` with the reason on the assertion. 15/15 green.

Your point about it having gained a second file type was the part I had not weighed: I widened a destructive path to `TmtDesign.txt` without re-examining whether it should exist at all.

trishorts:

You changed my mind on the delete rather than my defence of it. `441ab73c` stops deleting the design on both paths, and the prompt now reads "Continue without an experimental design?" matching #2780 word for word. The three tests that pinned the deletion are inverted, not removed. 15/15 green.

CI is still running on the last push.

acesnik (dismissed):
🤖 Approving. `441ab73c` is the whole of what I asked for, and I checked it against the head rather than reading the reply.

`File.Delete` is gone from `Program.cs` entirely — not moved or conditioned, removed. At `-v none` the method now returns 0 with the file untouched, and the prompt path continues only on `y`/`yes` and otherwise stops, both leaving the design alone. The three inverted tests are real and assert the right thing: each carries `File.Exists(designPath)` with a reason on the assertion, `DecliningStopsTheRunAndKeepsTheFile` covers `"n"`, `""` **and `null`**, so a closed stdin declines rather than falling through, and `AtVerbosityNoneAnUnreadableDesignIsSkippedNotDeleted` uses `WriteTmtDesign` — the second file type the old path had quietly gained, which is the part you said you had not weighed.

The end-to-end test driving `Program.Main` is the one I would have missed the value of. Using a *malformed* design rather than an absent one is what makes it discriminating: an absent design returns 5 from any directory, so a gate pointed at the wrong directory would still pass a test that only checked the exit code.

Two things left, neither of them code in this PR.

The description still does not say that the parsed design is discarded. `TmtExperimentalDesign.Read(...)` at `Program.cs:373` is called for its `out errors` and its return value is dropped, so the CLI validates a design it does not consume — reasonable as staging, but the title reads as though the design reaches the run.

And I can now be specific about the stale copy, having got it vague last time. This is a genuine stack: `43805109` is #2774's first commit and #2775 sits directly on it. What #2775 does not have is the three commits #2774 added afterwards — `b1a269ef` "Address review: two silent data losses, a rejected valid design", `873407d1` "normalize both sides of every design file comparison", and `7b1a5a1c` the platform-neutral fixture path. That is why the copy here is 638 lines against #2774's 715. So it is not a formatting drift to wave through: resolving that file toward this branch, or landing this one first, reverts a silent-data-loss fix — the same class of thing `441ab73c` just fixed on the CLI side. Rebasing on #2774 after it lands is the clean answer.

Nothing else changed since my last look. The only content commit is `441ab73c`; the `Product.wxs` line in a local diff arrived through the `cc405882` master merge and GitHub does not attribute it to this PR.

Unchanged from before: I have not run the CLI end to end, and `Test/Test.csproj` cannot execute on macOS, so the fixture count is CI's to confirm.

zhuoxinshi (approved):

nbollis (approved):

trishorts:
Your last review on this ended with two things left, neither of them code. Both are now closed, so this is a re-review request.

**The description now says the parsed design is discarded.** You asked for it twice and you were right that the title reads as though the design reaches the run. There is a section headed "What this PR does *not* do: consume the design" that names `Program.cs:373`, says the returned design is dropped and only an exit code comes back, and points at #2755 as the PR that consumes it. The opening sentence used to claim the opposite — that this "hands the parsed design to the search" — which was worse than silence. That is gone.

**The stale copy is resolved by #2774 landing.** #2774 merged 2026-09-02 as `d233c17b` and this branch was updated from master. `TmtExperimentalDesign.cs` and the two fixtures came back as add/add conflicts and were resolved **wholesale to master's side** — the rebase-after-it-lands answer you named, taken. Your worry was that resolving toward this branch would revert `b1a269ef`'s two silent-data-loss fixes; it did not, and I checked rather than assumed: all four shared files are byte-identical to `upstream/master`. The PR now shows only its own layer, `CMD/Program.cs` and `ResolveExperimentalDesignTests.cs`.

Nothing in `Program.cs` changed since the head you approved at — `441ab73c` is still the only content commit. Head is `d7d1694f3`.

Your approval was dismissed by the update-branch push, so this needs your click rather than your reading. @acesnik

MetaMorpheus/Test/ResolveExperimentalDesignTests.cs
```diff
@@ -0,0 +1,324 @@
...
+                ? WriteClassicDesign(folder, spectra, valid: false)
+                : WriteTmtDesign(folder, spectra, valid: false);
+
+            var written = new List<string>();
+            int code = Program.ResolveExperimentalDesign(
+                folder, new List<string> { spectra }, normalizationRequested: true,
+                reportToConsole: true, write: written.Add);
+
+            Assert.That(code == Stop);
+            Assert.That(written.Any(), "The errors are reported");
```
RayMSMS, line 183: **Minor** - `written.Any()` doesn't verify the reported error actually matches the broken design

**Problem:**
`AnUnreadableDesignStopsANormalizingRunAndIsKept` (line 169) writes a design with a specific, deliberate defect (an invalid Biological Replicate value) for both the classic and TMT cases, but the sole assertion on the console output at line 183, `Assert.That(written.Any(), "The errors are reported")`, only checks that *something* was written ΓÇö not that the written text relates to the actual parse failure.

**Impact:**
If a future change caused `ResolveExperimentalDesign` to print an unrelated or generic message (e.g. from a different code path being hit by mistake) while still returning `Stop` for some other reason, this test would still pass, masking the regression.

**Suggested fix:**
Assert on the actual error content, e.g. `written.Any(w => w.Contains("Biological Replicate"))`, mirroring the stronger assertion already used at line 316 for the `Main`-driven test.

**Affected test scenarios:**
- AnUnreadableDesignStopsANormalizingRunAndIsKept - classic/TMT invalid design - written output contains the specific parse error, not just "something"

RayMSMS (commented):
Automated code review - 1 finding(s) at severity >= Minor, posted as inline comments.

MetaMorpheus/Test/ResolveExperimentalDesignTests.cs
```diff
@@ -0,0 +1,324 @@
...
+
+        /// <summary>
+        /// Not normalizing, so an unreadable design is recoverable -- by not reading it. Agreeing
+        /// continues the run and LEAVES THE FILE ALONE. The prompt only ever asked about continuing,
+        /// so deleting was an answer to a question nobody was asked; see #2256, and #2780 for the
+        /// same change on the GUI side.
+        /// </summary>
+        [TestCase("y")]
+        [TestCase("YES")]
+        public static void AgreeingContinuesWithoutTheDesignAndKeepsTheFile(string answer)
```
RayMSMS, line 197: **Major** - Interactive-prompt tests never exercise a TMT design, and the assertion can't tell the two apart

**Problem:**
`Program.ResolveExperimentalDesign` (Program.cs:412) writes a design-type-specific message: `"An experimental design file"` for a classic design vs. `"A TMT design file"` for a TMT design, both followed by the shared text `"was found, but an error occurred reading it. Continue without an experimental design? y/n"`. `AgreeingContinuesWithoutTheDesignAndKeepsTheFile` (line 197) and `DecliningStopsTheRunAndKeepsTheFile` (line 219) only ever call `WriteClassicDesign` â€” no test in this fixture drives the interactive agree/decline path with a TMT design. Worse, the one assertion that touches the message (line 210) checks only `w.Contains("Continue without an experimental design?")`, a substring common to both branches, so even the classic case in front of it would not catch the ternary at Program.cs:412 being flipped or otherwise broken.

**Impact:**
A regression that swaps or breaks the classic/TMT prefix selection in the interactive-prompt branch, or that breaks the whole agree/decline flow specifically for TMT designs, would pass this entire fixture undetected. Given the PR's stated goal of aligning this exact prompt with the GUI's wording (#2780) for both design types, this leaves that claim only half-verified.

**Suggested fix:**
Parametrize `AgreeingContinuesWithoutTheDesignAndKeepsTheFile` and `DecliningStopsTheRunAndKeepsTheFile` over design type (as `AReadableDesignIsAcceptedAndNamed` and `AnUnreadableDesignStopsANormalizingRunAndIsKept` already do), and strengthen the message assertion to check the type-specific prefix (`"An experimental design file"` vs `"A TMT design file"`) rather than only the shared suffix.

_Reported as spanning lines 195-214; anchored here at the method signature since GitHub review comments require a single line._

MetaMorpheus/Test/ResolveExperimentalDesignTests.cs
```diff
@@ -0,0 +1,324 @@
...
+        /// continuing without the design, NOT by deleting it. This previously destroyed the file with
+        /// no prompt and no record, which meant `-v none` changed what the run did to the user's
+        /// input rather than only what it printed. A verbosity flag does not carry that authority.
+        /// </summary>
+        [Test]
+        public static void AtVerbosityNoneAnUnreadableDesignIsSkippedNotDeleted()
+        {
+            string folder = NewFolder("CmdDesignQuietSkip");
+            string spectra = Path.Combine(folder, "file.1.mzML");
+            string designPath = WriteTmtDesign(folder, spectra, valid: false);
```
RayMSMS, line 246: **Minor** - "-v none" skip is only tested for a TMT design, never for a classic one

**Problem:**
`AtVerbosityNoneAnUnreadableDesignIsSkippedNotDeleted` (line 242) exercises the "no console, no prompt, design kept" branch (Program.cs:407-410) only via `WriteTmtDesign`. No test drives this same branch with a classic `ExperimentalDesign.tsv`. Combined with the interactive-prompt gap above, the fixture covers "classic + interactive" and "TMT + verbosity-none" but never "classic + verbosity-none" or "TMT + interactive" as a complete matrix.

**Impact:**
A bug that only affects the "-v none" skip path for classic designs (e.g. an accidental branch on `hasClassicDesign` added to that guard) would not be caught by any test here.

**Suggested fix:**
Parametrize `AtVerbosityNoneAnUnreadableDesignIsSkippedNotDeleted` over design type, matching the pattern already used elsewhere in the fixture.

_Reported as spanning lines 242-258; anchored here at the method signature since GitHub review comments require a single line._

MetaMorpheus/Test/ResolveExperimentalDesignTests.cs
```diff
@@ -0,0 +1,324 @@
...
+
+            Assert.That(code == Stop);
+            Assert.That(written.Single().Contains("Only one design file type may be present"));
+
+            Directory.Delete(folder, true);
+        }
+
+        /// <summary>Refused whether or not normalization was asked for -- the ambiguity is the problem.</summary>
+        [Test]
+        public static void BothDesignFilesPresentIsRefusedEvenWithoutNormalization()
```
RayMSMS, line 80: **Minor** - Test name/doc claims normalization-independence but also flips `reportToConsole`

**Problem:**
`BothDesignFilesPresentIsRefusedEvenWithoutNormalization` is documented as proving the "both design files present" gate is "refused whether or not normalization was asked for", but relative to its sibling `BothDesignFilesPresentIsRefused` (normalizationRequested: false, reportToConsole: true), this test flips *two* parameters at once: `normalizationRequested: true` **and** `reportToConsole: false` (lines 89-90). Its only content-specific assertion, `written.Count == 0` (line 93), is actually validating verbosity suppression, not normalization independence â€” the two tests never hold `reportToConsole` fixed while only varying `normalizationRequested`.

**Impact:**
The combinations `(bothPresent, normalizationRequested: true, reportToConsole: true)` and `(bothPresent, normalizationRequested: false, reportToConsole: false)` are never exercised, so a regression that made this gate depend on both flags together (rather than on `hasClassicDesign && hasTmtDesign` alone) would not necessarily be caught, despite the test's name suggesting the independence is fully proven.

**Suggested fix:**
Split into two orthogonal tests: one that holds `reportToConsole` constant and varies only `normalizationRequested`, and a separate test (already effectively covered) for verbosity suppression.

_Reported as spanning lines 78-96; anchored here at the method signature since GitHub review comments require a single line._

MetaMorpheus/Test/ResolveExperimentalDesignTests.cs
```diff
@@ -0,0 +1,324 @@
...
+using System.IO;
+using System.Linq;
+using Nett;
+using TaskLayer;
+
+namespace Test
+{
+    /// <summary>
+    /// The command line's design-file gate. These branches decide whether a search starts at all --
+    /// and every one of them ends in either exit code 5 or a deleted file, so getting one wrong is
```
RayMSMS, line 15: **Minor** - Fixture-level doc comment contradicts every test's assertion

**Problem:**
The class summary states "every one of them ends in either exit code 5 or a deleted file". But per the PR's own description and every single test in this file, `File.Delete` was removed from this path entirely â€” no branch of `ResolveExperimentalDesign` deletes the design file anymore. The comment describes the pre-fix behavior this PR replaces, not what the code being tested actually does.

**Impact:**
A future maintainer reading this file's header could reasonably conclude a deletion path still exists and is exercised somewhere in the fixture, when in fact deletion was the exact bug (#2256) this PR removes â€” the opposite of what's documented.

**Suggested fix:**
Reword to state that every branch ends in exit code 0 or 5, and that the design file is never deleted regardless of outcome.

_Reported as spanning lines 13-17; anchored here at the specific contradicting sentence._

MetaMorpheus/Test/ResolveExperimentalDesignTests.cs
```diff
@@ -0,0 +1,324 @@
...
+
+            var written = new List<string>();
+            int code = Program.ResolveExperimentalDesign(
+                folder, new List<string> { spectra }, normalizationRequested: false,
+                reportToConsole: true, write: written.Add, readLine: () => "n");
+
+            Assert.That(code == Stop);
+            Assert.That(written.Single().Contains("Only one design file type may be present"));
+
+            Directory.Delete(folder, true);
```
RayMSMS, line 75: **Minor** - Temp-folder cleanup isn't guaranteed when an assertion fails

**Problem:**
Every test ends with a bare `Directory.Delete(folder, true)` after its assertions, with no `try`/`finally`. Since a failing `Assert.That` throws, any assertion failure skips this cleanup line and leaves the folder (and, for the `Main`-driven test, real files plus a potential `output` directory) on disk.

**Impact:**
Failed test runs leave stale folders under the test directory. This is self-healing only because `NewFolder` deletes an existing folder of the same name at the *start* of the next run â€” so a single failing run leaks disk state, and anyone inspecting a failure's artifacts on disk sees leftovers that the next CI run will silently overwrite.

**Suggested fix:**
Wrap each test body in a `try`/`finally` (or a shared IDisposable/teardown helper) so the folder is always removed regardless of assertion outcome.

_This pattern recurs at lines 75, 95, 122, 138, 160, 186, 213, 232, 257, 321 â€” flagging the first occurrence as representative; the fix applies to all of them._

PR #2776, opened 2026-08-29
Add the TMT experimental design GUI

trishorts:
**Third of three PRs splitting #2598.** #2774 (the model and its parser) merged 2026-09-02 as `d233c17b` and #2775 (the CMD front end) merged 2026-09-03 as `92dfdd6ca`, so this branch shows only its own layer: the `GUI/` folder.

Two windows: `TmtExperimentalDesignWindow` builds a design across data files and channels, and `AnnotatePlexWindow` edits one plex within it. `TmtDesign.txt` is consumed by `CMD/Program.cs`; wiring it into the run is #2755.

### The design window is the only thing that writes the design

`TmtExperimentalDesign` is the model behind the UI, as @nbollis asked on #2598, and the file is written through it — never by a hand-rolled `StreamWriter`. The window follows `ExperimentalDesignWindow`, which has owned `ExperimentalDesign.tsv` on the same terms all along:

- the grid's row set comes from `SpectraFiles` and **cannot be edited here** — files are added and removed on the main window, the only place that applies the extension allowlist and the Thermo RawFileReader licence agreement;
- the **whole** design is written **once, on Save, after it validates**, so annotating a second plex cannot truncate the first;
- the header comes from the model, including the `Sample Type` column, so a reference or bridge channel chosen from the window's own drop-down is not discarded and read back as `StudySample`.

Since the model's `Write` does not escape, the tab/newline scrubbing this window has always applied lives in the grid-to-model conversion.

### It cannot save a design the command line will not read

After writing, the window reads the design straight back through `TmtExperimentalDesign.Read` — the same parser `CMD/Program.cs` uses — and reports anything it says. A file this window reports as saved is one a run can load.

It also warns when `ExperimentalDesign.tsv` is already in the folder. Both writers target the first spectra file's directory and a run refuses to start with both files present, so with the two authoring buttons side by side the user could otherwise only discover that at run time.

### Defects found by testing rather than by inspection

Three simulated users were run against the real `Read` and `ToMzLibDesign`. Both paths this window offered produced a file MetaMorpheus could not read, while the Save dialog reported success either way:

| Path | Before | Now |
|---|---|---|
| Leave Biological Replicate alone (**the default**) | seeded `0`; `Read` rejects `< 1` → **12 errors, 0 usable files** | **0 errors, 11 channels** |
| Press the window's **Auto Fill BioRep** button | filled `0..10`; the `0` row was dropped → **11 channels in, 10 out** | **11/11 kept** |

`ValidateCurrent` checked `>= 0` while the parser checks `>= 1`, which is why neither was caught before the file was written. The seed, the auto-fill and the validation rule now all say 1.

### Round 3 — @pcruzparri's review

Four findings, three of them faces of one thing: the grid and the design file were separate objects and nothing reconciled them. Fixed as one shape, per the section above.

- **Save wrote nothing.** It validated, wrote two in-process statics, and closed. The only writer was the annotate dialog, so grid edits made after annotating never reached a file. The design window now writes.
- **A folder drop bypassed the extension allowlist and the Thermo licence gate**, and deduped on `FilePath` where `SpectraFileExists` dedupes on `FileName`. `AddPreRunFile` is `private` and both sibling dialogs annotate the file set rather than change it, so the drop is **removed** rather than plumbed through — which closes all three at once.
- **Switching either drop-down discarded what was typed**, and carried stale channel labels forward. Channel counts across the eight tag types are 6/10/11/18/4/8/4/12, so **4 is the only count that occurs twice** — iTRAQ4 is `114,115,116,117` and diLeu4 is `115,116,117,118`. The carry-over is now keyed on the tag *type*, and `Tag` always comes from the tag's own table.
- **Paste misaligned channels.** Rows split with `RemoveEmptyEntries`, so a blank row moved every sample below it onto the wrong channel — and an unused channel mid-plex is an ordinary TMT design. Cells split on tab *and* comma, so `HeLa, 2h` became a name and a condition. Rows are positional now and cells split on tab only.

Also taken: unplexed files are refused rather than dropped from the written file; plex names compare `OrdinalIgnoreCase` everywhere; `s_seedErrors` is cleared once shown; seeding runs from the **checked** files and the duplicate call outside the picker's `ShowDialog` guard is gone; the three `catch { /* ignore */ }` blocks are gone; the annotate window commits the cell being edited before reading the grid; and the XAML no longer promises drag-to-renumber-fractions, which nothing implemented.

The two integer columns bind with the default `UpdateSourceTrigger`, so `e.Cancel` in `DgTmt_CellEditEnding` genuinely rejects an edit rather than reverting a value the source had already taken.

Build warnings this PR had introduced are gone (CS0105, CS0169 ×2, CS8632 ×2, CS8073), and all cosmetic churn is reverted — **the PR is now +1248 / −0 against master**, and `MainWindow.xaml.cs` is `29+/0-`.

### Verified

- **Default path** — pick plex, pick TMT11, type names and conditions, never touch Biological Replicate, Save → `0 errors`, 11 annotations, projects with reporter m/z ascending.
- **Two-plex bridge** — annotate Plex1, then Plex2 → the file keeps **both** plexes, `Read` → 2 files / 0 errors, and `IsReferenceChannel` is true for exactly `131N` (bridge) and `131C` (reference). The shared bridge sample name across plexes is correctly **not** a duplicate error.
- **Fractionated plex via Auto Fill** — one plex over three fractions with a carrier channel → 3 files / 0 errors, **11 channels on every fraction**, `126` reads back as `Carrier`.

`Test.csproj` does not reference `GUI.csproj`, so the WPF handlers cannot be driven from a test — that behaviour is verified by reading and building, not by running, and I would rather say so than imply otherwise. `TmtDesignGuiContractTests` pins what is reachable: the iTRAQ4/diLeu4 collision and that 4 is the only shared count (so it fails if a colliding tag type is ever added), that a stale label is rejected by `ToMzLibDesign` rather than accepted, and that a whole two-plex design built the way the window builds it round-trips through `Read`, projects through `ToMzLibDesign`, and is accepted by `Program.ResolveExperimentalDesign`.

**Full suite green: 1962 passed, 0 failed, 1 skipped** (`dotnet test Test/Test.csproj`, 9m08s), including the 178 across the TMT design, sample-type, multiplex, mzLib-projection and `ResolveExperimentalDesign` fixtures. `GUI.csproj` builds with 0 errors, and none of the remaining warnings are in these files.

### Not in this PR

The GUI's run gate does not know about `TmtDesign.txt` and nothing in `TaskLayer` reads it. That wiring is **#2755**, which also carries this PR's GUI files byte-identically — the same fixes are on that branch, so its merge cannot resurrect them.

CMD's refusal of two design files is policy from merged #2775, so this warns rather than changes it. And `ValidateDesign` re-types the contiguity walk `EngineLayer.ExperimentalDesign.GetErrorsInExperimentalDesign` already implements; factoring those together means editing merged EngineLayer code, which belongs in its own PR.

### DIFF
--- a/MetaMorpheus/GUI/Views/AnnotatePlexWindow.xaml.cs
+++ b/MetaMorpheus/GUI/Views/AnnotatePlexWindow.xaml.cs
@@ -1,0 +1,278 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Windows;
+using System.Windows.Controls;
+using System.Windows.Input;
+using System.Windows.Media;
+using EngineLayer;
+
+namespace MetaMorpheusGUI
+{
+    public partial class AnnotatePlexWindow : Window
+    {
+        private readonly List<string> _plexNames;
+        private readonly Dictionary<string, List<PlexAnnotation>> _existingAnnotations;
+        private readonly Dictionary<string, List<PlexFileEntry>> _plexFileMap;
+
+        public string SavedPlexName { get; private set; } = "";
+        public List<PlexAnnotation> SavedAnnotations { get; private set; }
+
+        /// <summary>
+        /// The choices offered by the Sample Type column. Taken straight from
+        /// <see cref="EngineLayer.TmtExperimentalDesign.SampleTypeNames"/> so the drop-down cannot
+        /// offer a value the design-file parser would reject.
+        /// </summary>
+        public static IReadOnlyList<string> SampleTypeChoices =>
+            EngineLayer.TmtExperimentalDesign.SampleTypeNames;
+
+        private List<PlexAnnotation> _currentRows = new();
+        private Point _dragStartPoint;
+        private PlexAnnotation _dragSourceItem;
+
+        public AnnotatePlexWindow(List<string> plexNames,
+            Dictionary<string, List<PlexAnnotation>> existingAnnotations,
+            Dictionary<string, List<PlexFileEntry>> plexFileMap)
+        {
+            InitializeComponent();
+            _plexNames = plexNames ?? new();
+            _existingAnnotations = existingAnnotations ?? new();
+            _plexFileMap = plexFileMap ?? new();
+
+            PlexComboBox.ItemsSource = _plexNames;
+            TagTypeComboBox.ItemsSource = Enum.GetValues(typeof(IsobaricMassTagType)).Cast<IsobaricMassTagType>();
+        }
+
+        private void PlexComboBox_SelectionChanged(object sender, SelectionChangedEventArgs e) => RegenerateRows();
+        private void TagTypeComboBox_SelectionChanged(object sender, SelectionChangedEventArgs e) => RegenerateRows();
+
+        private void RegenerateRows()
+        {
+            if (PlexComboBox.SelectedItem == null || TagTypeComboBox.SelectedItem == null)
+            {
+                AnnotationGrid.ItemsSource = null;
+                return;
+            }
+
+            var plex = PlexComboBox.SelectedItem.ToString();
+            var tagType = (IsobaricMassTagType)TagTypeComboBox.SelectedItem;
+            var reporterLabels = GetReporterIonLabels(tagType);
+
+            if (_existingAnnotations.TryGetValue(plex, out var existing) &&
+                existing.Count == reporterLabels.Count)
+            {
+                _currentRows = existing
+                    .Select(a => new PlexAnnotation
+                    {
+                        Tag = a.Tag,
+                        SampleName = a.SampleName,
+                        Condition = a.Condition,
+                        BiologicalReplicate = a.BiologicalReplicate,
+                        SampleType = a.SampleType
+                    }).ToList();
+            }
+            else
+            {
+                _currentRows = reporterLabels
+                    .Select(lbl => new PlexAnnotation
+                    {
+                        Tag = lbl,
+                        SampleName = "",
+                        Condition = "",
+                        BiologicalReplicate = 0,
+                        SampleType = EngineLayer.TmtExperimentalDesign.ToDesignFileValue(
+                            EngineLayer.TmtSampleType.StudySample)
+                    }).ToList();
+            }
+
+            AnnotationGrid.ItemsSource = _currentRows;
+            AnnotationGrid.Items.Refresh();
+        }
+
+        /// <summary>
+        /// The channel labels for a tag, from the single table in
+        /// <see cref="IsobaricMassTag.GetReporterIonLabels(IsobaricMassTagType)"/>.
+        /// </summary>
+        /// <remarks>
+        /// This used to be a second copy of that table, and the two had drifted: TMT10 ended "131"
+        /// here against "131N" there, iTRAQ8 ended "121" against "120", and diLeu12 disagreed
+        /// throughout. A label typed into the design file by this window then failed to match any
+        /// channel of the tag, so the design could not be projected onto mzLib's
+        /// <see cref="IExperimentalDesign"/> at all.
+        ///
+        /// The iTRAQ8 disagreement was the one where this window had it right: 8-plex has no 120
+        /// channel -- it is skipped because the phenylalanine immonium ion sits at 120.081 -- and the
+        /// eighth reagent is 121, which is what the reporter ion at that index has always been.
+        /// IsobaricMassTag was corrected rather than this window changed to match it.
+        /// </remarks>
+        private static List<string> GetReporterIonLabels(IsobaricMassTagType type) =>
+            IsobaricMassTag.GetReporterIonLabels(type) ?? new List<string>();
+
+        private void Paste_CanExecute(object sender, CanExecuteRoutedEventArgs e)
+        {
+            e.CanExecute = _currentRows is { Count: > 0 };
+        }
+
+        private void Paste_Executed(object sender, ExecutedRoutedEventArgs e)
+        {
+            var raw = Clipboard.GetText();
+            if (string.IsNullOrWhiteSpace(raw) || AnnotationGrid.ItemsSource == null)
+                return;
+
+            int targetColumn = -1;
+            int startRow = 0;
+
+            if (AnnotationGrid.CurrentCell != null)
+            {
+                targetColumn = AnnotationGrid.CurrentCell.Column?.DisplayIndex ?? -1;
+                if (AnnotationGrid.CurrentCell.Item is PlexAnnotation pa)
+                {
+                    startRow = _currentRows.IndexOf(pa);
+                    if (startRow < 0) startRow = 0;
+                }
+            }
+
+            if (targetColumn < 0)
+                targetColumn = 1; // Sample Name column (Tag=0, Sample=1, Condition=2, BioRep=3)
+
+            var lines = raw
+                .Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries)
+                .Select(l => l.Trim())
+                .Where(l => !string.IsNullOrEmpty(l))
+                .ToList();
+
+            if (!lines.Any())
+                return;
+
+            for (int i = 0; i < lines.Count && (startRow + i) < _currentRows.Count; i++)
+            {
+                var row = _currentRows[startRow + i];
+                var cols = lines[i]
+                    .Split(new[] { '\t', ',' }, StringSplitOptions.RemoveEmptyEntries)
+                    .Select(c => c.Trim())
+                    .ToArray();
+
+                if (cols.Length == 0)
+                    continue;
+
+                switch (targetColumn)
+                {
+                    case 1: // Sample Name column
+                        row.SampleName = cols[0];
+                        if (cols.Length >= 2)
+                            row.Condition = cols[1];
+                        break;
+                    case 2: // Condition column
+                        row.Condition = cols[0];
+                        break;
+                    case 3: // Biological Replicate column (single integer)
+                        if (int.TryParse(cols[0], out int bio) && bio >= 0)
+                            row.BiologicalReplicate = bio;
+                        break;
+                    default:
+                        row.SampleName = cols[0];
+                        break;
+                }
+            }
+
+            AnnotationGrid.Items.Refresh();
+        }
+
+        private void AutoFillReplicates_Click(object sender, RoutedEventArgs e)
+        {
+            for (int i = 0; i < _currentRows.Count; i++)
+                _currentRows[i].BiologicalReplicate = i;
+
+            AnnotationGrid.Items.Refresh();
+        }
+
+        private void ValidateButton_Click(object sender, RoutedEventArgs e)
+        {
+            var error = ValidateCurrent();
+            MessageBox.Show(error ?? "Validation passed.",
+                error == null ? "OK" : "Validation Error");
+        }
+
+        private string ValidateCurrent()
+        {
+            if (!_currentRows.Any())
+                return "No rows to validate.";
+
+            foreach (var r in _currentRows)
+            {
+                if (r.BiologicalReplicate < 0)
+                    return $"Biological replicate must be >= 0 (row tag {r.Tag}).";
+            }
+            return null;
+        }
+
+        private void SaveButton_Click(object sender, RoutedEventArgs e)
+        {
+            var error = ValidateCurrent();
+            if (error != null)
+            {
+                MessageBox.Show(error, "Cannot Save");
+                return;
+            }
+
+            var plex = PlexComboBox.SelectedItem?.ToString();
+            if (string.IsNullOrEmpty(plex))
+            {
+                MessageBox.Show("Select a Plex before saving.");
+                return;
+            }
+
+            if (!_plexFileMap.TryGetValue(plex, out var fileEntries) || fileEntries.Count == 0)
+            {
+                MessageBox.Show($"No files mapped to plex '{plex}'.");
+                return;
+            }
+
+            SavedPlexName = plex;
+            SavedAnnotations = _currentRows.Select(a => a).ToList();
+
+            // ENSURE: fixed filename "TmtDesign.txt" is used for saving (overwrites if exists)
+            var defaultDir = Path.GetDirectoryName(fileEntries.First().FilePath) ?? Environment.CurrentDirectory;
+            var defaultName = "TmtDesign.txt";
+            var savePath = Path.Combine(defaultDir, defaultName);
+
+            try
+            {
+                using var writer = new StreamWriter(savePath, false);
+                writer.WriteLine("File\tPlex\tSample Name\tTMT Channel\tCondition\tBiological Replicate\tFraction\tTechnical Replicate");
+
+                foreach (var fe in fileEntries)
+                {
+                    int fractionValue = fe.Fraction; // USE USER-DEFINED FRACTION
+                    int techRepValue = fe.TechnicalReplicate;
+
+                    foreach (var ann in _currentRows)
+                    {
+                        writer.WriteLine(
+                            $"{fe.FilePath}\t{plex}\t{Escape(ann.SampleName)}\t{ann.Tag}\t{Escape(ann.Condition)}\t{ann.BiologicalReplicate}\t{fractionValue}\t{techRepValue}");
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                MessageBox.Show("Failed to write file: " + ex.Message);
+                return;
+            }
+
+            MessageBox.Show($"Saved TMT design (files x channels = {fileEntries.Count} x {_currentRows.Count}) to:\n{savePath}");
+            DialogResult = true;
+            Close();
+        }
+
+        private void CancelButton_Click(object sender, RoutedEventArgs e)
+        {
+            DialogResult = false;
+            Close();
+        }
+
+        private static string Escape(string s) =>
+            string.IsNullOrEmpty(s) ? "" :
+            s.Replace('\t', ' ').Replace('\r', ' ').Replace('\n', ' ');
+    }
+}
--- a/MetaMorpheus/GUI/Views/AnnotatePlexWindow.xaml.cs
+++ b/MetaMorpheus/GUI/Views/AnnotatePlexWindow.xaml.cs
@@ -1,11 +1,9 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Input;
-using System.Windows.Media;
 using EngineLayer;
 
 namespace MetaMorpheusGUI
@@ -13,11 +11,18 @@ namespace MetaMorpheusGUI
     public partial class AnnotatePlexWindow : Window
     {
         private readonly List<string> _plexNames;
-        private readonly Dictionary<string, List<PlexAnnotation>> _existingAnnotations;
-        private readonly Dictionary<string, List<PlexFileEntry>> _plexFileMap;
+        private readonly Dictionary<string, List<PlexAnnotation>> _annotations;
 
-        public string SavedPlexName { get; private set; } = "";
-        public List<PlexAnnotation> SavedAnnotations { get; private set; }
+        /// <summary>
+        /// Every plex this window holds annotations for, including ones the user never opened.
+        /// </summary>
+        /// <remarks>
+        /// This window no longer writes the design. <see cref="TmtExperimentalDesignWindow"/> owns the
+        /// file and writes the whole of it once, on its own Save, the way
+        /// <see cref="ExperimentalDesignWindow"/> owns ExperimentalDesign.tsv. Two writers was how a
+        /// design came to be written from a grid nothing had reconciled it with.
+        /// </remarks>
+        public IReadOnlyDictionary<string, List<PlexAnnotation>> SavedAnnotationsByPlex => _annotations;
 
         /// <summary>
         /// The choices offered by the Sample Type column. Taken straight from
@@ -28,17 +33,16 @@ namespace MetaMorpheusGUI
             EngineLayer.TmtExperimentalDesign.SampleTypeNames;
 
         private List<PlexAnnotation> _currentRows = new();
-        private Point _dragStartPoint;
-        private PlexAnnotation _dragSourceItem;
+        private string _currentPlex;
 
         public AnnotatePlexWindow(List<string> plexNames,
-            Dictionary<string, List<PlexAnnotation>> existingAnnotations,
-            Dictionary<string, List<PlexFileEntry>> plexFileMap)
+            Dictionary<string, List<PlexAnnotation>> existingAnnotations)
         {
             InitializeComponent();
             _plexNames = plexNames ?? new();
-            _existingAnnotations = existingAnnotations ?? new();
-            _plexFileMap = plexFileMap ?? new();
+            _annotations = new Dictionary<string, List<PlexAnnotation>>(
+                existingAnnotations ?? new Dictionary<string, List<PlexAnnotation>>(),
+                StringComparer.OrdinalIgnoreCase);
 
             PlexComboBox.ItemsSource = _plexNames;
             TagTypeComboBox.ItemsSource = Enum.GetValues(typeof(IsobaricMassTagType)).Cast<IsobaricMassTagType>();
@@ -47,11 +51,31 @@ namespace MetaMorpheusGUI
         private void PlexComboBox_SelectionChanged(object sender, SelectionChangedEventArgs e) => RegenerateRows();
         private void TagTypeComboBox_SelectionChanged(object sender, SelectionChangedEventArgs e) => RegenerateRows();
 
+        /// <summary>
+        /// Rebuilds the channel grid for the plex and tag type now selected, after storing whatever
+        /// the user had typed for the previous one.
+        /// </summary>
+        /// <remarks>
+        /// Nothing else wrote the grid back, and neither combo starts with a selection, so without the
+        /// stash annotating plex 1 and then selecting plex 2 discarded plex 1's samples with no
+        /// prompt -- and the design was then written with plex 1 as a file-only placeholder row.
+        ///
+        /// The Tag always comes from the tag type's own table rather than from the stored annotation.
+        /// Channel counts across the eight tag types are 6/10/11/18/4/8/4/12, so 4 is the only count
+        /// that occurs twice: iTRAQ4 is 114,115,116,117 and diLeu4 is 115,116,117,118. Reusing stored
+        /// tags whenever the counts matched therefore carried iTRAQ4's labels into a diLeu4 plex, and
+        /// <see cref="EngineLayer.TmtExperimentalDesign.ToMzLibDesign"/> rejects a tag that is not a
+        /// channel of the selected type -- so that design would not load.
+        /// </remarks>
         private void RegenerateRows()
         {
+            CommitPendingEdits();
+            StashCurrentRows();
+
             if (PlexComboBox.SelectedItem == null || TagTypeComboBox.SelectedItem == null)
             {
                 AnnotationGrid.ItemsSource = null;
+                _currentPlex = null;
                 return;
             }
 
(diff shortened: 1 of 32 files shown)


MetaMorpheus/GUI/Views/TmtExperimentalDesignWindow.xaml.cs
```diff
@@ -0,0 +1,471 @@
...
+            foreach (var dir in dirs)
+            {
+                var designPath = Path.Combine(dir!, GlobalVariables.TmtExperimentalDesignFileName);
+                if (!File.Exists(designPath))
+                    continue;
+
+                var filesInDir = rawSet.Where(f => string.Equals(Path.GetDirectoryName(f), dir, StringComparison.OrdinalIgnoreCase))
+                                       .ToList();
+
+                var tmtFiles = TmtExperimentalDesign.Read(designPath, filesInDir, out var _);
```
acesnik: `out var _` discards the whole error list, so this opens with a partial design and says nothing. Every failure mode #2774 added a message for lands here and stops: a channel described twice with disagreeing samples, a `Biological Replicate` below one, a row too short, an unrecognised `Sample Type`, and the "none of the rows name a file in this run" case — which is the one most likely to happen in the GUI, since the design travels with the data rather than with the run.

The list is already in hand; it needs somewhere to go. Whatever the window does with it, the case worth deciding is the last one: a design naming no file in the run currently opens as an empty grid, which looks identical to having no design at all.

acesnik (commented):
🤖 The window is more complete than I expected from a draft — `AnnotatePlexWindow`, the plex models, and a design window that reads and writes the file, wired into `MainWindow` in six lines of XAML. And it compiles: I built `GUI/GUI.csproj` with `EnableWindowsTargeting` on macOS and the markup compiler runs over all three new XAML files, 0 errors. Worth knowing because the `UbuntuMac` solution configuration skips the GUI entirely, so nothing in CI's Linux path would have told you that.

One housekeeping note rather than a finding: the copy of `TmtExperimentalDesign.cs` here is 638 lines against #2774's current 715, so it predates the seven fixes that went in yesterday. Presumably the rebase handles it — flagging only because resolved the wrong way it would revert two silent-data-loss fixes, and #2755 has the same shape of drift.

## The load path discards the parser's errors

Details inline at line 442. `TmtExperimentalDesign.Read(designPath, filesInDir, out var _)` throws away the error list, so opening a malformed `TmtDesign.txt` gives whatever partial result came back with no indication anything was wrong. The case I would most want handled is a design naming no file in the run: that currently opens as an empty grid, which looks identical to having no design at all.

To be clear about what this is not — the window does validate before saving, via `ValidateDesign()` and a "Cannot Save" refusal. This is the load path only. There is a related question about how those two sets of rules relate to each other, which I have put in a separate comment rather than here.

## Not verified

I cannot run the GUI, so layout, tab order, and the enable/disable behaviour of the annotate window are unchecked — compiling is not running. `Test/Test.csproj` cannot execute on macOS either, so the two carried test files are CI's to confirm.

acesnik:
🤖 Separate from the load-path note, and the more interesting half: `ValidateDesign()` is an independent validator, and its rules and the parser's barely intersect.

| `ValidateDesign()` enforces | `TmtExperimentalDesign.Read` enforces |
|---|---|
| fractions start at 1 and are contiguous | Bio / Fraction / TechRep >= 1 |
| tech reps start at 1 and contiguous per fraction | `Sample Type` is recognised |
| no duplicate (Fraction, TechnicalReplicate) | one channel not described twice with disagreeing samples |
| — | (Sample, Bio, Fraction, Tech) unique **per plex** |
| — | rows name a file in this run |

So each accepts designs the other refuses. Concretely: the window will save a design with two channels in one plex carrying different samples, or the same sample twice within a plex — both of which the parser now reports, because those were two of the seven fixes on #2774. A user authoring in the GUI can produce a file the CLI then refuses. In the other direction the parser is content with non-contiguous fractions the window will not let you save.

That is the part worth deciding, and it will keep coming up: the fixes go into the parser, and the GUI's save path cannot see any of them. Validating through the parser rather than alongside it — or calling `Read` on the written file and surfacing what comes back — would make the two agree, and would give the load path at line 442 somewhere to put its errors. Same fix twice.

The contiguity rules are a real addition worth keeping wherever validation ends up; the parser has no notion of a missing fraction number.

MetaMorpheus/GUI/Views/TmtExperimentalDesignWindow.xaml.cs
trishorts: Fixed in `9f2d07c9`. `out var _` is gone; `SeedFromDesignFiles` collects the messages into `s_seedErrors`, prefixed with the design path since seeding walks several folders.

On where they go: not at drag-drop time. Seeding runs from three `MainWindow` call sites while the user is still adding files, so a box per dropped folder is noise -- and all three sites wrap the call in `catch { }`, which is the habit that produced this in the first place. They surface when the TMT window opens, which is the moment the empty grid would otherwise be the only thing the user sees. That covers the case you singled out: a design naming no file in the run now says so instead of looking identical to having no design at all.

trishorts:

`9f2d07c9` replaces `out var _` with a collected error list. It surfaces when the TMT window opens rather than at drag-drop time, since seeding runs from three `MainWindow` call sites while the user is still adding files — and all three wrap the call in `catch { }`, which is the habit that produced the discard. That covers the case you singled out: a design naming no file in the run now says so instead of looking identical to having no design at all.

acesnik (commented):
🤖 The load-path fix is verified and the judgement in it is better than the fix I asked for. `out var _` is gone, `SeedFromDesignFiles` collects each message prefixed with the design path, and they surface in the window constructor rather than at drag-drop — which is right for the reason you give: seeding runs from three `MainWindow` sites while the user is still adding files, so a box per dropped folder would train people to dismiss it. The case I singled out is covered: a design naming no file in the run now says so instead of opening an empty grid that looks identical to having no design.

**The second comment is not addressed, though, and following it to the end turns it from a design question into a concrete defect.** I had described `ValidateDesign()` and the parser as enforcing overlapping-but-different rules. What I had not done was check what that costs on the simplest real design.

`ValidateDesign()` groups by `(Fraction, TechnicalReplicate)` and rejects any repeat. It does not scope that by plex — `_rows` spans plexes, since `TmtDesignRow` carries `Plex` — and `TmtDesignRow(filePath)` defaults `Fraction` and `TechnicalReplicate` to 1. So:

- Add two spectra files and open the window. Both rows are `(1, 1)`, the duplicate check fires, and **Save is refused** with "Duplicate Fraction/Technical Replicate combination: Fraction 1, Technical Replicate 1" before the user has done anything wrong. For a two-plex experiment `(1, 1)` on both files is the *correct* design — one fraction, one technical replicate, two plexes — so the only way past the validator is to enter a fraction or replicate number that is not true.
- Worse, it does not round-trip. `SeedFromDesignFiles` stores `Fraction`, `TechnicalReplicate` and `Plex` per file, and the constructor copies all three onto the row. So a two-plex design the parser **accepts** loads into the grid and then cannot be saved back out.

That is the same divergence from the other direction. I wrote before that a user authoring in the GUI can produce a file the CLI refuses; the reverse also holds, and the reverse is the one that blocks the feature's basic case. Adding `r.Plex` to the grouping key fixes the immediate refusal, and matches the parser, which already scopes its uniqueness per plex. The contiguity checks above it have the same unscoped shape and are worth the same look — two plexes with fractions 1-2 and 1-3 pass today only because the union happens to be contiguous.

For the record, the table from that comment still holds exactly at this head: `ValidateDesign` enforces fraction start and contiguity, tech-rep start and contiguity per fraction, and the duplicate pair; it enforces nothing about plex, channel, sample uniqueness, `Sample Type`, or biological replicate — all of which the parser does. So the window will still save a design with one channel carrying two different samples, which was one of #2774's seven fixes.

Smaller: `s_seedErrors` is static and cleared only at the top of `SeedFromDesignFiles`, so reopening the window without re-seeding shows the same errors again. Harmless, and arguably better than losing them.

Commenting rather than approving because the grouping key needs a push, and it is a draft besides. Unchanged from last time on what I cannot check: `GUI.csproj` compiles on macOS with `EnableWindowsTargeting` and the markup compiler runs over all three XAML files at 0 errors, but compiling is not running — so layout, tab order and the enable/disable behaviour are unverified, and I am reasoning about the Save refusal from the validator's code rather than from clicking the button.

MetaMorpheus/GUI/Views/TmtExperimentalDesignWindow.xaml.cs
```diff
@@ -0,0 +1,498 @@
...
+                var techs = grp.Select(r => r.TechnicalReplicate).Distinct().OrderBy(t => t).ToList();
+                if (techs.First() != 1)
+                    return $"Fraction {grp.Key}: technical replicates must start at 1.";
+                int maxTech = techs.Last();
+                for (int t = 1; t <= maxTech; t++)
+                    if (!techs.Contains(t))
+                        return $"Fraction {grp.Key}: missing technical replicate {t}.";
+            }
+
+            var duplicatePair = _rows.GroupBy(r => (r.Fraction, r.TechnicalReplicate))
```
acesnik: This grouping is not scoped by plex, and `_rows` spans plexes — `TmtDesignRow` carries `Plex`, and its constructor defaults `Fraction` and `TechnicalReplicate` to 1. So two spectra files give two `(1, 1)` rows and Save is refused before the user changes anything, even though `(1, 1)` on both files is the correct design for a two-plex run with one fraction and one technical replicate.

It also breaks the round trip: `SeedFromDesignFiles` stores fraction, replicate and plex per file and the constructor copies all three onto the row, so a two-plex design the parser accepts loads here and then cannot be saved back. Adding `r.Plex` to the key matches the parser, which already scopes uniqueness per plex. The contiguity checks above have the same unscoped shape.

trishorts: Fixed, including the contiguity checks — you were right that they have the same shape.

`ValidateDesign` now runs every check inside a `GroupBy(r => r.Plex ?? string.Empty, StringComparer.OrdinalIgnoreCase)`: fraction-starts-at-1, fraction contiguity, technical-replicate-starts-at-1 and contiguity per fraction, and the duplicate `(Fraction, TechnicalReplicate)` key. Ordinal-ignore-case to match how `TmtExperimentalDesign` keys its own per-file plex state. Every message now names the plex it is about (`Plex 2: duplicate Fraction/Technical Replicate combination: ...`, and `Files with no plex assigned: ...` for the unnamed group), since a global message is unactionable once the check is scoped.

The `TechnicalReplicate < 1` check stays global — it is a per-row domain check, not a per-plex one.

Not covered by a test: `ValidateDesign` is a private method on a WPF `Window` and `Test.csproj` references `CMD`, `EngineLayer`, `GuiFunctions` and `TaskLayer` but not `GUI`, so there is no seam to reach it from without lifting the rule out of the window. Worth doing — the same per-plex rule is the one `TmtExperimentalDesign` enforces on the parse side — but it is a bigger change than this PR, so flagging rather than folding it in.

nbollis (dismissed):

trishorts:
Re-review request — everything from your last two reviews is in at head (`6cf5003d1`), including the fix you had not seen yet.

**The grouping key.** You were right, and right that the contiguity checks above it had the same unscoped shape. `ValidateDesign` now runs every check inside `GroupBy(r => r.Plex ?? string.Empty, StringComparer.OrdinalIgnoreCase)` — fraction-starts-at-1, fraction contiguity, technical-replicate-starts-at-1 and contiguity per fraction, and the duplicate `(Fraction, TechnicalReplicate)` pair. Ordinal-ignore-case to match how `TmtExperimentalDesign` keys its own per-file plex state. Every message names the plex it is about, since a global message is unactionable once the check is scoped, and the unnamed group reads "Files with no plex assigned". The case you walked through — two spectra files, both rows `(1, 1)`, Save refused before the user has done anything wrong — no longer fires, and the round trip you found closes with it: a two-plex design the parser accepts now loads and saves back out. The `TechnicalReplicate < 1` check stays global, being a per-row domain check rather than a per-plex one. The reasoning is in a comment at the `foreach`, so the next person does not unscope it.

**The load path** you already verified at the previous head and it is unchanged.

**The stale copy.** #2774 merged 2026-09-02 as `d233c17b`; this branch was updated from master and `TmtExperimentalDesign.cs` plus the two fixtures resolved **wholesale to master's side**, so the two silent-data-loss fixes stand. All four shared files are byte-identical to `upstream/master`, and the four members this GUI calls — `Read`, `SampleTypeNames`, `ToDesignFileValue`, `TryParseSampleType` — all exist on that copy. The PR now shows only the `GUI/` folder.

**`s_seedErrors` staying static** I left as-is, on your reading that showing them again on a reopen without a re-seed beats losing them.

**Still not covered by a test, and flagging it rather than claiming otherwise:** `ValidateDesign` is a private method on a WPF `Window`, and `Test.csproj` references `CMD`, `EngineLayer`, `GuiFunctions` and `TaskLayer` but not `GUI` — so there is no seam to reach it from without lifting the rule out of the window. Worth doing, since this is the same per-plex rule `TmtExperimentalDesign` enforces on the parse side, but it is a bigger change than this PR.

It is no longer a draft. @acesnik

pcruzparri (commented):
Code review — 4 findings, 3 at severity >= High, posted as inline comments.

Framing worth stating up front: this is authoring UI for a file nothing in the GUI or TaskLayer reads yet — `TmtDesign.txt` is consumed only by `CMD/Program.cs`. So these are "get the authoring right before anything depends on the file" comments. Related: the GUI's run gate still checks only `ExperimentalDesign.tsv` (`MainWindow.xaml.cs:1001-1015`) and points a TMT user at the "Set Experimental Design" button, and `CMD/Program.cs:337-347` hard-refuses when both design files exist while this PR puts the two authoring buttons side by side writing into one folder.

Three of the four inline comments are faces of one thing — the grid and the design file are separate objects and nothing reconciles them — so they may be worth fixing as one shape rather than three patches.

Smaller things not worth their own threads:

- Unplexed files pass `ValidateDesign` (`:145-149`) but `plexFileMap` filters them out (`:313-315`), so they are absent from the written file. Either reject a blank plex or carry them through; `Write` already emits placeholder rows.
- Plex names are keyed with different comparers in four places: `_plexAnnotations` is `OrdinalIgnoreCase` (`:23`) but the dictionary handed to the annotate window is rebuilt without one (`:328`); `plexNames` dedupes `OrdinalIgnoreCase` (`:303`) while `plexFileMap` groups case-sensitively (`:316`). `PLEX1` in the file and `Plex1` in the grid opens blank, and Save writes that over the real annotations.
- `SeedFromDesignFiles` gets every spectra file (`MainWindow.xaml.cs:214/542/1762`) but the window lists only `Use == true` (`:49`), so deselecting a file produces a "design did not contain the file(s)" warning about a file the user deliberately excluded.
- `s_seedErrors` (`:449`) is cleared only at the top of `SeedFromDesignFiles` (`:454`), never after the constructor shows it (`:39-46`), so the same warning box reappears on every open.
- The three `catch { /* ignore */ }` blocks (`MainWindow.xaml.cs:217`, `:545`, `:1767`) still swallow thrown exceptions. `s_seedErrors` covers errors the parser returns, not ones it throws — appending in the catch would close that cheaply.
- `TmtExperimentalDesignWindow.xaml:49` tells users "Reorder rows by drag to renumber fractions automatically", but `Row_Drop` (`:216-236`) only moves the item in the collection and no consumer depends on row order. Implement it or drop the sentence and the handlers.
- New build warnings: CS0105 (duplicate `using System;`, `TmtExperimentalDesignWindow.xaml.cs:1-2`), CS0169 x2 (unused `_dragStartPoint`/`_dragSourceItem`, `AnnotatePlexWindow.xaml.cs:31-32`), CS8632 x2 (nullable annotations in `PlexModels.cs` without Nullable enabled), CS8073 (`:128`).
- Churn: BOM added to `MainWindow.xaml.cs:1`; stray space at `:2289`; "File-Specific Parameters" column narrowed 250 to 170 (`MainWindow.xaml:654`); whitespace-only change at `:601`; trailing blank lines at `:2371-2372`; the scaffolding comment at `:626` ("RawDataForDataGrid assumed to have a constructor taking file path (adjust if different)"); the `// NEW:` markers.

Two things I could not settle by reading, worth a manual check: whether `AnnotatePlexWindow` loses an in-progress cell edit on Save (it has no `CommitPendingEdits`, unlike `TmtExperimentalDesignWindow`, which calls one before every grid read); and whether `e.Cancel` in `DgTmt_CellEditEnding` actually reverts a value already written by `UpdateSourceTrigger=PropertyChanged` (`TmtExperimentalDesignWindow.xaml:28,31`).

MetaMorpheus/GUI/Views/TmtExperimentalDesignWindow.xaml.cs
```diff
@@ -0,0 +1,513 @@
...
+        #region Buttons
+        private void ValidateButton_Click(object sender, RoutedEventArgs e)
+        {
+            CommitPendingEdits();
+            var err = ValidateDesign();
+            MessageBox.Show(err ?? "Validation passed.",
+                err == null ? "OK" : "Validation Error");
+        }
+
+        private void SaveButton_Click(object sender, RoutedEventArgs e)
```
pcruzparri, line 229: **High** — Save validates, closes, and writes nothing

**Problem:**
`SaveButton_Click` validates, writes the in-process statics `s_fileState` / `s_fileToPlex` (`:364-369`), sets `DialogResult = true` and closes. No file is written. The only code that writes a design is `AnnotatePlexWindow` calling `TmtExperimentalDesign.Write` (`AnnotatePlexWindow.xaml.cs:249`).

On the receiving side `MainWindow` reads only `design.FilePath` from `GetResults()` (`MainWindow.xaml.cs:618-641`) — `Fraction`, `TechnicalReplicate` and `Plex` on the `TmtDesignResult` record (`:20`) are never read anywhere in the tree.

The in-memory copy isn't durable either. `SeedFromDesignFiles` overwrites `s_fileState` unconditionally (`:489`) and `MainWindow` calls it on every add-spectra action (`:215`, `:543`, `:1765`). At `MainWindow.xaml.cs:539-547` that call sits **outside** `if (openPicker.ShowDialog() == true)`, so even cancelling the file picker re-seeds from disk and clobbers unsaved edits — and it duplicates the seeding `AddPreRunFiles` already did at `:536`.

**Impact:**
Set fractions, technical replicates and plexes, click Save — which validates and closes exactly like a save that worked — and there is no design file. The work lives in static dictionaries that die with the process, and gets overwritten before then by the next add-spectra action, including a cancelled one. Edits made in the grid after annotating never reach the file at all, since the write only happens on the annotate dialog's Save.

**Suggested fix:**
Have this Save write the design. It has everything `TmtExperimentalDesign.Write` needs except annotations, and `Write` already emits file-only placeholder rows for a plex that has none, so a part-authored design round-trips. Failing that, relabel the button so it doesn't read as persistence and make Annotate Plex explicitly required. Separately, move the `:539-547` seeding block inside the `ShowDialog()` guard and drop the duplicate.

**Affected test scenarios:**
Save a design, reopen the app, confirm it survives. Save a design, open the spectra picker and cancel, confirm the edits are still there.

MetaMorpheus/GUI/Views/AnnotatePlexWindow.xaml.cs
```diff
@@ -0,0 +1,324 @@
...
+            _plexFileMap = plexFileMap ?? new();
+
+            PlexComboBox.ItemsSource = _plexNames;
+            TagTypeComboBox.ItemsSource = Enum.GetValues(typeof(IsobaricMassTagType)).Cast<IsobaricMassTagType>();
+        }
+
+        private void PlexComboBox_SelectionChanged(object sender, SelectionChangedEventArgs e) => RegenerateRows();
+        private void TagTypeComboBox_SelectionChanged(object sender, SelectionChangedEventArgs e) => RegenerateRows();
+
+        private void RegenerateRows()
```
pcruzparri, line 70: **High** — Changing either dropdown discards everything typed, and carries stale channel labels forward

**Problem:**
Both combos route here — `PlexComboBox_SelectionChanged` (`:47`) and `TagTypeComboBox_SelectionChanged` (`:48`) — and `RegenerateRows` unconditionally reassigns `_currentRows`, either deep-copying from `_existingAnnotations` (`:65-73`) or building blanks (`:77-88`).

`_existingAnnotations` has exactly four references in the tree: the declaration (`:16`), the constructor assignment (`:40`), and two reads (`:62`, `:278`). Nothing writes the in-progress grid back into it, and `AnnotatePlexWindow.xaml` has no `CellEditEnding`, no grid `SelectionChanged` and no `Closing` handler — only the paste bindings — so there's no other path either.

Neither combo is pre-selected (`:43-44` set `ItemsSource` with no `SelectedIndex`), so the user goes through the dropdown at least once, and Save closes the window (`:260`) — so the dropdown is the obvious way to try to annotate a second plex in one pass.

The reuse branch compounds it: it takes `Tag` from the stored annotation rather than from `reporterLabels`, and only requires the counts to match. iTRAQ4 is `114,115,116,117` and diLeu4 is `115,116,117,118` (`IsobaricMassTag.cs:196,204`) — same count. So switching an iTRAQ4-annotated plex to diLeu4 keeps the iTRAQ labels.

**Impact:**
Annotate Plex 1, switch to Plex 2, and Plex 1's names, conditions and sample types are gone with no prompt. Save makes it worse: `BuildWholeDesign` (`:276-278`) takes the current plex from `_currentRows` and every other plex from `_existingAnnotations`, so Plex 1 is written with no annotations — `Write` emits it as the file-only placeholder (`TmtExperimentalDesign.cs:402-405`) — while the confirmation box counts distinct plexes across *files* (`:257-258`) and reports "2 plex(es)". If Plex 1 was seeded from a design file, the seeded values survive and it's the user's new typing that vanishes, which looks even more like it saved.

The tag-label case at least fails loudly: `ToMzLibDesign` rejects a tag that isn't a channel of the selected type (`TmtExperimentalDesign.cs:510-514`), so that design won't load.

**Suggested fix:**
Flush the grid into `_existingAnnotations` before regenerating — capture the outgoing plex in `RegenerateRows`, or use `SelectionChanged`'s `e.RemovedItems`, and store `_currentRows` under it; `BuildWholeDesign` picks it up unchanged. In the reuse branch take `Tag` from `reporterLabels[i]` and carry over only the user-entered fields, and warn before discarding annotations when the counts differ.

**Affected test scenarios:**
Two plexes: annotate the first, switch the dropdown, save, read back with `TmtExperimentalDesign.Read`, assert the first plex kept its annotations. Seed a plex as iTRAQ4, select diLeu4, save, assert the written tags are the diLeu labels.

MetaMorpheus/GUI/Views/TmtExperimentalDesignWindow.xaml.cs
```diff
@@ -0,0 +1,513 @@
...
+            CommitPendingEdits();
+
+            var files = (string[])e.Data.GetData(DataFormats.FileDrop);
+            if (files == null) return;
+
+            foreach (var path in files)
+            {
+                if (Directory.Exists(path))
+                {
+                    foreach (var f in Directory.EnumerateFiles(path))
```
pcruzparri: **High** — Dropping a folder puts every file in it into the run, past the extension allowlist and the Thermo licence gate

**Problem:**
`Window_Drop` accepts any dropped path and enumerates a directory with no filter — every file in it becomes a grid row. On Save, `MainWindow` puts them straight into the run:

```csharp
var newRaw = new RawDataForDataGrid(design.FilePath) { Use = true };
SpectraFiles.Add(newRaw);          // MainWindow.xaml.cs:627-632
```

`RawDataForDataGrid`'s constructor validates nothing — `FileName`, `FilePath`, `Use = true` (`ForDisplayingInDataGrids/RawDataForDataGrid.cs:9-14`). Every other route into `SpectraFiles` goes through `MainWindow.AddPreRunFile` (`:1804`), a strict `switch (theExtension)` allowlist that also gates `.raw` behind the Thermo RawFileReader licence window (`:1843-1860`). This path skips both.

The dedupe is inconsistent too: `MainWindow.xaml.cs:621-622` matches on `FilePath` while `SpectraFileExists` (`:1742-1750`) matches on `FileName`, so this path can add two files with the same name from different folders — which `ToMzLibDesign` then rejects outright, since quantification keys channels by file name (`TmtExperimentalDesign.cs:500-505`).

**Impact:**
Dropping the data folder — the natural gesture, and the window advertises it with `AllowDrop="True"` — adds `TmtDesign.txt`, any `.toml`, `.fasta`, README, anything, to the spectra list marked `Use = true`. A `.raw` added this way skips the licence prompt every other entry point enforces.

**Suggested fix:**
Don't duplicate the allowlist — have Save hand the paths to `AddPreRunFile` rather than constructing `RawDataForDataGrid` directly, which fixes the licence gate and the dedupe mismatch at the same time. If the drop needs filtering earlier, filter `Directory.EnumerateFiles` to the extensions `AddPreRunFile` accepts.

**Affected test scenarios:**
Drop a folder holding one `.mzML` and one `.txt`; only the `.mzML` should arrive. Confirm a `.raw` added this way still hits the licence gate.

MetaMorpheus/GUI/Views/AnnotatePlexWindow.xaml.cs
```diff
@@ -0,0 +1,324 @@
...
+                    startRow = _currentRows.IndexOf(pa);
+                    if (startRow < 0) startRow = 0;
+                }
+            }
+
+            if (targetColumn < 0)
+                targetColumn = 1; // Sample Name column (Tag=0, Sample=1, Condition=2, BioRep=3)
+
+            var lines = raw
+                .Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries)
```
pcruzparri: **Medium** — Pasting into the annotation grid silently misaligns channels

**Problem:**
`Paste_Executed` splits pasted lines with `RemoveEmptyEntries` (`:142`), so a blank row inside a pasted range is dropped and every row below it moves up one channel. It splits cells on tab *and* comma, also with `RemoveEmptyEntries` (`:154`), so a leading empty cell shifts columns left and a sample name containing a comma — "HeLa, 2h" — splits into name and condition.

Three smaller things in the same method: `case 3:` accepts a Biological Replicate of `0` (`:172`) that both `ValidateCurrent` (`:208`) and the parser (`TmtExperimentalDesign.cs:213`) reject, and a non-numeric paste is dropped with no feedback; the column-mapping comment (`:139`) omits Sample Type, so pasting with the Tag or Sample Type cell current falls into `default:` and overwrites Sample Name; and `if (AnnotationGrid.CurrentCell != null)` (`:128`) is always true, since `DataGridCellInfo` is a struct — that's the CS8073 this file builds with.

**Impact:**
An unused channel mid-plex is an ordinary TMT design, so a pasted block with a gap in it is a normal thing to do — and it assigns every sample below the gap to the wrong channel, quietly, in a design that then looks entirely plausible. Comma-splitting corrupts any sample name containing a comma. All silent.

**Suggested fix:**
Drop `RemoveEmptyEntries` on both splits and pad short rows instead, so row *i* of the paste lands on row *i* of the grid. Split cells on tab only — the grid is populated from Excel, where tab is the column separator and a comma is data. Reject a non-numeric or `< 1` Biological Replicate with the same message `ValidateCurrent` uses rather than dropping it.

trishorts:
All four findings confirmed against the code and fixed, and your framing was right — three of them were the one thing, so they are fixed as one shape rather than three patches. Two places where I did something other than what you suggested, both with reasons, and two things you flagged as unsettleable by reading that turned out to be settleable.

Pushed as `795773d71`.

## The shape: one writer, and the file is the source of truth

`ExperimentalDesignWindow` has kept the invariant your comment is really about, all along, and it does it with three properties that hold together: the grid's row set is derived from `SpectraFiles` and is not editable in the dialog (`CanUserAddRows/CanUserDeleteRows="False"`), there is **no static state** — every open re-derives from `SpectraFiles` plus disk — and the whole collection is written **once, on Save, only after it validates** (`BtnSaveQuant_Click`, `ExperimentalDesignWindow.xaml.cs:63`, which `return`s and leaves the window open on a validation failure). This window broke all three at once, which is why the defects read as one.

So: the design window is now the only writer. `AnnotatePlexWindow` hands its annotations back and writes nothing; `TmtExperimentalDesignWindow.SaveButton_Click` builds the whole design and calls `TmtExperimentalDesign.Write`.

## Where I did not take the suggested fix

**The folder drop.** You suggested handing the paths to `AddPreRunFile` rather than constructing `RawDataForDataGrid`. That isn't reachable: `AddPreRunFile` is `private` (`MainWindow.xaml.cs:1718`) and so is `SpectraFiles` (`:35`) — `private` is type-scoped, so no other window can call either. More to the point, **both** existing dialogs that receive `SpectraFiles` — `ExperimentalDesignWindow` and `FileSpecificParametersWindow` — annotate the file set and never change it; the only bypass on master is `AddNewSpectraFileFromCalibration`, and that is inside `MainWindow`.

So the window no longer adds files at all. `AllowDrop`, `Window_Drop` and `AddFileIfNotExists` are gone, and `SetTmtExperimentalDesign_Click` reduces to exactly what `SetExperimentalDesign_Click` already is. That closes the licence gate and the allowlist in one move, and it also closes the `FilePath`/`FileName` dedupe mismatch you noted — which mattered more than it looks, since `SpectraFileExists` (`:1671`) matches on bare `FileName` and multi-plex runs routinely have identically-named fraction files in per-plex folders.

**Comma-splitting on paste.** I took your fix, and it is a deliberate deviation from the sibling: `ExperimentalDesignWindow.Paste` (`:194`) splits cells on `{'\t', ','}` too. But its editable columns are Condition/Biorep/Fraction/Techrep — short tokens — while this grid collects **Sample Name** and **Condition** as free text, where `HeLa, 2h` is ordinary. Tab only here. Worth knowing the two now differ on purpose.

On the row split I went further than the sibling for the same reason you gave: rows here **are** reporter channels and map positionally onto `ISpectralMatch.Intensities`, so a blank row in a pasted range has to stay a blank row. The sibling drops blank lines because its rows are files and a blank line means nothing there.

## The two you could not settle by reading

Both are real, and both are fixed.

**`AnnotatePlexWindow` did lose an in-progress cell edit on Save.** `AnnotationGrid` has no `CellEditEnding` and `SaveButton_Click` read `_currentRows` with no commit, while the design window calls `CommitPendingEdits()` before every grid read. Type a sample name, click Save without tabbing out, and that cell was not in the file. It now commits before every read, as the sibling does.

**`e.Cancel` did not revert.** With `UpdateSourceTrigger=PropertyChanged` the source is already written by the time `DgTmt_CellEditEnding` runs, so the "Fraction must be integer >= 1" refusal rejected nothing. The two integer columns now bind with the default trigger, so the cancel is real.

## The tag-type carry-over

Confirmed exactly, and it is narrower than it looks, which makes it pinnable. Channel counts across the eight tag types are **6 / 10 / 11 / 18 / 4 / 8 / 4 / 12** — **4 is the only count that occurs twice**. So iTRAQ4 and diLeu4 are the entire hazard. `RegenerateRows` now keys the carry-over on the tag **type**, takes `Tag` from `reporterLabels` always, carries over only user-entered fields, and warns before clearing annotations when the counts differ.

## Smaller items, all taken

Unplexed files are refused rather than silently dropped from the written file. Plex names compare `OrdinalIgnoreCase` in all four places. `s_seedErrors` is cleared once shown. Seeding runs from the **checked** files, and the duplicate call that sat outside the picker's `ShowDialog` guard is gone. The three `catch { /* ignore */ }` blocks are gone — rather than appending in the catch, `SeedFromDesignFiles` now reports a throw into the same list it already reports parse errors into, so the callers need no catch. The `Reorder rows by drag to renumber fractions` sentence is deleted along with the handlers, since nothing implemented it and no consumer depends on row order.

Build warnings gone: CS0105, CS0169 ×2, CS8632 ×2 (with `PlexFileEntry`, now unused), CS8073. All churn reverted — BOM, stray space, the narrowed column, trailing blanks, the scaffolding comment, the `// NEW:` markers.

**The PR is now +1248 / −0 against master**, and `MainWindow.xaml.cs` is `29+/0-` rather than `91+/2-`.

## One thing you gestured at that I closed

Both design writers target the first spectra file's directory, and a run refuses to start with both files present. With the two authoring buttons side by side, that stranded the user. Save now warns when `ExperimentalDesign.tsv` is already in the folder, rather than letting the run be where it is discovered.

And, because this is authoring UI for a file the command line consumes: after writing, the window **reads the design straight back through `TmtExperimentalDesign.Read`** — the same parser `CMD/Program.cs` uses — and reports anything it says. A design this window calls saved is one a run can load. That was the gap behind both of the defects you found in the earlier round, where the Save dialog reported success either way.

## Tests

`Test.csproj` does not reference `GUI.csproj`, so these handlers cannot be driven from a test, and I want to be explicit that the WPF behaviour above is verified by reading and building, not by running. `GUI.csproj` builds with 0 errors and none of the remaining warnings are in these files; 178 tests pass across the TMT design, sample-type, multiplex, mzLib-projection and `ResolveExperimentalDesign` fixtures. `TmtDesignGuiContractTests` pins what is reachable: that iTRAQ4 and diLeu4 collide on count but not labels, that 4 is the only shared count (so it fails if a tag type is ever added that collides), that a stale label is rejected by `ToMzLibDesign` rather than accepted, and that a whole two-plex design built the way the window builds it round-trips through `Read`, projects through `ToMzLibDesign`, and is accepted by `Program.ResolveExperimentalDesign`.

## Deliberately not here

The GUI run gate not knowing about `TmtDesign.txt`, and nothing in `TaskLayer` reading it — you are right on master, and that wiring **is** #2755, which carries `PostSearchAnalysisTask.cs` +152. Adding it here would re-merge the split. Worth knowing that #2755 carries all six of this PR's GUI blobs byte-identically; these fixes are going onto that branch too, so its merge cannot resurrect them.

CMD's refusal of two design files is policy from merged #2775, not a defect in this diff, so I have warned rather than changed it.

Last one, for a separate PR rather than this one: `ValidateDesign` re-types the contiguity-and-duplicate walk that `EngineLayer.ExperimentalDesign.GetErrorsInExperimentalDesign` (`ExperimentalDesign.cs:116`) already implements, minus the channel dimension. Factoring those together means editing merged EngineLayer code from a GUI PR, which I would rather do on its own terms.

pcruzparri (approved):

Alexander-Sol (approved):

Issue #2777, opened 2026-08-29
ModernSearch and NonSpecificEnzymeSearch score 79.43% on mutation testing: 131 unpinned behaviours, listed


trishorts:
Mutation testing over the two index-based search engines. Follows the pattern of mzLib #1236, which did the same for Deconvolution: the score is not the point, the enumerated list is.

Measured on master `d93e20b4` with `dotnet-stryker` 4.16.0, scoped with `-m "**/ModernSearchEngine.cs" -m "**/NonSpecificEnzymeSearchEngine.cs"`. 1h16m.

```
549 mutants tested
  Killed    479
  Survived   43
  Timeout    27
  NoCoverage 88     (not reached by any test in the suite)

final mutation score: 79.43 %
```

The score counts `NoCoverage` against, which is right — `(479 + 27) / (479 + 27 + 43 + 88)`. **131 behaviours are unpinned**: 43 that the tests execute and do not detect, and 88 they never reach.

The two engines are not in the same condition:

| | Survived | NoCoverage | Timeout |
|---|---|---|---|
| `NonSpecificEnzymeSearchEngine.cs` | **36** | 29 | 20 |
| `ModernSearchEngine.cs` | 7 | 59 | 7 |

Different problems, different fixes. `NonSpecificEnzymeSearchEngine` is *exercised* and not *asserted*. `ModernSearchEngine` is largely fine where it is tested, and has whole regions nothing enters.

## The survivors that would be real bugs

Not a full list — the ones where the mutant is a defect a user would feel, rather than an equivalent mutant or a boundary nobody hits.

**Comparisons that decide which match wins, inverted with nothing noticing.** `NonSpecificEnzymeSearchEngine.cs`:

```csharp
line 512:  if (majorPsm.Score > minorPsm.Score)              ->  <   , >=  , !(...)     all survive
line 530:  if (majorPsm.PsmFdrInfo.QValue > minorPsm...)     ->  <   , >=  , !(...)     all survive
line 504:  ...OrderByDescending(x => x.Score)                ->  OrderBy(...)           survives
```

Three mutants on line 512 and three on 530 all survive, and the ranking on 504 can be reversed. Whatever these are choosing between, no test distinguishes the better answer from the worse one.

**The scoring table can be decremented, or not written at all.**

```csharp
line 189:  scoringTable[bin[pep]]++;   ->  scoringTable[bin[pep]]--   survives
                                       ->  ;   (statement removed)    survives
```

[...]

PR #2778, opened 2026-08-29
Pin how NonSpecificEnzymeSearch picks between competing identifications

trishorts:
> **Project of origin:** none — standalone (Stryker mutation testing, issue #2777)

First tranche against #2777.

`NonSpecificEnzymeSearchEngine.ResolveFdrCategorySpecificPsms` is `public static` and had **no tests at all**, which is why mutation testing found the cluster it did. The comparisons that decide which spectral match survives can be inverted without the suite noticing — and a method that chooses between candidate identifications but cannot tell the better from the worse produces a *plausible answer* rather than an error, which is the worst shape a defect takes in this codebase.

### Four tests

- the highest-scoring match wins a q-value tie;
- **losing candidates are cleared from their categories** — the FDR recalculation at the end of the method runs over what is left, so a candidate that stays behind contributes to an FDR it was not chosen for;
- a category that matched nothing is skipped rather than treated as a scoreless candidate;
- the returned list is ordered by q ascending, then score descending.

### The fixture scores are 5, 9, 3, and that is the point

Both ways this can be got wrong look correct against a sorted fixture. Taking the first candidate unconditionally yields 5; taking the last yields 3; only actually comparing yields 9. An ascending or descending fixture would have passed against two of the three mutants below.

### Verified by installing each mutant #2777 lists

| mutant | result | |
|---|---|---|
| `&&` → `\|\|` | answer becomes **3** — every later candidate wins | FAILS |
| `==` → `!=` | answer becomes **5** — no later candidate wins | FAILS |
| `Score >` → `<` | answer becomes **3** — comparison inverted | FAILS |
| block removal at 565 | **2** PSMs left in the categories instead of 1 | FAILS |

All four restored afterwards. 32 green across this fixture and `FdrTest`.

### Not the whole of #2777

Deliberately. Still open:

- the q-value lifting between categories (lines 504–530) — needs a fixture where `FdrAnalysisEngine` produces *different* q-values per category, which is a bigger build than this;
- the 59 uncovered lines in `ModernSearchEngine` — those need someone to work out whether the regions are dead or merely unreached, which is a reading exercise before it is a testing one.

Neither is a reason to hold this up: it kills the mutants it claims to and leaves the issue's list correspondingly shorter.


MetaMorpheus/Test/NonSpecificEnzymeFdrCategoryTests.cs
```diff
@@ -0,0 +1,179 @@
...
+            }
+
+            var best = Resolve(new[] { psms });
+
+            Assert.That(best, Is.Not.Empty);
+            for (int i = 1; i < best.Count; i++)
+            {
+                double previousQ = best[i - 1].PsmFdrInfo.QValue;
+                double currentQ = best[i].PsmFdrInfo.QValue;
+                Assert.That(previousQ, Is.LessThanOrEqualTo(currentQ), "q-values must ascend");
```
acesnik, line 208: This assertion cannot fail. The fixture protein is a single target with no decoy, so `FdrAnalysisEngine` gives all five PSMs a q-value of 0 — measured: `best.Count == 5`, **one distinct q-value** — so this runs four comparisons of `0 <= 0`. Mutating the final line of the method:

- `ThenByDescending` → `ThenBy` — **fails**, caught
- `OrderBy` → `OrderByDescending` — **passes**, survives

So the score half of the contract is pinned and the q half is not, while the test name claims both. Either add a decoy so the q-values actually differ, or drop the q-clause from the name and the doc comment. I would lean to the decoy, since the same fixture would let you reach the q-lifting block that is still entirely uncovered — and a name claiming a covered branch is the part that costs the next person time.

MetaMorpheus/Test/NonSpecificEnzymeFdrCategoryTests.cs
```diff
@@ -0,0 +1,179 @@
...
+            var protein = new Protein("MNKNNKNNNKNNNNKPEPTIDEKPEPTIDER", "target");
+            var peptides = protein
+                .Digest(CommonParams.DigestionParams, new List<Modification>(), new List<Modification>())
+                .Cast<PeptideWithSetModifications>().ToList();
+            var dataFile = new TestDataFile(peptides.Cast<IBioPolymerWithSetMods>().ToList());
+
+            // One category, several spectra, scores out of order so the ordering has work to do.
+            var psms = new List<SpectralMatch>();
+            double[] scores = { 4, 11, 7, 2, 9 };
+            for (int i = 0; i < scores.Length && i < peptides.Count; i++)
```
acesnik: `OneSpectrumAcrossCategories` asserts `peptides.Count >= scoresByCategory.Length` as a premise; this test guards the same thing with `&& i < peptides.Count` and then only asserts `Is.Not.Empty`. Today the fixture digests to 11 peptides for the 5 needed, so there is margin — but a change to the default `DigestionParams` walks it down silently, and at 1 the ordering loop runs zero times and the test passes having compared nothing. `Has.Count.EqualTo(5)` after the call costs nothing and closes it.

MetaMorpheus/Test/NonSpecificEnzymeFdrCategoryTests.cs
```diff
@@ -0,0 +1,179 @@
...
+                {
+                    allPsms[category] = new List<SpectralMatch> { null };
+                    continue;
+                }
+
+                var peptide = peptides[category];
+                var scan = new Ms2ScanWithSpecificMass(
+                    dataFile.GetOneBasedScan(2), peptide.MonoisotopicMass.ToMz(1), 1, null, CommonParams);
+
+                // Scan number 1 for every category: these are competing identifications of ONE spectrum,
```
acesnik: Small: the PSMs' `ScanNumber` is 2, not 1 — `dataFile.GetOneBasedScan(2)` is the MS2 scan. The intent (one spectrum) holds and the assertions are unaffected; it is the number that is off.

MetaMorpheus/Test/NonSpecificEnzymeFdrCategoryTests.cs
```diff
@@ -0,0 +1,179 @@
...
+        [Test]
+        public static void ResultIsOrderedByQValueThenDescendingScore()
+        {
+            var protein = new Protein("MNKNNKNNNKNNNNKPEPTIDEKPEPTIDER", "target");
+            var peptides = protein
+                .Digest(CommonParams.DigestionParams, new List<Modification>(), new List<Modification>())
+                .Cast<PeptideWithSetModifications>().ToList();
+            var dataFile = new TestDataFile(peptides.Cast<IBioPolymerWithSetMods>().ToList());
+
+            // One category, several spectra, scores out of order so the ordering has work to do.
```
acesnik: Also small, but worth being accurate about because it could mislead a later edit: all five PSMs carry `ScanNumber` 2, so this is one spectrum with five candidate rows, not several spectra. They are distinguished only by peptide mass — and that mass difference is load-bearing, since it is what stops the `GroupBy((FullFilePath, ScanNumber, mass))` dedup at line 455 collapsing them into one. Someone "tidying" this to use distinct scan numbers would quietly change what the test exercises. The doc comment at 137 says the same thing.

acesnik (commented):
🤖 `ResolveFdrCategorySpecificPsms` is `public static`, it decides which identification survives when a non-specific search has produced one per FDR category for the same spectrum, and it had nothing asserting any of it. That is a genuinely uncomfortable place to have no coverage — a method that chooses between candidate identifications and cannot tell the better from the worse fails by producing a plausible answer, which is the failure mode you never find by reading output. Closing it is worth doing regardless of what the mutation score says.

I reproduced the results rather than taking them on trust: built `EngineLayer`/`TaskLayer` on macOS with a throwaway net10.0 harness compiling only this file plus `TestDataFile`, and hand-installed 22 mutants. All four of your claims hold, each mutant dying to exactly the test you name — 553's `&&`→`||`, `==`→`!=` and `Score >`→`<` to `HighestScoringMatchWinsWhenQValuesAreTied`, and both null-clearing removals at 558 and 566 to `LosingCandidatesAreClearedFromTheirCategories`. The 5/9/3 fixture does what you describe, with the first and last mutants both landing on the wrong answer.

What I did not expect is that **every one of the four tests earns at least one kill no other makes**, including `EmptyCategoriesAreSkipped`, which is the only thing catching `488 ranking >` → `<`. That is unusual in a batch of characterization tests — the common outcome is three tests that all die to the same mutant — and it suggests the fixtures were chosen rather than accumulated. **Test-only, and I confirmed it: zero production lines changed, so no produced output moves.**

I have left four notes on lines. The first is the one I would fix before merge; the rest are cheap.

**One framing fix for the description.** You write that these are the mutants #2777 lists for 553 and 565, and that the issue's list gets correspondingly shorter. #2777 actually lists `553-556` in its **NoCoverage** block, not among the survivors — its survivor cluster in this file is 504/512/530, and I measured all three still surviving here, consistent with your own "still open" section. So what shrinks is the NoCoverage list, not the survivor table. Since the survivor table is what a reader will check against, that is worth a sentence.

**One concrete defect in the block this leaves uncovered**, which I think argues for covering it better than a mutant count does. In the q-lifting wrap-up loop, `majorCategoryPsms[majorPsmIndex - 1]` is unguarded, and `majorPsmIndex` is still 0 whenever the preceding loop never iterated — which happens when `majorCategoryPsms` is empty. That is reachable in principle: `indexesOfInterest` admits any category whose list object is non-null, including one that is all nulls, and `majorCategoryIndex` falls back to `indexesOfInterest[0]` when every `ranking` is 0. So a search in which nothing clears 1% FDR and the first FDR category is entirely empty while a later one is not would index `[-1]`.

I could not reach it through your `OneSpectrumAcrossCategories` helper, and the reason is the same one behind my first line note: `FdrAnalysisEngine` gives every target a q of 0 in a target-only fixture, so `ranking` is at least 1 for any non-empty category and the major category is never the empty one. Constructing it needs a decoy-bearing fixture — the same fixture the q-half of test 4 needs, which is why I would build one rather than treat these as separate errands. Derived from the code, not observed, so I would not report it as a live bug.

For whoever takes the next tranche: **deleting the entire final `FdrAnalysisEngine` recalculation loop at 586 leaves all four tests green**, and that loop is the stated reason `LosingCandidatesAreClearedFromTheirCategories` matters. The reasoning is right; the assertion reaches the clearing but not its consequence. Beyond that, `455 First()`→`Last()`, `478 <=`→`<`, `488 >`→`>=`, `539 indexesOfInterest[0]`→last, and removing the `if (bestPsm != null)` guard at 570 all survive. Net over my 22: 11 killed, 11 survived — the winner-selection loop at 539-573 is now well pinned, the q-lifting block at 495-537 is not.

Two checks that came back clean, since they were my main worries. The pinned winner is genuinely determined rather than observed once: nothing in the method is parallel or dictionary-ordered, `indexesOfInterest` is built and iterated in ascending order, and no lifting fires at equal q. Twelve repeat runs, identical. And there is no interaction with #2761 — its hunks are all in lines 5-304 while this method spans 441-590; I merged both onto master, clean, 4/4 green. Its counting-sort stability question cannot reach these tests anyway, since they construct `SpectralMatch` objects by hand and never invoke the search engine.

I could not run `FdrTest` or the rest of `Test.csproj` — `net10.0-windows`, and the host needs `Microsoft.WindowsDesktop.App` — so the "32 green" figure is CI's to confirm. I also did not re-run Stryker, only re-derived the line mapping and installed mutants by hand.

MetaMorpheus/Test/NonSpecificEnzymeFdrCategoryTests.cs
trishorts: Fixed in `11bc1f67b` — added the decoy. You measured it exactly right, and this is the worst kind of test to have shipped: a name claiming a covered branch, over an assertion that cannot fail.

I took the decoy option for the reason you gave. `ResultIsOrderedByQValueThenDescendingScore` now also asserts the premise outright — more than one distinct q-value must come back — so if a future fixture change flattens them again the test says so instead of passing quietly.

Verified your mutant now dies:

```
OrderBy -> OrderByDescending
  q-values must ascend
  Expected: less than or equal to 0.75d
  But was:  1.0d
```

**Two traps in the fix itself, worth recording because both passed a first glance.** Concatenating the two digests and slicing the front takes only *targets* — the target protein alone yields more peptides than the slice — so the first attempt still had no decoys and one q-value, the exact condition it was written to remove. And a reversed-sequence decoy has peptides of **identical mass**, which the method dedups on `(file, scan, mass)`, so two candidates collapsed and 7 of 8 came back. The decoy is compositionally different now, and both are taken half and half explicitly with the counts asserted.

On the q-lifting block: still uncovered, and you are right that this fixture is the way in. Left for a follow-up rather than grown into this PR.

MetaMorpheus/Test/NonSpecificEnzymeFdrCategoryTests.cs
trishorts: Added in `11bc1f67b` — `Has.Count.EqualTo(wanted)` after the call, plus an assertion on the built list before it.

Your reasoning is the part I want to keep: the guard and the premise assertion were saying different things, and the weaker one silently won. At one peptide the loop compares nothing and the test passes having asserted nothing — which is the same failure as the thread above, reached by a different route.

The decoy work made this sharper rather than redundant: the count is now fixed by two explicit `Take(4)` calls with both asserted, so a change to the default `DigestionParams` fails at the premise instead of walking the coverage down.

MetaMorpheus/Test/NonSpecificEnzymeFdrCategoryTests.cs
trishorts: Fixed in `11bc1f67b`. You are right — `GetOneBasedScan(2)` is the MS2 scan, and the PSMs carry 2.

Corrected the comment rather than the number, since the intent (one spectrum) was what mattered and 2 is the value that makes it true.

MetaMorpheus/Test/NonSpecificEnzymeFdrCategoryTests.cs
trishorts: Fixed in `11bc1f67b`, and thank you for pushing on it — "small but could mislead a later edit" is exactly right, and the misleading direction is dangerous.

The comment now says these are candidate rows for one spectrum rather than several spectra, and states why the mass difference is load-bearing: it is what stops the `GroupBy((FullFilePath, ScanNumber, mass))` dedup collapsing them. Someone tidying this to distinct scan numbers would quietly change what is exercised.

That was not hypothetical — while adding the decoy I hit exactly that dedup from the other side. A reversed-sequence decoy has peptides of identical mass, two candidates collapsed, and 7 of 8 came back. The comment you asked for is the one that would have told me why.

trishorts:
@acesnik — all four threads addressed and resolved in `11bc1f67b`.

Your first finding was the one that mattered: **the q-value half of the ordering assertion could not fail.** Single target, no decoy, every candidate at q = 0, four comparisons of `0 <= 0` — so `OrderBy` → `OrderByDescending` survived while the test name claimed both halves of the contract. Worse than not having the test, since it tells the next person a branch is covered.

Added the decoy, and asserted the premise outright — more than one distinct q-value must come back — so a future fixture change flattens the test loudly rather than quietly. Your mutant now dies:

```
OrderBy -> OrderByDescending
  q-values must ascend
  Expected: less than or equal to 0.75d
  But was:  1.0d
```

Also took the definite count assertion, corrected the `ScanNumber` to 2, and rewrote the comment to say these are candidate rows for one spectrum rather than several — including *why* the mass difference is load-bearing.

**Two traps inside the fix, both of which passed a first glance:**

- Concatenating the target and decoy digests and slicing the front takes only **targets** — the target alone yields more peptides than the slice — so the first attempt still had no decoys and one q-value, the exact condition it was written to remove.
- A reversed-sequence decoy has peptides of **identical mass**, which the method dedups on `(file, scan, mass)`, so two candidates collapsed and 7 of 8 came back. That is the dedup from your fourth thread, hitting me from the other side while I was fixing it.

Both are now explicit — half and half with counts asserted, and a compositionally different decoy.

The q-lifting block is still uncovered. You are right that this fixture is the way in; leaving it for a follow-up rather than growing this PR.

Ready for another look.

acesnik (approved):
🤖 Approving. All four notes are addressed in `11bc1f67`, and the ordering fix is better than what I asked for.

I asked for a decoy so the q-values would actually differ. You added the decoy *and* an assertion that they do — `best.Select(b => b.PsmFdrInfo.QValue).Distinct().Count(), Is.GreaterThan(1)` at line 201, with the reason on the assertion. That is the difference between a test that happens to be meaningful today and one that says so when it stops being: a future fixture change that re-flattens the q-values now fails at the premise instead of quietly going back to comparing `0 <= 0` four times. The same pattern shows up in the count guard, where `Take(4)` on each side with both counts asserted replaces a margin that could walk down silently.

The two traps you hit while fixing it are the most useful thing in the thread. A concatenate-then-slice taking only targets reproduces the exact condition the fix was written to remove, and a reversed-sequence decoy has peptides of identical mass, which the `(file, scan, mass)` dedup then collapses. That second one is the comment I pushed for at line 149 arriving from the other direction — you hit the dedup from the inside, which is a better argument for documenting it than mine was.

**One thing still open, and it is the description rather than the code.** I checked this against the posted #2777 rather than my memory of it: `553-556` appears exactly once in that issue, in the **NoCoverage** block, and `565` does not appear anywhere in it at all. The survivor cluster for this file is 504, 512 and 530 — and those are precisely what your own "Still open" section leaves for later. So "verified by installing each mutant #2777 lists" and "leaves the issue's list correspondingly shorter" point a reader at the survivor table, which this PR does not change; what it shortens is the NoCoverage count. Worth saying plainly, because a reader who checks the survivor table will find it untouched and conclude the PR did not do what it says, when it did something different and still worth doing — covering lines nothing reached. Line numbers have also shifted since the Stryker run, so your 565 and the issue's 553-556 may well be the same code; that only reinforces which list is shrinking.

Still unguarded on master, and I would leave it with the q-lifting follow-up rather than grow this PR: `NonSpecificEnzymeSearchEngine.cs:528` is `majorCategoryPsms[majorPsmIndex - 1]` with `majorPsmIndex` still 0 whenever the preceding loop never iterated. Derived from the code, not observed — I could not reach it through the helper, for the same reason the old ordering test could not fail — so not a bug report, just the thing the fixture you have now built is the way into.

Test-only and confirmed as such: zero production lines changed, so nothing about produced output moves. `Test/Test.csproj` is `net10.0-windows` and cannot run here, so the 32-green figure is still CI's.

nbollis (approved):

PR #2779, opened 2026-08-30
MetaDraw: draw C-terminal and last-on-screen modifications in the stationary sequence

### LINKED ISSUE #2330: Metadraw crashes with PSMs that contains the fixed modification "Arginine amide on R, prot C-termini" Uniprot
When I run Child scan view in Metadraw with spectra files and corresponding PSM files with the modification Arginine amide on R, prot C-termini, Uniprot, the program crashes. When I use the same PSM files from a search without this modification, it doesn't crash, and I can the fragmentation. Can you help me? Best regards

acesnik:
🤖 Closes #2330. MetaDraw's stationary sequence — the one drawn over the spectrum in the Child Scan View — never draws a C-terminal modification, and silently drops the modification on the last residue on screen. This is what is left of #2330 now that the sequence-annotation crash reported there is fixed (#2579, released in 1.1.6). Closing the issue on this rather than on #2579: the crash is gone, but the modification the reporter was trying to see is still not drawn, so their scenario only works end to end once this lands.

**This changes where modification circles are drawn, for all modification types, in the stationary sequence only.** Three changes, all in `DrawStationarySequence`:

- a mod on the **last residue on screen** is now drawn (it used to be dropped);
- a mod on the residue **immediately to the left of the window** is no longer drawn (it used to be inserted at index 0 and rendered as an N-terminal circle at x = −12, i.e. attributed to a residue that is off screen);
- a **C-terminal** mod is now drawn one slot past the last residue when the window reaches the end of the peptide (it used to be drawn nowhere).

N-terminal mods and interior mods that are not at either edge of the window are unaffected. The scrollable sequence, the sequence-annotation view and the PTM legend are untouched.

## Cause

`AllModsOneIsNterminus` keys the N-terminus at 1, zero-based residue `r` at `r + 2`, and the C-terminus at `BaseSeq.Length + 2`. The filter used `p.Key - 1` against the on-screen residue window, which is the residue-mod convention off by one, so the window was shifted one residue left at both edges. `BaseSeq.Length + 2` fell outside it for every window, because the window can never extend past the last residue.

A C-terminal mod also cannot be `Insert`ed into the trimmed base sequence — its index is one past the end of the string, which is precisely how the #2330 crash happened in the segment loop of `DrawSequenceAnnotation`. It is appended as `-[type:id]` instead, the same representation `DrawSequenceAnnotation` uses and `IBioPolymerWithSetMods.GetModificationDictionaryFromFullSequence` already understands.

## Tests

`MetaDraw_StationarySequenceDrawsTerminalAndLastResidueMods` drives `DrawnSequence` at four scroll positions over `TestData/SequenceCoverageTestPSM.psmtsv` and counts the drawn circles. The PSM carries an N-terminal acetylation and an oxidation on residue 6; the test appends `Serine amide on S`, the protein-C-terminal sibling of the `Arginine amide on R` from the issue, to its C-terminal S. Three of the four expected counts differ from current behaviour.

`MetaDraw_TestStationarySequencePositioning` mirrors the production filter to compute its expected mod count, so its copy is updated to match; the two filters differ at `FirstAAonScreenIndex` 17 and 37 on that test's data.

## Not verified

The rendered result was not inspected — the WPF views cannot be run on macOS and `Test.csproj` cannot execute there, so the new test compiles but has never been executed and I did not confirm it fails against unfixed code. Its expected counts were derived from a headless harness that ran the old and new trimming side by side against real modification data and mzLib 1.0.587; three of the four differ from the old behaviour, so it should fail, but that is inference. CI is the first real run. The x-coordinates the circles land on were derived from parsed mod keys, not measured from a rendered canvas.

### DIFF
--- a/MetaMorpheus/GuiFunctions/MetaDraw/DrawnSequence.cs
+++ b/MetaMorpheus/GuiFunctions/MetaDraw/DrawnSequence.cs
@@ -263,16 +263,32 @@ namespace GuiFunctions
         {
             ClearCanvas(stationarySequence.SequenceDrawingCanvas);
             IBioPolymerWithSetMods peptide = sm.ToBioPolymerWithSetMods();
-            string baseSequence = sm.BaseSeq.Substring(MetaDrawSettings.FirstAAonScreenIndex, MetaDrawSettings.NumberOfAAOnScreen);
+            int firstResidueOnScreen = MetaDrawSettings.FirstAAonScreenIndex;
+            int residuesOnScreen = MetaDrawSettings.NumberOfAAOnScreen;
+            string baseSequence = sm.BaseSeq.Substring(firstResidueOnScreen, residuesOnScreen);
             string fullSequence = baseSequence;
 
+            // AllModsOneIsNterminus keys: 1 is the N-terminus, zero-based residue r is r + 2, and the C-terminus is BaseSeq.Length + 2
+            var cTerminalMod = peptide.AllModsOneIsNterminus
+                .Where(p => p.Key == sm.BaseSeq.Length + 2)
+                .Select(p => p.Value)
+                .FirstOrDefault();
+
             // Trim full sequences selectively based upon what is show in scrollable sequence
-            var modDictionary = peptide.AllModsOneIsNterminus.Where(p => p.Key - 1 >= MetaDrawSettings.FirstAAonScreenIndex 
-            && p.Key - 1 < (MetaDrawSettings.FirstAAonScreenIndex + MetaDrawSettings.NumberOfAAOnScreen)).OrderByDescending(p => p.Key);
+            var modDictionary = peptide.AllModsOneIsNterminus.Where(p => p.Key == 1
+                    ? firstResidueOnScreen == 0
+                    : p.Key - 2 >= firstResidueOnScreen && p.Key - 2 < firstResidueOnScreen + residuesOnScreen)
+                .OrderByDescending(p => p.Key);
             foreach (var mod in modDictionary)
             {
                 // if modification is within the visible region
-                fullSequence = fullSequence.Insert(mod.Key - 1 - MetaDrawSettings.FirstAAonScreenIndex, "[" + mod.Value.ModificationType + ":" + mod.Value.IdWithMotif + "]");
+                fullSequence = fullSequence.Insert(mod.Key - 1 - firstResidueOnScreen, "[" + mod.Value.ModificationType + ":" + mod.Value.IdWithMotif + "]");
+            }
+
+            // a C-terminal mod sits past the last residue, so it is appended rather than inserted
+            if (cTerminalMod != null && residuesOnScreen > 0 && firstResidueOnScreen + residuesOnScreen >= sm.BaseSeq.Length)
+            {
+                fullSequence += "-[" + cTerminalMod.ModificationType + ":" + cTerminalMod.IdWithMotif + "]";
             }
 
             List<MatchedFragmentIon> matchedIons = sm.MatchedIons.Where(p => p.NeutralTheoreticalProduct.ResiduePosition > MetaDrawSettings.FirstAAonScreenIndex &&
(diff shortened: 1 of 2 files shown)


nbollis (approved):

MetaMorpheus/GuiFunctions/MetaDraw/DrawnSequence.cs
```diff
@@ -263,16 +263,32 @@ public static void DrawStationarySequence(SpectrumMatchFromTsv sm, DrawnSequence
...
+            // AllModsOneIsNterminus keys: 1 is the N-terminus, zero-based residue r is r + 2, and the C-terminus is BaseSeq.Length + 2
+            var cTerminalMod = peptide.AllModsOneIsNterminus
+                .Where(p => p.Key == sm.BaseSeq.Length + 2)
+                .Select(p => p.Value)
+                .FirstOrDefault();
+
             // Trim full sequences selectively based upon what is show in scrollable sequence
-            var modDictionary = peptide.AllModsOneIsNterminus.Where(p => p.Key - 1 >= MetaDrawSettings.FirstAAonScreenIndex 
-            && p.Key - 1 < (MetaDrawSettings.FirstAAonScreenIndex + MetaDrawSettings.NumberOfAAOnScreen)).OrderByDescending(p => p.Key);
+            var modDictionary = peptide.AllModsOneIsNterminus.Where(p => p.Key == 1
```
trishorts, line 278: **The test now contains a second copy of this expression, which weakens what it can catch.**

`MetaDrawTest.MetaDraw_TestStationarySequencePositioning` reimplements this filter and the C-terminal append verbatim -- the same ternary on `p.Key == 1`, the same `p.Key - 2` bounds, the same `residuesOnScreen > 0 && first + residuesOnScreen >= BaseSeq.Length` guard -- and then compares the drawn output against what that copy produces.

This PR updates both copies identically, which is the only way to keep it green, but it means the test is now largely asserting that two expressions the same author wrote in the same commit agree with each other. If the convention is wrong (say the C-terminal on-screen condition should be `>` rather than `>=`), both copies are wrong together and the test still passes.

The new `MetaDraw_StationarySequenceDrawsTerminalAndLastResidueMods` test is much better precisely because it does not do this -- it states expected ellipse counts for four named scroll positions, which is an independent claim about behaviour. That one would catch a wrong convention; the mirrored one would not.

Not a merge blocker, and I would not ask you to rewrite `MetaDraw_TestStationarySequencePositioning` in this PR. But if the production expression is going to be duplicated in a test, it is worth extracting it to one internal helper that both call, so the duplication is visible rather than implied -- or trimming that test back to the parts that do not need a copy of the logic.

trishorts (commented):
The key-convention fix is correct. I worked through it rather than trusting the comment:

- `AllModsOneIsNterminus` keys are 1 for the N-terminus, `r + 2` for zero-based residue `r`, `BaseSeq.Length + 2` for the C-terminus. The old filter used `p.Key - 1` as the residue index, which is off by one; `p.Key - 2` is right.
- The insert position was **not** changed, and did not need to be: for residue index `r = Key - 2`, the mod belongs after that residue, i.e. at `r + 1 - first == Key - 1 - first`, which is what the existing expression already computed. Easy to misread as a leftover bug, so worth a word in the comment.
- Special-casing `Key == 1` to `firstResidueOnScreen == 0` is right, since an N-terminal mod is only on screen when the sequence is scrolled to the start.

The four-position test (whole peptide / scrolled to end / modified residue last on screen / scrolled one past) covers the boundaries that actually matter, and counting `Ellipse` children is a reasonable proxy for "was it drawn".

One structural concern, below.

trishorts (approved):
Checked the index arithmetic against the `AllModsOneIsNterminus` key convention (N-terminus 1, zero-based residue `r` at `r + 2`, C-terminus `BaseSeq.Length + 2`):

- the old filter compared `Key - 1` against a residue index, so it was off by one for every residue mod; `Key - 2 >= first && Key - 2 < first + count` is the correct window.
- the insert offset `Key - 1 - first` puts a residue mod immediately after its residue and the N-terminal mod at index 0, and the `Key == 1 ? first == 0` branch is what keeps that offset from going negative.
- descending `OrderByDescending(p => p.Key)` keeps earlier inserts from shifting later ones, and appending the C-terminal mod after the loop is correctly ordered.

My earlier note about `MetaDraw_TestStationarySequencePositioning` re-deriving the production expression still stands, but the new `MetaDraw_StationarySequenceDrawsTerminalAndLastResidueMods` asserts drawn `Ellipse` counts across four scroll positions, which is the independent oracle I was asking for. Non-blocking.

Approving.

PR #2780, opened 2026-08-30
Read the experimental design of the checked spectra files only, and stop deleting it

### LINKED ISSUE #2256: Error Reading Experimental Design File when Re-Running a Search

acesnik:
🤖 Fixes #2256, and removes a way to lose a file that the issue exposed but did not report.

## The reported bug

Re-running a Calibrate → GPTMD → Search set from the GUI raised "An experimental design file was found, but an error occurred reading it … The experimental design did not contain the file(s): …-calib.mzML", naming the calibrated files, even though the design MetaMorpheus had just written for them was correct.

`RunAllTasks_Click` located and validated `ExperimentalDesign.tsv` from the whole spectra-file list, ignoring the `Use` checkbox, while the run itself used `SpectraFiles.Where(b => b.Use)`. `AddNewSpectraFileFromCalibration` leaves the uncalibrated files in the grid deselected and appends the calibrated ones, so the second run looked for the design next to the first *unchecked* `.raw` file — the user's original design, which names `.raw` files — and validated it against a list that also held the calibrated paths. Neither `SpectraFiles.First()` at `:926` nor `SpectraFiles.Select(p => p.FilePath)` at `:940` filtered on `.Use`.

The fix hoists the checked-file list once and uses it for the design path, the validation and the `EverythingRunnerEngine` construction, so the check can no longer disagree with the run. A run with no spectra file checked is now refused rather than dereferencing the first unchecked one.

Reproduced through the CLI before the fix, with the same mixed list the GUI passes:

```
Error: The experimental design did not contain the file(s): .../SmallCalibratible_Yeast-calib.mzML
```

and with the checked file alone: `Read ExperimentalDesign.tsv successfully`.

## Second commit: the prompt no longer deletes the file

The dialog above offers "Do you wish to continue with an empty experimental design?" with Yes/No. Answering **Yes** ran `File.Delete(pathToExperDesign)`.

That is three problems at once. The wording never disclosed it — "continue with an empty experimental design" reads as *ignore the file for this run*, not *remove it from disk*. It is irreversible, with no backup or rename. And because of the bug above, the path it targeted was the user's **original** design in the raw-data folder, which was neither the wrong file nor invalid.

It also achieved nothing for the run in progress. `PostSearchAnalysisTask.cs:290` resolves the experimental design from the used-file list independently, so the run reads the calibration folder's design whether or not the raw-folder file exists. The delete only suppressed the prompt on subsequent runs — destroying user data to avoid re-showing a dialog.

Now it proceeds without reading the file, and the prompt says "continue **without** an experimental design". Nothing depends on the file's absence: all ten `ExperimentalDesignFileName` sites either read or write it.

## Tests

`ExperimentalDesignIsReadFromTheSelectedSpectraFilesFolder` runs a real `CalibrationTask`, then asserts the calibrated-only list validates against the design calibration wrote, and that including the deselected uncalibrated file reproduces this error.

**Neither change is reachable from a test, and I want to be explicit about that.** Both edits are in `MainWindow.xaml.cs`, and `Test.csproj` does not reference `GUI.csproj` — so the `.Use` filter and the removed `File.Delete` are not testable here, and the new test cannot fail against unfixed code. It is a characterization test of the two lists' behaviour. What was verified against unfixed code is the mechanism, end to end through `CMD.dll`. `GUI.csproj` builds clean with `-p:EnableWindowsTargeting=true`.

## Deliberately unchanged

The default output-folder textbox still derives from `SpectraFiles.First()` rather than the first checked file; changing it would move where output lands for existing users and is not part of this bug.

Two things noticed and not touched: the "You need to add at least one protein database!" branch notifies and then falls through without `return`; and `CalibrationTask.WriteNewExperimentalDesignFile` matches unsuccessfully-calibrated files with `fn.Contains(originalFileName)`, so `sample1` also matches a path containing `sample11`.


MetaMorpheus/GUI/MainWindow.xaml.cs
```diff
@@ -918,12 +918,20 @@ private void RunAllTasks_Click(object sender, RoutedEventArgs e)
                 NotificationHandler(null, new StringEventArgs("You need to add at least one protein database!", null));
             }
 
+            // only checked spectra files are run, so only they may define the experimental design
+            List<string> spectraFilePathsToUse = SpectraFiles.Where(p => p.Use).Select(p => p.FilePath).ToList();
+            if (!spectraFilePathsToUse.Any())
```
trishorts, line 935: This early return is a real behaviour change and it is not in the PR title, so it is worth being deliberate about.

Two things about it:

**It is stricter than its neighbour.** The database check immediately above (`:918`) notifies and then falls through -- `NotificationHandler(...)` with no `return`. So "no databases" lets the run continue to the experimental-design logic, while "no spectra files" now aborts. That asymmetry is pre-existing on the database side and I would not ask you to fix it here, but the two now read as though they were written to different rules.

**`EverythingRunnerEngine` already refuses this.** `EverythingRunnerEngine.cs:70` warns `"Cannot proceed. No spectra files selected."` and stops the run. So the case was already handled, just further downstream and with a different message. This check is still worth having -- it prevents the `Directory.GetParent(...).First()` on the next line from throwing, which is presumably why you added it -- but the justification is "so the line below has something to work with", not "the user could otherwise run with no files". Worth saying so in the comment, and worth checking the two messages do not contradict each other.

MetaMorpheus/GUI/MainWindow.xaml.cs
```diff
@@ -948,15 +956,15 @@ private void RunAllTasks_Click(object sender, RoutedEventArgs e)
...
-                            "occurred reading it. Do you wish to continue with an empty experimental design?" +
+                            "occurred reading it. Do you wish to continue without an experimental design?" +
                             "\nThe error was: " + errors.First(), "Error", MessageBoxButton.YesNo);
 
-                        if (result == MessageBoxResult.Yes)
-                        {
-                            File.Delete(pathToExperDesign);
-                        }
-                        else
+                        if (result != MessageBoxResult.Yes)
```
trishorts, line 979: Good change, and the reasoning in the comment is the right level of detail.

**But the CMD front end still deletes it.** `CMD/Program.cs:307-311` has the same prompt and the same `File.Delete(pathToExperDesign)` on the same error path:

```csharp
if (result.ToLowerInvariant() == "y" || result.ToLowerInvariant() == "yes")
{
    File.Delete(pathToExperDesign);
}
```

Every argument in your comment applies there verbatim -- the prompt only ever offered to continue, the deletion is undisclosed and irreversible, and `PostSearchAnalysisTask` resolves the design from the used-file list regardless. After this PR the GUI and the CLI disagree about whether reading a malformed experimental design destroys it, which is a worse state than both being wrong in the same way.

I would fold the CMD change in here rather than leaving it -- it is the same three lines, and splitting it means the CLI keeps the data-loss bug with no issue tracking it. (Note #2724 is also editing `Program.cs` in the same method, so there is a small conflict to sequence.)

trishorts (commented):
Both halves of this are right, and the second one is overdue -- a dialog that offers to "continue" and then deletes the user's file is the kind of thing that only gets found by someone losing data.

The reworded prompt matches the new behaviour, which matters: the old text promised an *empty* experimental design and the code delivered a *deleted* one, so the fix and the wording had to move together.

Two things, inline.

MetaMorpheus/GUI/MainWindow.xaml.cs
trishorts: Addressed upstream, so clearing this. #2775 ("Wire the TMT experimental design into the CMD front end", `92dfdd6ca`, merged 2026-09-02 — the day after this comment) replaced both `File.Delete(pathToExperDesign)` calls in `CMD/Program.cs` with `ResolveExperimentalDesign`, which reports and returns an exit code and deletes nothing. `git show upstream/master:MetaMorpheus/CMD/Program.cs | grep File.Delete` is now empty, so the two front ends agree again.

trishorts (approved):
Both halves check out.

Reading the design from the checked files only is right, and it also removes a live crash: `spectraFilePathsToUse.First()` on an all-unchecked list would throw, which is why the new "you need to check at least one spectra file" guard has to be there. Master only guards `SpectraFiles.Any()` (`MainWindow.xaml.cs:922`), not `.Use`, so today an all-unchecked run reaches `EverythingRunnerEngine` with an empty file list. My earlier note about that being an undocumented behaviour change stands, but the new behaviour is the defensible one and I would keep it.

On deleting the design: agreed, and I have resolved my other thread — #2775 removed the matching `File.Delete` from the CMD front end on 2026-09-02, so the two agree again.

Approving.

nbollis (approved):

Alexander-Sol (approved):

Issue #1921, opened 2020-07-07
gi|16128008|ref|NP_414555.1| formatted proteins in FASTA


mriffle:
More a clarifying question: We're attempted an open mod search using a non-uniprot database (a gi| formatted database and some other custom protein names) and getting odd results. Does MetaMorpheus require/expect us to use a uniprot-formatted protein database for searches? Is there a way to turn off this expectation and just grab everything up to the first white space as the protein name in the FASTA file?

acesnik:
Thanks for the questions. Here are the regexes, which you could test at a website like http://regexr.com/

Accession: "([A-Z0-9_.]+)"
Full Name: "\s(.*?)\sOS="
Name: "\|([^\|][A-Z0-9_]+)", gets the second match
Gene Name: "GN=([^ ]+)"
Organism: "OS=(.*?)\sGN="

So based on the entry above, the accession would be 16128008 and the name would be NP_414555.1. Is that what you see?

I think including an option for changing the regexes would be a good idea (maybe in the global settings; MaxQuant has something like this).

mriffle:
We do get "16128008" as the accession, but nothing for the name. Here is the entry in the mzID file:

```
    <DBSequence id="DBS_16128008" length="638" searchDatabase_ref="SDB_0" accession="16128008">
[...]

PR #2782, opened 2026-08-30
Let the user choose or supply the FASTA header regexes

### LINKED ISSUE #1936: Allow user to change the regular expressions for FASTA fields
See this issue https://github.com/smith-chem-wisc/MetaMorpheus/issues/1921

acesnik:
🤖 MetaMorpheus parsed every `.fasta` with the UniProt header regexes, unconditionally. There was no way to change that, so a non-UniProt defline produced an accession that is not an accession — and accession is the join key in every output file we write. This adds a per-task setting, in the TOML and in the GUI, that picks a header format or supplies your own regexes.

Closes #1936.

`ProteinDbLoader.LoadProteinFasta` has always accepted caller-supplied field regexes, so **this is a MetaMorpheus-only change: no mzLib release, no pin bump.** The setting lives on `CommonParameters`, which `DatabaseLoadingEngine` already receives, so nothing had to be threaded anywhere.

## What you get

`[CommonParameters.FastaHeaderParsing]` in every task TOML — one enum and six strings, no nested object graph:

```toml
[CommonParameters.FastaHeaderParsing]
HeaderFormat = "Custom"
CustomAccessionRegex = ">.*\\|(.*)\\|"
CustomFullNameRegex = ""
CustomNameRegex = ""
CustomGeneNameRegex = ""
CustomOrganismRegex = "\\[([^\\]]+)\\]\\s*$"
CustomOrganismIdRegex = ""
```

Formats are `UniProt` (default), `Ensembl`, `Gencode`, `Ncbi`, `Auto` and `Custom`. The three mzLib presets are referenced by name rather than copied, so if mzLib's UniProt accession regex is improved we inherit it. `Ncbi` is new here because mzLib has no protein-path preset for it. An empty custom pattern means that field is skipped, and `Custom` requires an accession pattern — see `Auto` below for why that argument being null is significant.

The GUI panel is one `UserControl` added to the five task windows that load a database. Each gains one XAML line and two code-behind lines, so the panel exists once rather than five times. The CMD front end needs no new flag: `-t` already carries the TOML.

## `Auto`, and the root cause it exposes

`LoadProteinFasta` calls `DetectFastaHeaderFormat` **only when the caller passes no accession regex**. MetaMorpheus has always passed one, so it opted out of detection entirely — which is why an Ensembl or Gencode database is mangled here while mzLib's own tests, which pass nothing, parse those files correctly. That is the actual root cause of #1921, and it is a bigger effect than the regex itself.

`Auto` returns an empty regex set, because the empty set is how detection is requested. It is the only option that copes with a Gencode or Ensembl database without the user having to know which one they have.

Two things make it opt-in rather than the default.

**`Auto` is not a synonym for the `UniProt` preset.** Detection assigns `UniprotNameRegex` to `Name` where MetaMorpheus hardcodes `UniprotFullNameRegex`, and does `organismIdRegex ??= UniprotOrganismIdRegex`. So on a UniProt database `Auto` changes `Name` on every entry — `Aspartate aminotransferase, mitochondrial` becomes `AATM_RABIT` — and starts populating `NcbiTaxonomyId`, which is null today. Both are arguably more correct; both move output. A test pins the difference so it is documented rather than discovered.

**Detection throws on a header it cannot classify**, and `Unknown` is easy to reach: a plain `>rat_reductase` defline is none of UniProt, Ensembl or Gencode, and `MzLibException("Unknown fasta header format: …")` reaches a GUI user as a crash-report prompt. The runner gate therefore classifies the sampled deflines up front and refuses with a warning naming the offending header, reusing the deflines it already reads for regex validation. `FastaHeaderType` has no NCBI case, so `Auto` does not help legacy NCBI — the `Ncbi` preset and the companion mzLib PR (smith-chem-wisc/mzLib#1251) cover that.

## Nothing changes by default, and here is the measurement

Both vignette databases — `uniprot-mouse-reviewed-2-5-2018.fasta` (16,954 entries) and `uniprot-cRAP-2-2-2018.fasta` (115) — loaded twice in one process, once through the old literal argument list and once through the new UniProt preset, comparing accession, name, full name, organism, taxonomy database references, gene names and sequence on every entry: **0 differences out of 17,069.** Two tests also pin the preset to mzLib's own constants by reference identity, so a future divergence in either repo fails rather than drifting.

Two existing behaviours are preserved deliberately rather than tidied, and both are worth a decision:

- **`Name` is parsed with the *full-name* regex**, which is what `LoadProteinDb` has always passed. Passing `UniprotNameRegex` instead — presumably what was meant — changes `Name` on **17,069 of 17,069** entries, from `Alcohol dehydrogenase 1` to `ADH1_YEAST`. That is a wholesale output change and does not belong here.
- **Organism id is not parsed.** mzLib's auto-detect branch defaults `organismIdRegex` to `UniprotOrganismIdRegex`, but that branch never runs from MetaMorpheus because we always pass an accession regex, so `Protein.NcbiTaxonomyId` is null today. Passing it changes nothing on those two files — but only because neither carries an `OX=` field (0 of 17,069 headers do; UniProt added `OX=` after 2018). So it is inert on that input and **not** provably inert on a current download, which is why the preset leaves it out.

Opting *in* to `Auto` or to any non-UniProt format changes accessions and therefore every downstream file — `.psmtsv`, protein groups, quantification, `.mzID`, pruned databases. Results are not comparable across the setting.

## Modomics is not where the issue text says it is, and this PR does not reach it

This is the finding that changed the design. **Modomics FASTA is already parsed — by mzLib's `RnaDbLoader`, on the oligo path, not `ProteinDbLoader`.** `RnaFastaHeaderType.Modomics` and an eight-entry `ModomicsFieldRegexes` both exist in the pinned 1.0.587 package, and `DetectRnaFastaHeaderType` selects them on any line starting `>id`. So Modomics headers were never unsupported; they were supported somewhere this setting cannot reach.

`DatabaseLoadingEngine.LoadOligoDb` and `LoadProteinDb` are separate methods. An oligo run reaches `RnaDbLoader.LoadRnaFasta`, which in 1.0.587 takes **no regex parameters at all** — only location, decoy and terminus arguments — with a closed format list and hardcoded `public static readonly` regex dictionaries. **Making the oligo path customizable is therefore an mzLib change and cannot be done from here.** It also throws `MzLibException("Unknown fasta header format: …")` on an unrecognised RNA header, which reaches a GUI user as the crash-report prompt — the same failure mode the `Warn` gate below avoids on the protein path.

An earlier revision of this branch carried a `Modomics` preset for the *protein* path. It has been removed: since a Modomics database is loaded as an oligo, the preset could only ever fire on someone loading RNA into a protein search, which means digesting ACGU with a protease. A dropdown entry named for the motivating case that cannot serve it is worse than no entry, because it looks like the feature works. A test still records what the UniProt default does to such a header, so the gap is documented rather than silently absent:

```
>id:1|Name:tdbR00000010|SOterm:SO:0000254|Type:tRNA|Subtype:Ala|Feature:VGC|Cellular_Localization:prokaryotic cytosol|Species:Escherichia coli
```

yields the accession `Name:tdbR00000010|SOterm:SO:0000254|…|Cellular_Localization:prokaryotic cytosol`.

## A bad configuration is a warning, not a crash

The gate is in `EverythingRunnerEngine.Run`, inside the per-task loop next to the existing database and spectra gates — not before it, since `CurrentXmlDbFilenameList` grows as tasks chain. It reports with `Warn` + `FinishedAllTasks` + `return`. A `MetaMorpheusException` thrown from the task would instead reach a GUI user as "Would you like to report this crash?" with a mailto to mm_support, which is the wrong answer to a typo in a textbox. The throw inside `GetFieldRegexes` stays as a backstop for anything calling `RunTask` directly.

`Validate` refuses a pattern that does not compile, a pattern with no capture group, and a pattern that is too slow to match. That last one needs care: mzLib's `FastaHeaderFieldRegex` constructor is `Regex = new Regex(pattern)` with **no match timeout**, so a 250 ms ceiling applied only to canned sample headers would prove nothing about the user's own file. The runner therefore reads the first few deflines out of each FASTA in the run and validates against those as well as the canned ones. It narrows the window rather than closing it — a timeout on the load itself would need an mzLib change — and the doc comment says so rather than implying more.

An unrecognised format *name* is handled too. The TOML converter used `Enum.Parse`, so `HeaderFormat = "Uniprot2"` failed during the TOML read with a bare "Requested value was not found", before the gate exists to catch it. It now names the legal values. Treating a config typo as a message is the whole premise here; the enum was the one field where it wasn't.

## Tests

Eighteen tests in `Test/FastaHeaderParsingTest.cs`, all green. Fixtures are written at test time, so no `CopyToOutputDirectory` entry was needed, and every path is built with `Path.Combine`.

The gate test asserts on `engine.Warnings` and on an empty list of engines collected from `MetaMorpheusEngine.StartingSingleEngineHander`. It deliberately does not check that the output folder is empty: every result file is written by `PostSearchAnalysisTask`, so that assertion would pass whether or not the search ran. One test strips the new table out of a written TOML to confirm an older TOML still reads as UniProt — as do all five 2019-vintage vignette configs. Reverting any of the three mechanisms added in the last two commits fails exactly the tests written for them. The Gencode assertion is stated as an invariant rather than a value, because the pinned mzLib 1.0.587 and the pending accession fix mangle that header differently and neither yields the Gencode accession — which is the point.

Tests were run on macOS through a throwaway net10.0 NUnit harness, since `Test/Test.csproj` targets `net10.0-windows` and cannot execute here. `Test.csproj` and `GUI.csproj` both compile clean with `EnableWindowsTargeting`, and that does run the WPF markup compiler over the new XAML — but the GUI cannot be *run* on macOS, so **the panel's layout, tooltips and enable/disable behaviour are unverified.** The GUI also does not validate on OK; a bad pattern surfaces at Run, via the warning above.

## Related, separate

`ProteinDbLoader.UniprotAccessionRegex` is `[|](.+)[|]`, unanchored, so on a four-pipe `gi|` defline it captures across all of them. That is a defect in the default itself and ships separately as smith-chem-wisc/mzLib#1251; this PR does not touch it. The two are complementary — because the presets here reference mzLib's regex by name rather than copying it, that fix improves the `UniProt` preset with no change here. The `Ncbi` preset gives affected users an answer in the meantime.

Not touched: the `|` protein-group separator, which is a format-versioning decision of its own. Custom regexes for the oligo path are the natural follow-up and belong in mzLib.

### DIFF
--- a/MetaMorpheus/EngineLayer/CommonParameters.cs
+++ b/MetaMorpheus/EngineLayer/CommonParameters.cs
@@ -1,4 +1,4 @@
-using MassSpectrometry;
+﻿using MassSpectrometry;
 using MzLibUtil;
 using Omics.Fragmentation;
 using Proteomics.ProteolyticDigestion;
@@ -11,6 +11,7 @@ using Transcriptomics.Digestion;
 using EngineLayer.DIA;
 using Transcriptomics;
 using EngineLayer.FdrAnalysis;
+using EngineLayer.DatabaseLoading;
 
 namespace EngineLayer
 {
@@ -65,7 +66,8 @@ namespace EngineLayer
             DIAparameters diaParameters = null,
             IFragmentationParams fragmentationParams = null,
             PrecursorMassMatchMode precursorMassMatchMode = PrecursorMassMatchMode.Monoisotopic,
-            string rtPredictorName = RTPredictorNames.Chronologer)
+            string rtPredictorName = RTPredictorNames.Chronologer,
+            FastaHeaderParsingParameters fastaHeaderParsing = null)
 
         {
             TaskDescriptor = taskDescriptor;
@@ -136,6 +138,7 @@ namespace EngineLayer
             }
 
             RTPredictorName = rtPredictorName;
+            FastaHeaderParsing = fastaHeaderParsing ?? new FastaHeaderParsingParameters();
 
             CustomIons = digestionParams.ProductsFromDissociationType()[DissociationType.Custom];
 
@@ -220,6 +223,11 @@ namespace EngineLayer
         public IFragmentationParams FragmentationParameters { get; set; }
         public string RTPredictorName { get; private set; }
 
+        /// <summary>
+        /// How protein FASTA deflines are split into accession, name, gene and organism.
+        /// </summary>
+        public FastaHeaderParsingParameters FastaHeaderParsing { get; private set; }
+
         public CommonParameters Clone()
         {
             CommonParameters c = new CommonParameters();
@@ -293,7 +301,8 @@ namespace EngineLayer
                                 DIAparameters,
                                 FragmentationParameters,
                                 PrecursorMassMatchMode,
-                                RTPredictorName);
+                                RTPredictorName,
+                                FastaHeaderParsing);
         }
 
         public void SetCustomProductTypes()
--- a/MetaMorpheus/EngineLayer/DatabaseLoading/DatabaseLoadingEngine.cs
+++ b/MetaMorpheus/EngineLayer/DatabaseLoading/DatabaseLoadingEngine.cs
@@ -168,9 +168,11 @@ public class DatabaseLoadingEngine(
         if (theExtension.Equals(".fasta") || theExtension.Equals(".fa"))
         {
             um = null;
+            var headerRegexes = (commonParameters.FastaHeaderParsing ?? new FastaHeaderParsingParameters()).GetFieldRegexes();
             proteinList = ProteinDbLoader.LoadProteinFasta(fileName, generateTargets, decoyType, isContaminant, out var dbErrors,
-                ProteinDbLoader.UniprotAccessionRegex, ProteinDbLoader.UniprotFullNameRegex, ProteinDbLoader.UniprotFullNameRegex, ProteinDbLoader.UniprotGeneNameRegex,
-                ProteinDbLoader.UniprotOrganismRegex, commonParameters.MaxThreadsToUsePerFile, addTruncations: commonParameters.AddTruncations);
+                headerRegexes.Accession, headerRegexes.FullName, headerRegexes.Name, headerRegexes.GeneName,
+                headerRegexes.Organism, commonParameters.MaxThreadsToUsePerFile, addTruncations: commonParameters.AddTruncations,
+                organismIdRegex: headerRegexes.OrganismId);
         }
         else
         {
--- a/MetaMorpheus/EngineLayer/DatabaseLoading/FastaHeaderParsingParameters.cs
+++ b/MetaMorpheus/EngineLayer/DatabaseLoading/FastaHeaderParsingParameters.cs
@@ -1,0 +1,250 @@
+#nullable enable
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using UsefulProteomicsDatabases;
+
+namespace EngineLayer.DatabaseLoading;
+
+/// <summary>
+/// Named FASTA header layouts that MetaMorpheus can parse. <see cref="FastaHeaderFormat.Custom"/>
+/// takes the regexes from <see cref="FastaHeaderParsingParameters"/> instead of a preset.
+/// </summary>
+public enum FastaHeaderFormat
+{
+    UniProt,
+    Ensembl,
+    Gencode,
+    Ncbi,
+    Modomics,
+    Custom
+}
+
+/// <summary>
+/// The six regexes handed to <see cref="ProteinDbLoader.LoadProteinFasta"/>. A null member means the
+/// field is not extracted.
+/// </summary>
+public class FastaHeaderFieldRegexSet
+{
+    public FastaHeaderFieldRegex? Accession { get; init; }
+    public FastaHeaderFieldRegex? FullName { get; init; }
+    public FastaHeaderFieldRegex? Name { get; init; }
+    public FastaHeaderFieldRegex? GeneName { get; init; }
+    public FastaHeaderFieldRegex? Organism { get; init; }
+    public FastaHeaderFieldRegex? OrganismId { get; init; }
+}
+
+/// <summary>
+/// How to pull accession, name, gene and organism out of a protein FASTA defline.
+/// Serialized as a table under [CommonParameters.FastaHeaderParsing]; every member is a string or an
+/// enum so Nett needs no custom converter for the fields themselves.
+/// </summary>
+public class FastaHeaderParsingParameters
+{
+    /// <summary>
+    /// Ceiling on a single validation match. The user supplies these patterns, so a pathological one
+    /// must be rejected at configuration time rather than stalling a load.
+    /// </summary>
+    public static readonly TimeSpan ValidationMatchTimeout = TimeSpan.FromMilliseconds(250);
+
+    /// <summary>
+    /// Deflines the validator matches each pattern against. Only used to smoke out catastrophic
+    /// backtracking; a pattern that matches nothing here is still accepted.
+    /// </summary>
+    internal static readonly string[] ValidationHeaders =
+    [
+        ">sp|P12345|AATM_RABIT Aspartate aminotransferase OS=Oryctolagus cuniculus OX=9986 GN=GOT2 PE=1 SV=2",
+        ">gi|16128008|ref|NP_414555.1| thr operon leader peptide [Escherichia coli str. K-12 substr. MG1655]",
+        ">id:1|Name:tdbR00000010|SOterm:SO:0000254|Type:tRNA|Subtype:Ala|Feature:VGC|Cellular_Localization:prokaryotic cytosol|Species:Escherichia coli"
+    ];
+
+    public FastaHeaderFormat HeaderFormat { get; set; } = FastaHeaderFormat.UniProt;
+
+    public string CustomAccessionRegex { get; set; } = "";
+    public string CustomFullNameRegex { get; set; } = "";
+    public string CustomNameRegex { get; set; } = "";
+    public string CustomGeneNameRegex { get; set; } = "";
+    public string CustomOrganismRegex { get; set; } = "";
+    public string CustomOrganismIdRegex { get; set; } = "";
+
+    public FastaHeaderParsingParameters()
+    {
+    }
+
+    public FastaHeaderParsingParameters(FastaHeaderFormat format,
+        string? accessionRegex = null, string? fullNameRegex = null, string? nameRegex = null,
+        string? geneNameRegex = null, string? organismRegex = null, string? organismIdRegex = null)
+    {
+        HeaderFormat = format;
+        CustomAccessionRegex = accessionRegex ?? "";
+        CustomFullNameRegex = fullNameRegex ?? "";
+        CustomNameRegex = nameRegex ?? "";
+        CustomGeneNameRegex = geneNameRegex ?? "";
+        CustomOrganismRegex = organismRegex ?? "";
+        CustomOrganismIdRegex = organismIdRegex ?? "";
+    }
+
+    public FastaHeaderParsingParameters Clone() => new(HeaderFormat, CustomAccessionRegex,
+        CustomFullNameRegex, CustomNameRegex, CustomGeneNameRegex, CustomOrganismRegex, CustomOrganismIdRegex);
+
+    /// <summary>
+    /// Checks the custom patterns without touching a database. Returns false and fills
+    /// <paramref name="errors"/> with user-facing messages when the configuration cannot be used.
+    /// </summary>
+    public bool Validate(out List<string> errors)
+    {
+        errors = new List<string>();
+        if (HeaderFormat != FastaHeaderFormat.Custom)
+            return true;
+
+        // mzLib falls back to header auto-detection when the accession regex is null, and that
+        // detection throws on a header it does not recognize. Require one instead.
+        if (string.IsNullOrWhiteSpace(CustomAccessionRegex))
+            errors.Add("Custom FASTA header format requires an accession regex. Set 'CustomAccessionRegex' " +
+                       "(the first capture group is used as the accession), or pick a preset FASTA header format.");
+
+        foreach (var (fieldName, pattern) in EnumerateCustomPatterns())
+        {
+            if (string.IsNullOrWhiteSpace(pattern))
+                continue;
+            if (!TryValidatePattern(pattern, out string? reason))
+                errors.Add($"The custom FASTA header regex for {fieldName} is not usable: {reason} Pattern: {pattern}");
+        }
+
+        return errors.Count == 0;
+    }
+
+    /// <summary>
+    /// Builds the regexes to hand to mzLib. Call <see cref="Validate"/> first; this throws on a
+    /// pattern that does not compile, as a backstop for callers that bypass the runner.
+    /// </summary>
+    public FastaHeaderFieldRegexSet GetFieldRegexes()
+    {
+        switch (HeaderFormat)
+        {
+            case FastaHeaderFormat.UniProt:
+                // FullName is deliberately used for both FullName and Name: that is what MetaMorpheus
+                // has always passed, and OrganismId has never been parsed. Changing either moves output.
+                return new FastaHeaderFieldRegexSet
+                {
+                    Accession = ProteinDbLoader.UniprotAccessionRegex,
+                    FullName = ProteinDbLoader.UniprotFullNameRegex,
+                    Name = ProteinDbLoader.UniprotFullNameRegex,
+                    GeneName = ProteinDbLoader.UniprotGeneNameRegex,
+                    Organism = ProteinDbLoader.UniprotOrganismRegex,
+                };
+
+            case FastaHeaderFormat.Ensembl:
+                return new FastaHeaderFieldRegexSet
+                {
+                    Accession = ProteinDbLoader.EnsemblAccessionRegex,
+                    FullName = ProteinDbLoader.EnsemblFullNameRegex,
+                    GeneName = ProteinDbLoader.EnsemblGeneNameRegex,
+                };
+
+            case FastaHeaderFormat.Gencode:
+                return new FastaHeaderFieldRegexSet
+                {
+                    Accession = ProteinDbLoader.GencodeAccessionRegex,
+                    FullName = ProteinDbLoader.GencodeFullNameRegex,
+                    GeneName = ProteinDbLoader.GencodeGeneNameRegex,
+                };
+
+            case FastaHeaderFormat.Ncbi:
+                return new FastaHeaderFieldRegexSet
+                {
+                    Accession = NcbiAccessionRegex,
+                    FullName = NcbiFullNameRegex,
+                    Organism = NcbiOrganismRegex,
+                };
+
+            case FastaHeaderFormat.Modomics:
+                return new FastaHeaderFieldRegexSet
+                {
+                    Accession = ModomicsAccessionRegex,
+                    FullName = ModomicsFullNameRegex,
+                    Name = ModomicsNameRegex,
+                    Organism = ModomicsOrganismRegex,
+                };
+
+            case FastaHeaderFormat.Custom:
+                return new FastaHeaderFieldRegexSet
+                {
+                    Accession = Build("accession", CustomAccessionRegex),
+                    FullName = Build("fullName", CustomFullNameRegex),
+                    Name = Build("name", CustomNameRegex),
+                    GeneName = Build("geneName", CustomGeneNameRegex),
+                    Organism = Build("organism", CustomOrganismRegex),
+                    OrganismId = Build("organismId", CustomOrganismIdRegex),
+                };
+
+            default:
+                throw new MetaMorpheusException($"Unrecognized FASTA header format: {HeaderFormat}");
+        }
+    }
+
+    /// <summary>
+    /// Legacy and modern NCBI deflines: >gi|16128008|ref|NP_414555.1| ... and >ref|NP_414555.1| ...
+    /// The greedy prefix pushes the opening pipe as late as possible, so the capture lands between the
+    /// last two pipes rather than spanning all of them.
+    /// </summary>
+    public static readonly FastaHeaderFieldRegex NcbiAccessionRegex = new("accession", @">.*\|(.*)\|", 0, 1);
+    public static readonly FastaHeaderFieldRegex NcbiFullNameRegex = new("fullName", @">[^ ]*\s(.*?)(\s\[|$)", 0, 1);
+    public static readonly FastaHeaderFieldRegex NcbiOrganismRegex = new("organism", @"\[([^\[\]]+)\]\s*$", 0, 1);
+
+    // Patterns copied verbatim from mzLib RnaDbLoader.ModomicsFieldRegexes; only the mapping onto the
+    // protein fields is ours. Accession uses Name (unique per entry) rather than the repeating SOterm.
+    public static readonly FastaHeaderFieldRegex ModomicsAccessionRegex = new("accession", @"Name:(.+?)\|", 0, 1);
+    public static readonly FastaHeaderFieldRegex ModomicsNameRegex = new("name", @"SOterm:(.+?)\|", 0, 1);
+    public static readonly FastaHeaderFieldRegex ModomicsFullNameRegex = new("fullName", @"Type:(.+?)\|", 0, 1);
+    public static readonly FastaHeaderFieldRegex ModomicsOrganismRegex = new("organism", @"Species:(.+?)$", 0, 1);
+
+    private static FastaHeaderFieldRegex? Build(string fieldName, string pattern) =>
+        string.IsNullOrWhiteSpace(pattern) ? null : new FastaHeaderFieldRegex(fieldName, pattern, 0, 1);
+
+    private IEnumerable<(string FieldName, string Pattern)> EnumerateCustomPatterns()
+    {
+        yield return ("accession", CustomAccessionRegex);
+        yield return ("full name", CustomFullNameRegex);
+        yield return ("name", CustomNameRegex);
+        yield return ("gene name", CustomGeneNameRegex);
+        yield return ("organism", CustomOrganismRegex);
+        yield return ("organism id", CustomOrganismIdRegex);
+    }
+
+    private static bool TryValidatePattern(string pattern, out string? reason)
+    {
+        Regex compiled;
+        try
+        {
+            compiled = new Regex(pattern, RegexOptions.None, ValidationMatchTimeout);
+        }
+        catch (ArgumentException e)
+        {
+            reason = e.Message;
+            return false;
+        }
+
+        if (compiled.GetGroupNumbers().Length < 2)
+        {
+            reason = "it has no capture group, so no field value can be taken from it.";
+            return false;
+        }
+
+        foreach (string header in ValidationHeaders)
+        {
+            try
+            {
+                compiled.Match(header);
+            }
+            catch (RegexMatchTimeoutException)
+            {
+                reason = $"matching it against a sample header took longer than {ValidationMatchTimeout.TotalMilliseconds} ms.";
+                return false;
+            }
+        }
+
+        reason = null;
+        return true;
+    }
+}
--- a/MetaMorpheus/EngineLayer/DatabaseLoading/FastaHeaderParsingParameters.cs
+++ b/MetaMorpheus/EngineLayer/DatabaseLoading/FastaHeaderParsingParameters.cs
@@ -9,7 +9,8 @@ namespace EngineLayer.DatabaseLoading;
 
 /// <summary>
 /// Named FASTA header layouts that MetaMorpheus can parse. <see cref="FastaHeaderFormat.Custom"/>
-/// takes the regexes from <see cref="FastaHeaderParsingParameters"/> instead of a preset.
+/// takes the regexes from <see cref="FastaHeaderParsingParameters"/> instead of a preset;
+/// <see cref="FastaHeaderFormat.Auto"/> passes none and lets mzLib detect the format per file.
 /// </summary>
 public enum FastaHeaderFormat
 {
@@ -17,6 +18,7 @@ public enum FastaHeaderFormat
     Ensembl,
     Gencode,
     Ncbi,
+    Auto,
     Custom
 }
 
@@ -96,6 +98,25 @@ public class FastaHeaderParsingParameters
     public bool Validate(out List<string> errors, IEnumerable<string>? sampleHeaders = null)
     {
         errors = new List<string>();
+
+        if (HeaderFormat == FastaHeaderFormat.Auto)
+        {
+            // mzLib throws on a header it cannot classify, which reaches a GUI user as a crash
+            // report. Refuse it here instead, while there is still a message to give.
+            foreach (string header in sampleHeaders ?? [])
+            {
+                if (ProteinDbLoader.DetectFastaHeaderFormat(header) != FastaHeaderType.Unknown)
+                    continue;
+
+                errors.Add("The FASTA header format could not be detected automatically for this "
+                           + $"header: {header}. Pick a preset FASTA header format, or use Custom "
+                           + "and supply an accession regex.");
+                break;
+            }
+
+            return errors.Count == 0;
+        }
+
         if (HeaderFormat != FastaHeaderFormat.Custom)
             return true;
 
@@ -159,6 +180,11 @@ public class FastaHeaderParsingParameters
                     FullName = NcbiFullNameRegex,
                     Organism = NcbiOrganismRegex,
                 };
+
+            // All null on purpose: LoadProteinFasta only calls DetectFastaHeaderFormat when the
+            // accession regex is null, so an empty set is how detection is requested.
+            case FastaHeaderFormat.Auto:
+                return new FastaHeaderFieldRegexSet();
 
             case FastaHeaderFormat.Custom:
                 return new FastaHeaderFieldRegexSet
--- a/MetaMorpheus/EngineLayer/DatabaseLoading/FastaHeaderParsingParameters.cs
+++ b/MetaMorpheus/EngineLayer/DatabaseLoading/FastaHeaderParsingParameters.cs
@@ -1,6 +1,7 @@
 #nullable enable
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text.RegularExpressions;
 using UsefulProteomicsDatabases;
 
@@ -16,7 +17,6 @@ public enum FastaHeaderFormat
     Ensembl,
     Gencode,
     Ncbi,
-    Modomics,
     Custom
 }
 
@@ -42,14 +42,17 @@ public class FastaHeaderFieldRegexSet
 public class FastaHeaderParsingParameters
 {
     /// <summary>
-    /// Ceiling on a single validation match. The user supplies these patterns, so a pathological one
-    /// must be rejected at configuration time rather than stalling a load.
+    /// Ceiling on a single validation match. Applies to validation only: mzLib's
+    /// <see cref="FastaHeaderFieldRegex"/> compiles the pattern with no timeout, so a pattern that is
+    /// fast on the headers checked here can still stall a load. Passing real deflines to
+    /// <see cref="Validate"/> is what narrows that gap.
     /// </summary>
     public static readonly TimeSpan ValidationMatchTimeout = TimeSpan.FromMilliseconds(250);
 
     /// <summary>
-    /// Deflines the validator matches each pattern against. Only used to smoke out catastrophic
-    /// backtracking; a pattern that matches nothing here is still accepted.
+    /// Deflines every validation run matches each pattern against, on top of any real ones supplied.
+    /// Only used to smoke out catastrophic backtracking; a pattern that matches nothing here is still
+    /// accepted, since a preset for one database format will not match another.
     /// </summary>
     internal static readonly string[] ValidationHeaders =
     [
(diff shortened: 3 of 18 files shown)


MetaMorpheus/EngineLayer/DatabaseLoading/FastaHeaderParsingParameters.cs
```diff
@@ -0,0 +1,262 @@
...
+        }
+
+        return errors.Count == 0;
+    }
+
+    /// <summary>
+    /// Builds the regexes to hand to mzLib. Call <see cref="Validate"/> first; this throws on a
+    /// pattern that does not compile, as a backstop for callers that bypass the runner.
+    /// </summary>
+    public FastaHeaderFieldRegexSet GetFieldRegexes()
```
trishorts, line 144: **mzLib already has this table -- it is just not public.**

`ProteinDbLoader.cs:300-322` contains a `switch (HeaderType)` mapping UniProt/Ensembl/Gencode onto precisely these regex constants, inline in the read loop and private. So this method is not adding a concept; it is re-expressing an existing mzLib switch as a MetaMorpheus type because mzLib forgot to expose it. Two copies of the same mapping in two repos will drift, and the description's own goal -- "if mzLib's UniProt accession regex is improved we inherit it" -- only holds for the *constants*, not for the *grouping*.

There is also a ready-made precedent for the fix upstream. `RnaDbLoader.cs:69-127` already declares five `public static readonly Dictionary<string, FastaHeaderFieldRegex>` presets -- `ModomicsFieldRegexes`, `EnsemblFieldRegexes`, `NcbiAssemblyFieldRegexes`, `NcbiRefSeqGeneFieldRegexes`, `MzLibRegexes` -- each a named, public, per-format bundle selected by a switch at `:190-217`. That is this class, on the RNA half of the same assembly. The protein side simply never got it.

The "referenced by name rather than copied" claim also only fully holds for UniProt. `EnsemblAccessionRegex`/`FullName`/`GeneName` (`ProteinDbLoader.cs:56-58`) and the Gencode trio (`:60-62`) are three fields each, not six -- which is why the Ensembl and Gencode cases below leave `Name`, `Organism` and `OrganismId` null. If anyone later fills those in, the authoritative Ensembl organism pattern will live in MetaMorpheus, where `LoadProteinFasta`'s own auto-detect path can never use it and FlashLFQ / ProteaseGuru / pyMzLib never see it.

I would split this: `FastaHeaderFieldRegexSet` + the preset table + public accessors go upstream to `ProteinDbLoader` as the mirror of the RNA presets, and MetaMorpheus keeps the enum-to-preset binding, the `Custom` free-text entry, validation, and the GUI. That does impose the usual ordering (mzLib PR merges and releases, then this consumes it), which is a real cost -- but the alternative is a permanent second home for a mapping mzLib already owns.

MetaMorpheus/EngineLayer/DatabaseLoading/FastaHeaderParsingParameters.cs
```diff
@@ -0,0 +1,262 @@
...
+
+            case FastaHeaderFormat.Gencode:
+                return new FastaHeaderFieldRegexSet
+                {
+                    Accession = ProteinDbLoader.GencodeAccessionRegex,
+                    FullName = ProteinDbLoader.GencodeFullNameRegex,
+                    GeneName = ProteinDbLoader.GencodeGeneNameRegex,
+                };
+
+            case FastaHeaderFormat.Ncbi:
```
trishorts, line 176: **`Ncbi` is an mzLib gap rather than a MetaMorpheus feature, and adding it here makes the gap worse rather than better.**

The description says "mzLib has no protein-path preset for NCBI", which is literally true and somewhat misleading: mzLib already knows how to parse NCBI FASTA headers. `RnaFastaHeaderType` (`RnaDbLoader.cs:18-26`) has `NcbiRefSeq` and `NcbiAssembly` members, `RnaDbLoader.cs:44,46,47` hold the detection regexes, and `:95-115` hold accession/name/gene/organism field bundles for both variants.

So this defines a third, protein-flavoured NCBI header spec that cannot be reconciled with the two mzLib already ships. And because `ProteinDbLoader.FastaHeaderType` (`:21-22`) still has only `{ UniProt, Ensembl, Gencode, Unknown }`, `DetectFastaHeaderFormat` remains unable to classify an NCBI protein FASTA -- meaning every other mzLib consumer, *and this PR's own `Auto` option*, still hits `throw new MzLibException("Unknown fasta header format: " + line)` at `ProteinDbLoader.cs:297` on the very file MetaMorpheus can now parse.

Upstream this is one enum member, one detection branch and one regex bundle, and everyone gains it. I see #1251 is referenced in the description as the companion -- if that is where this is going, worth saying so explicitly here so a reviewer knows the `Ncbi` case is a temporary local copy rather than the intended permanent home.

trishorts (commented):
This is a real gap -- parsing every FASTA with the UniProt regexes and producing an "accession" that is not an accession, when accession is the join key in every output file, is worth fixing. And the measurement discipline in the description is exemplary: loading both vignette databases twice in one process and comparing every field on all 17,069 entries is the right way to claim "nothing changes by default", and it is far more convincing than a green test suite.

I also want to credit two things the description gets right that are easy to get wrong:

- **The `Name`/`FullName` question is left alone deliberately and documented.** `GetFieldRegexes()` preserves `Name = ProteinDbLoader.UniprotFullNameRegex`, which reproduces the pre-existing duplicate argument at the old `LoadProteinFasta` call site. That looks like a bug being cemented, and it *is* a bug -- but changing it moves `Name` on 17,069 of 17,069 entries, so keeping it here and flagging it is the correct call.
- **The mechanical integration is complete**, which for a new `CommonParameters` property is the part people miss. The ctor parameter (`:70`), the non-null default (`:141`), `CloneWithNewTerminus` passing it positionally (`:305`), `SetAllFileSpecificCommonParams` threading it (`MetaMorpheusTask.cs:645`), and the Nett `ConfigureType<...>.CreateInstance` registration (`MetaMorpheusTask.cs:192`) are all present. Both hand-maintained rebuild paths would have silently reset the property, and neither does.

My substantive concern is placement, not correctness -- I think most of this belongs in mzLib. Details inline, but the short version: the preset table you are adding already exists in mzLib as a private `switch`, and mzLib already has a public idiom for exactly this on the RNA side.

MetaMorpheus/EngineLayer/DatabaseLoading/FastaHeaderParsingParameters.cs
```diff
@@ -0,0 +1,262 @@
...
+        CustomOrganismIdRegex = organismIdRegex ?? "";
+    }
+
+    /// <summary>
+    /// Checks the custom patterns. Returns false and fills <paramref name="errors"/> with user-facing
+    /// messages when the configuration cannot be used. <paramref name="sampleHeaders"/> should be real
+    /// deflines from the databases about to be loaded where the caller can get them: the canned
+    /// headers cannot prove a pattern is fast on the user's own file.
+    /// </summary>
+    public bool Validate(out List<string> errors, IEnumerable<string>? sampleHeaders = null)
```
trishorts, line 98: A sharp edge in `LoadProteinFasta` that is worth pinning with a test, since `Validate` is where you would catch it.

Auto-detection is gated on `accessionRegex == null` **alone** (`ProteinDbLoader.cs:291`), not on all six being null. And within that branch, lines 303-320 assign the detected preset with `=`, not `??=` -- the single exception being `organismIdRegex ??= UniprotOrganismIdRegex` at `:307`.

The consequence: a caller that supplies, say, only a gene-name regex and leaves accession null gets detection run *and* gets its gene-name regex silently overwritten by the detected preset. Your description already says `Custom` requires an accession pattern, which is exactly what avoids this -- but that invariant is currently load-bearing in a way the code does not state. A test that asserts `Custom` with an empty accession regex is refused would stop someone relaxing that rule later without realising it silently clobbers the other five fields.

Also worth knowing: `HeaderType` is latched from the first `>` line only, so a mixed-format FASTA is parsed entirely with the first header's preset. That is mzLib's behaviour, not yours, but it bounds what `Auto` can honestly promise.

MetaMorpheus/EngineLayer/CommonParameters.cs
```diff
@@ -220,6 +223,11 @@ public int DeconvolutionMaxAssumedChargeState
         public IFragmentationParams FragmentationParameters { get; set; }
         public string RTPredictorName { get; private set; }
 
+        /// <summary>
+        /// How protein FASTA deflines are split into accession, name, gene and organism.
+        /// </summary>
+        public FastaHeaderParsingParameters FastaHeaderParsing { get; private set; }
```
trishorts, line 229: **Placement question: this is a property of a database, but it now lives on the per-spectra-file object.**

`CommonParameters` is what `FileSpecificParameters` overrides per spectra file -- that is its whole purpose. Databases are carried separately, as `DbForTask`. So after this change a user can in principle write a per-spectra-file `.toml` that sets a different FASTA header format for `sample1.mzML` than for `sample2.mzML`, which is not a meaningful thing to say: the databases are loaded once, for the task, before any spectra file is considered.

You have threaded it through `SetAllFileSpecificCommonParams` correctly so it does not get *reset*, which is the bug that would have bitten. But that also means the nonsensical override is now expressible rather than rejected.

Two alternatives worth weighing:

- **`DbForTask`** -- header format is genuinely a per-database property, and it would let a user load a UniProt database and an Ensembl one in the same run, which the current design cannot express at all. This is the most correct home and the most work.
- **`GuiGlobalParams`** (`GuiFunctions/GuiGlobalParams.cs:11`) -- the established home for run-level settings that are not per-spectra-file.

If `CommonParameters` stays, it would be worth having `FileSpecificParameters` explicitly reject a `FastaHeaderParsing` key rather than leaving the override silently available. Its `default:` case already throws `MetaMorpheusException("Unrecognized parameter ...")` for unknown keys, so the machinery is there.

MetaMorpheus/EngineLayer/DatabaseLoading/DatabaseLoadingEngine.cs
```diff
@@ -168,9 +168,11 @@ public static IEnumerable<Protein> LoadProteinDb(string fileName, bool generateT
         if (theExtension.Equals(".fasta") || theExtension.Equals(".fa"))
         {
             um = null;
+            var headerRegexes = (commonParameters.FastaHeaderParsing ?? new FastaHeaderParsingParameters()).GetFieldRegexes();
```
trishorts, line 171: The oligo path a few lines above (`LoadOligoDb`, `DatabaseLoadingEngine.cs:144`) gets none of this, and the new setting is silently inert there.

`RnaDbLoader.LoadRnaFasta` takes no regex parameters at all -- it is auto-detect-only, using `DetectRnaFastaHeaderType` (`RnaDbLoader.cs:50-64`) and its own five public preset dictionaries. So a user who sets `HeaderFormat = Ncbi` on an RNA task gets no error, no warning, and no effect.

Given the GUI panel is added to five task windows and those windows are shared between protein and oligo work, someone will set this and wonder why nothing happened. Worth either warning when a non-default header format is configured for an oligo database, or noting the limitation in the panel's tooltip.

It also sharpens the previous point: the two mzLib loaders have divergent contracts for the same concern -- the protein one takes caller-supplied regexes and only auto-detects when accession is null, the RNA one always auto-detects and accepts nothing. Reconciling those upstream is the change that would make this setting mean the same thing on both paths.

MetaMorpheus/GUI/Views/FastaHeaderParsingControl.xaml.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            InitializeComponent();
+
+            foreach (FastaHeaderFormat format in Enum.GetValues<FastaHeaderFormat>())
+                FastaHeaderFormatComboBox.Items.Add(format);
+
+            SetFromParameters(new FastaHeaderParsingParameters());
+        }
+
+        public void SetFromParameters(FastaHeaderParsingParameters parameters)
```
trishorts, line 22: Convention note on the control, not the parsing logic.

The house pattern for a shared task-window panel is DataContext-driven with a view model in `GuiFunctions/ViewModels/`. `FragmentationParamsControl` -- which is bound into exactly the same five task windows this PR targets -- does `if (DataContext is FragmentationParamsViewModel viewModel)` and keeps the code-behind thin, with `FragmentationParamsViewModel.cs` holding the logic. `SpectralAveragingParametersViewModel`, `MassDifferenceAcceptorSelectionViewModel` and `GptmdFilterViewModel` follow the same shape.

This control instead exposes imperative `SetFromParameters` / `ToParameters` methods that each window calls. It works, and it is less code -- but it puts the enum-to-visibility logic and the regex round-trip inside a WPF `UserControl`, where it cannot be unit-tested and cannot move. That is worth avoiding right now in particular: #2729 and #2733 are actively extracting the UI-free half of `GuiFunctions` into a `net10.0` `GuiFunctions.Core` so it can be tested on Linux and macOS, and a view model would land on the right side of that line while this code-behind will not.

Not a blocker, but a `FastaHeaderParsingViewModel` would match the neighbours, would make `Validate`'s error surface testable without a window, and would cost roughly what the two methods already cost.

MetaMorpheus/EngineLayer/DatabaseLoading/FastaHeaderParsingParameters.cs
```diff
@@ -0,0 +1,262 @@
...
+                throw new MetaMorpheusException($"Unrecognized FASTA header format: {HeaderFormat}");
+        }
+    }
+
+    /// <summary>
+    /// Legacy and modern NCBI deflines: >gi|16128008|ref|NP_414555.1| ... and >ref|NP_414555.1| ...
+    /// The greedy prefix pushes the opening pipe as late as possible, so the capture lands between the
+    /// last two pipes rather than spanning all of them.
+    /// </summary>
+    public static readonly FastaHeaderFieldRegex NcbiAccessionRegex = new("accession", @">.*\|(.*)\|", 0, 1);
```
trishorts, line 210: **Major -- the `Ncbi` preset cannot match a current NCBI defline, and the failure is silent and destructive.**

`>.*\|(.*)\|` requires at least two pipes. NCBI retired GI numbers in 2016; RefSeq and GenBank have emitted pipe-less deflines ever since. I ran the three preset regexes against four real defline shapes:

| defline | accession | fullName | organism |
|---|---|---|---|
| `>gi\|16128008\|ref\|NP_414555.1\| thr operon leader peptide [E. coli str. K-12]` | `NP_414555.1` | ok | ok |
| `>ref\|NP_414555.1\| thr operon leader peptide [E. coli str. K-12]` | `NP_414555.1` | ok | ok |
| `>NP_414555.1 thr operon leader peptide [E. coli str. K-12]` | **NO MATCH** | ok | ok |
| `>CAA23472.1 thrL [Escherichia coli]` | **NO MATCH** | ok | ok |

The other two fields still match, so nothing looks wrong. What happens to the unmatched accession is the problem -- `ProteinDbLoader.cs:336-339`:

```csharp
if (accession == null || accession == "")
{
    accession = line.Substring(1).TrimEnd();
}
```

The accession silently becomes **the entire defline**. Pick `Ncbi` on a database downloaded from NCBI any time in the last nine years and every protein in `AllPSMs.psmtsv`, `AllProteinGroups.tsv` and the `.mzID` is keyed by `NP_414555.1 thr operon leader peptide [Escherichia coli str. K-12]`. No exception, no warning, and the search still runs to completion.

`NcbiPresetParsesALegacyDefline` (`FastaHeaderParsingTest.cs:93`) uses `LegacyNcbiHeader`, so the one shape under test is the one shape that works.

Something like `>(?:.*\|)?([^\s|]+?)\|?(?:\s|$)` would cover both, but the class of bug matters more than the pattern: a preset that half-matches is worse than one that throws, because the fallback turns it into data corruption rather than an error. It is also why I would still rather this preset lived in mzLib beside the other four (my other thread) -- there it would sit under mzLib's own header tests instead of a single MetaMorpheus case.

MetaMorpheus/EngineLayer/DatabaseLoading/FastaHeaderParsingParameters.cs
```diff
@@ -0,0 +1,262 @@
...
+        {
+            reason = "it has no capture group, so no field value can be taken from it.";
+            return false;
+        }
+
+        foreach (string header in ValidationHeaders.Concat(sampleHeaders ?? []))
+        {
+            try
+            {
+                compiled.Match(header);
```
trishorts, line 250: **Major -- the gate requires an accession regex but never checks that it matches anything, and the same defline-as-accession fallback is waiting behind it.**

`compiled.Match(header)` is called and the result discarded. The loop exists only to catch `RegexMatchTimeoutException`, so what is actually validated is "does this pattern compile, have a capture group, and not catastrophically backtrack" -- not "does it match the user's headers".

A regex that is valid and matches nothing therefore passes `Validate` cleanly. It reaches `ProteinDbLoader`, misses on every line, and hits `accession = line.Substring(1).TrimEnd()` at `ProteinDbLoader.cs:336-339` -- whole defline as accession, for the entire database, silently. Which is the outcome this gate exists to prevent. The doc comment two methods up says as much:

> mzLib falls back to header auto-detection when the accession regex is null, and that detection throws on a header it does not recognize. Require one instead.

It requires one to *exist*. Nothing requires it to *work*. And a user reaching for Custom is by definition someone whose headers no preset handles -- the population most likely to get the regex wrong.

`sampleHeaders` is already threaded all the way down to here, so this is using a return value you are already computing:

```csharp
var samples = (sampleHeaders ?? []).ToList();
if (samples.Count > 0 && !samples.Any(h => compiled.Match(h).Success))
{
    reason = "it does not match any of the sampled FASTA headers.";
    return false;
}
```

Check `Success` against `sampleHeaders` only, not `ValidationHeaders` -- the built-ins are backtracking bait, and a custom pattern has no reason to match them.

MetaMorpheus/TaskLayer/EverythingRunnerEngine.cs
```diff
@@ -142,6 +154,44 @@ public void Run()
...
+        private static List<string> SampleFastaHeaders(List<DbForTask> databases, int perFile = 5)
+        {
+            var headers = new List<string>();
+            foreach (var db in databases ?? new List<DbForTask>())
+            {
+                if (db.IsSpectralLibrary)
+                    continue;
+
+                string extension = Path.GetExtension(db.FilePath).ToLowerInvariant();
+                if (extension != ".fasta" && extension != ".fa")
```
trishorts, line 171: **Major -- gzipped FASTA is skipped here, so the `Auto` gate passes vacuously on exactly the databases it exists to protect.**

`Path.GetExtension("uniprot_sprot.fasta.gz")` returns `".gz"`, so this `continue` drops every compressed database. `SampleFastaHeaders` returns an empty list, and the `Auto` branch of `Validate` iterates `sampleHeaders ?? []`:

```csharp
foreach (string header in sampleHeaders ?? [])
{
    if (ProteinDbLoader.DetectFastaHeaderFormat(header) != FastaHeaderType.Unknown)
        continue;
    errors.Add(...);
    break;
}
return errors.Count == 0;   // true -- nothing was ever examined
```

Zero headers means zero errors means the gate passes. The mzLib crash it was written to pre-empt (`DetectFastaHeaderFormat` throwing `Unknown fasta header format:`) then happens anyway, arriving as the GUI crash report the comment at `FastaHeaderParsingParameters.cs:103` says it is there to avoid.

`.fasta.gz` is not an edge case -- it is how UniProt and NCBI ship, mzLib loads it, and `TestDbReadingDifferentExtensions("20mer1.fasta.gz")` already pins that it does.

Sampling through the same decompression the loader uses would fix it. Failing that, the gate should report that it could not sample rather than return `true`: an empty sample and a clean sample must not produce the same answer.

MetaMorpheus/Test/FastaHeaderParsingTest.cs
```diff
@@ -0,0 +1,367 @@
...
+                "the accession contains the character MetaMorpheus uses to separate multiple values");
+
+            var withCustom = Load(path, new FastaHeaderParsingParameters(FastaHeaderFormat.Custom,
+                accessionRegex: @">.*\|(.*)\|")).Single();
+            Assert.That(withCustom.Accession, Is.EqualTo("NP_414555.1"));
+            Assert.That(withCustom.Organism, Is.Null.Or.Empty, "no organism regex was given, so the field is skipped");
+        }
+
+        [Test]
+        public static void NcbiPresetParsesALegacyDefline()
```
trishorts, line 93: **Minor -- three of the four presets have no test that goes through `GetFieldRegexes()`.**

- `Ncbi`: covered, but only by `LegacyNcbiHeader`, the pipe-bearing shape. That is the only shape the regex handles (see my comment on `FastaHeaderParsingParameters.cs:210`), so this test cannot fail for the reason it ought to.
- `Gencode`: reached only through *Auto* detection, in `AutoDetectsGencodeWhereTheUniProtPresetManglesIt`. `FastaHeaderFormat.Gencode` is never constructed, so `case FastaHeaderFormat.Gencode:` in `GetFieldRegexes()` is never executed.
- `Ensembl`: appears once, in a comment on line 334. The preset arm has no coverage at all.

Those arms are one-line delegations to mzLib's `EnsemblAccessionRegex` / `GencodeAccessionRegex`, so the risk is not the regexes themselves -- it is that the wiring is unpinned, and this PR is what promotes the enum to a user-facing choice. One parameterised case per preset over a representative defline would close all three, and would have caught the NCBI one:

```csharp
[TestCase(FastaHeaderFormat.Ensembl, EnsemblHeader, "ENSP00000493376.2")]
[TestCase(FastaHeaderFormat.Gencode, GencodeHeader, "ENSP00000493376.2")]
[TestCase(FastaHeaderFormat.Ncbi, ModernNcbiHeader, "NP_414555.1")]
```

trishorts (commented):
Ran the full review pipeline over this one rather than reading it by hand -- 52 findings, 10 of them Major, collapsing to four distinct issues once the parallel passes are de-duplicated. I verified each of the four myself against head `7e986ad` before posting; what follows is my wording and my measurements, not the engine's.

Three of them are the same decision seen from three directions, and it is `ProteinDbLoader.cs:336-339`:

```csharp
if (accession == null || accession == "")
    accession = line.Substring(1).TrimEnd();
```

An accession regex that fails to match does not raise -- it makes the whole defline the accession. So every "this regex does not fit these headers" case becomes silent data corruption that survives into `AllPSMs.psmtsv`, `AllProteinGroups.tsv` and the `.mzID`. It is reachable three ways here: the `Ncbi` preset on any post-2016 defline, Custom with a regex that matches nothing, and Auto on a gzipped database. Worth settling once what this PR wants to happen when an accession regex matches no sampled header, since all three findings fall out of that one answer.

My six threads from 2026-09-01 are all still open and unaddressed; I have not repeated them here.

trishorts (commented):
Went back through the 43 Minor/Informational findings from the same run. They de-duplicate to about twenty distinct items, nine of which are whitespace-revert noise; I am not posting those, and one (`FastaHeaderParsingTest.cs:174`, "backwards-compat test passes vacuously") I checked and it is a false positive — the `Does.Not.Contain` guard on line 174 makes it non-vacuous.

What is below is the residue: the five I think are worth acting on now because they are things this code will make someone deal with later, rather than style. Each is verified against head `7e986ad`.

The first one is not really "later" at all — **two test assertions are already scheduled to fail on this PR's next required branch update**, because mzLib #1251 is merged and shipped in 1.0.589 while this branch still pins 1.0.587. Details in that thread; it is the one I would fix before anything else here, purely so the person doing the bump is not the one who has to work out why.

MetaMorpheus/Test/FastaHeaderParsingTest.cs
```diff
@@ -0,0 +1,367 @@
...
+            Assert.That(protein.GeneNames.Single().Item2, Is.EqualTo("GOT2"));
+        }
+
+        [Test]
+        public static void UniProtDefaultGluesLegacyNcbiFieldsTogetherAndACustomRegexDoesNot()
+        {
+            string path = WriteFasta(Path.Combine("FastaHeaderParsing", "legacyNcbi.fasta"), LegacyNcbiHeader, "PEPTIDEK");
+
+            var withDefault = Load(path, new FastaHeaderParsingParameters()).Single();
+            Assert.That(withDefault.Accession, Is.EqualTo("16128008|ref|NP_414555.1"));
```
trishorts, line 82: **These two assertions are already scheduled to fail — mzLib #1251 is merged and shipped, and this branch is the only reason they are still green.**

Both literals are consequences of the *old* `UniprotAccessionRegex`, `[|](.+)[|]`. mzLib #1251 ("Capture the last pipe-delimited field as the UniProt FASTA accession", merged 2026-09-01) replaced it with `^>.*\|([^|]*)\|`, and that shipped in **1.0.589** on 2026-09-02. This PR pins **1.0.587**; MetaMorpheus master is already on **1.0.589**. Master requires up-to-date branches, so the pin moves before this can merge.

I ran both regexes over the three fixture headers:

| header | 1.0.587 | 1.0.589 |
|---|---|---|
| `LegacyNcbiHeader` | `16128008\|ref\|NP_414555.1` | `NP_414555.1` |
| `ModomicsHeader` | `Name:tdbR00000010\|SOterm:...\|Cellular_Localization:prokaryotic cytosol` | `Cellular_Localization:prokaryotic cytosol` |
| `GencodeHeader` | `ENSMUST...\|...\|Mmp20` | `Mmp20` |

So:

- line 82, `Assert.That(withDefault.Accession, Is.EqualTo("16128008|ref|NP_414555.1"))` — **fails**, gets `NP_414555.1`.
- line 109, `Assert.That(protein.Accession, Does.StartWith("Name:tdbR00000010|"))` — **fails**, gets `Cellular_Localization:prokaryotic cytosol`.

Whoever does the branch update gets two red tests in a file about FASTA header parsing, immediately after a PR about FASTA header parsing, and will reasonably look for the bug here first. It is not here — it is an intended upstream improvement propagating through a by-reference preset.

`AutoDetectsGencodeWhereTheUniProtPresetManglesIt` (line 299) already got this right: it asserts the accession is *not* the correct one rather than pinning the mangled string, which is why it survives the bump. These two want the same treatment — assert the accession is not `NP_414555.1` / `tdbR00000010`, or that it still contains a `|`, and the test keeps meaning "the UniProt default mangles this header" under either regex.

Worth doing in this PR rather than leaving it for the bump, since right now you are the only person who knows why they went red.

MetaMorpheus/TaskLayer/EverythingRunnerEngine.cs
```diff
@@ -87,6 +87,18 @@ public void Run()
                
                 var ok = TaskList[i];
 
+                // A malformed user-supplied header regex is a configuration mistake, not a crash.
+                // Real deflines are passed in because mzLib compiles the pattern without a match
+                // timeout, so only the user's own headers can show it is fast enough on them.
+                if (ok.Item2.CommonParameters?.FastaHeaderParsing is { } headerParsing
```
trishorts, line 93: **Two things about this call site that will be felt later, both cheap now.**

**1. The gate runs inside the task loop, so a config typo aborts after earlier tasks have already run.** This sits at `TaskList[i]`, evaluated immediately before task `i` executes. On a Calibrate → GPTMD → Search chain, a malformed custom regex on the Search task is not reported until calibration and GPTMD have finished — potentially hours — and the run then stops with `Cannot proceed`, leaving a half-populated output folder.

Most of what `Validate` checks is pure configuration: does the pattern compile, does it have a capture group, is an accession regex present for `Custom`. None of that depends on anything a previous task produced, and all of it is knowable before task 1 starts. Hoisting that half into a pre-flight pass over `TaskList` — the same place `PreRunTasks`-style checks already live — turns "three hours in, wrong regex" into "did not start, wrong regex".

The header-dependent half genuinely has to stay in the loop, because `CurrentXmlDbFilenameList` is reassigned from `myTaskResults.NewDatabases` and a GPTMD task changes the databases under later tasks. Worth splitting on that line rather than leaving the whole check late.

(Related: after a GPTMD the new database is `.xml`, which `SampleFastaHeaders` also filters out, so downstream tasks in a chain get an empty sample and the `Auto` gate passes vacuously — same root as my `.fasta.gz` comment.)

**2. `SampleFastaHeaders(...)` is an eager argument, so every FASTA is opened for every task even when the format cannot use the result.** C# evaluates the argument before the call, but `Validate` returns at

```csharp
if (HeaderFormat != FastaHeaderFormat.Custom)
    return true;
```

for `UniProt`, `Ensembl`, `Gencode` and `Ncbi` — without ever touching `sampleHeaders`. `UniProt` is the default, so on the overwhelmingly common path every database is opened and read once per task purely to build a list that is discarded. It is only five deflines, so it is not dramatic, but it scales with tasks × databases and it is entirely avoidable — take a `Func<List<string>>`, or check the format before sampling.

MetaMorpheus/GUI/Views/FastaHeaderParsingControl.xaml.cs
```diff
@@ -0,0 +1,79 @@
...
+            CustomGeneNameRegexTextBox.Text = parameters.CustomGeneNameRegex;
+            CustomOrganismRegexTextBox.Text = parameters.CustomOrganismRegex;
+            CustomOrganismIdRegexTextBox.Text = parameters.CustomOrganismIdRegex;
+
+            UpdateCustomEnabledState();
+        }
+
+        public FastaHeaderParsingParameters ToParameters()
+        {
+            var format = FastaHeaderFormatComboBox.SelectedItem is FastaHeaderFormat selected
```
trishorts, line 39: **Three silent fallbacks in this control, and the enum they fall back from is the thing this PR exists to make user-editable.**

```csharp
var format = FastaHeaderFormatComboBox.SelectedItem is FastaHeaderFormat selected
    ? selected
    : FastaHeaderFormat.UniProt;          // (1)
```

If `SelectedItem` is anything other than a `FastaHeaderFormat` — null before the combo is populated, or a `ComboBoxItem` wrapper if the XAML is ever changed to use one — the user's saved format is silently rewritten to `UniProt` on save. No warning, and the toml then genuinely says `UniProt`, so the evidence is gone by the time anyone looks.

Two more of the same shape below:

```csharp
if (CustomRegexGrid == null || PresetSummaryTextBlock == null)
    return;                               // (2) leaves the Custom grid in whatever state it had
...
_ => ""                                   // (3) DescribePreset, for any format not listed
```

(3) is the one that is certain to bite: `DescribePreset` enumerates all five current members, so the next member added to `FastaHeaderFormat` gets an empty preset summary in the GUI and nothing fails, anywhere. Given that my other thread argues `Ncbi` belongs in mzLib and that presets will keep being added, the enum growing is not hypothetical — it is the roadmap.

Cheapest durable fix for (1) and (3) is to make the absent case loud rather than plausible: throw or `Debug.Fail` on the unexpected `SelectedItem`, and give `DescribePreset` a `_ => throw new ArgumentOutOfRangeException(...)` so a new member fails at the point it is added instead of shipping a blank label. A silent `UniProt` and a silent `""` are both indistinguishable from working.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -46,6 +46,20 @@ public enum MyTask
 
     public abstract class MetaMorpheusTask
     {
+        /// <summary>
+        /// Nett hands a bare ArgumentException to the caller on an unrecognized enum name, which does
+        /// not say what the legal values are.
+        /// </summary>
+        private static FastaHeaderFormat ParseFastaHeaderFormat(string value)
+        {
+            if (Enum.TryParse<FastaHeaderFormat>(value, true, out var parsed))
```
trishorts, line 55: **Minor, but it defeats the message this method exists to give.** `Enum.TryParse` returns `true` for any numeric string, whether or not it names a defined member — so a toml carrying

```toml
FastaHeaderFormat = "7"
```

parses to `(FastaHeaderFormat)7` and returns `true` here, skipping the friendly "Valid values are: ..." error entirely.

It does not go unnoticed forever: `Validate` returns `true` for it (it is neither `Auto` nor `Custom`, so it takes the early `return true`), and it finally throws from `GetFieldRegexes()`'s `default:` arm — deep in the run, with `Unrecognized FASTA header format: 7`, which is the unhelpful shape this method was written to replace.

One line closes it:

```csharp
if (Enum.TryParse<FastaHeaderFormat>(value, true, out var parsed) && Enum.IsDefined(parsed))
    return parsed;
```

Worth having because tomls get hand-edited and copied between versions, and a member removed or reordered later turns an old toml's number into a different format rather than an error.

MetaMorpheus/Test/FastaHeaderParsingTest.cs
```diff
@@ -0,0 +1,367 @@
...
+            var readBack = Toml.ReadFile<SearchTask>(tomlPath, MetaMorpheusTask.tomlConfig);
+            Assert.That(readBack.CommonParameters.FastaHeaderParsing, Is.Not.Null);
+            Assert.That(readBack.CommonParameters.FastaHeaderParsing.HeaderFormat, Is.EqualTo(FastaHeaderFormat.UniProt));
+        }
+
+        [Test]
+        public static void PresetsNeedNoValidationAndAMalformedCustomRegexFailsIt()
+        {
+            foreach (FastaHeaderFormat format in Enum.GetValues<FastaHeaderFormat>().Where(f => f != FastaHeaderFormat.Custom))
+                Assert.That(new FastaHeaderParsingParameters(format).Validate(out _), Is.True, format.ToString());
```
trishorts, line 185: **This test currently pins the empty-sample behaviour as correct, which is the same behaviour I flagged as a bug on `EverythingRunnerEngine.cs:171`.**

The loop covers every format except `Custom` — so it includes `Auto` — and calls the two-argument overload, leaving `sampleHeaders` null. For `Auto`, `Validate` then iterates `sampleHeaders ?? []`, examines nothing, and returns `true`. The assertion passes for `Auto` precisely because no headers were supplied.

So the only test touching `Auto`'s validation asserts that validating with no headers succeeds. If the `.fasta.gz` gap gets fixed by making an unsampleable database an error rather than a pass, this test goes red — and it will read like the fix broke something rather than that the test encoded the gap.

Two changes make it say what it means:

- exclude `Auto` from the "presets need no validation" loop, since `Auto` is not a preset in this sense — its whole behaviour is header-dependent, and grouping it with the four static presets is what makes the vacuous pass look intentional;
- add a case that gives `Auto` real headers both ways: a recognizable defline (passes) and an unrecognizable one (fails with the "could not be detected automatically" message). Right now no test shows `Auto` ever rejecting anything, so the error path added at `FastaHeaderParsingParameters.cs:106-115` is entirely uncovered.

Issue #1992, opened 2020-11-23
Cannot create a spectral library from mzID generated by MM


yishai99:
Hi, when trying to import search results from MetaMorpheus into Skyline or trying to create a library in Skyline using the mzID files generated by MM, we get an error. The Skyline team says its a none standard mzID format. Can you look into it?
Thanks,

Yishai

trishorts:
Right now we're working on a spectral library writer in MetaMorpheus. We should at least have something to test out this week. we'll leave this open and let you know. It will be .msp format. I hope that is okay. I don't know how to right the dlib format.

I think Rob figured out how to do spectral libraries with MM output in skyline. I'll ask him and see if we can explain it. I think it probably has to do with the modifications if memory serves.

rmillikin:
I will copy+paste my response to an email from another user that had similar issues – can you try this solution and see if it works out for you? I will update this page https://github.com/smith-chem-wisc/MetaMorpheus/wiki/Import-Results-Into-Skyline w/ this info also.



> Thanks for your patience. I was able to get our .mzID file imported into Skyline (I’m using version 19.1) by following these two steps:
> 
> 1. Make sure the .mzML spectra file you performed the search with is in the same directory as the .mzID file. If you used MetaMorpheus’s calibration, you will need to look at the Calibration Task results directory and move the calibrated .mzML spectra files in that folder to the folder with your .mzIDs, or vice versa.
> 2. Skyline will let you try to open files labeled “.mzID” but it will crash upon trying to open them, saying that “.mzID” is not a recognized file type. Simply rename the file extension to “.mzid” (lower-case) and it will import successfully.
> 
> I will post about issue #2 on Skyline’s support forum when I get a chance. It might have more weight if you do it in addition to me, since you’re a user of both softwares :)
> 
[...]

PR #2783, opened 2026-08-31
chore(mzlib): bump the pin to 1.0.588

trishorts:
mzLib released **1.0.588**. This repository pinned `1.0.587`.

### Pin sites updated

- `MetaMorpheus/CMD/CMD.csproj`
- `MetaMorpheus/EngineLayer/EngineLayer.csproj`
- `MetaMorpheus/GUI/GUI.csproj`
- `MetaMorpheus/GuiFunctions/GuiFunctions.csproj`
- `MetaMorpheus/TaskLayer/TaskLayer.csproj`
- `MetaMorpheus/Test/Test.csproj`

### What to look at before merging

The checks on this pull request are the gate: they say whether this repository still
builds and still passes its own tests against the new package. What they cannot say is
whether mzLib **changed a value it reports**, because a re-baselined expectation is
indistinguishable from a correct one until someone reads the release notes.

A version bump here is only sometimes a version string. Of the four consumers bumped
for mzLib 1.0.585, one was; the others needed a framework migration, a namespace fix
after drifting nineteen releases, and thirteen test updates. **If CI is red, read it
as work to do rather than as a broken release.**

Release notes: https://github.com/smith-chem-wisc/mzLib/releases/tag/1.0.588

---
Opened by `.github/workflows/upstream-watch.yml`.

### DIFF
--- a/MetaMorpheus/CMD/CMD.csproj
+++ b/MetaMorpheus/CMD/CMD.csproj
@@ -24,7 +24,7 @@
     <PackageReference Include="Microsoft.ML" Version="3.0.1" />
     <PackageReference Include="Microsoft.ML.CpuMath" Version="3.0.1" />
     <PackageReference Include="Microsoft.ML.FastTree" Version="3.0.1" />
-    <PackageReference Include="mzLib" Version="1.0.587" />
+    <PackageReference Include="mzLib" Version="1.0.588" />
     <PackageReference Include="Nett" Version="0.15.0" />
     <PackageReference Include="SQLite.Interop.dll" Version="1.0.103" />
     <PackageReference Include="System.Data.SQLite.Core" Version="1.0.118" />
--- a/MetaMorpheus/EngineLayer/EngineLayer.csproj
+++ b/MetaMorpheus/EngineLayer/EngineLayer.csproj
@@ -30,7 +30,7 @@
     <PackageReference Include="Microsoft.ML" Version="3.0.1" />
     <PackageReference Include="Microsoft.ML.CpuMath" Version="3.0.1" />
     <PackageReference Include="Microsoft.ML.FastTree" Version="3.0.1" />
-    <PackageReference Include="mzLib" Version="1.0.587" />
+    <PackageReference Include="mzLib" Version="1.0.588" />
     <PackageReference Include="Nett" Version="0.15.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.4" />
     <PackageReference Include="SQLite.Interop.dll" Version="1.0.103" />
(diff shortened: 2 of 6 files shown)


acesnik (approved):
🤖 Approving. The diff is exactly six version strings and nothing else, and your own description names the one thing the checks cannot answer — whether mzLib changed a value it reports — so I went and did that part rather than reading the green ticks.

1.0.588's notes say no library behaviour changes, and the release is packaging rather than code: XML docs, the modification data files, and one Thermo assembly removed. Three things in it touch MetaMorpheus specifically, and two are absence claims *about this repository* made in mzLib's notes, which seemed worth checking rather than accepting.

**The `Test on M` rename holds.** `#1185` renames that protease cleavage modification to `Homoserine lactone N-terminal on M`, and the notes assert nothing here referenced the old name. `git grep "Test on M"` over `upstream/master` returns nothing. Confirmed.

**`mzLib.targets` is inert here.** The package now ships `build\mzLib.targets`, which NuGet imports into every consuming project automatically. It gates the ~6.8 MB of modification data on `MzLibIncludeModificationData`, defaults that to `false` itself, and MetaMorpheus never sets it — `git grep` returns nothing. So nothing is copied out. Worth knowing that the 6.8 MB is in the `.nupkg` regardless of the property; the opt-in controls only whether the files land beside the application.

**The `MassPrecisionEstimator` claim needs one qualification.** mzLib's notes say nothing referenced it, which is true of mzLib — but MetaMorpheus does name it, at `.github/workflows/InstallRunAndArtifact.yml:278`, in `$whitelistedMissing`. That is an allowed-to-be-absent list, so a stale entry is inert, and `Create Installer Artifact` and `Search with Installed` both passing here is the empirical version of that rather than my reading of the script.

It is stale now, though, and this PR is precisely the trigger its own comment names:

> This entry becomes dead and should be swept when MetaMorpheus moves to a post-#1156 mzLib.

1.0.588 is that mzLib. One precision while you are in there: the four EntityFramework entries in the same block carry the same instruction keyed to mzLib#1142, and #1142's merge commit is already an ancestor of 1.0.585 — so those four have been dead since an earlier bump, not this one. Five dead entries total, one of them newly dead. A follow-up rather than a blocker, since inert is inert.

**One thing worth stating out loud, because it is easy to assume otherwise.** This bump does not bring the mzIdentML root-namespace fix. mzLib#1171 is still open, so the `.mzID` MetaMorpheus emits continues to declare `http://psidev.info/psi/pi/mzIdentML/1.1.0` where the schema's target namespace is `.../1.1`, and continues to fail validation on that one point. Anyone reading "pin bumped to 1.0.588" as "mzID conformance resolved" would be wrong, and that matters because it is the live defect behind #1992.

nbollis (approved):

zhuoxinshi (approved):

PR #2520, opened 2025-06-13
Support for timsTOF .d data

Alexander-Sol:
This PR enables MetaMorpheus support for timsTOF .d data file. Calibration and quant aren't currently supported, and will be quietly disabled or skipped if the user tries to quantify/calibrate a timsTOF data file. 

Search, GPTMD, and MetaDraw will work and are supported

Issue #2192, opened 2022-10-19
quantification error due to the experimental design


AmirWIS:
Running a set of 21 samples, I got the following error: "Error reading experimental design file: Error: The experimental design did not contain the file(s): [_directory and file name_]. Skipping quantification."
As a result, MM generated the AllQuantifiedPproteinGroups, AllPSMs and AllPeptide files, but without intensity values. 
I tried processing the data on several computers, but got the same error message. 
Only when removing that sample from the set, the process has ended successfully. 
The data file is not corrupted. Using another data processing software, I did not have any issues. 
I am using MM version 0.0.320
The data file is QEP2_ZE_17202_TECPR_3_29092022.raw (line#19 in the experimental design)
The experimental design (screen capture) is attached.
Please let me know what you think and how to solve it.
Thanks,
Amir
![image](https://user-images.githubusercontent.com/93867179/196707068-2f8d92b6-9cc3-40e4-92ed-ef3b4ed2dca3.png)

trishorts:
can you search that one file by itself successfully in metamorpheus?

AmirWIS:
Yes. since it is one file, the search took only few minutes. It ended successfully.

acesnik:
🤖 Sorry this sat so long. Your answer to trishorts' question — that the file searched fine on its own — is what makes it diagnosable now, so here is what was happening.

One of your 21 files failed calibration. That part was fine and expected; MetaMorpheus copies such a file through uncalibrated and carries on. The bug was in the experimental design it then rewrote. Calibration kept a list of the files it could not calibrate, and in the version you were running it added a **bare filename with no extension** to that list, while the code that rewrote the design checked the list for a **full path with extension**. Those can never be equal, so the check was always false and every row of the new design was rewritten to point at `<name>-calib.mzML` — including the row for the one file where no calibrated file had been written. Quantification then looked for a file that did not exist, reported the one name you saw, and skipped.

That accounts for the shape of what you saw: exactly one file named, that same file being the one whose removal fixed the run, and a single-file search succeeding because a run with nothing to reconcile never hits the mismatch.

It was fixed as part of the timsTOF `.d` support work in #2520 — the title looks unrelated, but that PR reworked how calibration tracks files it cannot calibrate, and the list now stores full paths on both sides. First release carrying it is **1.1.0** (June 2025). You reported this on 0.0.320, which was already a year old at the time, so an upgrade is the fix here rather than any change to your data or design.

[...]

PR #2784, opened 2026-08-31
Make the rewritten experimental design name the files that were written

acesnik:
🤖 #2192 reported quantification skipping with "the experimental design did not contain the file(s)". Its own cause was fixed as part of the timsTOF work in #2520, and I closed it on that basis. While verifying, three further routes to the same symptom turned out to be live on master, and one of them is worse than the original: instead of naming a file that does not exist, it names the wrong file that does.

Calibration and averaging both rewrite the experimental design for the next task. The invariant is simple — every row must name the file the run loop actually wrote — and each task breaks it differently.

## Averaging never recorded a skipped file

The list is declared `unsuccessfulyAveragedFilePaths` at `SpectralAveragingTask.cs:45` (one `l`), passed to the design writer at `:125`, received under the correctly spelled parameter at `:136`, read via `Contains` at `:161` — and never added to. So a `.mgf`, `.d`, `.msalign` or Bruker file, which averaging does not support, got a design row naming an `-averaged.mzML` that was never written. Calibration records exactly this case at `CalibrationTask.cs:414`.

The spelling mismatch is probably why it lasted: grep the correct spelling and you find the parameter, not the field.

## Calibration paired by substring, and mispaired

`FirstOrDefault(fn => fn.Contains(originalFileName))` matches `run_1` against a stored path for `run_10`. Measured on a two-file design where `run_10` failed calibration:

```
old design:                     new design:
run_1.mzML    CondA             run_10.mzML   CondA
run_10.mzML   CondB             run_10.mzML   CondB
```

`run_1`'s row names `run_10.mzML` under `run_1`'s condition; `run_1-calib.mzML` appears nowhere despite having been written; and `run_10.mzML` appears twice under two conditions. That is silent mislabelling rather than the loud error #2192 produced, which makes it the one I would most want out of the tree.

Now compared on whole filenames. The list holds two forms — an original input path from `:414`, an output path from `:461` — so the filename without extension is what both have in common.

## Both design writers omitted `ToSafeOutputPath`

Their run loops apply it (`CalibrationTask.cs:89`, `SpectralAveragingTask.cs:63`); the design writers did not. A name the run loop trimmed to fit the path cap was named untrimmed in the design. Measured: the run loop wrote a 133-character name while the design named the 261-character one. Reachable through a long filename, and more plausibly through a deep output folder, since the cap is on the whole path.

## The symmetry worth noticing

Averaging pairs by exact equality, so it was never exposed to the substring problem. Calibration populates its list, so it was never exposed to the empty one. Each task had half of the correct implementation, which is why neither showed up as a pattern until both were read side by side.

## Output

Where these fire: a run that previously skipped quantification now quantifies, and a run that previously mispaired now pairs correctly. Otherwise unaffected — the trimming path differs only for names at the cap, and the substring path only for filenames that are prefixes of each other.

## Tests

`Test/ExperimentalDesignRewriteTests.cs` drives the private design writer by reflection, which is much cheaper than provoking a real calibration failure and tests the unit whose contract is actually at issue. `Test/AveragingTests.cs` covers the skip-list case end to end through the averaging task, and asserts the negative too — that no `-averaged.mzML` was written for the skipped file.

All three confirmed failing against the unfixed code. The prefix test reports `run_10.mzML, run_10.mzML`; the trimming test reports 133 against 261. Verified by reverting the production changes and rerunning, not by inspection.

`Test/Test.csproj` is `net*.0-windows` and cannot execute on macOS, so these ran through a throwaway net10.0 harness inside the repo tree, since deleted. Solution builds clean on `-c UbuntuMac`, and `Test.csproj` and `GUI.csproj` both compile with `EnableWindowsTargeting`. CI is the first real run.

## Deliberately not included

`Test/CalibrationTests.cs`'s `ExperimentalDesignCalibrationAndSearchWithOneCalibratibleAndOneNoncalibratible` is this area's own test and asserts only that `ExperimentalDesign.tsv` exists — under the comment "test to see if quantification ran correctly". `PostSearchAnalysisTask` copies that file at `:295` before the read at `:306` whose failure path is `Warn` + `return`, so it passes whether or not quantification ran, and it would have stayed green throughout #2192's life. Worth fixing, but #2773 and #2780 are both open against that file, so it belongs after they land.

### DIFF
--- a/MetaMorpheus/TaskLayer/AveragingTask/SpectralAveragingTask.cs
+++ b/MetaMorpheus/TaskLayer/AveragingTask/SpectralAveragingTask.cs
@@ -155,7 +155,9 @@ namespace TaskLayer
             {
                 var originalUnaveragedFilePath = unaveragedSpectraFile.FullFilePathWithExtension;
                 var originalUnaveragedFilenameWithoutExtension = GlobalVariables.GetFilenameWithoutExtension(originalUnaveragedFilePath);
-                string averagedFilePath = Path.Combine(outputFolder, originalUnaveragedFilenameWithoutExtension + AveragingSuffix + ".mzML");
+                // ToSafeOutputPath here too, or a name the run loop trimmed is named untrimmed.
+                string averagedFilePath = Path.Combine(outputFolder, originalUnaveragedFilenameWithoutExtension + AveragingSuffix + ".mzML")
+                    .ToSafeOutputPath(AveragingSuffix + ".mzML");
 
                 var averagedSpectraFile = new SpectraFileInfo(averagedFilePath,
                     unaveragedSpectraFile.Condition, unaveragedSpectraFile.BiologicalReplicate, unaveragedSpectraFile.TechnicalReplicate, unaveragedSpectraFile.Fraction);
--- a/MetaMorpheus/TaskLayer/AveragingTask/SpectralAveragingTask.cs
+++ b/MetaMorpheus/TaskLayer/AveragingTask/SpectralAveragingTask.cs
@@ -64,6 +64,8 @@ namespace TaskLayer
                 var originalFileExtension = GlobalVariables.GetFileExtension(originalUnaveragedFilepath);
                 if (originalFileExtension.Equals(".mgf", StringComparison.OrdinalIgnoreCase) || originalFileExtension.Equals(".d", StringComparison.OrdinalIgnoreCase) || originalFileExtension.Equals(".msalign", StringComparison.OrdinalIgnoreCase) || BrukerDataDirectory.IsInnerFileExtension(originalFileExtension))
                 {
+                    // record it, or the new experimental design names a -averaged.mzML that was never written
+                    unsuccessfulyAveragedFilePaths.Add(originalUnaveragedFilepath);
                     Warn("Averaging for " + originalFileExtension + " files is not supported.");
                     FinishedDataFile(originalUnaveragedFilepath, new List<string> { taskId, "Individual Spectra Files", originalUnaveragedFilepath });
                     ReportProgress(new ProgressEventArgs(100, "Done!", new List<string> { taskId, "Individual Spectra Files", originalUnaveragedFilepathWithoutExtenstion }));
--- a/MetaMorpheus/TaskLayer/CalibrationTask/CalibrationTask.cs
+++ b/MetaMorpheus/TaskLayer/CalibrationTask/CalibrationTask.cs
@@ -551,14 +551,18 @@ namespace TaskLayer
             foreach (SpectraFileInfo originalSpectraFile in oldExperDesign)
             {
                 var originalFileName = originalSpectraFile.FilenameWithoutExtension;
-                var k = unsuccessfullyCalibratedFilePaths.FirstOrDefault(fn => fn.Contains(originalFileName));
+                // Compare whole names, not substrings: "run_1" is a substring of "run_10", so a
+                // substring match hands run_1's design row the file belonging to run_10.
+                var k = unsuccessfullyCalibratedFilePaths.FirstOrDefault(
+                    fn => GlobalVariables.GetFilenameWithoutExtension(fn) == originalFileName);
                 if (k != null)
                 {
                     newExperDesign.Add(new SpectraFileInfo(k, originalSpectraFile.Condition, originalSpectraFile.BiologicalReplicate, originalSpectraFile.TechnicalReplicate, originalSpectraFile.Fraction));
                 }
                 else
                 {
-                    SpectraFileInfo calibratedSpectraFile = new(Path.Combine(outputFolder, originalFileName + CalibSuffix + ".mzML"),
+                    // ToSafeOutputPath here too, or a name the run loop trimmed is named untrimmed.
+                    SpectraFileInfo calibratedSpectraFile = new(Path.Combine(outputFolder, originalFileName + CalibSuffix + ".mzML").ToSafeOutputPath(CalibSuffix + ".mzML"),
                     originalSpectraFile.Condition, originalSpectraFile.BiologicalReplicate, originalSpectraFile.TechnicalReplicate, originalSpectraFile.Fraction);
                     newExperDesign.Add(calibratedSpectraFile);
                 }
--- a/MetaMorpheus/Test/AveragingTests.cs
+++ b/MetaMorpheus/Test/AveragingTests.cs
@@ -238,6 +238,57 @@ namespace Test
             Directory.Delete(testPath, true);
         }
 
+        /// <summary>
+        /// A file type the averaging task cannot average is skipped, and EverythingRunnerEngine hands
+        /// the next task its original path. Its experimental design row must keep that path rather than
+        /// naming a -averaged.mzML that was never written. Issue #2192.
+        /// </summary>
+        [Test]
+        public static void SkippedFileKeepsItsOriginalPathInTheNewExperimentalDesign()
+        {
+            string testPath = Path.Combine(TestFolder, "AveragingSkippedFileDesign");
+            string outputFolder = Path.Combine(testPath, "Task1-Average");
+            Directory.CreateDirectory(testPath);
+            Directory.CreateDirectory(outputFolder);
+
+            string averagable = Path.Combine(testPath, "sample1.mzML");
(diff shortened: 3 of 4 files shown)


Alexander-Sol (approved):

trishorts (commented):
Two genuine bugs, both of the same shape -- the rewritten experimental design naming a file that was never written -- and both fixed at the right level. Nothing blocking from me.

The substring-versus-whole-name bug is the better catch of the two, and the `run_1` / `run_10` test is exactly the right way to pin it: it fails on the old code for the stated reason rather than incidentally, and `Assert.That(names, Is.Unique, ...)` states the actual invariant (no design row may claim another row's file) rather than a golden string.

Reaching `WriteNewExperimentalDesignFile` by reflection is justified here and the comment says why. The `Assert.That(method, Is.Not.Null, "CalibrationTask.WriteNewExperimentalDesignFile was renamed")` guard is the detail that makes it acceptable -- without it a rename turns the test into a silent no-op, which is the usual reason to distrust reflection in tests.

Two small notes inline.

MetaMorpheus/TaskLayer/AveragingTask/SpectralAveragingTask.cs
```diff
@@ -64,6 +64,8 @@ protected override MyTaskResults RunSpecific(string OutputFolder, List<DbForTask
                 var originalFileExtension = GlobalVariables.GetFileExtension(originalUnaveragedFilepath);
                 if (originalFileExtension.Equals(".mgf", StringComparison.OrdinalIgnoreCase) || originalFileExtension.Equals(".d", StringComparison.OrdinalIgnoreCase) || originalFileExtension.Equals(".msalign", StringComparison.OrdinalIgnoreCase) || BrukerDataDirectory.IsInnerFileExtension(originalFileExtension))
                 {
+                    // record it, or the new experimental design names a -averaged.mzML that was never written
+                    unsuccessfulyAveragedFilePaths.Add(originalUnaveragedFilepath);
```
trishorts, line 68: This is the fix, and it is correct -- without recording the skipped file, `WriteNewExperimentalDesignFile` falls into the `else` branch and names a `-averaged.mzML` that was never written.

One thing worth checking: the surrounding condition tests `.mgf`, `.d`, `.msalign` and `BrukerDataDirectory.IsInnerFileExtension(...)`. `unsuccessfulyAveragedFilePaths` now has to be a complete record of *every* path that leaves this loop without producing an averaged file. If any later `continue`, exception handler, or early exit in the same loop can also skip a file, it needs the same line, or the same bug comes back through a different door.

The `AveragingTests` case covers the extension route specifically. If there is a failure route as well -- averaging attempted and throwing, rather than declined up front -- it would be worth a second case, since that is the one a user is more likely to hit on real data.

(Also, `unsuccessfulyAveragedFilePaths` is misspelled -- pre-existing, and not worth churning the diff for, but if it ever gets renamed this is the PR that made it load-bearing.)

MetaMorpheus/TaskLayer/CalibrationTask/CalibrationTask.cs
```diff
@@ -551,14 +551,18 @@ private static void WriteNewExperimentalDesignFile(string pathToOldExperDesign,
             foreach (SpectraFileInfo originalSpectraFile in oldExperDesign)
             {
                 var originalFileName = originalSpectraFile.FilenameWithoutExtension;
-                var k = unsuccessfullyCalibratedFilePaths.FirstOrDefault(fn => fn.Contains(originalFileName));
+                // Compare whole names, not substrings: "run_1" is a substring of "run_10", so a
+                // substring match hands run_1's design row the file belonging to run_10.
+                var k = unsuccessfullyCalibratedFilePaths.FirstOrDefault(
+                    fn => GlobalVariables.GetFilenameWithoutExtension(fn) == originalFileName);
```
trishorts, line 557: Good fix. `Contains` on a filename is the kind of thing that works on every test fixture and fails on real data, because real data is where someone has `run_1` through `run_12`.

Worth noting the fix's own assumption, since it is now the thing holding the invariant up: `GlobalVariables.GetFilenameWithoutExtension(fn) == originalFileName` requires that both sides strip extensions identically. `originalFileName` comes from `SpectraFileInfo.FilenameWithoutExtension` while the other side goes through `GlobalVariables.GetFilenameWithoutExtension`. For a double extension such as `.mzML.gz`, or for a `.d` directory, those two could disagree -- and if they do, the `FirstOrDefault` silently returns null and the row falls into the `else` branch, naming a calibrated file that does not exist. Which is precisely the failure mode this PR exists to remove, just reached from the other side.

A case with a `.d` or a gzipped input in `ExperimentalDesignRewriteTests` would close that off. `StringComparison.OrdinalIgnoreCase` is probably also wanted here, since the rest of the file-extension handling in this task is case-insensitive and Windows paths are.

MetaMorpheus/TaskLayer/AveragingTask/SpectralAveragingTask.cs
trishorts, line 68: **The other route out of this loop without an averaged file is still open, and it is the one real data hits.**

I raised this as a "worth checking" last round; I have now followed it and it is real. `SpectralAveragingTask.cs:94` catches everything `AverageSpectraFile` / `ExportAsMzML` can throw, `Warn`s, and falls through:

```csharp
catch (Exception e)
{
    Warn($"Averaging Failure! Could not average spectra for file {...} with exception {e.Message}");
}
myFileManager.DoneWithFile(originalUnaveragedFilepath);
...
FinishedWritingFile(averagedFilepath, ...);          // line 112
MyTaskResults.NewSpectra.Add(averagedFilepath);      // line 113
```

So when averaging throws: no `-averaged.mzML` is written, the file is **not** added to `unsuccessfulyAveragedFilePaths`, and `WriteNewExperimentalDesignFile` takes the `else` branch and names the file that was never written. That is #2192 again, reached through the failure door rather than the unsupported-extension door. The task also reports the missing file as written and adds it to `NewSpectra`, so the next task is handed a path that is not on disk.

The extension case this PR fixes is the one a user meets on their first Bruker run; the exception case is the one they meet on a bad file halfway through a real batch.

The same line inside the `catch` closes it:

```csharp
catch (Exception e)
{
    unsuccessfulyAveragedFilePaths.Add(originalUnaveragedFilepath);
    Warn($"Averaging Failure! ...");
    continue;
}
```

The `continue` matters as much as the `Add` -- without it, lines 112-113 still announce and register a file that does not exist.

`GlobalVariables.StopLoops` at line 54 is a third route: every file after the stop keeps its `-averaged.mzML` row. I would leave that one -- the user cancelled the run -- but it is worth a line of comment saying so, since the invariant this PR establishes is "every row names a file that was actually written".

Both changes in the PR are correct and I have said so in the other two threads. I am holding approval only because this is the same invariant, in the same method, one branch over.

trishorts (commented):
Two threads from last round still stand (both non-blocking). One new finding, verified this time rather than flagged: the exception path out of the averaging loop reintroduces #2192.

nbollis (approved):

trishorts:
A related route to the same symptom, for TMT, sits next to the code this PR changes. It may be in scope here or better as its own PR, your call.

**Calibration and averaging carry `ExperimentalDesign.tsv` forward, but not `TmtDesign.txt`.**
- `CalibrationTask.UpdateExperimentalDesignFile` (`CalibrationTask.cs:176-190`, `:534-567`) and the averaging equivalent only read and rewrite `GlobalVariables.ExperimentalDesignFileName`.
- TMT quantification looks for the design beside the files it is handed: `Directory.GetParent(Parameters.CurrentRawFileList.First())` + `TmtExperimentalDesignFileName` (`PostSearchAnalysisTask.cs:290-299`).
- After `EverythingRunnerEngine` swaps in the `-calib` / `-averaged` files (`:127-141`), that folder is the calibration task's output folder, which has no `TmtDesign.txt`.

So a Calibrate → Search TMT run hits `"No TmtDesign.txt found next to the spectra files ... not quantified per channel"` and returns without per-channel quantification, even though the user supplied a design.

Copying the file forward isn't enough on its own: its `File` column names the original paths, so the rows need the same rename to the written files that this PR makes reliable for `ExperimentalDesign.tsv`, including the skipped-file case.

We found it while working on SDRF output (#2816/#2817), which reads the same design and loses its channel rows in the same chain.

PR #2787, opened 2026-09-01
One seeding recipe for every user-editable custom file

trishorts:
🤖 MetaMorpheus ships seven files a user is expected to edit, and every one of them was handled differently. #2752 is what that costs: a `CopyToOutputDirectory` rule plus an installer component overwrote a user's edited custom monosaccharides on install, repair and upgrade. That was fixed for monosaccharides alone. This applies the same contract to the rest, and adds a test that a new custom file has to join.

## The recipe

Documented on `CustomDataFile`, and taken from the **custom-protease seeding**, which already did the right thing and is the model for the others:

1. **Never touch a file that exists.** Seeding is guarded by `!File.Exists`. An empty file counts as the user's — truncating one on purpose is a thing people do, and *"it was empty so I replaced it"* is how data gets lost.
2. **Seed a documented template**: the shipped sibling's comment banner plus its header row, and **no data rows**. The user opens a file that explains its own format and has nothing to delete first. `BannerAndHeaderFrom` derives it from the shipped file, so the template cannot drift from what it documents.
3. **The template ships inside an assembly.** Never a `<File>` in `Product.wxs`, never a `<None Update ... CopyToOutputDirectory>` rule. Those two are exactly what let a build or an installer overwrite the user's copy, and both were what #2752 removed.
4. **A failure to seed is reported**, naming the file, rather than swallowed.

## What changes, per file

| File | Before | After |
|---|---|---|
| `proteases_custom.tsv` | 30 lines of inline seeding | calls the helper |
| `rnase_custom.tsv` | 30 near-identical lines | calls the helper |
| `CustomCrosslinkers.tsv` | **never seeded** | seeded with a template |
| `CustomModifications.txt` | **never seeded** | seeded with a template |
| `RnaCustomModifications.txt` | **never seeded** | seeded with a template |
| `MonosaccharidesCustom.tsv` | already canonical | unchanged |
| `CustomAminoAcids.txt` | full A–Z dump | unchanged, documented as the deliberate exception |

The two digestion-agent blocks were near-identical copies of each other, and carried the mzLib resource names inline annotated `// private hard-coded string path in MzLib`. Those are now named constants.

The three that were never seeded required a user to know both the filename and the format before they could start. The **crosslinker template is the header row alone**, deliberately: `LoadCrosslinkers` skips only line 1 and parses everything after it, so a banner there would be read as a crosslinker.

`CustomAminoAcids.txt` is left as it is and the reason is written down. It seeds a full A–Z dump of the existing residues rather than a header because its purpose is *adjusting a residue that already exists*; a bare header would make the common case harder, not easier.

## Two things found on the way

**`LoadCrosslinkers` crashed on a hand-edited file.** A blank line or a `#` note in `CustomCrosslinkers.tsv` reached `ParseCrosslinkerFromString` and came back as an unhandled `IndexOutOfRangeException` during startup — before any window opened, with nothing naming the file. I hit this by appending a comment to the seeded file to check it would not be overwritten. Blank and comment lines are now skipped, and a genuinely short row throws a `MetaMorpheusException` naming the file, the line number and the line.

**mzLib's collision reports had no consumer.** `LoadAndMergeCustomProteases` and `LoadAndMergeCustomRnases` refuse to let a custom entry shadow a built-in and report it through `CustomDigestionAgentLoadResult.Skipped` *rather than throwing, specifically so the caller can tell the user* — mzLib documents that as a guarantee. Both results were discarded. Someone who named a custom protease `trypsin` got silence and a protease that was not theirs. Those now reach the user through `GlobalVariables.StartupWarnings`, surfaced beside `ErrorsReadingMods` by both the GUI and the CLI. It is a separate list because `ErrorsReadingMods`' name is a promise about what is in it.

## Verification

Against a real `DataDir`, not a mock:

- All seven custom files are seeded on a clean start.
- After appending a sentinel to each and restarting, **zero were rewritten**.
- Protein modification, RNA modification and crosslinker counts are **identical** whether the custom files are freshly seeded, user-edited, or deleted and reseeded — so the templates contribute nothing. (4325 / 35 / 6 in all three states, with 0 errors and 0 warnings.)

`CustomDataFileTests` pins the contract: an existing file is never rewritten, an empty one counts as the user's, the template is not even built when the file is present, a seeding failure names the file, the banner/header extraction drops every data row, the real shipped crosslinker file yields a one-line template, every known custom file is seeded by startup, and the crosslinker loader tolerates blank and comment lines while naming the file and line for a malformed row.

55 digestion-agent, custom-file and crosslink tests pass. The solution builds with 0 errors.

## Two tests changed, and why

`TestCustomProteaseCreationCatchIsHitWhenCustomPathIsDirectory` and its rnase twin asserted the exact text `"Error creating default custom protease file with error message:"`. The new message names the file that could not be created, which the old one did not. Both assertions now check for that instead — they are stronger than before, not weaker.

## Not in this PR

`RnaCustomModifications.txt`, `CustomModifications.txt` and `CustomCrosslinkers.tsv` live under `Mods\` and `Data\`, whose WiX components carry `<RemoveFolder On="both"/>` — removal on *install* as well as uninstall. MSI only removes an empty directory, so this is latent rather than a live bug, and none of those files has a `<File>` component that could overwrite it. But it is the same "the installer owns this folder, the user's file is in it" arrangement that became #2752, and it deserves its own issue rather than being bundled here.

https://claude.ai/code/session_01UbCbEbFSY76K9CnjeLZff2

### DIFF
--- a/MetaMorpheus/EngineLayer/CrosslinkSearch/Crosslinker.cs
+++ b/MetaMorpheus/EngineLayer/CrosslinkSearch/Crosslinker.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.IO;
 using MassSpectrometry;
 using System.Globalization;
@@ -75,6 +76,13 @@ namespace EngineLayer
             return cleaveDissociationTypes;
         }
 
+        /// <summary>
+        /// Columns in a crosslinker tsv row: Name, CrosslinkAminoAcid, CrosslinkerAminoAcid2, Cleavable,
+        /// DissociationType, CrosslinkerTotalMass, CrosslinkerShortMass, CrosslinkerLongMass, QuenchMassH2O,
+        /// QuenchMassNH2, QuenchMassTris.
+        /// </summary>
+        private const int ExpectedColumnCount = 11;
+
         public static IEnumerable<Crosslinker> LoadCrosslinkers(string CrosslinkerLocation)
         {
             using (StreamReader crosslinkers = new StreamReader(CrosslinkerLocation))
@@ -88,6 +96,20 @@ namespace EngineLayer
                     if (lineCount == 1)
                     {
                         continue;
+                    }
+
+                    // CustomCrosslinkers.tsv is edited by hand, so a blank line or a '#' note in it is a
+                    // thing that happens. Both used to reach ParseCrosslinkerFromString and come back as
+                    // an unhandled IndexOutOfRangeException during startup, before any window opened.
+                    if (string.IsNullOrWhiteSpace(line) || line.StartsWith("#", StringComparison.Ordinal))
+                    {
+                        continue;
+                    }
+
+                    if (line.Split('\t').Length < ExpectedColumnCount)
+                    {
+                        throw new MetaMorpheusException($"Line {lineCount} of {Path.GetFileName(CrosslinkerLocation)} has "
+                            + $"{line.Split('\t').Length} tab-separated column(s); {ExpectedColumnCount} are required. The line was: {line}");
                     }
 
                     yield return ParseCrosslinkerFromString(line);
--- a/MetaMorpheus/EngineLayer/GlobalVariables.cs
+++ b/MetaMorpheus/EngineLayer/GlobalVariables.cs
@@ -1,4 +1,4 @@
-global using obo = Omics.Modifications.IO.obo;
+﻿global using obo = Omics.Modifications.IO.obo;
 using Chemistry;
 using Easy.Common.Extensions;
 using EngineLayer.GlycoSearch;
@@ -46,6 +46,37 @@ namespace EngineLayer
 
         public static List<string> ErrorsReadingMods;
 
+        /// <summary>
+        /// Non-fatal things the user should know about that were noticed while starting up, and that are
+        /// not about reading a modification. Surfaced beside <see cref="ErrorsReadingMods"/> by both front
+        /// ends. Kept separate because that list's name is a promise about what is in it.
+        /// </summary>
+        public static List<string> StartupWarnings { get; private set; } = new List<string>();
+
+        // mzLib keeps these as private constants, so the names are repeated rather than referenced.
+        // They are the shipped files whose banner and header row seed the custom counterparts.
+        private const string EmbeddedProteasesResourceName = "Proteomics.ProteolyticDigestion.proteases.tsv";
+        private const string EmbeddedRnasesResourceName = "Transcriptomics.Digestion.rnases.tsv";
+
+        /// <summary>Template seeded into Mods\CustomModifications.txt and Mods\RnaCustomModifications.txt.</summary>
+        /// <remarks>
+        /// The first line is the title CustomModWindow writes when it creates the file itself, so the GUI's
+        /// append path and this template agree. The rest are '#' banner lines, which the modification format
+        /// treats as comments -- see the shipped Mods.txt, which opens the same way.
+        /// </remarks>
+        private static string CustomModificationsTemplate(string analyte) =>
+            "Custom Modifications" + Environment.NewLine +
+            "################################## " + analyte + " modifications you add are stored here." + Environment.NewLine +
+            "################################## Modifications added through the GUI are appended below this banner." + Environment.NewLine +
+            "################################## One entry per modification, terminated by a line containing only //" + Environment.NewLine +
+            "##################################   ID   <name>            the modification's name" + Environment.NewLine +
+            "##################################   TG   <residues>        target, e.g. S or T or Y" + Environment.NewLine +
+            "##################################   PP   <location>        Anywhere. / N-terminal. / C-terminal. / Peptide N-terminal. / Peptide C-terminal." + Environment.NewLine +
+            "##################################   CF   <formula>         chemical formula, e.g. H1 N1 O2" + Environment.NewLine +
+            "##################################   MM   <mass>            monoisotopic mass, if no formula is given" + Environment.NewLine +
+            "##################################   MT   <type>            the group it is listed under" + Environment.NewLine +
+            "################################## See Mods.txt in this folder for worked examples." + Environment.NewLine;
+
         // File locations
         public static string DataDir { get; private set; }
         public static string UserSpecifiedDataDir { get; set; }
@@ -85,6 +116,8 @@ namespace EngineLayer
             _InvalidAminoAcids = new char[] { 'X', 'B', 'J', 'Z', ':', '|', ';', '[', ']', '{', '}', '(', ')', '+', '-' };
             ExperimentalDesignFileName = "ExperimentalDesign.tsv";
             SeparationTypes = new List<string> { { "HPLC" }, { "CZE" } };
+
+            StartupWarnings = new List<string>();
 
             SetMetaMorpheusVersion();
             SetUpDataDirectory();
@@ -394,6 +427,13 @@ namespace EngineLayer
 
             // load custom crosslinkers
             string customCrosslinkerLocation = Path.Combine(DataDir, @"Data", @"CustomCrosslinkers.tsv");
+
+            // Header row only, with no banner: LoadCrosslinkers skips line 1 and parses every line after
+            // it, so a comment banner here would be read as a crosslinker and throw on its columns.
+            CustomDataFile.EnsureExists(customCrosslinkerLocation,
+                () => CustomDataFile.BannerAndHeaderFromFile(crosslinkerLocation, "Name\t"),
+                "custom crosslinker");
+
             if (File.Exists(customCrosslinkerLocation))
             {
                 AddCrosslinkers(Crosslinker.LoadCrosslinkers(customCrosslinkerLocation));
@@ -411,6 +451,11 @@ namespace EngineLayer
             PsiModDeserialized = Loaders.LoadPsiMod(Path.Combine(DataDir, @"Data", @"PSI-MOD.obo.xml"));
             var formalChargesDictionary = Loaders.GetFormalChargesDictionary(PsiModDeserialized);
             UniprotDeseralized = Loaders.LoadUniprot(Path.Combine(DataDir, @"Data", @"ptmlist.txt"), formalChargesDictionary).ToList();
+
+            // Seeded before the sweep below picks it up. The template is a title line plus a '#' banner,
+            // so it contributes no modifications until the user or the GUI adds one.
+            CustomDataFile.EnsureExists(Path.Combine(DataDir, @"Mods", "CustomModifications.txt"),
+                () => CustomModificationsTemplate("Protein"), "custom modification");
 
             foreach (var modFile in Directory.GetFiles(Path.Combine(DataDir, @"Mods")))
             {
@@ -456,6 +501,8 @@ namespace EngineLayer
             }
 
             var customModsPath = Path.Combine(DataDir, @"Mods", "RnaCustomModifications.txt");
+            CustomDataFile.EnsureExists(customModsPath,
+                () => CustomModificationsTemplate("RNA"), "custom RNA modification");
             if (File.Exists(customModsPath))
             {
                 AddMods(ModificationLoader.ReadModsFromFile(customModsPath, out var errorMods), false, true);
@@ -591,49 +638,30 @@ namespace EngineLayer
 
         private static void LoadDigestionAgents()
         {
+            // Seed first, then load: the template is header-only, so a freshly seeded file contributes
+            // nothing and the two steps are independent. See CustomDataFile for the recipe every custom
+            // file follows.
+            CustomDataFile.EnsureExists(CustomProteasePath,
+                () => CustomDataFile.BannerAndHeaderFrom(typeof(ProteaseDictionary).Assembly,
+                    EmbeddedProteasesResourceName, "Name\t"),
+                "custom protease");
+
+            CustomDataFile.EnsureExists(CustomRnasePath,
+                () => CustomDataFile.BannerAndHeaderFrom(typeof(RnaseDictionary).Assembly,
+                    EmbeddedRnasesResourceName, "Name\t"),
+                "custom rnase");
+
             if (File.Exists(CustomProteasePath))
             {
                 try
                 {
                     var mods = ProteaseDictionary.LoadEmbeddedProteaseMods();
                     var result = ProteaseDictionary.LoadAndMergeCustomProteases(CustomProteasePath, mods);
+                    ReportSkippedCustomEntries(result.Skipped, "protease", CustomProteasePath);
                 }
                 catch (Exception e)
                 {
                     throw new MetaMorpheusException($"Error loading custom proteases with error message: {e.Message}", e);
-                }
-            }
-            else
-            {
-                try
-                {
-                    var assembly = typeof(ProteaseDictionary).Assembly;
-
-                    // private hard-coded string path in MzLib
-                    string EmbeddedProteaseResourceName = "Proteomics.ProteolyticDigestion.proteases.tsv"; 
-
-                    var stream = assembly.GetManifestResourceStream(EmbeddedProteaseResourceName);
-                    var reader = new StreamReader(stream);
-
-                    string fileContent = reader.ReadToEnd();
-                    string[] lines = fileContent.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
-
-                    bool foundHeader = false;
-                    using var sw = new StreamWriter(File.Create(CustomProteasePath));
-                    foreach (string line in lines) 
-                    {
-                        if (!foundHeader && !line.StartsWith("#") && line.TrimStart().StartsWith("Name\t"))
-                        {
-                            sw.WriteLine(line);
-                            foundHeader = true;
-                            break;
-                        }
-                        sw.WriteLine(line);
-                    }
-                }
-                catch (Exception e)
-                {
-                    throw new MetaMorpheusException($"Error creating default custom protease file with error message: {e.Message}", e);
                 }
             }
 
@@ -642,45 +670,31 @@ namespace EngineLayer
                 try
                 {
                     var result = RnaseDictionary.LoadAndMergeCustomRnases(CustomRnasePath);
+                    ReportSkippedCustomEntries(result.Skipped, "rnase", CustomRnasePath);
                 }
                 catch (Exception e)
                 {
                     throw new MetaMorpheusException($"Error loading custom rnases with error message: {e.Message}", e);
                 }
             }
-            else
+        }
+
+        /// <summary>
+        /// mzLib refuses to let a custom digestion agent shadow one of its own, and reports the collision
+        /// through <c>CustomDigestionAgentLoadResult.Skipped</c> rather than throwing, specifically so the
+        /// caller can tell the user. Nothing consumed that before, so a user who named a custom protease
+        /// "trypsin" got silence and a protease that was not theirs.
+        /// </summary>
+        private static void ReportSkippedCustomEntries(IReadOnlyList<string> skipped, string kind, string path)
+        {
+            if (skipped == null || skipped.Count == 0)
             {
-                try
-                {
-                    var assembly = typeof(RnaseDictionary).Assembly;
-
-                    // private hard-coded string path in MzLib
-                    string EmbeddedProteaseResourceName = "Transcriptomics.Digestion.rnases.tsv";
-
-                    var stream = assembly.GetManifestResourceStream(EmbeddedProteaseResourceName);
-                    var reader = new StreamReader(stream);
-
-                    string fileContent = reader.ReadToEnd();
-                    string[] lines = fileContent.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
-
-                    bool foundHeader = false;
-                    using var sw = new StreamWriter(File.Create(CustomRnasePath));
-                    foreach (string line in lines)
-                    {
-                        if (!foundHeader && !line.StartsWith("#") && line.TrimStart().StartsWith("Name\t"))
-                        {
-                            sw.WriteLine(line);
-                            foundHeader = true;
-                            break;
-                        }
-                        sw.WriteLine(line);
-                    }
-                }
-                catch (Exception e)
-                {
-                    throw new MetaMorpheusException($"Error creating default custom rnase file with error message: {e.Message}", e);
-                }
+                return;
             }
+
+            StartupWarnings.Add($"{skipped.Count} custom {kind}(s) in {Path.GetFileName(path)} were ignored because "
+                + $"a built-in {kind} already uses the same name: {string.Join(", ", skipped.Select(p => "'" + p + "'"))}. "
+                + $"Rename them in {path} if you meant to define your own.");
         }
     }
 }
--- a/MetaMorpheus/EngineLayer/Util/CustomDataFile.cs
+++ b/MetaMorpheus/EngineLayer/Util/CustomDataFile.cs
@@ -1,0 +1,151 @@
+using System;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+
+namespace EngineLayer
+{
+    /// <summary>
+    /// The one place MetaMorpheus creates a data file that the USER is expected to edit.
+    ///
+    /// <para>
+    /// Every such file -- custom proteases, rnases, crosslinkers, modifications, RNA modifications,
+    /// monosaccharides -- follows the same recipe, and it exists because breaking it produced a
+    /// data-loss bug (#2752: a user's edited custom monosaccharides were overwritten on install,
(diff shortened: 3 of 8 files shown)


MetaMorpheus/EngineLayer/CrosslinkSearch/Crosslinker.cs
```diff
@@ -90,6 +98,20 @@ public static IEnumerable<Crosslinker> LoadCrosslinkers(string CrosslinkerLocati
                         continue;
                     }
 
+                    // CustomCrosslinkers.tsv is edited by hand, so a blank line or a '#' note in it is a
+                    // thing that happens. Both used to reach ParseCrosslinkerFromString and come back as
+                    // an unhandled IndexOutOfRangeException during startup, before any window opened.
+                    if (string.IsNullOrWhiteSpace(line) || line.StartsWith("#", StringComparison.Ordinal))
```
pcruzparri, line 104: **High** — Header is detected by position, so a comment or blank line at the top of `CustomCrosslinkers.tsv` crashes startup

**Problem:** The loop skips `lineCount == 1` unconditionally, which assumes the header is physically the first line. The new blank/`#` skip runs *after* that rule. So a leading comment — or a stray blank line — is consumed by the line-1 skip, and the header row shifts to line 2 where it is parsed as data. The header has exactly 11 tab-separated columns, so it passes the new `ExpectedColumnCount` guard and reaches `double.Parse(split[5])`.

**Impact:** `FormatException: The input string 'CrosslinkerTotalMass' was not in a correct format.`, thrown out of `GlobalVariables.SetUpGlobalVariables()`. `MainWindow()` calls that from its constructor before `InitializeComponent()`, and `App.xaml`/`App.xaml.cs` register no `DispatcherUnhandledException` handler — so the GUI dies with a raw .NET crash naming neither the file nor the line, before any window opens. That is the failure this PR sets out to remove, reached by the route this PR opens: the file is now seeded for every user, and the same change tells them blank and comment lines are tolerated.

**Suggested fix:** Detect the header by content rather than position — skip blank/comment lines first, then treat the first line matching `line.TrimStart().StartsWith("Name\t")` as the header wherever it lands, instead of `if (lineCount == 1)`.

**Affected test scenarios:** `LoadCrosslinkers_SkipsBlankAndCommentLines` asserts the new tolerance but places every blank and comment *below* the header, so it passes while the defect stands. Add the mirrored case — a `#` note and a blank line *above* the header row — plus one asserting `SetUpGlobalVariables()` survives a `CustomCrosslinkers.tsv` written that way.

MetaMorpheus/EngineLayer/Util/CustomDataFile.cs
```diff
@@ -0,0 +1,151 @@
...
+
+            using (var reader = new StreamReader(shipped))
+            {
+                string line;
+                while ((line = reader.ReadLine()) != null)
+                {
+                    bool isComment = line.StartsWith("#", StringComparison.Ordinal);
+                    bool isHeader = !isComment && line.TrimStart().StartsWith(headerPrefix, StringComparison.Ordinal);
+
+                    if (!isComment && !isHeader)
```
pcruzparri, line 112: **Medium** — `BannerAndHeaderFrom` can emit a headerless or empty template, and `EnsureExists` then seeds it permanently

**Problem:** The loop classifies anything that is neither a `#`-prefixed comment nor the header as a data row and breaks. Two shapes defeat it. A blank line inside the banner ends the scan before the header is reached. And because `isComment` uses `StartsWith("#")` while `isHeader` uses `TrimStart()`, an *indented* comment matches neither and breaks on the first line, returning `""`. `EnsureExists` then writes that with `File.WriteAllText`, which creates a 0-byte file without complaint (line 82) — and rule 1 (`!File.Exists`) protects it on every subsequent startup, so it is never repaired.

**Impact:** Latent against mzLib 1.0.588 — I read both embedded resources and neither has a blank line before its header (`proteases.tsv` header at line 98, `rnases.tsv` at line 84). But the templates are derived from a third-party package that is re-pinned regularly, and one cosmetic reformat upstream silently hands every new user a blank, undocumented custom file. That contradicts this class's own rule 4, "failing to seed is reported": nothing fails and nothing is reported.

**Suggested fix:** In `BannerAndHeaderFrom`, `continue` on `string.IsNullOrWhiteSpace(line)` and apply `TrimStart()` before the `#` test so the two classifications agree. In `EnsureExists`, refuse a null or whitespace-only template rather than writing it.

**Affected test scenarios:** `BannerAndHeader_KeepsCommentsAndHeader_AndDropsEveryDataRow` uses a clean banner and cannot reach either shape. Add a shipped-file fixture with a blank line inside the banner, and one with an indented comment, both asserting the header survives. The rnase path has no template test at all — mirror `TheProteaseTemplateComesFromMzLibsEmbeddedFile_AndCarriesNoProteases` for `Transcriptomics.Digestion.rnases.tsv`, since that resource name is a hard-coded copy of an mzLib private with nothing pinning it.

MetaMorpheus/Test/CustomFileCollisionTests.cs
```diff
@@ -0,0 +1,131 @@
...
+            _hadFile = File.Exists(GlobalVariables.CustomProteasePath);
+            if (_hadFile)
+            {
+                _backup = GlobalVariables.CustomProteasePath + ".bak";
+                File.Copy(GlobalVariables.CustomProteasePath, _backup, true);
+            }
+        }
+
+        [TearDown]
+        public static void TearDown()
```
pcruzparri, line 36: **Low-Medium** — TearDown cannot undo the custom protease, so the fixture is not idempotent and leaks into a process-wide static

**Problem:** `ProteaseDictionary.LoadAndMergeCustomProteases` merges into mzLib's static `ProteaseDictionary.Dictionary` and never resets it. TearDown restores the file and re-runs `SetUpGlobalVariables()`, which cannot remove an already-merged entry. Measured: `Dictionary.Count` goes 23 to 24 and `NotAShadowingName_ForTest` is still present after TearDown returns.

**Impact:** Two effects. `ACustomProteaseWithItsOwnNameIsNotReported` fails on a second run in the same process — mzLib now reports its own leftover as a built-in collision, so the "nothing to report" assertion trips. And every GUI protease dropdown built from `ProteaseDictionary.Dictionary.Values` carries the phantom protease for the rest of the run. No current test asserts a protease count, so nothing else breaks today, but `TestToml.TestBadFileSpecificProtease` already reserves an "arbitrary" name against exactly this class of assumption.

**Suggested fix:** Remove the key in TearDown (`ProteaseDictionary.Dictionary.Remove("NotAShadowingName_ForTest")`). Add `[NonParallelizable]` to this fixture and to `CustomDataFileTests` — this repo already marks state-mutating fixtures that way (`LoadDigestionAgentsTest`, `PsmTsvWriterTests`, `FdrTest`, `MyTaskTest`), and both new fixtures write to the real `DataDir` and re-run `SetUpGlobalVariables()`.

**Affected test scenarios:** Running this fixture twice in one process, and any future test that asserts on the protease dictionary's size or contents.

> Out of scope / non-blocking: the assembly has no `[assembly: Parallelizable]`, so NUnit runs it sequentially today. The `[NonParallelizable]` half is a convention gap, not a live flake; the non-idempotence is real now.

MetaMorpheus/Test/CustomDataFileTests.cs
```diff
@@ -0,0 +1,239 @@
...
+        /// <remarks>
+        /// Runs the startup itself rather than reading whatever the process happens to have lying around.
+        /// Test order is not fixed, and CustomAminoAcidsTest deletes CustomAminoAcids.txt when it finishes
+        /// ("Delete so it doesn't crash the next time") without seeding it again -- so asserting on ambient
+        /// state passes or fails depending on what ran first. Re-running SetUpGlobalVariables is what the
+        /// application does on launch, and is the pattern LoadDigestionAgentTest already uses to put the
+        /// globals back.
+        /// </remarks>
+        [Test]
+        public static void EveryKnownCustomFileIsSeededByStartup()
```
pcruzparri, line 173: **Low** — `EveryKnownCustomFileIsSeededByStartup` is not hermetic, and its remark overstates what it checks

**Problem:** The remark says it "runs the startup itself rather than reading whatever the process happens to have lying around". But `MySetUpClass.GlobalSetup` (`Test/SetUpTests.cs:24`) already ran `SetUpGlobalVariables()` for the whole assembly, and on a developer machine all seven files already exist in `Test\bin\Debug\net10.0-windows\` from earlier runs. The test's own `SetUpGlobalVariables()` call therefore asserts against state it did not create.

**Impact:** The test would pass with every `EnsureExists` call in the PR deleted, on any machine that has run the suite before. It still catches a regression on a clean CI checkout, so this is a weakness rather than a dead test — but it is the test the PR nominates as "the inventory that a new custom file has to join", so it is worth making it actually prove that.

**Suggested fix:** Delete the seven paths inside the test before calling `SetUpGlobalVariables()`, then assert they came back.

**Affected test scenarios:** Any local re-run; the test is only meaningful on a cold checkout as written.

MetaMorpheus/EngineLayer/Util/CustomDataFile.cs
```diff
@@ -0,0 +1,151 @@
+using System;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+
+namespace EngineLayer
+{
+    /// <summary>
+    /// The one place MetaMorpheus creates a data file that the USER is expected to edit.
```
pcruzparri, line 10: **Low** — The policy doc contradicts the code it documents in three places

**Problem:** This class is explicitly the written-down contract, so the wording matters more than usual.

1. "The one place MetaMorpheus creates a data file that the USER is expected to edit" — it is one of three; `GlycanDatabase.EnsureCustomMonosaccharideFileExists` and `GlobalVariables.WriteAminoAcidsFile` still do it, and monosaccharides are listed as following "the same recipe" while never touching this class.
2. Rule 2 says the template is derived from the shipped sibling "which keeps the two from drifting" — but `CustomModifications.txt` and `RnaCustomModifications.txt` use the hand-written `CustomModificationsTemplate` literal, which is derived from nothing and free to drift from `Mods.txt`, and `CustomCrosslinkers.tsv` gets a header with no banner. The doc names one exception; there are three.
3. Rule 3 says "The template ships inside an assembly, never as a file on disk" — but `BannerAndHeaderFromFile` exists solely to read the crosslinker template from `Data\Crosslinkers.tsv`, which is a `<File>` at `Product.wxs:699` and carries `CopyToOutputDirectory` at `EngineLayer.csproj:48`.

**Impact:** Documentation only — the intent holds in every case (no custom file is installer-managed, and I confirmed the append-a-sentinel-and-restart contract survives for all seven). But the next person to add a custom file will read these rules as literal and find they do not describe the code.

**Suggested fix:** Say "the one helper for seeding", enumerate the three exceptions rather than one, and restate rule 3 as the thing it actually guarantees — the *custom* file is never installer- or build-managed — rather than a claim about where templates live.

**Affected test scenarios:** None.

pcruzparri (commented):
Two-pass code review (independent audit of the first pass) — 1 finding at severity >= High, posted as inline comments.

RayMSMS (dismissed):

MetaMorpheus/EngineLayer/GlobalVariables.cs
```diff
@@ -46,6 +46,37 @@ public static class GlobalVariables
 
         public static List<string> ErrorsReadingMods;
 
+        /// <summary>
+        /// Non-fatal things the user should know about that were noticed while starting up, and that are
+        /// not about reading a modification. Surfaced beside <see cref="ErrorsReadingMods"/> by both front
+        /// ends. Kept separate because that list's name is a promise about what is in it.
+        /// </summary>
+        public static List<string> StartupWarnings { get; private set; } = new List<string>();
```
nbollis, line 54: Is this necessary? I understand the goal is to allow issues that normally go through the event handling system to be displayed on the UI, are now also displayed in the commandline. However, I believe the solution would be to alter the event endpoints to either send to the UI or the the CMD depending on the case. This adds more state based complexity to our already complex startup system.

MetaMorpheus/GUI/MainWindow.xaml.cs
```diff
@@ -1633,6 +1633,13 @@ private void PrintErrorsReadingMods()
                 NotificationHandler(null, new StringEventArgs(error, null));
             }
             GlobalVariables.ErrorsReadingMods.Clear();
+
+            // and anything else startup noticed, such as a custom protease that collided with a built-in
+            foreach (var warning in GlobalVariables.StartupWarnings)
```
nbollis, line 1638: The point of this is to surface errors to CMD that were previously GUI only. This implementation will miss any warnings that take place after startup.

MetaMorpheus/EngineLayer/GlobalVariables.cs
trishorts: Agreed, it's gone. `GlobalVariables` now raises a static `WarnHandler` event (same shape as `MyFileManager`'s), and CMD and GUI each route it to their own output. No list, no drain step. 4e98d18ac

MetaMorpheus/GUI/MainWindow.xaml.cs
trishorts: Fixed by the same change. Both front ends subscribe before `SetUpGlobalVariables` and stay subscribed, so warnings raised at startup and any raised later both arrive. The GUI queues the write, since the notifications box doesn't exist until `InitializeComponent` has run. 4e98d18ac

MetaMorpheus/EngineLayer/CrosslinkSearch/Crosslinker.cs
trishorts: Fixed in b3e815448. `LoadCrosslinkers` now treats the first line that isn't blank or a `#` note as the header, wherever it lands. A file that starts with its header row loads exactly as before. New tests put a note and a blank line above the header, through the loader and through `SetUpGlobalVariables()`.

MetaMorpheus/EngineLayer/Util/CustomDataFile.cs
trishorts: Fixed in b3e815448. `BannerAndHeaderFrom` keeps blank and indented `#` lines as banner, and throws when a data row comes before any header instead of returning a headerless template. `EnsureExists` refuses an empty or whitespace template, so nothing gets written and the next startup tries again. Added tests for both banner shapes, a missing header, an empty template, and the rnase template.

MetaMorpheus/Test/CustomFileCollisionTests.cs
trishorts: Fixed in b3e815448. TearDown now removes `NotAShadowingName_ForTest` from `ProteaseDictionary.Dictionary`, and both fixtures are `[NonParallelizable]`.

MetaMorpheus/Test/CustomDataFileTests.cs
trishorts: Fixed in b3e815448. The test backs up the seven files, deletes them, runs startup, asserts they came back, then restores the originals.

MetaMorpheus/EngineLayer/Util/CustomDataFile.cs
trishorts: Fixed in b3e815448. The doc now calls the class the seeding helper, and restates rule 3 as "the custom file is never installer- or build-managed". It also lists the exceptions: amino acids, monosaccharides, the hand-written mod banner, and the header-only crosslinker template.

RayMSMS (dismissed):

trishorts:
### Backward compatibility — do existing custom files get rewritten, and do they still work?

This came up in review, so here it is in one place. Short answer: **an existing file is never touched, never gains the new header, and still loads.**

#### Nothing rewrites an existing file

`CustomDataFile.EnsureExists` opens with `if (File.Exists(path)) { return; }`. No rewrite, no reformat, no migration, no header injection, no validation-on-startup. An empty file counts as the user's — truncating one on purpose is a thing people do, and "it was empty so I replaced it" is how #2752 happened. There is no migration path in this PR at all; `EnsureExists` is the only writer it adds.

Per file, for someone upgrading:

| File | Before | After, if they already have one |
|---|---|---|
| `proteases_custom.tsv`, `rnase_custom.tsv` | seeded at first launch by two near-identical inline blocks | untouched — keeps whatever banner the old code wrote |
| `CustomCrosslinkers.tsv`, `CustomModifications.txt`, `RnaCustomModifications.txt` | never seeded; you had to know the filename and the format | untouched. Only a user with *no* such file gets the new template |
| `MonosaccharidesCustom.tsv` | already had the pattern (the #2752 fix) | not touched by this PR at all |
| `CustomAminoAcids.txt` | full A–Z residue dump | deliberately left alone, documented as the exception |

The accepted cost: someone who has had MetaMorpheus for years keeps their banner-less `proteases_custom.tsv` and never sees the documented format. Their only route to the new template is deleting the file, which reseeds it. That is rule 1 working as intended — the alternative is touching a file a user may have edited, which is the bug this PR exists to prevent.

#### And they still load

**Only one parser changed in the whole PR** — `Crosslinker.LoadCrosslinkers`. The protease, rnase, modification and RNA-modification read paths are untouched; `LoadAndMergeCustomProteases` / `LoadAndMergeCustomRnases` / `ModificationLoader.ReadModsFromFile` are called exactly as before.

**The one new throw cannot reject a file that used to work.** `ParseCrosslinkerFromString` unconditionally indexes `split[10]`, so a row with fewer than 11 tab columns *already* died — with an unhandled `IndexOutOfRangeException` naming nothing, during startup, before any window opened. `ExpectedColumnCount = 11` is exactly max-index + 1, so the guard converts an existing crash into a message naming the file, the line number and the line. It does not tighten what is accepted. The blank-and-`#` skip above it *widens* acceptance: a hand-edited file that used to crash startup now loads.

**`LoadDigestionAgents` changed shape, not outcome.** It went from `if (exists) load; else create` to `seed; if (exists) load`. For a user who has the file, identical. For one who does not, the old code created it and did not read it that session; the new code reads the freshly seeded file — which is header-only, so it yields nothing. Same end state.

**The only thing an existing file can newly trigger** is a warning: a custom protease named `trypsin` now says it was skipped instead of silently losing to the built-in. mzLib was already skipping it and reporting it through `CustomDigestionAgentLoadResult.Skipped`; nothing consumed that. Behaviour unchanged, silence removed.

Backing it up:

- `EnsureExists_NeverRewritesAnExistingFile`, `EnsureExists_TreatsAnEmptyExistingFileAsTheUsers` — content preserved, including a zero-byte file.
- `EnsureExists_DoesNotBuildTheTemplateWhenTheFileExists` — the template builder is not even *invoked* when the file is there, so no path exists by which template text could reach an old file.
- `LoadCrosslinkers_SkipsBlankAndCommentLines`, `LoadCrosslinkers_NamesTheFileAndLineForAShortRow`.
- Verified against a real DataDir: a sentinel appended to all seven custom files, then restart — zero rewritten. Protein modification, RNA modification and crosslinker counts identical whether the files were freshly seeded, user-edited, or deleted and reseeded.

Scope note: this covers the readers **this** PR touches. #2791 adds comment tolerance to the glycan `.gdb` loaders on top of it, and that one *does* change how an existing `.gdb` is read — it starts tolerating `#` lines that previously threw.

pcruzparri (dismissed):

pcruzparri:
A few non-blocking things from a read through this branch. None of it blocks the merge.

**In this PR:**

- `CustomDataFile.EnsureExists` — `File.WriteAllText` truncates the destination before it streams, so an interrupted write (disk full being the realistic one) leaves a truncated file that rule 1 then protects forever. The `IsNullOrWhiteSpace` guard closes the empty-template route into that state but not this one; `.tmp` + `File.Move` would close it. Reasoned from the API, not reproduced.
- `ReportSkippedCustomEntries` — the message says a built-in owns the name, but mzLib documents `Skipped` as deliberately conflating "collides with a built-in", "loaded by an earlier call" and "added by an earlier file in the batch". Only bites if `SetUpGlobalVariables` is ever called twice in one process, which production doesn't do, so it's wording rather than a bug.

**Pre-existing, noting so they don't get re-found:**

- `LoadCrosslinkers` — `headerSeen` is a skip-once flag rather than a check that a header exists, so a headerless `CustomCrosslinkers.tsv` silently loses its first crosslinker. The old `lineCount == 1` skip did the same. One narrow variant does shift here: a headerless file starting with a `#` comment used to load every row and now loses the first (measured: old `[MyLinker1, MyLinker2]`, new `[MyLinker2]`). It needs a hand-written headerless file with a leading comment, and you now seed a header, so nobody new can land there — not asking for a change.
- `Cleavable` doesn't round-trip: `ToString(true)` writes `True`/`False`, `ParseCrosslinkerFromString` tests `split[3] == "F"`. A non-cleavable crosslinker saved through the GUI comes back cleavable on the next start. Untouched lines, and the shipped `Crosslinkers.tsv` uses `T`/`F`, so it's only custom files.

trishorts:
Correcting my comment above: I wrote it against `7cb5442dd`, not the current head `b3e815448`, so two of its claims are wrong. My apologies — @pcruzparri's comment is right and mine is the stale one.

1. **It describes `GlobalVariables.StartupWarnings`.** That list no longer exists; `4e98d18ac` replaced it with a static `WarnHandler` event that each front end subscribes to. The user-visible outcome is the same, the mechanism in my description is not.

2. **It claims the crosslinker change "only widens acceptance", and that is not true at the head.** The measured counterexample is the one in the comment below: a hand-written headerless `CustomCrosslinkers.tsv` that begins with a `#` note used to load every row, and now loses the first one — *silently*, which is worse than the crash it replaced. Under `lineCount == 1` the note was consumed by the position skip; under `headerSeen` the note is skipped as a note and the first real row becomes the header. My "nothing that loaded before fails after" does not survive that case.

What still holds, re-checked against `b3e815448`: nothing rewrites an existing file, the per-file table is accurate, `LoadCrosslinkers` is still the only parser the PR changes, and `LoadDigestionAgents` changed shape without changing outcome.

Fixes for that and for the other points are on the way in a follow-up commit.

trishorts:
@pcruzparri @nbollis @RayMSMS — heads up that I pushed `9bf2771cc` after your approvals, so the diff has moved. Here is exactly what changed and why, in case you want another look.

**1. Seeding is now atomic.** You were right that the `IsNullOrWhiteSpace` guard closes the empty-template route and not this one. `EnsureExists` writes a sibling `.tmp` and moves it into place, so the only step that creates the real path is a rename, and a failure cleans up after itself. The reason it is worth three lines rather than a note: rule 1 makes whatever sits at the destination permanent, so a partial file written there is never repaired.

**2. The collision message no longer names a cause mzLib did not give it.** Your reading of `Skipped` is right — I pulled mzLib's own doc to check, and it says the three cases are "intentionally not distinguished". I went slightly further than "only bites if `SetUpGlobalVariables` runs twice": that is exactly what the **test suite** does, and `ProteaseDictionary.Dictionary` is process-wide static, so on a second run every one of the user's own proteases comes back skipped and the old wording would have told a developer a built-in owned their names. The message now says the name was already taken and that the definition already loaded wins, and the reasoning is in a `<remarks>` so the next person does not "fix" it back.

**3. The headerless case — I took this one further than you asked, because you are right that it is a silent loss.** Your measurement reproduces exactly: I reverted the rule locally and got `[MyLinker2]` where the old code gave `[MyLinker1, MyLinker2]`. The header is now the first line that actually starts with `Name\t`, rather than the first line that is not a note. That keeps every row of a headerless file *and* fixes the pre-existing variant you noted — the headerless file with no comment, which lost its first row under `lineCount == 1` too. So the skip-once flag is gone rather than documented.

That change opened one more route I had to close with it: a header this loader does **not** recognise now reaches `ParseCrosslinkerFromString`, and a row with eleven columns but an unreadable mass always did — both die in `double.Parse` with a `FormatException` naming neither file nor line, out of `SetUpGlobalVariables` before any window opens. Parsing is now wrapped to report a `MetaMorpheusException` naming the file, the line number and the line, which is the contract the other custom-file readers already follow. That is the last raw-crash route I can find in this path.

**4. `Cleavable` not round-tripping** is real and live — `CustomCrosslinkerWindow` writes `ToString(true)`, which emits `True`/`False`, and the parser tests `== "F"`. Filed as #2831 along with a second one I hit while confirming it: that same save does `File.Delete` then `File.WriteAllLines` over the whole file, so an interrupted save loses every custom crosslinker the user had. Both pre-existing, both out of scope here.

Three tests came with the commit — the headerless file keeps both crosslinkers (verified to fail against the previous rule, with your exact symptom), an unparsable row names the file and line, and a seed that cannot finish leaves no `.tmp` behind. 99 custom-file, crosslink, digestion-agent and toml tests pass locally, 0 build errors.

I have also merged this branch forward into #2791 and #2792, which needed it anyway — they were still sitting on the pre-`WarnHandler` version of this PR and had not seen the glycan box-mass cap from master.

Separately: my backward-compatibility comment above was written against the stale branch and I have corrected it in-thread. The correction is the comment directly above this one.

pcruzparri (approved):

nbollis (approved):

Issue #1204, opened 2018-06-05
add jenkins test for commandline version


trishorts:
just to be safe. not sure we're testing this continuously since we don't really use it in-house

rmillikin:
Jenkins tests the windows command line version. We could add a test for the .NET core one somehow. If Jenkins exists for Linux we could test that. Not sure how to test the GUI automatically except with an autoit script

PR #2790, opened 2026-09-03
Add --auditSdrf, so a user can see what an SDRF says about quantification

trishorts:
Adds `--auditSdrf`, a read-only command that prints what one SDRF document says about
quantification and runs nothing else. All the reading is mzLib's `SdrfQuantAuditor`, which shipped in
1.0.589; this is the surface that lets a user run it without writing code.

## Why

The failure this catches is silent. An SDRF whose `comment[label]` names the reagent kit —
`TMT10plex` on every row — rather than the channel parses cleanly, validates cleanly, and then yields
**one channel for a ten-channel experiment**. So does a file with two rows per (data file, channel),
or one that declares no isobaric modification at all. None of that is a structural defect, so the
SDRF validator is right not to report it: it is a defect in what the annotation *means*. Today the
only way a user finds out is a wrong number downstream.

Deliberately scoped to every SDRF rather than to TMT — "this is not an isobaric experiment" is a
useful answer to get.

## Using it

```
CMD.exe --auditSdrf <sdrf file> [--auditData <folder of downloaded files>]
```

```
SDRF quantification audit: demo.sdrf.tsv
  design            : ChannelLevel
  channels          : 4 (TMT126, TMT127N, TMT127C, TMT128N)
  plex              : Column -- characteristics[biological replicate batch] -> 1
  isobaric mod      : NOT DECLARED
  label form        : 4 bare, 0 accessioned
  data files        : 1 found, 0 missing
  facts:
     1. data file               ok
     2. channel                 ok
     3. reporter m/z           ABSENT  SDRF has no column for it; it is derived from the channel
     4. sample                  ok
     5. plex                    ok
     6. condition              ABSENT  no factor value column
     ...
```

and the case it exists for:

```
  design            : KitOnly
  kit               : TMT6PLEX -- channels are never enumerated
```

## Shape

It follows `-g` and `--acceptThermoLicence`: an option on `CommandLineSettings` and an early return
in `Run`, not a `CommandLine` verb. Verbs would mean every existing invocation had to grow one, and
this is not worth breaking every command line anyone has written.

- **Runs on its own.** Validation returns early for it, so no task, database, spectra file or output
  folder is required for a command that reads one text file and prints a report.
- **Its inputs are validated up front**, so a mistyped path comes back as a settings error (exit 2)
  before any setup happens, rather than as a read failure later.
- **`--auditData` without `--auditSdrf` is refused** rather than ignored. Silently dropping it looks
  identical to honouring it.
- **The report prints at `-v none`.** `-v` suppresses MetaMorpheus talking about itself; the report
  is the artifact that was asked for, and an audit that produced it silently would have produced
  nothing. Error messages stay verbosity-gated like every other error in `Program.cs`.
- **Exit codes** are the existing ones: 0 with a report, 2 for a settings error, 4 for a document
  that cannot be read.

`Readers` is aliased rather than imported, because `using Readers` collides with
`IO.ThermoRawFileReader` over `ThermoRawFileReaderLicence` and CS0104 is not worth a namespace.

## Tests

Nine, in `TestCmd.cs`, covering the verb rather than the parsing — mzLib owns the parsing and tests
it there. Channel-level and kit-only designs, data files counted found and missing, the report
surviving `-v none`, exit 4 on an unreadable document, and the four validation rules.

Two of them pin behaviour that surprised me while writing this, so the next reader does not take
either for a bug:

- **A file that is not an SDRF does not fail.** The auditor's contract is to describe a document, not
  reject one, so a text file with no SDRF columns comes back as a report saying it is not isobaric
  and every fact is absent. That is the honest answer and it is how a user learns they pointed at the
  wrong file.
- **`[Option(Default = ...)]` only applies when the parser builds the object.** An object initializer
  leaves `Verbosity` at the enum default, which is `none` — worth knowing before writing any test
  that expects a message.

Checked by running the built CLI, not only by unit test: the report above is real output, and the
exit codes were confirmed at the shell (2 for a missing file, 2 for `--auditData` alone, 0 with a
report).

## The mzLib pin

**Not here any more.** This PR carried the 1.0.588 → 1.0.589 bump when it was opened, alongside
#2789's identical edit to the same six lines. #2801 has since put 1.0.589 on master, so both
collapsed to no-ops and this PR is now three files: the verb, its settings, and its tests.
`SdrfQuantAuditor` comes from the pin master already has.

## Current with master

Full suite **1989 passed / 0 failed / 1 skipped**, re-run after #2789 landed. That run is worth more
than it was: master picked up #2775's rewrite of `CMD/Program.cs` after this branch was opened, and
this verb edits the same file, so this is the first time the two have been exercised together.

---

https://claude.ai/code/session_01HzqSuQ9kjYT2XBCubqNQtM

### DIFF
--- a/MetaMorpheus/CMD/CommandLineSettings.cs
+++ b/MetaMorpheus/CMD/CommandLineSettings.cs
@@ -43,6 +43,12 @@ namespace MetaMorpheusCommandLine
         [Option("acceptThermoLicence", HelpText = "[Optional] Agree to the Thermo RawFileReader licence, which is required to read .raw files, and record the agreement. Prints the licence and does not prompt, so it can be used where no console is available to answer one. May be given on its own as a setup step, or alongside a run.")]
         public bool AcceptThermoLicence { get; set; }
 
+        [Option("auditSdrf", HelpText = "[Optional] Path to an SDRF file to audit for what it says about quantification. Read-only: prints a report and exits without running anything. Reports whether the design is channel-level, kit-only or not isobaric at all; the channels found and the plex they imply; where the plex came from; whether an isobaric modification is declared; and which of the eleven per-channel facts are present, absent or unparseable. Given on its own, so no task, database or spectra file is required.")]
+        public string AuditSdrf { get; set; }
+
+        [Option("auditData", HelpText = "[Optional] Folder of downloaded data files to check the SDRF's comment[data file] entries against, reported as found or missing. Only meaningful with --auditSdrf.")]
+        public string AuditDataDirectory { get; set; }
+
         public enum VerbosityType { none, minimal, normal };
 
         public void ValidateCommandLineSettings()
@@ -59,6 +65,32 @@ namespace MetaMorpheusCommandLine
             if (GenerateDefaultTomls || RunMicroVignette)
             {
                 return;
+            }
+
+            // --auditSdrf reads one file and prints what it found. It runs nothing, writes nothing and
+            // needs no output folder, so holding it to a run's requirements would only stop it working.
+            // Its own inputs are checked here rather than at the point of use, so a typo comes back as
+            // a settings error before any setup happens.
+            if (AuditSdrf != null)
+            {
+                if (!File.Exists(AuditSdrf))
+                {
+                    throw new MetaMorpheusException("The SDRF file to audit was not found: " + AuditSdrf);
+                }
+
+                if (AuditDataDirectory != null && !Directory.Exists(AuditDataDirectory))
+                {
+                    throw new MetaMorpheusException("The data folder to audit against was not found: " + AuditDataDirectory);
+                }
+
+                return;
+            }
+
+            // --auditData without --auditSdrf says to check data files against nothing. Rejected rather
+            // than ignored, because silently dropping it looks identical to auditing with it.
+            if (AuditDataDirectory != null)
+            {
+                throw new MetaMorpheusException("--auditData is only meaningful with --auditSdrf.");
             }
 
             // --acceptThermoLicence on its own is a setup step - record the agreement and stop - so it
--- a/MetaMorpheus/CMD/CommandLineSettings.cs
+++ b/MetaMorpheus/CMD/CommandLineSettings.cs
@@ -82,6 +82,16 @@ namespace MetaMorpheusCommandLine
                 }
             }
 
+            // --auditData without --auditSdrf says to check data files against nothing. Rejected rather
+            // than ignored, because silently dropping it looks identical to auditing with it. Checked
+            // here, above the early return below, so that it holds for -g and --test too: those return
+            // before any later rule runs, so a guard placed after them refuses the bare flag and lets
+            // the same flag through unmentioned beside them.
+            if (AuditDataDirectory != null && AuditSdrf == null)
+            {
+                throw new MetaMorpheusException("--auditData is only meaningful with --auditSdrf.");
+            }
+
             if ((GenerateDefaultTomls || RunMicroVignette) && OutputFolder == null)
             {
                 throw new MetaMorpheusException("An output path must be specified with the -o parameter.");
@@ -109,13 +119,6 @@ namespace MetaMorpheusCommandLine
                 }
 
                 return;
-            }
-
-            // --auditData without --auditSdrf says to check data files against nothing. Rejected rather
-            // than ignored, because silently dropping it looks identical to auditing with it.
-            if (AuditDataDirectory != null)
-            {
-                throw new MetaMorpheusException("--auditData is only meaningful with --auditSdrf.");
             }
 
             // --acceptThermoLicence on its own is a setup step - record the agreement and stop - so it
--- a/MetaMorpheus/CMD/CommandLineSettings.cs
+++ b/MetaMorpheus/CMD/CommandLineSettings.cs
@@ -43,7 +43,7 @@ namespace MetaMorpheusCommandLine
         [Option("acceptThermoLicence", HelpText = "[Optional] Agree to the Thermo RawFileReader licence, which is required to read .raw files, and record the agreement. Prints the licence and does not prompt, so it can be used where no console is available to answer one. May be given on its own as a setup step, or alongside a run.")]
         public bool AcceptThermoLicence { get; set; }
 
-        [Option("auditSdrf", HelpText = "[Optional] Path to an SDRF file to audit for what it says about quantification. Read-only: prints a report and exits without running anything. Reports whether the design is channel-level, kit-only or not isobaric at all; the channels found and the plex they imply; where the plex came from; whether an isobaric modification is declared; and which of the eleven per-channel facts are present, absent or unparseable. Given on its own, so no task, database or spectra file is required.")]
+        [Option("auditSdrf", HelpText = "[Optional] Path to an SDRF file to audit for what it says about quantification. Read-only: prints a report and exits without running anything. Reports whether the design is channel-level, kit-only or not isobaric at all; the channels found and the plex they imply; where the plex came from; whether an isobaric modification is declared; and which of the eleven per-channel facts are present, absent or unparseable. Runs on its own: no task, database or spectra file is required, and it cannot be given alongside a run.")]
         public string AuditSdrf { get; set; }
 
         [Option("auditData", HelpText = "[Optional] Folder of downloaded data files to check the SDRF's comment[data file] entries against, reported as found or missing. Only meaningful with --auditSdrf.")]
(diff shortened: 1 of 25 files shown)


nbollis (dismissed):

acesnik (commented):
🤖 What this catches is genuinely invisible today. An SDRF with `TMT10plex` in `comment[label]` on every row is structurally perfect and semantically a ten-channel experiment collapsed to one, and nothing between the validator and the quant table has any reason to mention it — so `KitOnly` printed beside "channels are never enumerated" is the whole value of the verb, and scoping it to every SDRF rather than to TMT is right for the same reason. The up-front validation earns more than the body claims for it, too: mzLib sets `SearchedForDataFiles` from `dataDirectory != null` and returns empty match lists when the folder is absent (`SdrfQuantAudit.cs:405`, `:552`), so without your `Directory.Exists` check at `CommandLineSettings.cs:81` a typo'd `--auditData` would print a confident `0 found, 0 missing`.

Which is what makes the one problem worth a push rather than a note: `--auditSdrf` alongside anything else accepts both flags and honours one, at exit 0. Built this branch and ran it —

```
-t SearchTask.toml -d SmallYeast.fasta -s SmallCalibratible_Yeast.mzML -o out
    → 11 files, 41 proteins at 1% FDR, exit 0
... same argv plus --auditSdrf minimal.sdrf.tsv
    → 0 files, report only, exit 0
```

The search is discarded without a word. `ValidateCommandLineSettings()` returns at `CommandLineSettings.cs:86` with `Tasks`, `Databases` and `Spectra` still populated — I checked their contents at that point, they are not cleared — and `Program.cs:128` then returns the audit's exit code before any task setup.

It is not one combination, and the precedence flips between them. `-g -o F --auditSdrf S` writes six tomls and prints no report; `--test -o F --auditSdrf S` prints the report and leaves `F` empty where the vignette writes 26 files; `--acceptThermoLicence --auditSdrf S` prints the report and leaves `UserHasAgreedToThermoRawFileReaderLicence` at `false`. So `-g` beats the audit, and the audit beats `--test` and `--acceptThermoLicence` — there is no one rule to learn. `-g` also wins early enough to skip the audit's own path check: `-g -o F --auditSdrf /nonexistent` exits 0 with no error, because validation returns for `-g` at `:65`, before your `File.Exists` at `:76`.

Your own comment two lines under that early return is the argument for fixing it:

> `// --auditData without --auditSdrf says to check data files against nothing. Rejected rather`
> `// than ignored, because silently dropping it looks identical to auditing with it.`

That holds identically for `-t/-d/-s`, `-g`, `--test` and `--acceptThermoLicence`, and it is stronger for a run than for `--auditData`: exit 0 with an empty output folder is indistinguishable from success to any wrapper checking `$?`. Refusing the combination is the smaller change and matches the rule you already wrote; auditing and then continuing into the run is the other defensible answer, and `README.md:58` has `--acceptThermoLicence` doing exactly that. Either beats the current third answer. None of the ten added tests exercises a combined invocation — `TestAuditSdrfNeedsNoTaskDatabaseOrSpectra` pins only the opposite direction — so whichever you pick, that is where it goes.

Two small things. `AuditSdrf` has picked up `AgreeToThermoLicence`'s doc comment: `Program.cs:491-497` still describes printing the licence and "the data directory and the output writer are parameters", your own `<summary>` then opens at `:498`, so the method carries two and `AgreeToThermoLicence` at `:542` has none. Nothing ships from it — `GenerateDocumentationFile` is set nowhere in this repo and CMD is an `Exe` — so this is the tooltip and the next reader, not the `Readers.xml` problem from mzLib #1204. And `README.md:48` gave `--acceptThermoLicence` a section when it landed, while `--auditSdrf` currently exists only in `--help`.

Not verified by me: the test project needs `Microsoft.WindowsDesktop.App` and will not execute here, so your 1989 green is CI's to confirm. Everything above is a Release build of `CMD.csproj` driven from the shell on this branch.

trishorts:
🤖 All three addressed in `c20ca56fa`.

**The combination.** Took the refusal rather than audit-then-continue — it is the smaller change, and it is the rule already written two lines under that early return for `--auditData`. The check now runs at the top of `ValidateCommandLineSettings()`, before every other rule, so it also holds for the flags that return early: `-g` no longer wins ahead of the audit's own `File.Exists`.

```
-t … -d … -s … -o out --auditSdrf demo.sdrf.tsv
    → exit 2: "--auditSdrf reads one file and runs nothing else, so it cannot be given with -t, -d, -s. Run the audit on its own."
-g -o out --auditSdrf <nonexistent>        → exit 2, names -g                     (was: exit 0, six tomls, silence)
--test -o out --auditSdrf demo.sdrf.tsv    → exit 2, names --test                 (was: report, empty folder)
--acceptThermoLicence --auditSdrf …        → exit 2, names --acceptThermoLicence  (was: report, licence unrecorded)
```

`-o`, `-v` and `--mmsettings` stay allowed. They modify a run rather than asking for one, and the audit simply does not use them — pinned by a test so the next reader does not take it for an oversight.

**Tests.** Five in `TestCmd.cs`, where you said they would go: one per refused combination, plus the `-o` acceptance. `TestAuditSdrfNeedsNoTaskDatabaseOrSpectra` still pins the opposite direction, unchanged. The four invocations above are shell output from a build of this branch, not only the unit tests.

**Doc comment.** `AgreeToThermoLicence`'s `<summary>` is back on it; `AuditSdrf` no longer carries two.

**README.** `--auditSdrf` now has a section beside the `--acceptThermoLicence` one — why the failure is invisible, both invocations, the report shape, the `KitOnly` case, the exit codes. The `data files` line only appears in the `--auditData` example, since a run without that flag does not print it.

Thanks for building it and running the combinations — the `-g` precedence and the silent skip of the path check were both news to me.

trishorts:
@acesnik ready for another look when you have a minute — `c20ca56fa` has all three, and the combination now refuses rather than picking a winner.

MetaMorpheus/CMD/CommandLineSettings.cs
```diff
@@ -61,6 +92,32 @@ public void ValidateCommandLineSettings()
...
+                {
+                    throw new MetaMorpheusException("The data folder to audit against was not found: " + AuditDataDirectory);
+                }
+
+                return;
+            }
+
+            // --auditData without --auditSdrf says to check data files against nothing. Rejected rather
+            // than ignored, because silently dropping it looks identical to auditing with it.
+            if (AuditDataDirectory != null)
```
pcruzparri: **Non-blocking** — `--auditData` is silently ignored with `-g` and `--test`

**Problem:** This guard sits below the `if (GenerateDefaultTomls || RunMicroVignette) return;` at `:90`, so it never runs for those two flags. The `--auditSdrf` guard at `:67` was deliberately hoisted above that early return — its comment at `:65` says *"Checked before every other rule so that it holds even for the flags that return early"* — but this guard, the one whose reasoning that comment borrows, stayed below it.

**Impact:** `--auditData` is accepted and dropped without a word, which is the failure the comment at `:114-115` says it is refused to prevent. It contradicts the help text at `:49` ("Only meaningful with `--auditSdrf`") and the README's "`--auditData` without `--auditSdrf` is refused". Reproduced against a Release build of c20ca56:

```
dotnet CMD.dll --auditData <dir> -g -o <out>              -> exit 0, six task tomls written
dotnet CMD.dll --auditData <dir> --test -o <out> -v none  -> exit 0, 26 files written
dotnet CMD.dll --auditData <dir>                          -> exit 2  (correct)
```

**Suggested fix:** Move the guard above the `-o` requirement at `:85` and condition it on `AuditSdrf == null`, so it holds for the flags that return early:

```csharp
if (AuditDataDirectory != null && AuditSdrf == null)
{
    throw new MetaMorpheusException("--auditData is only meaningful with --auditSdrf.");
}
```

I ran this: both invocations above become exit 2 with nothing written, `-g` on its own still writes its six tomls, and the `TestCmd` fixture stays at 36 passed / 1 skipped.

**Affected test scenarios:** `TestAuditDataWithoutAuditSdrfIsRefused` pins the bare case only. Two beside it would cover the gap — `--auditData` with `GenerateDefaultTomls` + `OutputFolder`, and `--auditData` with `RunMicroVignette` + `OutputFolder`, both expecting `MetaMorpheusException`.

pcruzparri (commented):
Co-review — 1 non-blocking finding, posted as an inline comment.

MetaMorpheus/CMD/CommandLineSettings.cs
trishorts: Fixed in `f6b90987f`, as you wrote it: above the `-o` requirement, conditioned on `AuditSdrf == null` so the both-flags case still falls through to its own path check.

Reproduced both invocations here before changing anything, and the diagnosis is exact — the guard that was hoisted above that early return, and the guard whose reasoning it borrowed, ended up on opposite sides of it. The comment at `:65` was written for this case and then only applied to the other one.

Falsified by reverting the production change alone, which is the part that shows the two new tests are pinning the gap rather than the flags:

```
with the fix      Passed: 38, Failed: 0, Skipped: 1
guard put back    Passed: 36, Failed: 2, Skipped: 1
                    TestAuditDataWithoutAuditSdrfIsRefusedAlongsideGeneratingDefaultTomls
                    TestAuditDataWithoutAuditSdrfIsRefusedAlongsideTheMicroVignette
                    both  Expected: <EngineLayer.MetaMorpheusException>  -- nothing thrown
                    TestAuditDataWithoutAuditSdrfIsRefused still passes
```

The bare case passing while the two new ones fail is the shape of the bug. 38/1 against the 36/1 you measured, so nothing else in the fixture moved; full suite 1997 passed / 0 failed / 1 skipped at this tip.

Two tests rather than three: the accepting direction is already pinned by `TestAuditSdrfCountsDataFilesFoundAndMissing`, which calls `ValidateCommandLineSettings()` with both flags set and would throw if the `AuditSdrf == null` condition were dropped. Adding a `DoesNotThrow` beside it would have been a second assertion of the same fact.

Thanks for building it and running the combinations. Marking it non-blocking was generous -- it is the same invisible-failure class the flag's own refusal exists to prevent, reached through the flags that were supposed to make it impossible.

Alexander-Sol (approved):

pcruzparri (approved):

PR #2791, opened 2026-09-03
Let a user build their own O-glycan database

trishorts:
> **Project of origin:** `customGlycanDb`

🤖 **Stacked on #2787** — it uses `CustomDataFile`, so #2787 must merge first and this diff shows its commits until it does. #2792 (custom N-glycan database) stacks on this one.

`MonosaccharidesCustom.tsv` already tells users, in the file MetaMorpheus hands them, that their custom monosaccharides exist *"so they can be used inside custom glycan database entries (**NGlycan_Custom.gdb / OGlycan_Custom.gdb**, or any user-supplied .gdb file)"*.

Neither file has ever existed. This adds the O-glycan one.

## What the user gets

`OGlycan_Custom.gdb`, seeded once into their DataDir and offered in the GlycoSearch task's O-glycan list beside the shipped databases — which stay untouched. A **Create new glycan** button on the main window validates a glycan and appends it, or they can open the file and type, because it documents its own format.

It follows #2787's recipe exactly: never touched once it exists, template shipped inside `EngineLayer`, no `<File>` in `Product.wxs`, no `CopyToOutputDirectory`. It joins `EveryKnownCustomFileIsSeededByStartup`, which is the inventory a new custom file has to join.

## Where it lives, and why that costs something

At the **DataDir root**, not under `Glycan_Mods\OGlycan\`, alongside `proteases_custom.tsv`, `rnase_custom.tsv` and `MonosaccharidesCustom.tsv`. `Product.wxs` gives `Glycan_Mods` **and both of its subfolders** a `<RemoveFolder On="both"/>`, so the installer owns those folders; the #2752 fix moved the monosaccharide file out of one of them for that reason and this should not move back in.

The cost is that `LoadGlycans` discovers databases by an unfiltered `Directory.GetFiles` over those two subfolders only, so a root-level file is safe but invisible. It is added to `OGlycanDatabasePaths` by name, guarded on `File.Exists` so a failed seeding cannot put a phantom path in the list the task window offers.

## The seeding could not land on its own

`LoadGlycans` reads **every** database in both folders eagerly at startup, so a template the loaders could not parse would throw inside `SetUpGlobalVariables()` — crashing MetaMorpheus on first launch, before any window opened. All three entry points choked on a documented file:

| | before | after |
|---|---|---|
| `LoadGlycan` | sniffed kind-vs-structure from **line 1**, so a `#` banner sent a composition database to the structure parser | sniffs from the first **data** line |
| `LoadStructureGlycan` | ran `ValidateStructureLine` over the `#` and threw | skips comments and blanks |
| `LoadKindGlycan` | skipped only lines lacking `Hex`, so a comment quoting `HexNAc(2)Hex(5)` reached `String2Kind` and died on a dictionary lookup | skips comments explicitly |

`IsCommentOrBlank` already existed and already handled the Excel `"#` quirk — it just had exactly one caller. All three now use it.

`ValidateStructureLine` threw a bare `FormatException` naming neither the file nor the line. It now throws `MetaMorpheusException` naming the file, the line number, the line, the offending character, and the fact that an unknown monosaccharide has to be declared in `MonosaccharidesCustom.tsv` first. Nothing asserted the old type.

## Adding a glycan

`GlycanDatabase.PersistCustomGlycan` validates with the same parser the search will use, so a glycan the user is told was added cannot fail to load later. It refuses:

- text that is neither a structure nor a composition;
- unbalanced parentheses, in either direction;
- a monosaccharide name or code MetaMorpheus does not know, pointing at `MonosaccharidesCustom.tsv`;
- the same monosaccharide counted twice in one composition;
- a duplicate of something already in the file;
- **an entry whose format disagrees with the rest of the file.** A database is read entirely as one format — inferred from its first data line — so mixing one line of the other kind silently changes how every *other* line is read. A freshly seeded file is all banner, so its format is still undecided and either is accepted.

Its return contract deliberately differs from `PersistCustomMonosaccharide`: it throws for every failure and returns only once the glycan is on disk. There is no registered-but-not-saved middle state, because a glycan database is not held in memory the way a monosaccharide is — the engine reads the file when it is built, which is also why a glycan added now is picked up by a search in the same session.

It carries the missing-trailing-newline guard from its monosaccharide sibling. Hand-edited `.gdb` files usually lack one; so does the shipped `NGlycan_ForNoSearch.gdb` test fixture.

## Two bugs found on the way

**An empty glycan database was searched in silence.** `BuildOGlycanBoxes` over zero glycans builds an **empty box array** without complaint, and the failure surfaced much later and elsewhere — `GlycanBoxes.First().Mass` inside the parallel search loop, as `InvalidOperationException: Sequence contains no elements`, naming nothing. If no scan reached that branch, the run instead returned zero results and looked like a search that simply found nothing. This mattered before; it matters more now that a user can be handed an empty database, because a freshly seeded one is all banner until they add a glycan.

**A selected database that is no longer in the folder** threw the identical unhelpful message, from the `.First()` on the path lookup.

Both now throw `MetaMorpheusException` naming the file — and for the missing one, listing what is available. The four bare `.First()` lookups in the constructor became one helper.

## Verification

`CustomGlycanDatabaseTests` — 21 tests covering the template shape (every line a comment or blank, and the resource resolves by the name the seeding asks for), the seeded file loading to zero glycans without throwing, all three loader-tolerance fixes including the comment-quotes-a-composition trap, every validation refusal, the trailing-newline append, a declared custom monosaccharide being legal in both formats, the startup wiring, and both guards.

**147 glyco and glycan tests pass**; 39 custom-file and custom-glycan tests pass. EngineLayer, GUI and Test build with 0 errors.

## Not in this PR

`GlycanDatabase.EnsureCustomMonosaccharideFileExists` predates #2787 and still hand-rolls `CustomDataFile.EnsureExists` plus a legacy-location carry-over. Folding it in is right and is #2787's own thesis, but it swaps a `File.Copy` for a read-and-rewrite on a path whose whole point is not losing a user's data, so it wants its own PR and its own verification rather than a ride on this one.

https://claude.ai/code/session_01Kr37s1DXoTUnMcn3htxnR4

### DIFF
--- a/MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
+++ b/MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
@@ -397,9 +397,16 @@ namespace EngineLayer
                 // Parsed exactly as LoadStructureGlycan would parse it: Struct2Glycan is the only thing that
                 // knows whether the nesting describes a tree it can actually build.
                 List<Glycan> parsed = Glycan.Struct2Glycan(entry, 1, isOGlycan);
-                if (parsed == null || parsed.Count == 0)
+
+                // "()" parses happily into a glycan of nothing, whose mass is zero. Searching for it is
+                // meaningless and it would widen every box it landed in, so it is refused here rather than
+                // left for the user to wonder about. The empty case is folded in: Struct2Glycan either
+                // throws or yields glycans, so "nothing came back" and "nothing in what came back" are the
+                // same answer to the user.
+                if (parsed == null || parsed.Count == 0 || parsed.All(g => g.Kind.Sum(count => (int)count) == 0))
                 {
-                    throw new MetaMorpheusException($"Could not add the glycan \"{entry}\": it did not parse into any glycan.");
+                    throw new MetaMorpheusException(
+                        $"Could not add the glycan \"{entry}\": it contains no monosaccharides.");
                 }
             }
             catch (MetaMorpheusException)
@@ -428,6 +435,7 @@ namespace EngineLayer
             CaptureCollection names = match.Groups["name"].Captures;
             CaptureCollection counts = match.Groups["count"].Captures;
             HashSet<string> seen = new HashSet<string>(StringComparer.Ordinal);
+            int total = 0;
 
             for (int i = 0; i < names.Count; i++)
             {
@@ -449,6 +457,14 @@ namespace EngineLayer
                     throw new MetaMorpheusException(
                         $"Could not add the glycan \"{entry}\": the count for '{name}' must be a whole number between 0 and 255.");
                 }
+                total += count;
+            }
+
+            // Same reason as the structure case: a composition that totals nothing is a zero-mass glycan.
+            if (total == 0)
+            {
+                throw new MetaMorpheusException(
+                    $"Could not add the glycan \"{entry}\": it contains no monosaccharides.");
             }
         }
 
--- a/MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
+++ b/MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 
 namespace EngineLayer
 {
@@ -21,13 +22,22 @@ namespace EngineLayer
         /// <returns> A glycan object collection </returns>
         public static IEnumerable<Glycan> LoadGlycan(string filePath, bool ToGenerateIons, bool IsOGlycan)
         {
+            // The format is inferred from the first DATA line; it is never declared. Comment and blank lines
+            // are skipped while sniffing because a documented database -- such as the seeded custom one --
+            // opens with a '#' banner, and judging the format from that banner routes the whole file to the
+            // wrong parser. A file with no data line is a legal empty database: either parser yields nothing
+            // for it, so which one is chosen does not matter.
             bool isKind = true;
             using (StreamReader lines = new StreamReader(filePath))
             {
                 while(lines.Peek() != -1)
                 {
                     string line = lines.ReadLine();
-                    if (!line.Contains("HexNAc"))  // use the first line to determine the format (kind / structure) of glycan database.
+                    if (IsCommentOrBlank(line))
+                    {
+                        continue;
+                    }
+                    if (!line.Contains("HexNAc"))  // use the first data line to determine the format (kind / structure) of glycan database.
                     {
                         isKind = false;
                     }
@@ -233,6 +243,262 @@ namespace EngineLayer
             }
         }
 
+
+        /// <summary>
+        /// The two ways a glycan database line can be written. A database file is read entirely as one or
+        /// the other -- the format is inferred from its first data line and never declared -- so an entry
+        /// being added has to agree with whatever is already in the file.
+        /// </summary>
+        public enum GlycanLineFormat
+        {
+            /// <summary>Nested single-character codes, e.g. <c>(N(H(A)))</c>. What OGlycan.gdb uses.</summary>
+            Structure,
+
+            /// <summary>Name-and-count, e.g. <c>HexNAc(2)Hex(5)</c>. What NGlycan.gdb uses.</summary>
+            Composition
+        }
+
+        /// <summary>
+        /// Validates one glycan the user has typed and appends it to their custom glycan database, in the
+        /// format that database is already written in.
+        ///
+        /// <para>
+        /// Return contract, and it differs deliberately from <see cref="PersistCustomMonosaccharide"/>:
+        /// this method throws <see cref="MetaMorpheusException"/> for every failure and returns normally
+        /// only once the glycan is on disk. There is no registered-but-not-saved middle state to report,
+        /// because a glycan database is not held in memory the way a monosaccharide is -- the search reads
+        /// the file when the engine is built. Which is also why an entry added now is picked up by a search
+        /// run in this same session; only MetaDraw's glycan list waits for a restart.
+        /// </para>
+        /// </summary>
+        /// <param name="glycanText">The glycan, as typed: a structure or a composition.</param>
+        /// <param name="databasePath">The custom database to append to. Created if it is not there.</param>
+        /// <param name="isOGlycan">
+        /// Whether this is an O-glycan database, so the entry is validated the same way the search will
+        /// read it and a glycan that parses here cannot fail to parse there.
+        /// </param>
+        public static void PersistCustomGlycan(string glycanText, string databasePath, bool isOGlycan)
+        {
+            string entry = (glycanText ?? string.Empty).Trim();
+            string fileName = Path.GetFileName(databasePath);
+
+            if (entry.Length == 0)
+            {
+                throw new MetaMorpheusException("Could not add the glycan: no glycan was given.");
+            }
+            if (IsCommentOrBlank(entry))
+            {
+                throw new MetaMorpheusException(
+                    $"Could not add the glycan: \"{entry}\" is a comment, not a glycan. Lines beginning with '#' are ignored when the database is read.");
+            }
+            if (entry.IndexOfAny(new[] { '\t', '\n', '\r' }) >= 0)
+            {
+                throw new MetaMorpheusException(
+                    "Could not add the glycan: one glycan per line, with no tabs or line breaks in it.");
+            }
+
+            GlycanLineFormat format = ValidateGlycanLine(entry, fileName, isOGlycan);
+
+            // The format is a property of the whole file, so an entry that disagreed with what is already
+            // there would silently change how every OTHER line in the file is read.
+            GlycanLineFormat? existingFormat = FormatOfExistingEntries(databasePath);
+            if (existingFormat.HasValue && existingFormat.Value != format)
+            {
+                throw new MetaMorpheusException(
+                    $"Could not add the glycan to '{fileName}': that file is written in {existingFormat.Value.ToString().ToLowerInvariant()} format " +
+                    $"and \"{entry}\" is {format.ToString().ToLowerInvariant()} format. A glycan database is read entirely as one format or the other, " +
+                    "so the two cannot be mixed in one file. Convert the entry, or keep it in a database of its own.");
+            }
+
+            if (ContainsEntry(databasePath, entry))
+            {
+                throw new MetaMorpheusException($"Could not add the glycan: '{fileName}' already contains \"{entry}\".");
+            }
+
+            try
+            {
+                string directory = Path.GetDirectoryName(databasePath);
+                if (!string.IsNullOrEmpty(directory))
+                {
+                    Directory.CreateDirectory(directory);
+                }
+
+                if (!File.Exists(databasePath))
+                {
+                    File.WriteAllLines(databasePath, new[] { entry });
+                }
+                else
+                {
+                    // AppendAllLines never checks whether the file already ends in a newline. A hand-edited
+                    // .gdb very often does not -- NGlycan_ForNoSearch.gdb in the test data does not -- and
+                    // without this the new glycan is glued onto the end of the last one, corrupting both.
+                    string existing = File.ReadAllText(databasePath);
+                    if (existing.Length > 0 && !existing.EndsWith("\n"))
+                    {
+                        File.AppendAllText(databasePath, Environment.NewLine);
+                    }
+                    File.AppendAllLines(databasePath, new[] { entry });
+                }
+            }
+            catch (Exception ex)
+            {
+                throw new MetaMorpheusException($"Could not save the glycan to '{databasePath}': {ex.Message}", ex);
+            }
+        }
+
+        /// <summary>
+        /// Works out which format a glycan line is written in, and checks that the parser the search will
+        /// use actually accepts it -- so a glycan the user is told was added cannot fail to load later.
+        /// </summary>
+        /// <exception cref="MetaMorpheusException">The line is neither format, or does not parse.</exception>
+        public static GlycanLineFormat ValidateGlycanLine(string entry, string fileName, bool isOGlycan)
+        {
+            if (entry.StartsWith("(", StringComparison.Ordinal))
+            {
+                ValidateStructureEntry(entry, fileName, isOGlycan);
+                return GlycanLineFormat.Structure;
+            }
+
+            ValidateCompositionEntry(entry);
+            return GlycanLineFormat.Composition;
+        }
+
+        private static void ValidateStructureEntry(string entry, string fileName, bool isOGlycan)
+        {
+            // The same character check the loader runs, so an unknown monosaccharide code is rejected here
+            // rather than being silently counted as zero mass at search time.
+            ValidateStructureLine(entry, fileName, 1);
+
+            int depth = 0;
+            foreach (char c in entry)
+            {
+                if (c == '(')
+                {
+                    depth++;
+                }
+                else if (c == ')')
+                {
+                    depth--;
+                    if (depth < 0)
+                    {
+                        throw new MetaMorpheusException(
+                            $"Could not add the glycan \"{entry}\": the parentheses do not balance -- a ')' closes a branch that was never opened.");
+                    }
+                }
+            }
+            if (depth != 0)
+            {
+                throw new MetaMorpheusException(
+                    $"Could not add the glycan \"{entry}\": the parentheses do not balance -- {depth} branch(es) are left open.");
+            }
+
+            try
+            {
+                // Parsed exactly as LoadStructureGlycan would parse it: Struct2Glycan is the only thing that
+                // knows whether the nesting describes a tree it can actually build.
+                List<Glycan> parsed = Glycan.Struct2Glycan(entry, 1, isOGlycan);
+                if (parsed == null || parsed.Count == 0)
+                {
+                    throw new MetaMorpheusException($"Could not add the glycan \"{entry}\": it did not parse into any glycan.");
+                }
+            }
+            catch (MetaMorpheusException)
+            {
+                throw;
+            }
+            catch (Exception ex)
+            {
+                throw new MetaMorpheusException($"Could not add the glycan \"{entry}\": {ex.Message}", ex);
+            }
+        }
+
+        private static void ValidateCompositionEntry(string entry)
+        {
+            // Name(count), repeated -- e.g. HexNAc(2)Hex(5)NeuAc(1). Checked here rather than by handing it
+            // to String2Kind, which throws KeyNotFoundException on an unknown name and quietly accepts a
+            // trailing unclosed group.
+            Match match = Regex.Match(entry, @"^(?:(?<name>[A-Za-z][A-Za-z0-9]*)\((?<count>\d+)\))+$");
+            if (!match.Success)
+            {
+                throw new MetaMorpheusException(
+                    $"Could not add the glycan \"{entry}\": it is neither a structure -- which starts with '(', e.g. (N(H(A))) -- " +
+                    "nor a composition, which is a name and a count repeated, e.g. HexNAc(2)Hex(5).");
+            }
+
+            CaptureCollection names = match.Groups["name"].Captures;
+            CaptureCollection counts = match.Groups["count"].Captures;
+            HashSet<string> seen = new HashSet<string>(StringComparer.Ordinal);
+
+            for (int i = 0; i < names.Count; i++)
+            {
+                string name = names[i].Value;
+                if (!Glycan.NameCharDic.ContainsKey(name))
+                {
+                    string allowed = string.Join(", ", Glycan.NameCharDic.Keys);
+                    throw new MetaMorpheusException(
+                        $"Could not add the glycan \"{entry}\": '{name}' is not a monosaccharide MetaMorpheus knows. Known: {allowed}. " +
+                        "A monosaccharide it does not ship with must be declared in MonosaccharidesCustom.tsv first.");
+                }
+                if (!seen.Add(name))
+                {
+                    throw new MetaMorpheusException(
+                        $"Could not add the glycan \"{entry}\": '{name}' appears more than once. Give each monosaccharide a single total count.");
+                }
+                if (!byte.TryParse(counts[i].Value, NumberStyles.Integer, CultureInfo.InvariantCulture, out byte count))
+                {
+                    throw new MetaMorpheusException(
+                        $"Could not add the glycan \"{entry}\": the count for '{name}' must be a whole number between 0 and 255.");
+                }
+            }
+        }
+
+        /// <summary>
+        /// The format the entries already in a database are written in, or null when it holds no entries
+        /// yet -- a file that is missing, or one that is all banner, which is what a freshly seeded custom
+        /// database is.
+        /// </summary>
+        private static GlycanLineFormat? FormatOfExistingEntries(string databasePath)
+        {
+            if (!File.Exists(databasePath))
+            {
+                return null;
+            }
+
+            foreach (string line in File.ReadLines(databasePath))
+            {
+                if (IsCommentOrBlank(line))
+                {
+                    continue;
+                }
+                return line.TrimStart().StartsWith("(", StringComparison.Ordinal)
+                    ? GlycanLineFormat.Structure
+                    : GlycanLineFormat.Composition;
+            }
+
+            return null;
+        }
+
+        /// <summary>Whether the database already holds this exact glycan, so it is not added twice.</summary>
+        private static bool ContainsEntry(string databasePath, string entry)
+        {
+            if (!File.Exists(databasePath))
+            {
+                return false;
+            }
+
+            foreach (string line in File.ReadLines(databasePath))
+            {
+                // Split on tab first: the shipped .txt databases carry name and mass columns after the
+                // glycan, and it is the glycan that has to be unique.
+                if (!IsCommentOrBlank(line) && line.Split('\t')[0].Trim().Equals(entry, StringComparison.Ordinal))
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
         /// <summary>
         /// Ensure the MonosaccharidesCustom.tsv exists in the directory. If the file is missing, 
         /// write the embedded full 85-line documented template—instructions, column spec, the built-in name/code table,
@@ -292,7 +558,17 @@ namespace EngineLayer
                 int id = 1;
                 while (lines.Peek() != -1)
                 {
-                    string line = lines.ReadLine().Split('\t').First();
+                    string rawLine = lines.ReadLine();
+
+                    // Skipped explicitly rather than left to the Hex test below: a comment that quotes a
+                    // composition -- which the documented template does, to show the format -- would otherwise
+                    // reach String2Kind and die on a dictionary lookup naming neither the file nor the line.
+                    if (IsCommentOrBlank(rawLine))
+                    {
+                        continue;
+                    }
+
+                    string line = rawLine.Split('\t').First();
 
                     if (!(line.Contains("HexNAc") || line.Contains("Hex"))) // Make sure the line is a glycan line. The line should contain HexNAc or Hex.
                     {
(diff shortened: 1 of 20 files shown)


MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
```diff
@@ -233,6 +243,278 @@ public static string PersistCustomMonosaccharide(string name, string codeText, s
...
+        }
+
+        /// <summary>
+        /// Works out which format a glycan line is written in, and checks that the parser the search will
+        /// use actually accepts it -- so a glycan the user is told was added cannot fail to load later.
+        /// </summary>
+        /// <exception cref="MetaMorpheusException">The line is neither format, or does not parse.</exception>
+        public static GlycanLineFormat ValidateGlycanLine(string entry, string fileName, bool isOGlycan)
+        {
+            if (entry.StartsWith("(", StringComparison.Ordinal))
```
RayMSMS, line 365: <html>
<body>
<h3 dir="ltr">🐛 Composition glycans without <code>HexNAc</code> are accepted on add, then crash the app on next launch</h3>
<h4 dir="ltr">Steps to reproduce</h4>
<ol dir="ltr">
<li>Add a custom monosaccharide.</li>
<li>Add a composition glycan built on it.</li>
</ol>
<p dir="ltr">It saves without complaint, but the <strong>next launch throws before any window opens</strong>. This affects any composition without the literal <code>HexNAc</code>, not just custom sugars.</p>
<h4 dir="ltr">Root cause — two format detectors disagree</h4>
<div dir="ltr">
Path | Rule for "structure vs composition"
-- | --
ValidateGlycanLine / FormatOfExistingEntries | structure iff it starts with (, else composition
LoadGlycan (format sniff) | composition iff first data line contains "HexNAc", else structure

</div>
<p dir="ltr">A HexNAc-less composition validates as <strong>composition</strong> and gets written, but <code>LoadGlycan</code> reads it as <strong>structure</strong> → <code>LoadStructureGlycan</code> throws. This breaks the core invariant of this PR (<em>validates here ⇒ loads there</em>).</p>
<p dir="ltr">Because <code>GlobalVariables.LoadGlycans</code> eagerly loads every DB inside <code>SetUpGlobalVariables</code>, it surfaces as a <strong>startup crash</strong> — the app won't launch, and the only recovery is hand-deleting the line from the <code>.gdb</code>.</p>
<h4 dir="ltr">Suggested fix</h4>
<p dir="ltr">Guard in <code>ValidateGlycanLine</code> before the composition branch. Every O-/N-glycan has a HexNAc core, so "no <code>HexNAc</code> and not a <code>(</code>-structure" is invalid — and using <code>Contains("HexNAc")</code> makes the accept-set <strong>exactly match</strong> what <code>LoadGlycan</code> treats as composition, so the two can't diverge.</p>
<div><div role="group" aria-label="csharp code" tabindex="0"><div><div></div></div><div>csharp</div><div><pre style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono);"><code class="language-csharp" style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono); white-space: pre;"><span><span><span style="color: rgb(129, 0, 194);">if</span> <span style="color: rgb(43, 48, 59);">(</span>entry<span style="color: rgb(43, 48, 59);">.</span><span style="color: rgb(0, 81, 194);">StartsWith</span><span style="color: rgb(43, 48, 59);">(</span><span style="color: rgb(0, 128, 0);">"("</span><span style="color: rgb(43, 48, 59);">,</span> StringComparison<span style="color: rgb(43, 48, 59);">.</span>Ordinal<span style="color: rgb(43, 48, 59);">))</span>
</span><span><span style="color: rgb(43, 48, 59);">{</span>
</span><span>    <span style="color: rgb(0, 81, 194);">ValidateStructureEntry</span><span style="color: rgb(43, 48, 59);">(</span>entry<span style="color: rgb(43, 48, 59);">,</span> fileName<span style="color: rgb(43, 48, 59);">,</span> isOGlycan<span style="color: rgb(43, 48, 59);">);</span>
</span><span>    <span style="color: rgb(129, 0, 194);">return</span> GlycanLineFormat<span style="color: rgb(43, 48, 59);">.</span>Structure<span style="color: rgb(43, 48, 59);">;</span>
</span><span><span style="color: rgb(43, 48, 59);">}</span>
</span><span>
</span><span><span style="color: rgb(110, 118, 135);">// Not a structure, so it must be a composition. Every O-/N-glycan is built on a HexNAc core,</span>
</span><span><span style="color: rgb(110, 118, 135);">// and LoadGlycan sniffs composition-vs-structure by this same test (Contains("HexNAc")).</span>
</span><span><span style="color: rgb(110, 118, 135);">// Keeps the accept-set aligned with the loader, so a composition that validates cannot later</span>
</span><span><span style="color: rgb(110, 118, 135);">// be read as a structure and crash during the eager load in LoadGlycans.</span>
</span><span><span style="color: rgb(129, 0, 194);">if</span> <span style="color: rgb(43, 48, 59);">(</span>!entry<span style="color: rgb(43, 48, 59);">.</span><span style="color: rgb(0, 81, 194);">Contains</span><span style="color: rgb(43, 48, 59);">(</span><span style="color: rgb(0, 128, 0);">"HexNAc"</span><span style="color: rgb(43, 48, 59);">))</span>
</span><span><span style="color: rgb(43, 48, 59);">{</span>
</span><span>    <span style="color: rgb(129, 0, 194);">throw</span> <span style="color: rgb(129, 0, 194);">new</span> <span style="color: rgb(179, 74, 0);">MetaMorpheusException</span><span style="color: rgb(43, 48, 59);">(</span>
</span><span>        <span style="color: rgb(0, 128, 0);">$"Could not add the glycan </span><span style="color: rgb(0, 128, 128);">\"</span><span style="color: rgb(43, 48, 59);">{</span><span style="color: rgb(0, 128, 0);">entry</span><span style="color: rgb(43, 48, 59);">}</span><span style="color: rgb(0, 128, 128);">\"</span><span style="color: rgb(0, 128, 0);">: it is neither a structure -- which starts with '(', "</span> +
</span><span>        <span style="color: rgb(0, 128, 0);">"e.g. (N(H(A))) -- nor a composition. Every O- and N-glycan is built on a HexNAc core, so a "</span> +
</span><span>        <span style="color: rgb(0, 128, 0);">"composition must contain HexNAc, e.g. HexNAc(2)Hex(5)."</span><span style="color: rgb(43, 48, 59);">);</span>
</span><span><span style="color: rgb(43, 48, 59);">}</span>
</span><span>
</span><span><span style="color: rgb(0, 81, 194);">ValidateCompositionEntry</span><span style="color: rgb(43, 48, 59);">(</span>entry<span style="color: rgb(43, 48, 59);">);</span>
</span><span><span style="color: rgb(129, 0, 194);">return</span> GlycanLineFormat<span style="color: rgb(43, 48, 59);">.</span>Composition<span style="color: rgb(43, 48, 59);">;</span></span></span></code></pre></div></div></div>
<h4 dir="ltr">Checklist</h4>
<ul dir="ltr">
<li><input disabled="" type="checkbox"> Add the <code>HexNAc</code> guard in <code>ValidateGlycanLine</code></li>
<li><input disabled="" type="checkbox"> Round-trip test: HexNAc-less composition → rejected; valid composition → <code>LoadGlycan</code> reads it back as composition</li>
<li><input disabled="" type="checkbox"> <strong>Follow-up:</strong> guard only covers the GUI path — a hand-edited <code>.gdb</code> can still crash startup, so the eager load in <code>LoadGlycans</code> should degrade a broken custom DB to a <code>StartupWarnings</code> entry and skip it (as done for protease/rnase collisions in this PR)</li></ul>
</body>
</html>

nbollis (approved):

nbollis (dismissed):

Alexander-Sol:
## Agent review — Request changes

**Summary.** Seeds a user-editable `OGlycan_Custom.gdb` at the DataDir root, offers it in GlycoSearch, adds a "Create new O-glycan" window, and hardens the glycan loaders (comment/blank-line tolerance, better empty/missing-database messages). Engineering is careful and the 21 tests are genuinely good. One confirmed bug falsifies the PR's headline guarantee.

### 🔴 HIGH (confirmed) — validator and loader disagree on format detection; a HexNAc-less composition is accepted, written, then fails to load
- **Validator** `GlycanDatabase.cs:356` (`ValidateGlycanLine`) and `:489` (`FormatOfExistingEntries`) decide structure-vs-composition by whether the line **starts with `(`**.
- **Loader** `GlycanDatabase.cs:40` (`LoadGlycan`) decides by **`!line.Contains("HexNAc")` → structure**. These disagree for any composition lacking the literal substring `HexNAc`.

**Scenario:** user opens "Create new O-glycan", types `Hex(1)` (a legitimate single-hexose O-glycan) into the freshly-seeded (empty) database. `ValidateCompositionEntry` passes it, `PersistCustomGlycan` writes it, the GUI shows *"…was added. Select that database in a GlycoSearch task…"*. The user selects the DB and searches. `LoadGlycan` sniffs the first data line `Hex(1)`, sees no `HexNAc`, routes the whole file to `LoadStructureGlycan`, which throws `MetaMorpheusException: Unrecognized character 'e' …`. Trigger condition: the *first* data line lacks `HexNAc` (e.g. `Hex(1)`, `NeuAc(1)`, `Fuc(1)Hex(2)`) — which makes it intermittent and confusing.

This breaks `PersistCustomGlycan`'s stated contract ("validates with the same parser the search will use"). **It also affects the stacked PR #2792**, which shares `LoadGlycan`.

**Fix:** unify format detection — make `LoadGlycan` classify structure by leading `(` (matching the validator) instead of by the `HexNAc` substring. No test covers "persist a HexNAc-less composition, then load it"; every composition test uses `HexNAc(2)Hex(5)` or expects a refusal, which is why this slipped through.

### Minor
- **Low (plausible):** N-glycan mass filter in `GlycoSearchEngine` (`.Where(p => p.Mass/1E5 <= maxGlycanBoxMass)`) can reduce a non-empty database back to zero glycans, re-introducing the "Sequence contains no elements" case the new empty-db guard was meant to close. (Adjacent to this PR's scope.)
- **Low:** dedup (`ContainsEntry`, `:498`) is exact-ordinal string compare, so `HexNAc(2)Hex(5)` and `Hex(5)HexNAc(2)` both persist as duplicates.
- **Nit:** the diff prepends a UTF-8 BOM to `CustomDataFile.cs` and `MainWindow.xaml.cs`.

### Verified correct
Culture-safe parsing (InvariantCulture), trailing-newline append guard, balanced-paren / unknown-code / repeated-monosaccharide / count>255 / zero-mass / injection cases (all tested), custom-monosaccharide integration, DataDir-root seeding recipe consistent with #2787, and the improved empty/missing-database messages.

trishorts:
@Alexander-Sol Thank you — the HIGH is real, and it falsifies the sentence this PR is built on. Fixed in 93cb58026, with a follow-up BOM sweep in 0c83c7410.

### 🔴 The format-detection bug — confirmed, reproduced verbatim

Your trace was exactly right, including the character. Before writing the fix I wrote the test and ran it against the old code:

```
EngineLayer.MetaMorpheusException : Could not parse glycan structure in 'OGlycan_HexOnly.gdb'
at line 61: "Hex(1)". Unrecognized character 'e'. Allowed: parentheses and one of HNAGFPSYCXK.
  at EngineLayer.GlycanDatabase.ValidateStructureLine(...) GlycanDatabase.cs:line 728
  at EngineLayer.GlycanDatabase.LoadStructureGlycan(...)+MoveNext() GlycanDatabase.cs:line 688
```

`LoadGlycan` now uses the leading-`(` test — the one `ValidateGlycanLine` and `FormatOfExistingEntries` already use, which is what makes them agree by construction rather than by coincidence.

Before changing a sniff that every glycan database in the repo goes through, I checked the first data line of all eleven — six shipped, five test fixtures:

| | first data line |
|---|---|
| `OGlycan.gdb`, `OGlycan_withIsobaric.gdb`, `NGlycan_struct.gdb`, `OGlycan.gdb` (test) | `(N)` |
| `NGlycan.gdb`, the five `Olgycan Database *.txt`, `NGlycan_ForNoSearch.gdb`, `NGlycan_NOBoxesTesting.gdb`, `NGlycan_fakeN-glycan.gdb` | `HexNAc(…)` |

Every structure file opens with `(`; every composition file opens with `HexNAc(`. So classification is bit-for-bit unchanged everywhere except the case that was broken. You're right that it reaches #2792 as well — that branch has the merge.

Your diagnosis of why no test caught it was also right, and I've used it as the test: `AHexNAcLessCompositionLoadsBackAsWhatItWasValidatedAs` persists a `HexNAc`-less composition into a freshly seeded database and loads it back. It fails against the old code with the message above.

### The N-glycan mass filter — half right, and the half that's wrong points at a real one

The N-glycan path is already safe: `GlycoSearchEngine.cs:948` reads `if (NGlycans.Length == 0 || possibleGlycanMassHigh < …First().Mass …)`, so the `.First()` is guarded where it's used.

But following it up found the same hole one branch over, and that one is open. `GlycanBoxes.First().Mass` at `:1010` has **no** length guard, and `BuildOGlycanBoxes` starts at `i = 1` and `continue`s past every box over `maxGlycanBoxMass`. So a database that holds glycans can still produce zero boxes — every glycan heavier than the cap — and that is precisely the `Sequence contains no elements` the new empty-database guard was written to close, arriving one step later. The O-glycan and N+O branches now refuse it up front by name:

> No combination of glycans from the O-glycan database 'X.gdb' is within the maximum glycan box mass of 4000 Da, so there is nothing to search for. Raise that maximum, or choose a database of lighter glycans.

Tested by `EveryGlycanBeingHeavierThanTheBoxMassCapIsRefusedByName`, also verified to fail against the old code. So: the finding was worth raising even though the file and line were off.

### Dedup — agreed, and deliberately left

`ContainsEntry` is an ordinal compare, so `HexNAc(2)Hex(5)` and `Hex(5)HexNAc(2)` both persist. It only affects compositions — `(N(H))` and `(H(N))` are genuinely different glycans — and the consequence is a duplicate line in the user's own file, which the search handles. Canonicalising compositions for the comparison is the right fix but it's new behaviour rather than a repair, so I'd rather not fold it into a PR that's already had its approvals reset once. Happy to add it here if you'd prefer it not be deferred.

### BOMs — fixed, and there was a third

`CustomDataFile.cs` and `MainWindow.xaml.cs` are back to no BOM. Auditing every changed file rather than only the two named turned up `Test/CustomDataFileTests.cs` as well; that's 0c83c7410. The two genuinely new files keep theirs — a BOM on a file that didn't exist before isn't a change to anything.

---

2191 glyco, custom-file and calibration tests pass; solution builds with 0 errors.

@nbollis — sorry, this resets your approval for the second time. The short version: a user typing `Hex(1)` into a freshly seeded custom database was told it was added and then could not search it, because the validator and the loader disagreed about what format the line was in. The fix is one line plus a test; the rest is a guard and three BOMs.

trishorts:
Two corrections to my comment above, both from the agent review on #2792, which I should have read before writing it.

**1. I understated the severity.** I described the misclassified line as failing "on the next search." It is worse than that: `GlobalVariables.LoadGlycans` calls `LoadGlycan` eagerly on every path in `OGlycanDatabasePaths` inside `SetUpGlobalVariables`, so once `Hex(1)` is in the custom database the next launch throws before any window opens. A user who adds a legitimate single-hexose O-glycan through the new window ends up unable to start MetaMorpheus, and the only way out is to hand-edit a file they don't know is the cause. The fix already shipped covers it; the description was too mild.

**2. There is a second bug in the same area, and 93cb58026 did not fix it.** `LoadKindGlycan` decided what counted as a data line with `line.Contains("HexNAc") || line.Contains("Hex")`. A composition built from neither — `NeuAc(2)Fuc(1)` — is accepted by `ValidateCompositionEntry`, written by `PersistCustomGlycan`, reported to the user as added, and then dropped by the loader in silence. No error, no glycan. It is not first-line-dependent, so unifying the format sniff does nothing for it.

That substring test predates skipping comments explicitly and is now redundant as a "is this a data line" guard. It is a shape test instead, against the same pattern `ValidateCompositionEntry` uses — one shared constant, since the validator and the loader holding separate opinions about what a composition looks like is exactly what the HIGH finding was:

```csharp
private const string CompositionShape = @"^(?:(?<name>[A-Za-z][A-Za-z0-9]*)\((?<count>\d+)\))+$";
```

Tolerance for a line that genuinely is not a composition is unchanged — still skipped, now by shape rather than substring, and pinned by a test. A line that *is* composition-shaped but will not parse now raises a named error rather than a bare `KeyNotFoundException` out of `SetUpGlobalVariables`.

Three more tests: the silent drop, the named parse failure, and the unchanged skip tolerance.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
```diff
@@ -290,16 +588,52 @@ public static IEnumerable<Glycan> LoadKindGlycan(string filePath, bool ToGenerat
...
+                    string line = rawLine.Split('\t').First().Trim();
+
+                    // Whether the line has the SHAPE of a composition -- a name and a count, repeated. The
+                    // test used to be "does it contain Hex or HexNAc", a stand-in for "is this a data line"
+                    // from before comments were skipped above. It also threw away, without a word, every
+                    // legitimate composition built from neither: NeuAc(2)Fuc(1) is one, and the entry
+                    // validator accepts it, so a user could be told a glycan had been added and then have
+                    // it silently vanish from the search. Tolerance for a line that is not a composition at
+                    // all is unchanged -- it is still skipped, now by shape rather than by substring.
+                    if (!Regex.IsMatch(line, CompositionShape))
```
pcruzparri, line 614: **High** — a composition line with anything after the composition is now dropped without a word

**Problem:**
The old gate was `line.Contains("HexNAc") || line.Contains("Hex")`, and `String2Kind` then ignored whatever followed the last `)` — its loop stops at `i < x.Length - 1`, so trailing junk fell off the end. `CompositionShape` is anchored, so the whole line now has to be composition and nothing else. Trailing text that is not behind a tab fails the match and the line is `continue`d silently. I ran the old gate and the new one over the same inputs:

```
HexNAc(1)Hex(1)                          old=2 glycans  new=2 glycans
HexNAc(1)Hex(1) # core 1                 old=2 glycans  new=0 glycans
HexNAc(1)Hex(1) N1H1F0S0 365.1322        old=2 glycans  new=0 glycans
HexNAc(1)Hex(1)\tN1H1F0S0\t365.1322      old=2 glycans  new=2 glycans
```

The tab-separated form is the one the shipped `.txt` databases use, so nothing in the repo changes — I diffed glycan counts and masses for all thirteen `.gdb`/`.txt` databases under `Glycan_Mods` and `GlycoTestData`, before and after, and they are identical. It is the hand-edited file that loses rows: a user who annotated a line, or who lined the columns up with spaces instead of tabs, had those glycans loaded before this branch and does not now.

**Impact:**
The glycan is missing from the search space with no error, no warning and no entry in the output — the search just finds fewer glycopeptides. That is the same silent drop this commit set out to close, and the comment above the check says the opposite: "Tolerance for a line that genuinely is not a composition is unchanged." The line in question *is* a composition; it just has a note after it.

**Suggested fix:**
Cut the annotation off before the shape test rather than letting it disqualify the line — strip from the first `#` and take the first whitespace-delimited field, the way the tab split already does — and if what remains still is not composition-shaped, say so instead of skipping. The `Split('\t').First()` on the line above is already doing exactly this job for one delimiter.

**Affected test scenarios:**
`ALineThatIsNotACompositionAtAllIsStillSkipped` pins the skip, but only for a line that never was a composition. Worth a case per row above: an inline `# note`, and space-separated trailing columns, both asserting the glycan still loads.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
```diff
@@ -233,6 +252,285 @@ public static string PersistCustomMonosaccharide(string name, string codeText, s
...
+            GlycanLineFormat? existingFormat = FormatOfExistingEntries(databasePath);
+            if (existingFormat.HasValue && existingFormat.Value != format)
+            {
+                throw new MetaMorpheusException(
+                    $"Could not add the glycan to '{fileName}': that file is written in {existingFormat.Value.ToString().ToLowerInvariant()} format " +
+                    $"and \"{entry}\" is {format.ToString().ToLowerInvariant()} format. A glycan database is read entirely as one format or the other, " +
+                    "so the two cannot be mixed in one file. Convert the entry, or keep it in a database of its own.");
+            }
+
+            if (ContainsEntry(databasePath, entry))
```
pcruzparri, line 322: **Non-blocking** — the duplicate guard compares text, so the same glycan written two ways goes in twice

**Problem:**
`ContainsEntry` does an ordinal string compare against each line, but a glycan has more than one spelling. `Fuc` and `dHex` are both in `Glycan.NameCharDic` and both map to index 4, and a composition is a set of counts, not an ordered one. Persisting each of these pairs into a freshly seeded database: both members accepted, both written.

```
Fuc(1)            then dHex(1)           -> 4 glycans, one distinct mass (14605791), one composition (F1)
HexNAc(1)Hex(1)   then Hex(1)HexNAc(1)   -> 4 glycans, one distinct mass (36513219)
```

The method's own summary says it "refuses duplicates", and for the case it can see it does.

**Impact:**
The database holds the same glycan twice. `BuildOGlycanBoxes` combines over glycan indices, so a duplicate entry is not inert — it enlarges the index set the boxes are drawn from and adds boxes that are the same glycan set under two names. The user is also told the second one was added, which reads as confirmation that it was something new.

**Suggested fix:**
Compare on what the entry parses to rather than on its text — `ValidateGlycanLine` has already produced a parse by this point, so the `kind[]` (or `Glycan.Composition`) of the new entry against those of the existing lines catches the alias and the reordering together.

**Affected test scenarios:**
`TheSameGlycanIsNotAddedTwice` covers the literal repeat. An alias pair and a reordered composition are the two cases it does not reach.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
```diff
@@ -233,6 +252,285 @@ public static string PersistCustomMonosaccharide(string name, string codeText, s
...
+
+            ValidateCompositionEntry(entry);
+            return GlycanLineFormat.Composition;
+        }
+
+        private static void ValidateStructureEntry(string entry, string fileName, bool isOGlycan)
+        {
+            // The same character check the loader runs, so an unknown monosaccharide code is rejected here
+            // rather than being silently counted as zero mass at search time.
+            ValidateStructureLine(entry, fileName, 1);
```
pcruzparri, line 379: **Non-blocking** — the typed-entry path reports a line number the entry does not have

**Problem:**
`ValidateStructureEntry` borrows `ValidateStructureLine`, whose message is written for a file being read, and passes `1` for the line number. Typing `(Z)` into the Create-new-O-glycan window gives, verbatim:

```
Could not parse glycan structure in 'OGlycan_Custom.gdb' at line 1: "(Z)". Unrecognized character 'Z'. Allowed: parentheses and one of HNAGFPSYCXK. ...
```

Line 1 of that file is the `# ====` rule at the top of the banner. Nothing was written, so there is no line to look at.

**Impact:**
The user is sent to a file, and to a line of it that has nothing to do with what they typed, for an entry that was refused before it reached disk. Every other refusal on this path names the entry instead (`Could not add the glycan "..."`), so this one is the odd one out as well as the wrong one.

**Suggested fix:**
Give the character check a caller-supplied prefix, or wrap it here, so the typed-entry path produces `Could not add the glycan "(Z)": unrecognized character 'Z' ...` and the loader keeps the file-and-line form it needs.

**Affected test scenarios:**
`AnUnknownMonosaccharideCodeNamesTheFileAndTheLine` pins the loader's message. Nothing pins the message the window shows, which is why the two could diverge.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
```diff
@@ -233,6 +252,285 @@ public static string PersistCustomMonosaccharide(string name, string codeText, s
...
+                if (!byte.TryParse(counts[i].Value, NumberStyles.Integer, CultureInfo.InvariantCulture, out byte count))
+                {
+                    throw new MetaMorpheusException(
+                        $"Could not add the glycan \"{entry}\": the count for '{name}' must be a whole number between 0 and 255.");
+                }
+                total += count;
+            }
+
+            // Same reason as the structure case: a composition that totals nothing is a zero-mass glycan.
+            if (total == 0)
```
pcruzparri, line 480: **Non-blocking** — counts have a floor but no ceiling

**Problem:**
A total of zero is refused here, and the comment gives the reason: a glycan that means nothing would widen every box it landed in. The other end is open — each count only has to fit in a byte, and there is no cap on the total. `HexNAc(255)Hex(255)NeuAc(255)` passes, is written, and the window reports it added. Loading it back with ion generation on, which is what `GlycoSearchEngine` does:

```
HexNAc(60)Hex(60)NeuAc(60)       4.3 s        223,260 child ions
HexNAc(100)Hex(100)NeuAc(100)   11.4 s      1,020,100 child ions
HexNAc(255)Hex(255)NeuAc(255)  229.0 s     16,711,680 child ions
```

Startup is unaffected — `LoadGlycans` passes `ToGenerateIons: false` — so this lands on the first GlycoSearch that selects the database.

**Impact:**
The search does not fail, it stops responding, for minutes, with no indication that the glycan just added is the reason. Out-of-scope note: the combinatorial part is `OGlycanCompositionCombinationChildIons`, which this PR does not touch and which a hand-edited database could always reach. What is new is a supported way in, and a validator that has taken a view on one end of the range and not the other.

**Suggested fix:**
Cap the total the same way the zero case is handled — a ceiling on the summed count, or on the resulting mass, refused with the same kind of message.

**Affected test scenarios:**
`ACountTooLargeToStoreIsRefused` covers one count above 255. A legal-but-enormous total has nothing on it.

pcruzparri (changes requested):
Opus 5 max-effort code review — 1 finding(s) at severity >= High, posted as inline comments.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
```diff
@@ -233,6 +252,285 @@ public static string PersistCustomMonosaccharide(string name, string codeText, s
...
+                throw new MetaMorpheusException($"Could not add the glycan \"{entry}\": {ex.Message}", ex);
+            }
+        }
+
+        /// <summary>
+        /// A composition line: a monosaccharide name and a count, repeated, e.g. HexNAc(2)Hex(5). Shared
+        /// so the loader and the entry validator cannot drift apart about what one looks like -- which is
+        /// exactly how the format-detection bug got in.
+        /// </summary>
+        private const string CompositionShape = @"^(?:(?<name>[A-Za-z][A-Za-z0-9]*)\((?<count>\d+)\))+$";
```
RayMSMS, line 436: Problem: order is never checked, so duplicates get in. Hex(1)HexNAc(1), HexNAc(1)Hex(1) and HexNAc(1)Hex(1)Fuc(0) all pass. ContainsEntry compares exact strings, so a user can add all three to the same file. That puts three copies of one glycan in the database, which means duplicate glycan boxes and duplicate candidates. If HexNAc(1)Hex(1) is the correct format, the PR should either:
(1) compare entries by their parsed kind[] instead of the raw string, or
(2) rewrite each entry into the standard order before saving

MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
```diff
@@ -415,16 +759,18 @@ private static bool IsCommentOrBlank(string line)
         // MonosaccharidesCustom.tsv). Struct2Glycan silently miscounts unknown chars (no entry in
         // CharMassDic), so we pre-validate to give the user a clear error instead of a silently
         // wrong glycan mass.
-        private static void ValidateStructureLine(string trimmedLine)
+        private static void ValidateStructureLine(string trimmedLine, string filePath, int lineNumber)
         {
             foreach (char c in trimmedLine)
             {
                 if (c == '(' || c == ')') continue;
```
RayMSMS, line 766: Multiple roots aren't rejected. (N)(H) balances, so it may pass validation. That's worth one guard or pop up a exception message. The potential bug existed for a while. But if we provide more way for user to typo the new glycan. We need more exception handling.

PR #2792, opened 2026-09-03
Let a user build their own N-glycan database

trishorts:
> **Project of origin:** `customGlycanDb`

🤖 **Stacked on #2791, which is stacked on #2787.** Merge order is #2787 → #2791 → this. The diff shows their commits until they land.

The N-glycan half of the pair `MonosaccharidesCustom.tsv` already promises the user: *"so they can be used inside custom glycan database entries (**NGlycan_Custom.gdb** / OGlycan_Custom.gdb, or any user-supplied .gdb file)"*.

Everything this rests on came with #2791 — the seeding recipe, the comment-tolerant loaders, `PersistCustomGlycan` and its validation, the empty-database guard. What is left is genuinely small, and that is the point of stacking them.

## What changes

`NGlycan_Custom.gdb` is embedded in `EngineLayer`, seeded once into the DataDir root through `CustomDataFile.EnsureExists`, and added to `NGlycanDatabasePaths` by name — root-level and added explicitly for the same reasons as its O-glycan sibling: `Product.wxs` gives `Glycan_Mods` and both subfolders a `<RemoveFolder On="both"/>`, and `LoadGlycans` only sweeps the subfolders. It joins `EveryKnownCustomFileIsSeededByStartup`.

Its template invites **composition** format — what `NGlycan.gdb` uses, and what most published N-glycan lists are distributed in — and documents structure format beside it, noting that structure carries the branching so the Y-ion series comes from the tree rather than from the counts. Either is legal: the loader picks the parser from the first data line, not from which database it is. There is a test for that.

## One change to #2791's GUI

`CustomGlycanWindow` gained a database selector rather than growing a second window. So the main window keeps **one** button covering both glycan classes, and no grid row had to be renumbered — those rows are hard-numbered, and adding a second button would have shifted every element below it.

## Verification

`CustomNGlycanDatabaseTests` — 7 tests: the template is embedded under the name the seeding asks for and has no data rows; a freshly seeded file loads to zero glycans without throwing (the first-launch case, since `LoadGlycans` reads every database eagerly); a composition added to it reads back on **both** sequons with the composition it was written with; a structure is legal in an N-glycan database too; startup offers the file; and the empty-database guard covers the N-glycan search branch, which loads through its own call.

**147 glyco and glycan tests pass**; 39 custom-file and custom-glycan tests pass. EngineLayer, GUI and Test build with 0 errors.

https://claude.ai/code/session_01Kr37s1DXoTUnMcn3htxnR4

### DIFF
--- a/MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
+++ b/MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
@@ -397,9 +397,16 @@ namespace EngineLayer
                 // Parsed exactly as LoadStructureGlycan would parse it: Struct2Glycan is the only thing that
                 // knows whether the nesting describes a tree it can actually build.
                 List<Glycan> parsed = Glycan.Struct2Glycan(entry, 1, isOGlycan);
-                if (parsed == null || parsed.Count == 0)
+
+                // "()" parses happily into a glycan of nothing, whose mass is zero. Searching for it is
+                // meaningless and it would widen every box it landed in, so it is refused here rather than
+                // left for the user to wonder about. The empty case is folded in: Struct2Glycan either
+                // throws or yields glycans, so "nothing came back" and "nothing in what came back" are the
+                // same answer to the user.
+                if (parsed == null || parsed.Count == 0 || parsed.All(g => g.Kind.Sum(count => (int)count) == 0))
                 {
-                    throw new MetaMorpheusException($"Could not add the glycan \"{entry}\": it did not parse into any glycan.");
+                    throw new MetaMorpheusException(
+                        $"Could not add the glycan \"{entry}\": it contains no monosaccharides.");
                 }
             }
             catch (MetaMorpheusException)
@@ -428,6 +435,7 @@ namespace EngineLayer
             CaptureCollection names = match.Groups["name"].Captures;
             CaptureCollection counts = match.Groups["count"].Captures;
             HashSet<string> seen = new HashSet<string>(StringComparer.Ordinal);
+            int total = 0;
 
             for (int i = 0; i < names.Count; i++)
             {
@@ -449,6 +457,14 @@ namespace EngineLayer
                     throw new MetaMorpheusException(
                         $"Could not add the glycan \"{entry}\": the count for '{name}' must be a whole number between 0 and 255.");
                 }
+                total += count;
+            }
+
+            // Same reason as the structure case: a composition that totals nothing is a zero-mass glycan.
+            if (total == 0)
+            {
+                throw new MetaMorpheusException(
+                    $"Could not add the glycan \"{entry}\": it contains no monosaccharides.");
             }
         }
 
--- a/MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
+++ b/MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 
 namespace EngineLayer
 {
@@ -21,13 +22,22 @@ namespace EngineLayer
         /// <returns> A glycan object collection </returns>
         public static IEnumerable<Glycan> LoadGlycan(string filePath, bool ToGenerateIons, bool IsOGlycan)
         {
+            // The format is inferred from the first DATA line; it is never declared. Comment and blank lines
+            // are skipped while sniffing because a documented database -- such as the seeded custom one --
+            // opens with a '#' banner, and judging the format from that banner routes the whole file to the
+            // wrong parser. A file with no data line is a legal empty database: either parser yields nothing
+            // for it, so which one is chosen does not matter.
             bool isKind = true;
             using (StreamReader lines = new StreamReader(filePath))
             {
                 while(lines.Peek() != -1)
                 {
                     string line = lines.ReadLine();
-                    if (!line.Contains("HexNAc"))  // use the first line to determine the format (kind / structure) of glycan database.
+                    if (IsCommentOrBlank(line))
+                    {
+                        continue;
+                    }
+                    if (!line.Contains("HexNAc"))  // use the first data line to determine the format (kind / structure) of glycan database.
                     {
                         isKind = false;
                     }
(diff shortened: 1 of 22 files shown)


nbollis (dismissed):

Alexander-Sol:

**Summary.** The small, clean top of a 3-PR stack (#2787 → #2791 → #2792). Its own diff is one commit: embed `NGlycan_Custom.gdb`, seed it at the DataDir root, add it to `NGlycanDatabasePaths` by name, and generalize the O-glycan GUI window into a two-item (O/N) selector. Wiring is correct, the settings grid isn't renumbered, serialized settings are untouched, and the 7 new tests plus shared coverage are thorough.

### 🟠 HIGH (confirmed, inherited from shared code — more reachable here)
Validation and load-time format detection use different heuristics. `ValidateGlycanLine`/`FormatOfExistingEntries` classify by `StartsWith("(")`; `LoadGlycan` (`GlycanDatabase.cs:40`) sniffs the first data line with `!line.Contains("HexNAc")` → structure.

**Scenario (reachable through the new N-glycan GUI):** user picks "N-glycan", types `Hex(5)`. Validation accepts it, it's written, the GUI reports success. On next launch `LoadGlycans` enumerates `NGlycanDatabasePaths` **eagerly** (`GlobalVariables.cs:618-629`); the first data line lacks `HexNAc` → routed to `LoadStructureGlycan` → `ValidateStructureLine("Hex(5)")` throws. That exception propagates out of `SetUpGlobalVariables` — i.e. a **startup crash**, not just a search-time one, and later N databases never populate. Second manifestation: a composition lacking both `Hex` and `HexNAc` (e.g. `NeuAc(2)Fuc(1)`) that isn't the first line is silently dropped by the `Contains("HexNAc") || Contains("Hex")` filter (`:589`).

Same root cause as #2791 — **fix `LoadGlycan`'s sniff once** (classify composition by "not starting with `(`") and both PRs are resolved. Reachability caveat: real N-glycans carry the HexNAc₂ chitobiose core, so a HexNAc-free entry is atypical — hence nits, not blocking — but the GUI enforces no N-glycan-ness and there is zero coverage for it.

**Nit:** add a test asserting `PersistCustomGlycan("Hex(5)", path, false)` then `LoadGlycan(path,...)` round-trips.

### Verified correct
GUI wiring (`IsOGlycanSelected => SelectedIndex == 0`, `SelectedDatabasePath`, handler rename matched in XAML), no grid-row renumbering, no serialized-settings back-compat risk, DataDir-root seeding + `File.Exists` guard, empty-database guard for the N branch, both-sequon (Nxs/Nxt) round-trip, `Hex(300)` byte-overflow refused at validation, InvariantCulture parsing, all-comment template contributes zero glycans.

trishorts:
The agent review on #2791 turned up a confirmed bug in shared code — `LoadGlycan` sniffed structure-vs-composition by whether the line contains "HexNAc", while the validator that tells the user their entry was accepted sniffs it by whether the line opens with `(`. A composition without that literal substring (`Hex(1)`, `NeuAc(1)`) was validated, written, and then routed to the structure parser on the next search.

That is `LoadGlycan`, so it hit this PR too. The fix is merged forward here — 93cb58026 plus the BOM cleanup in 0c83c7410. Full reasoning, the reproduction, and the classification check across all eleven glycan databases in the repo are on #2791: https://github.com/smith-chem-wisc/MetaMorpheus/pull/2791#issuecomment-5767900041

2198 glyco, custom-file and calibration tests pass on this branch; the solution builds with 0 errors.

trishorts:
@Alexander-Sol Apologies — my earlier comment here answered #2791's review and not this one. Read properly now, and both of your points were right, including the one that says this is worse than #2791's review made it sound.

### You're right that it's a startup crash, not a search failure

I checked `LoadGlycans`. It calls `LoadGlycan` eagerly on **both** path lists inside `SetUpGlobalVariables`:

```csharp
foreach (var path in OGlycanDatabasePaths) { var oGlycans = GlycanDatabase.LoadGlycan(path, false, true); … }
foreach (var path in NGlycanDatabasePaths) { var nGlycans = GlycanDatabase.LoadGlycan(path, false, false); … }
```

So a misclassified line doesn't fail a search — it stops MetaMorpheus opening, on both sides of the stack, for a glycan the user was told had been added. My write-up on #2791 described it as a search-time failure; that understated it and I've corrected it there.

Fixed once in `LoadGlycan`, as you said, so #2791 and this PR are both resolved by 93cb58026.

### The silent drop is a second bug, and it wasn't fixed by that

This is the part I'd have missed entirely without your review. `LoadKindGlycan` decided what counted as a data line with:

```csharp
if (!(line.Contains("HexNAc") || line.Contains("Hex"))) continue;
```

so `NeuAc(2)Fuc(1)` — which `ValidateCompositionEntry` accepts, and which `PersistCustomGlycan` writes and reports as added — is dropped by the loader without an error, a warning or a glycan. It isn't first-line-dependent, so unifying the format sniff does nothing for it.

That substring test was a stand-in for "is this a data line" from before comments were skipped explicitly, and comments are handled above it now. It's a shape test instead, against the same pattern the entry validator uses:

```csharp
private const string CompositionShape = @"^(?:(?<name>[A-Za-z][A-Za-z0-9]*)\((?<count>\d+)\))+$";
```

One constant, shared by `LoadKindGlycan` and `ValidateCompositionEntry` — the two drifting apart is what the HIGH finding *was*, so they no longer get to hold separate opinions.

Tolerance for a line that genuinely isn't a composition is unchanged: still skipped, now by shape rather than by substring, pinned by `ALineThatIsNotACompositionAtAllIsStillSkipped`. What does change is that a line which *is* composition-shaped but won't parse — `Nonsense(1)` — now raises a named error instead of a bare `KeyNotFoundException` out of `SetUpGlobalVariables`. Given where that runs, an unnamed startup crash was the worse of the two.

### The nit you asked for

`ACompositionOfNeitherHexNorHexNAcIsNotSilentlyDropped` does the round-trip on the N-glycan side: seed a composition database, `PersistCustomGlycan("NeuAc(2)Fuc(1)", path, false)`, load it back, assert both compositions survive on both sequons. Plus `ACompositionShapedLineThatWillNotParseNamesTheFileAndLine`.

### On reachability

Agreed that a HexNAc-free N-glycan is atypical — the chitobiose core means real ones carry HexNAc₂. But the window enforces no N-glycan-ness, the user is told the entry was added either way, and the consequence is a program that won't start. So I've treated it as blocking rather than as a nit, which I think is where you were pointing.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
```diff
@@ -233,6 +252,285 @@ public static string PersistCustomMonosaccharide(string name, string codeText, s
...
+            {
+                string name = names[i].Value;
+                if (!Glycan.NameCharDic.ContainsKey(name))
+                {
+                    string allowed = string.Join(", ", Glycan.NameCharDic.Keys);
+                    throw new MetaMorpheusException(
+                        $"Could not add the glycan \"{entry}\": '{name}' is not a monosaccharide MetaMorpheus knows. Known: {allowed}. " +
+                        "A monosaccharide it does not ship with must be declared in MonosaccharidesCustom.tsv first.");
+                }
+                if (!seen.Add(name))
```
pcruzparri, line 466: **Non-blocking** — A composition entry is checked as text, not as the composition it parses to

**Problem:** This guard dedupes on the monosaccharide *name*, and `ContainsEntry` compares the typed string, but the loader reduces both to a `Kind` slot. `Fuc` and `dHex` are two names for the same slot — `Glycan.NameCharDic["Fuc"].Item2` and `Glycan.NameCharDic["dHex"].Item2` are both 4 — so `Fuc(1)dHex(1)` clears `seen.Add(name)`, and `String2Kind`'s second assignment then overwrites the first. Same root cause one level up: `HexNAc(2)Hex(5)` and `Hex(5)HexNAc(2)` are the same glycan written two ways, so `ContainsEntry` lets both into one file.

**Impact:** Both run on this head. `PersistCustomGlycan("Fuc(1)dHex(1)", path, false)` returns normally — the GUI reports success — and the entry loads back as `kind=[0,0,0,0,1,0,0,0,0,0,0]`, mass 146.05791, identical to `Fuc(1)` on its own. The user is told a two-residue glycan was added and searches a one-residue one, with nothing said. The reorder case writes a second copy of a glycan already in the file: `HexNAc(2)Hex(5)` followed by `Hex(5)HexNAc(2)` gives four identical glycans out of `LoadGlycan` where there should be two. Worth noting alongside: the message just above at line 461 advertises `dHex` in its `Known:` list while the seeded template's name list does not mention it, so the one place a user learns the alias exists is the message that steers them into this.

**Suggested fix:** Compare the parsed composition rather than the text — key `seen` on `Glycan.NameCharDic[name].Item2` instead of `name`, and have `ContainsEntry` compare `String2Kind(line)` against `String2Kind(entry)` for a composition-format file. Then either add `dHex` to the template's name list or drop it from the `Known:` message.

**Affected test scenarios:** `CustomGlycanDatabaseTests` — `PersistCustomGlycan("Fuc(1)dHex(1)", path, false)` should be refused rather than silently collapsed; appending `Hex(5)HexNAc(2)` to a file that already holds `HexNAc(2)Hex(5)` should be refused as a duplicate.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanDatabase.cs
```diff
@@ -290,16 +588,52 @@ public static IEnumerable<Glycan> LoadKindGlycan(string filePath, bool ToGenerat
...
+                    string line = rawLine.Split('\t').First().Trim();
+
+                    // Whether the line has the SHAPE of a composition -- a name and a count, repeated. The
+                    // test used to be "does it contain Hex or HexNAc", a stand-in for "is this a data line"
+                    // from before comments were skipped above. It also threw away, without a word, every
+                    // legitimate composition built from neither: NeuAc(2)Fuc(1) is one, and the entry
+                    // validator accepts it, so a user could be told a glycan had been added and then have
+                    // it silently vanish from the search. Tolerance for a line that is not a composition at
+                    // all is unchanged -- it is still skipped, now by shape rather than by substring.
+                    if (!Regex.IsMatch(line, CompositionShape))
```
pcruzparri, line 614: **Non-blocking** — The new shape test silently drops composition lines the old one loaded

**Problem:** The substring test this replaces handed the whole line to `String2Kind`, which splits on `(` and `)` and stops after the last pair — so anything trailing the composition was ignored rather than fatal. `CompositionShape` is anchored at both ends, so a line carrying anything after the final `)` no longer matches and is dropped by the `continue` below, without a word. That is the same silent-drop failure the commit above this one set out to remove.

**Impact:** Verified on this head: `String2Kind("HexNAc(2)Hex(5)   # high mannose")` still returns `[5,2,0,…]`, but a two-line database whose first line is that yields 2 glycans instead of 4 — the annotated glycan is gone, with no message and no count anywhere for the user to notice. Same for a hand-made database that puts the mass in a space-separated rather than tab-separated column (`HexNAc(2)Hex(5) 1216.4228`), a shape a user's own `.gdb` under `Glycan_Mods\NGlycan` plausibly has. Nothing that ships is affected — I classified every `.gdb` in the repo under both the old and the new rule and the kept-line sets are identical — so this only reaches databases users brought themselves, which is why it is not blocking. A malformed line is equally quiet: `HexNAc(2)Hex5)`, `HexNAc(2) Hex(5)` and `HexNAc[2]Hex[5]` all just disappear, while `LoadStructureGlycan` names file and line for an unrecognized character.

**Suggested fix:** Match against the part of the line up to and including the last `)` so trailing text is tolerated the way `String2Kind` already tolerates it. For a line that is neither a comment nor composition-shaped, raise it through `GlobalVariables.WarnHandler` naming the file and the line number rather than dropping it, which is the same contract `LoadStructureGlycan` gives its own bad lines.

**Affected test scenarios:** `CustomGlycanDatabaseTests` — a composition line with a trailing `#` note and one with a space-separated mass column should both load; a composition-shaped-but-malformed line should surface a warning naming the file and line rather than vanishing.

pcruzparri (commented):
Opus 5 max-effort code review — 0 finding(s) at severity >= High, posted as inline comments.

pcruzparri (approved):

PR #2794, opened 2026-09-03
Let a user pick individual glycans instead of a whole glycan database

trishorts:
> **Project of origin:** `glycanGui`

🤖 **Stacked on #2792** — which stacks on #2791 and #2787. Those must merge first, and this diff shows their commits until they do.

The Glyco Search task offers a whole glycan database and nothing smaller. This adds a tree beside the two database combo boxes: one group per database file, every glycan in it as a checkable row.

Checking nothing means the whole database, so every task written before this behaves exactly as it did, and a user who never opens the section sees no change.

## Why it is not only convenience

`BuildOGlycanBoxes` is combinations-with-repetition over the whole array, so box count grows as C(n+k-1, k), and `BuildNOGlycanBoxes` multiplies that by `GlobalNGlycans.Count + 1`. The default `MaximumOGlycanAllowed` is 4. Picking six glycans out of twenty-four collapses the search space rather than trimming it.

## The control is not new

`GlycoSearchTaskWindow` already hosts two filter-box + checkbox trees, for fixed and variable mods. A `Glycan` **is** a `Modification`, so this is a third instance of a pattern already in the file: `ModTypeForTreeViewModel` and `ModForTreeViewModel` are reused as they are, and there is no glycan-specific view-model.

Two things had to be added rather than reused.

**`GlobalVariables` knew every glycan but not where it came from.** `LoadGlycans` flattens all databases into `AllModsKnown` and discards the file each glycan was read from — which is exactly the grouping this needs. It now also records the glycans per database, ordered by mass.

**`FilterTree` matched only `ModName`.** That is enough for modifications, whose identifiers are English words: `ribosyl` finds `ADP-ribosylation on S`. A glycan's identifier is a composition code, so `HexNAc` would match **nothing** — even in `NGlycan.gdb`, which spells `HexNAc(1)` on disk but parses to `N1`. `FilterTree` gains an overload taking the text to search, defaulting to today's behaviour so the five existing search boxes are untouched, and each glycan row carries an expanded label the tree searches:

```
H5N4A2 on Nxs   —   Hex(5)HexNAc(4)NeuAc(2)   —   2204.77 Da
```

`HexNAc`, `NeuAc`, `2204` and `H5N4A2` all find that row.

## Two decisions worth stating

**The selection is stored as (database file name, `IdWithMotif`), never by `GlyId`.** `GlyId` is a positional index into the loaded array, so a saved index would quietly point at a different glycan if the `.gdb` were edited. The property is typed `List<(string, string)>` deliberately: `MetaMorpheusTask.tomlConfig` already registers a converter for that type, so it round-trips with no new serialization code.

**The subset is applied before the glycans reach `GlycanBox.GlobalOGlycans`**, and before any box is built, because `GlycanBox` identifies a glycan by its position in that array — `BuildOGlycanBoxes` enumerates `Range(0, GlobalOGlycans.Length)`. Filtering later would re-point every box at a different glycan; filtering here means the dense re-index falls out for free. It is deliberately **not** applied inside `GlycanDatabase.LoadGlycan`, which `GlobalVariables` shares to populate `AllModsKnown` for MetaDraw and which must keep seeing every glycan.

A selection naming none of a given database means that whole database, so choosing individual O-glycans does not silently narrow the N-glycan side too.

## What driving the window turned up

The tree was exercised through UI automation — ticking glycans across two databases, running the search box, saving the task. It reported its own state wrongly in three ways, all now fixed.

| | Before | After |
|---|---|---|
| Summary line | permanently "none checked", even after ticking a whole 24-glycan group | `6 of 830 entries checked, across 2 databases`, live |
| Database checkbox | stayed `Off` with children ticked | `Off` → `Indeterminate` → `On` |
| Expand 830 entries | ~4 s, UI blocked | 0.06 s |

The first two share a cause: `ModForTreeViewModel.Use` notifies only itself, and `VerifyCheckState` is a manual pull the window has to make. The modification trees only ever make it while **restoring** a saved task, which is why a group there still reads as fully checked right after you uncheck something inside it. Subscribing to the children fixes both, and does it in the window rather than in the shared view-model, which MetaDraw also uses.

The re-entrancy guard on that handler is load-bearing, not defensive: `VerifyCheckState` assigns the parent's `Use`, and a non-null assignment cascades back down to every child, each raising `PropertyChanged` again. Ticking the last box in a group would otherwise recurse forever.

The third was the odd one. The two custom databases hold no glycans, and they were the only groups rendering as partly selected — exactly backwards. `ModTypeForTreeViewModel` takes `bad` as its second constructor argument, **not** `use`, so `Use` is never initialized and a `bool?` defaults to `null`, which is the indeterminate box. A group with children is corrected by the `VerifyCheckState` pass; an empty one has nothing to inspect, so it has to be said outright.

Also from that run: the tree virtualizes now, and `AutomationProperties.Name` is set on the item container rather than on controls inside the header — a `TreeViewItem` takes its automation name from the container, so every row had been announcing `GuiFunctions.ModForTreeViewModel` to a screen reader.

The count is described as **entries** rather than glycans because a glycan is listed once per attachment site: the 12 lines of `OGlycan.gdb` are 24 rows here, and calling them glycans invites the reader to compare the number with the file and find it wrong.

## Where the code lives

`Test.csproj` references CMD, EngineLayer, GuiFunctions and TaskLayer — **not GUI** — so the test assembly cannot see `MetaMorpheusGUI` types and coverlet emits no rows for that project. Code left in a `.xaml.cs` is neither covered nor uncovered; it is absent from the report. So the logic is in **GuiFunctions**, where it can be tested, and the window keeps the WPF.

- **`GuiFunctions/ViewModels/GlycanSelectionViewModel.cs`** — builds the tree, settles the database checkboxes, maintains the summary, restores a saved selection, and reads one back through `ToSelectedGlycans()`. It takes its glycans as a constructor argument rather than reading `GlobalVariables`, so a test can supply its own. This follows `FragmentationParamsViewModel`, which takes the task's parameter objects in and gives a domain object back, and whose control code-behind is 38 lines.
- **`GuiFunctions/ModTreeFilter.cs`** — the filtering itself. `SearchModifications` keeps only the WPF: read the `TextBox`, assign the `TreeView`'s `DataContext`.
- **`GlycoSearchTaskWindow.xaml.cs`** grows by **27 lines**: one field, one `new GlycanSelectionViewModel(...)`, two `DataContext` assignments, one read-back, one filter call.

## Tests

`Test/GuiTests/GlycanSelectionViewModelTests.cs` (28) and `Test/GuiTests/ModTreeFilterTests.cs` (14) — 42 tests, `[TestFixture]`, instance methods, `Assert.That`, hand-written fixtures, no `GlobalVariables` dependency and so no static-state hazard.

They pin the two behaviours previously confirmed only by driving the window: that a database checkbox goes `Off` → `Indeterminate` → `On` as its children are ticked, and that ticking the last box in a group **terminates at all** — settling the parent cascades back down and re-raises `PropertyChanged`, so without the guard it recurses forever.

They also caught a bug the GUI had been showing. `Glycan.NameCharDic` carries aliases as well as canonical names — `dHex` and `Fuc` share a code and a slot — so inverting it has to choose between them, and it was choosing by enumeration order and landing on the alias. Ordering by slot then ordinally by name is deterministic and picks the canonical spelling, which is also what the shipped databases write on disk and therefore what a user is likely to type.

Two more fixtures cover the other side of the boundary, which is EngineLayer and TaskLayer rather than GUI:

- **`Test/GlycoSearchParametersTomlTests.cs`** (6) — the selection surviving save-and-reopen. The one that matters is the compatibility case, tested the way it will actually be met: every GlycoSearch toml already on a user''s disk has no `SelectedGlycans` line at all, so the test writes a task with a selection, strips the line back out, and reads it again. It asserts first that the line was there to remove, because a test that strips nothing would pass for the wrong reason.
- **`Test/GlycoSearchEngineSelectedGlycansTests.cs`** (10) — the engine honouring a selection. That no selection, an empty one, and one naming only the *other* database each leave this database whole; that a selection matching nothing falls back rather than handing `BuildOGlycanBoxes` an empty array; that narrowing collapses the box count; and — the reason the filter runs before `GlobalOGlycans` is assigned — that every box resolves to a glycan the selection actually named. The N side gets the same through `GlobalNGlycans`, whose negative keys must come out `-1..-m` with no gaps. `[NonParallelizable]`, and TearDown restores both databases, because `GlobalVariables` is static and never reset between fixtures and other glyco fixtures assert baked-in counts.

**58 tests for this feature; 500 pass** across the glyco, GUI and toml suites.

## Also here

`mzLib` bumped to **1.0.589** in all six projects (its own commit).

## Not in this PR

**The same stale-parent-checkbox behaviour in the modification trees.** It is fixed here for the glycan tree without touching `ModForTreeViewModel`. Fixing it at the source would improve the Search and GPTMD windows too, but it is a hot type MetaDraw shares and `MetaDrawSettingsAndViewsTest.cs` pins its semantics, so it wants its own PR.

**Known conflict:** #2541 deletes `GlycanBox.cs` and rewrites `GlycoSearchEngine.cs:63-95`. Whichever lands first, the other needs rework.

https://claude.ai/code/session_01SYPSBFRWvhAPXaR6RktVuR

### DIFF
--- a/MetaMorpheus/CMD/CMD.csproj
+++ b/MetaMorpheus/CMD/CMD.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
@@ -24,7 +24,7 @@
     <PackageReference Include="Microsoft.ML" Version="3.0.1" />
     <PackageReference Include="Microsoft.ML.CpuMath" Version="3.0.1" />
     <PackageReference Include="Microsoft.ML.FastTree" Version="3.0.1" />
-    <PackageReference Include="mzLib" Version="1.0.588" />
+    <PackageReference Include="mzLib" Version="1.0.589" />
     <PackageReference Include="Nett" Version="0.15.0" />
     <PackageReference Include="SQLite.Interop.dll" Version="1.0.103" />
     <PackageReference Include="System.Data.SQLite.Core" Version="1.0.118" />
--- a/MetaMorpheus/CMD/Program.cs
+++ b/MetaMorpheus/CMD/Program.cs
@@ -207,6 +207,16 @@ namespace MetaMorpheusCommandLine
                 }
             }
 
+            // anything else startup noticed, such as a custom protease that collided with a built-in
+            foreach (var warning in GlobalVariables.StartupWarnings)
+            {
+                if (settings.Verbosity == CommandLineSettings.VerbosityType.minimal || settings.Verbosity == CommandLineSettings.VerbosityType.normal)
+                {
+                    Console.WriteLine(warning);
+                }
+            }
+            GlobalVariables.StartupWarnings.Clear();
+
             List<(string, MetaMorpheusTask)> taskList = new List<(string, MetaMorpheusTask)>();
 
             var tasks = settings.Tasks.ToList();
--- a/MetaMorpheus/EngineLayer/CrosslinkSearch/Crosslinker.cs
+++ b/MetaMorpheus/EngineLayer/CrosslinkSearch/Crosslinker.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.IO;
 using MassSpectrometry;
 using System.Globalization;
@@ -75,6 +76,13 @@ namespace EngineLayer
             return cleaveDissociationTypes;
         }
 
+        /// <summary>
+        /// Columns in a crosslinker tsv row: Name, CrosslinkAminoAcid, CrosslinkerAminoAcid2, Cleavable,
+        /// DissociationType, CrosslinkerTotalMass, CrosslinkerShortMass, CrosslinkerLongMass, QuenchMassH2O,
+        /// QuenchMassNH2, QuenchMassTris.
+        /// </summary>
+        private const int ExpectedColumnCount = 11;
+
         public static IEnumerable<Crosslinker> LoadCrosslinkers(string CrosslinkerLocation)
         {
             using (StreamReader crosslinkers = new StreamReader(CrosslinkerLocation))
@@ -88,6 +96,20 @@ namespace EngineLayer
                     if (lineCount == 1)
                     {
                         continue;
+                    }
+
+                    // CustomCrosslinkers.tsv is edited by hand, so a blank line or a '#' note in it is a
+                    // thing that happens. Both used to reach ParseCrosslinkerFromString and come back as
+                    // an unhandled IndexOutOfRangeException during startup, before any window opened.
+                    if (string.IsNullOrWhiteSpace(line) || line.StartsWith("#", StringComparison.Ordinal))
+                    {
+                        continue;
+                    }
+
+                    if (line.Split('\t').Length < ExpectedColumnCount)
+                    {
+                        throw new MetaMorpheusException($"Line {lineCount} of {Path.GetFileName(CrosslinkerLocation)} has "
+                            + $"{line.Split('\t').Length} tab-separated column(s); {ExpectedColumnCount} are required. The line was: {line}");
                     }
 
                     yield return ParseCrosslinkerFromString(line);
--- a/MetaMorpheus/EngineLayer/EngineLayer.csproj
+++ b/MetaMorpheus/EngineLayer/EngineLayer.csproj
@@ -38,7 +38,7 @@
     <PackageReference Include="Microsoft.ML" Version="3.0.1" />
     <PackageReference Include="Microsoft.ML.CpuMath" Version="3.0.1" />
     <PackageReference Include="Microsoft.ML.FastTree" Version="3.0.1" />
-    <PackageReference Include="mzLib" Version="1.0.588" />
+    <PackageReference Include="mzLib" Version="1.0.589" />
     <PackageReference Include="Nett" Version="0.15.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.4" />
     <PackageReference Include="SQLite.Interop.dll" Version="1.0.103" />
--- a/MetaMorpheus/EngineLayer/EngineLayer.csproj
+++ b/MetaMorpheus/EngineLayer/EngineLayer.csproj
@@ -20,10 +20,16 @@
 
   <ItemGroup>
     <None Remove="Glycan_Mods\MonosaccharidesCustom.tsv" />
+    <None Remove="Glycan_Mods\OGlycan_Custom.gdb" />
   </ItemGroup>
   
+  <!-- Templates for the user-editable custom files. Embedded, deliberately: they must NOT carry a
+       CopyToOutputDirectory rule and must NOT appear in Product.wxs as a <File>, because either one
+       lets a build or an installer overwrite the copy the user has edited. That was #2752. The whole
+       recipe is written down on EngineLayer\Util\CustomDataFile.cs. -->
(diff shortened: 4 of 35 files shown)


acesnik:
🤖 The stack itself is sound — the four branches nest exactly. #2791 adds three commits on top of #2787's three, #2792 adds one on top of those, and this PR adds six on top of that; `a297cddf...822e68e2` is ahead 10, behind 0, so nothing has drifted between them.

What's missing is only GitHub's native stacked-PR view, and two separate things block it. The heads are all on the `trishorts` fork, and the feature requires every branch in a stack to live in the same repository. Independently of that, each PR here bases on `master`, where a stack needs each PR to base on the branch below it.

Getting it would mean pushing the four branches to `smith-chem-wisc/MetaMorpheus` and setting each base to the one beneath: #2787 stays on `master`, #2791 onto `feat/custom-file-policy`, #2792 onto `feat/custom-oglycan-database`, this one onto `feat/custom-nglycan-database`. `gh extension install github/gh-stack` manages it from there.

One snag if you do. `feat/custom-file-policy`'s tip is `7cb5442d`, a merge of `master` that the upper three don't carry — they fork from `a297cddf`, the commit before it. So #2791 would sit 3 behind its new base and wants a merge-down before the chain is flush.

The moment is cheap for it: #2791, #2792 and this one carry no reviews yet, and #2787 stays on `master` so its approval isn't disturbed. In exchange each PR would show only its own diff, and GitHub would retarget the next one as the one below merges, rather than the order being maintained by hand in the bodies.

trishorts:
@acesnik Sorry this sat unanswered — thank you for actually measuring the stack rather than assuming it. The nesting you described was right at the time, and the `ahead 10, behind 0` check is the one I'd have wanted.

I'm going to decline the restructure, and the reasons have both grown since you wrote it.

**The proposed base is gone.** #2787 was squash-merged into `master` on 2026-09-17 as `e38ec2cb`, so `feat/custom-file-policy` no longer exists as a base for #2791 to sit on. The stack is two deep now rather than four, and #2791 bases on plain `master` — which removes the `7cb5442d` snag you flagged along with the tier it belonged to.

**The window you correctly identified as cheap has closed.** You noted the moment was cheap precisely because "#2791, #2792 and this one carry no reviews yet." That stopped being true — both have been reviewed since, and #2791 has been approved and reset twice inside four days. To be precise about what a re-base would and wouldn't cost: the review record on those two is issue-level rather than inline, so it would survive; what wouldn't is the approval state, which any force-push dismisses again. Having just reset @nbollis's approval twice in one day for a real bug, I'd rather not spend a third on presentation.

The rule behind it isn't hypothetical either. This stack already faced the same choice from the other side: after the squash-merge, the tidier option was rebasing to drop #2787's now-duplicated commits, and that one *would* have orphaned inline threads, so it was rejected. Merge forward, never rebase — applied consistently, including when it costs a cleaner history.

The cost you named is real and I'm accepting it: each PR shows the stack's cumulative diff rather than only its own, and the ordering is maintained by hand in the bodies rather than by GitHub. For a two-deep stack that's a tolerable amount of bookkeeping.

Worth flagging while you're here: this PR can no longer be merged forward mechanically. Upstream reshaped `GlycoSearchEngine`'s constructor — `peptideIndex` is now `IEnumerable<IBioPolymerWithSetMods>`, `fragmentIndex` is `Indexing.FragmentIndex`, `maxGlycanBoxMass` and `candidates` were added — and dropped the `selectedGlycans` parameter this PR exists to add, while `GlycoSearchTask` moved to `MakeSearchEngine` instead of constructing inline. Re-threading the selection through both is feature work rather than a merge. A merge was attempted on 2026-09-17 and aborted clean, so this branch is untouched and still sits on #2792.

PR #2795, opened 2026-09-03
Bugfixes: Rna custom fragments and Read in Multiple Decon Toml

nbollis:
RNA custom fragments were prevented from being selected in the popup due to bad logic. This has fixed it. 

Multiple Decon runs fine, but could not read the toml back in. Now we can.


trishorts (commented):
Reviewed the RNA custom-fragments and multiple-decon-toml changes. The toml half follows the existing `ConfigureType` convention exactly and I have no concerns about it. One blocker (red required check), one Major in the GUI, two minors.

MetaMorpheus/GUI/Views/CustomFragmentationWindow.xaml.cs
```diff
@@ -39,6 +33,15 @@ public CustomFragmentationWindow(List<ProductType> list, bool isRna = false)
                 }
             }
 
+            // Update options on mode change
+            GuiGlobalParamsViewModel.Instance.PropertyChanged += (sender, args) =>
```
trishorts: **This subscribes a lambda to a process-lifetime singleton and never detaches it, and a new `CustomFragmentationWindow` is constructed every time a task window opens.**

Five call sites each build one in their own constructor -- `SearchTaskWindow.xaml.cs:346`, `CalibrateTaskWindow.xaml.cs:103`, `GPTMDTaskWindow.xaml.cs:117`, `GlycoSearchTaskWindow.xaml.cs:211`, `XLSearchTaskWindow.xaml.cs:173` -- so the handler list on `GuiGlobalParamsViewModel.Instance` grows by one per task window the user opens, for the life of the process. Nothing ever removes them: the lambda is not stored, so there is no `-=` to write, and `OnClosing` cancels the close and hides the window instead, so the instance is never finalised either. Each captured window keeps its whole visual tree alive through the singleton.

The functional half matters more than the memory: after N opens, one RNA-mode toggle calls `PopulateChoices()` on N stale windows, each of which rebuilds its `TheList` and discards whatever the user had checked.

Both go away by keeping the handler and detaching it:

```csharp
private readonly PropertyChangedEventHandler _modeChanged;
...
_modeChanged = (s, args) => { if (args.PropertyName == nameof(GuiGlobalParamsViewModel.IsRnaMode)) PopulateChoices(); };
GuiGlobalParamsViewModel.Instance.PropertyChanged += _modeChanged;
```

and `-=` in `OnClosing` on the branch that actually closes.

Worth noting this is the same defect #2799 is fixing for CMD's static engine events this week -- an undetached subscription to something that outlives the subscriber. Same shape, different singleton.

MetaMorpheus/Test/Transcriptomics/TestCalibration.cs
```diff
@@ -47,73 +48,84 @@ public static void Setup()
     }
 
     [Test]
+    [NonParallelizable]
```
trishorts, line 51: **`Test on Windows` is red on this PR, and the failure is in a file the PR changes.**

```
Failed CalibrationTask_ReducesMassError [6 s]
  Expected more OligoSpectralMatches after calibration.
  Assert.That(secondOsms.Count, Is.GreaterThan(firstOsms.Count))
  Expected: greater than 175
  But was:  175
```
(`TestCalibration.cs:106`, run 34265929075)

`Test on Windows` is the only required status check on master, so this blocks regardless of the rest.

Worth saying that I think the change here is the right one and the failure is a *result* of it. Pinning `GlobalVariables.AnalyteType = AnalyteType.Oligo` and marking the fixture `[NonParallelizable]` is what made this test deterministic; before that it was reading whatever analyte type a neighbouring test had left behind. So the honest reading is not "this PR broke calibration" but "this PR removed the accident that made the assertion pass", and RNA calibration does not in fact add OSMs on `RnaStandard_Subset.mzML`.

Which of the two that means -- the assertion was always too strong for this fixture, or calibration genuinely is not helping here -- is worth deciding deliberately rather than by whichever makes CI green. The second assertion (`QValue <= 0.01`) and the mass-error one are the claims that matter; the raw count is the weakest of the three.

MetaMorpheus/GuiFunctions/MetaDraw/PlotModelStat.cs
```diff
@@ -582,8 +582,8 @@ private Dictionary<string, int> GetCategoryDictForGroup(int plotType, Observable
                 Dictionary<string, Dictionary<string, ObservableCollection<SpectrumMatchFromTsv>>> groupedPsmsBySourceFile,
                 bool isGroupingEnabled)
         {
-            int end = dictsBySourceFile.Values.Max(p => p.Max(v => int.Parse(v.Key)));
-            int start = dictsBySourceFile.Values.Min(p => p.Min(v => int.Parse(v.Key)));
+            int end = dictsBySourceFile.Values.Max(p => p.Count == 0 ? 0 : p.Max(v => int.Parse(v.Key)));
```
trishorts, line 585: **Minor, and out of the stated scope of the PR.** `0` is a real bin value here, not a neutral element, so an empty dictionary does not drop out of the `Max`/`Min` -- it votes.

If the real keys are `1..10`, `start` becomes `0` and you get a leading empty bin; if they are all negative, `end` becomes `0` and you get a run of trailing empty ones. `numBins = end - start + 1` then drives the label spacing, so an empty source file quietly changes the axis of the other files' plot.

`SelectMany` over the non-empty ones sidesteps the sentinel:

```csharp
var keys = dictsBySourceFile.Values.SelectMany(p => p).Select(v => int.Parse(v.Key)).ToList();
if (keys.Count == 0) return new Dictionary<string, int>();
int end = keys.Max(), start = keys.Min();
```

Also: nothing in the diff explains what made this reachable. If it is the multiple-deconvolution work, a one-line comment or a test would help; if it is unrelated, it may be cleaner in its own PR.

MetaMorpheus/Test/TestToml.cs
```diff
@@ -602,6 +602,24 @@ public static void TestToml_IsoDecWithOxyRiboAverageResidueModel()
...
+            var searchTaskLoaded = Toml.ReadFile<SearchTask>(tomlPath, MetaMorpheusTask.tomlConfig);
+
+            Assert.That(searchTaskLoaded.CommonParameters.PrecursorDeconvolutionParameters, Is.TypeOf<MultipleDeconParameters>());
+            var multipleParams = (MultipleDeconParameters)searchTaskLoaded.CommonParameters.PrecursorDeconvolutionParameters;
+
+            Assert.That(multipleParams.Parameters.Count(), Is.EqualTo(2));
+            Assert.That(multipleParams.Parameters.First(), Is.TypeOf<ClassicDeconvolutionParameters>());
+            Assert.That(multipleParams.Parameters.Last(), Is.TypeOf<IsoDecDeconvolutionParameters>());
+            Assert.That(searchTaskLoaded.CommonParameters.ProductDeconvolutionParameters, Is.TypeOf<ClassicDeconvolutionParameters>());
+            GlobalVariables.AnalyteType = AnalyteType.Peptide;
```
trishorts: Two things about this test, both small.

**The restore is not in a `finally`.** If either `Assert` above it throws, `GlobalVariables.AnalyteType` is left on `Oligo` for every test that runs afterwards in the same process -- which is precisely the cross-test leakage the `TestCalibration.cs` half of this PR adds a `try`/`finally` to fix. Same treatment here, and `[NonParallelizable]` too, since it mutates a global while other fixtures read it.

**It asserts the shape but not the values.** The `CreateInstance` seed for `MultipleDeconParameters` hard-codes charges `1..20`, `Polarity.Positive`, `Averagine` and a proton mass -- all of which are wrong for oligos, and all of which this toml happens to override. `Is.TypeOf<...>` and `Count == 2` would still pass if any one of them stopped round-tripping. One line pins it:

```csharp
Assert.That(multipleParams.Polarity, Is.EqualTo(Polarity.Negative));
Assert.That(multipleParams.MinAssumedChargeState, Is.EqualTo(-40));
```

The `"Multiple"` arm and the `CreateInstance`/`IgnoreProperty` shape itself match the existing `ClassicDeconvolution` / `IsoDecDeconvolution` entries exactly, which is the right call.

MetaMorpheus/GuiFunctions/MetaDraw/PlotModelStat.cs
nbollis: Not multiple decon work, just a latent bug that caused a plotting crash.

MetaMorpheus/Test/Transcriptomics/TestCalibration.cs
nbollis: Good note. Made the original search tolerance exceptionally small to assure that we get more search results. The test is really about the mass errors being reduced.

MetaMorpheus/Test/TestToml.cs
nbollis: Added more assertions as well as moving the analyte type restoration to a finally block

MetaMorpheus/GUI/Views/CustomFragmentationWindow.xaml.cs
nbollis: implemented this suggestion

MetaMorpheus/GuiFunctions/MetaDraw/PlotModelStat.cs
trishorts: Understood on scope — a latent crash fix riding along is fine by me, and I'm not re-opening this thread. But the scope answer and my actual point are different things, so recording the detail here so it isn't lost.

The crash is fixed. The substance of my comment isn't: `0` is still a real bin value, not a neutral element, so an empty source file still *votes* in the `Min`/`Max`.

```csharp
int end   = dictsBySourceFile.Values.Max(p => p.Count == 0 ? 0 : p.Max(v => int.Parse(v.Key)));
int start = dictsBySourceFile.Values.Min(p => p.Count == 0 ? 0 : p.Min(v => int.Parse(v.Key)));
```

With real notches `1..10` and one empty file, `start` becomes `0` rather than `1`, `numBins` goes 10 -> 11, and you get a leading empty bin. All-negative notches do the same at the other end. Not a crash, just a spurious bin and slightly-off label spacing — which is why I filed it Minor and why I'm happy to leave it.

**The part worth knowing:** `TestPlotModelStat_Notch_GroupsAcrossFiles_HandlesEmptySourceFile` can't catch it. The fixture uses notches `0.000` / `1.003` / `2.007`, so the real keys are `{0, 1, 2}` — `start` is `0` either way and the substituted `0` is indistinguishable from a real key. Change those three to `1.003` / `2.007` / `3.010` and the leading empty bin shows up in the series.

If you do want it tightened later, filtering before the aggregate is the shape — with the caveat that a naive `.Where(p => p.Count > 0)` reintroduces your original crash when *every* file is empty, so that case needs its own branch. Your current version at least degrades to `start == end == 0` there. Genuinely fine as a follow-up or never.

MetaMorpheus/GUI/Views/CustomFragmentationWindow.xaml.cs
```diff
@@ -130,6 +135,8 @@ private void OnClosing(object sender, CancelEventArgs e)
                 this.Visibility = Visibility.Hidden;
                 e.Cancel = true;
             }
+
+            GuiGlobalParamsViewModel.Instance.PropertyChanged -= _modeChanged;
```
trishorts, line 139: **Minor, non-blocking — the unsubscribe sits outside the guard, so it also fires on the path where the window *survives*.**

The detach itself is right and it fixes what I filed. But `OnClosing` has two outcomes and this line runs on both:

```csharp
if (this.Visibility == Visibility.Visible)
{
    this.Visibility = Visibility.Hidden;
    e.Cancel = true;          // <- window is NOT closing, it's being hidden
}

GuiGlobalParamsViewModel.Instance.PropertyChanged -= _modeChanged;   // <- runs anyway
```

Clicking the window's X while it's visible takes the cancel path: hidden, close cancelled, instance alive — and now unsubscribed for the rest of its life. A later `.Show()` (e.g. `SearchTaskWindow.xaml.cs:1104`) brings back a window that no longer repopulates on mode change.

**I want to be straight about the severity: I could not construct a user-visible consequence today, so treat this as latent, not a bug report.** `IsRnaMode` can't change while the popup exists — every task window is opened with `ShowDialog()` (`MainWindow.xaml.cs:2089`, and the pre-run path at `:2193`), so the main-window toggle at `MainWindow.xaml:938` is blocked, and the programmatic flip at `MainWindow.xaml.cs:2011-2016` runs in `AddPreRunTask`, after the dialog returns. The real fix for this PR's bug is the `isRna` field becoming `GuiGlobalParamsViewModel.Instance.IsRnaMode`, which is correct regardless.

So this only bites if task windows ever go non-modal or a mode toggle lands inside one. One line either way:

```csharp
if (this.Visibility == Visibility.Visible)
{
    this.Visibility = Visibility.Hidden;
    e.Cancel = true;
}
else
{
    GuiGlobalParamsViewModel.Instance.PropertyChanged -= _modeChanged;
}
```

Take it or leave it — the leak I filed is fixed either way, since the real closes (`Save`/`Cancel` set `Hidden` first, then the task window calls `.Close()`) all land on the else branch.

trishorts (approved):
Re-reviewed at `728528c7`. **All four of my earlier findings are addressed and `Test on Windows` is green** — that was my only blocker, so nothing is blocking now.

| # | Finding | Status |
|---|---|---|
| 1 | Handler subscribed to a process-lifetime singleton, never detached | Fixed — handler stored and detached in `OnClosing` |
| 2 | `CalibrationTask_ReducesMassError` red on a file the PR changes | Fixed at the root, not papered over |
| 3 | `0` votes in the bin `Min`/`Max` | Crash fixed; the original point still stands. Minor, non-blocking |
| 4 | `AnalyteType` restore not in a `finally` | Fixed, with a lot more assertions |

**On #1** — I checked the thing that would have made the fix cosmetic: whether `OnClosing` ever actually runs. It does. All five task windows call `CustomFragmentationWindow.Close()` from several places each (`SearchTaskWindow.xaml.cs:465`, `:506`, `:1148`, and the equivalents in the other four), and because `Save_Click`/`Cancel_Click` set `Hidden` first, those closes take the real-close branch. The leak is genuinely gone. One latent wart in the placement of the detach, commented inline — explicitly non-blocking, and I say there why I could not make it bite today.

**On #2** — this is the one I want to say something about, because "the test went green after the tolerances changed" is a shape that deserves suspicion, and it holds up:

- The root cause was global state. The test never set `GlobalVariables.AnalyteType`, so it was reading whatever ran before it. It now sets `Oligo` explicitly and restores in a `finally`.
- The tolerance change makes the assertion *more* discriminating, not less. At 20/10/5 ppm the window was already wider than the error, so calibration could not move the count — hence the 175 vs 175 that failed. At 10/3/1 ppm the uncalibrated search misses what the calibrated one recovers, which is the effect the test claims to measure.
- **All three original assertions are retained**, including the one that is actually the point (`secondAverage < firstAverage` on `MassDiffDa`). Nothing was deleted or loosened to get green.
- Verified it runs rather than being skipped: `Passed CalibrationTask_ReducesMassError [6 s]` in `Test on Windows (core)`, 1950 tests.

**On #4** — the `finally` is in and the test grew from a couple of assertions to about forty, including proof that the nested `Parameters` list round-trips as two entries of *different* types. That matters more than it looks: it's what shows Nett replaces the seed list in your new `CreateInstance` rather than appending to it, which was the one thing I'd have wondered about in the `ConfigureType<MultipleDeconParameters>` block.

One correction to my own comment there: **I also asked for `[NonParallelizable]`, and that ask was unnecessary.** There is no `[assembly: Parallelizable]` or `LevelOfParallelism` anywhere in the Test project, so NUnit runs fixtures sequentially by default and the attribute is a no-op — on `TestToml` where it's absent, and equally on `TestCalibration` where you added it. The `try`/`finally` was the half that mattered, and that's done.

**On #3** — leaving it resolved and not re-opening, but the reply there records what's still open: the crash is fixed, while the substance (an empty source file contributing a real `0` to the `Min`/`Max`) isn't. Worth knowing that the new `HandlesEmptySourceFile` test can't catch it either, because its notches start at `0` — so the substituted `0` is indistinguishable from a real key. Minor, and fine as a follow-up or never.

**Approving.** For the record this is 1 of the 3 approvals master requires here.

zhuoxinshi (approved):

RayMSMS (approved):

Issue #2797, opened 2026-09-04
MaxModificationIsoforms truncates peptidoform enumeration silently


trishorts:
### Summary

When `MaxModificationIsoforms` truncates a peptide's peptidoform enumeration, nothing reports it — no warning, no counter, no line in `results.txt`. The truncation is also *ordered*, not random, so the hypotheses it removes are removed systematically.

This is easy to hit on a GPTMD-annotated database with `MaxModsForPeptide` ≥ 3, and when it happens the affected peptide is searched against a fraction of its own hypothesis space with no indication in any output.

Observed with MetaMorpheus 1.1.5 / mzLib 1.0.589.

### Where it happens

`mzLib` — `Proteomics/ProteolyticDigestion/ProteolyticPeptide.cs`, `GetModifiedPeptides`:

```csharp
variable_modification_isoforms++;
if (variable_modification_isoforms == maximumVariableModificationIsoforms)
    yield break;
```

A hard `yield break` on an ordered enumeration. Two orderings decide which isoforms survive:

1. **Across modification counts — ascending.** `Omics/Digestion/DigestionProduct.cs`, `GetVariableModificationPatterns`:
   ```csharp
   for (int variable_modifications = 0; variable_modifications <= maxVariableMods; variable_modifications++)
   ```
   The 0-mod form is yielded first, then all 1-mod forms, then 2-mod, and so on — so **the cap always deletes from the top of the ladder**. The heavily modified isoforms that `MaxModsForPeptide` was raised to obtain are exactly the first ones cut. In effect the two parameters interact, and only their combination is meaningful.

2. **Within a position — alphabetical.** Candidates at one position live in a `SortedSet<Modification>`, and `Omics/Modifications/Modification.cs` `CompareTo` orders by `IdWithMotif` (ordinal string comparison) first, with `ModificationType` only as a tiebreaker. So when the cap bites, the surviving modification is decided by where its **name** falls in the alphabet.

### Worked example

Search of MassIVE `MSV000083304` (Jurkat, multi-protease), 12 files, 194,487 MS2 scans, Calibrate → GPTMD → Search, GPTMD database with 8,625 annotations, `MaxModsForPeptide = 4`, MC 0, ±5 ppm precursor.

`Q13885` (tubulin β-2A) residues 163–213 is a 51-mer carrying **33 GPTMD annotations at 24 positions**. The isoform count per modification tier (elementary symmetric polynomials of the per-position candidate counts) is:

[...]

PR #2798, opened 2026-09-04
Do not treat a failed retention-time prediction as a hydrophobicity of zero

trishorts:
> **Project of origin:** `phred`

### Summary

PEP builds a `HydrophobicityZScore` for every candidate from a predicted retention time. Two places coerced a **failed** prediction to zero:

```csharp
predictor.PredictRetentionTimeEquivalent(pep, out _) ?? 0
```

Zero is a real and extreme value on that scale, not a sentinel. So *"could not predict"* became *"disagrees with the observed retention time as badly as possible"* — the z-score saturates at its maximum of 10 and the model reads it as strong evidence against the candidate. The failure reason was discarded with `out _`, so nothing downstream could tell the two situations apart.

### What changes

**1. The reference distribution no longer absorbs failed predictions.** `ComputeRetentionTimeEquivalentValues` builds the per-file, per-RT-bin median and standard deviation that *every other* peptide's z-score is measured against. An unpredictable peptidoform did not merely mis-score itself — it entered that reference as a predicted hydrophobicity of zero and dragged the statistics for every peptide in its bin. Those peptides are now skipped.

**2. `PsmData.HasHydrophobicity`**, a companion feature in exactly the relationship `HasSpectralAngle` already has with `SpectralAngle`. The model can now distinguish *"predicted, and disagrees"* from *"could not be predicted"*, instead of being handed a value that means neither.

**3. Chronologer is constructed with `SequenceConversionHandlingMode.ReturnNull`** rather than the default `RemoveIncompatibleElements`. **This is the behavioural change worth your attention.** Under the default, a modification outside Chronologer's vocabulary is silently deleted and the residue encoded bare — so the model predicts the retention time of a *different molecule* and reports success. Chronologer supports roughly twenty modifications; a G-PTM-D search can enable many times that.

### How this surfaced

Within a single scan, two peptidoforms that the spectrum **cannot** separate received different predictions purely because of vocabulary membership:

| candidate | Chronologer encoding |
|---|---|
| `[N-term acetyl]SKGPAVGID…` | `^SKGPAVGID…` — N-terminal acetylation is a supported N-terminus state |
| `S[Acetylation on S]KGPAVGID…` | `-SKGPAVGID…` — `(42.01, 'S')` is not in the table, silently removed |

Both place +42.011 **inside residue 1**, so every b/y/a and immonium ion carries it identically and the `Mass Diff (ppm)` cell is a single value. Walking the `"standard"` feature list, *every* other feature is identical for such a pair — the matched ions are the same, so all ion-derived features are the same; the rest are PSM-level. **Hydrophobicity alone decided between them**, and PEP then deleted the "loser" via `GetIndicesOfPeptidesToRemove`.

On a 12-file G-PTM-D search (MassIVE `MSV000083304`, calibrate → GPTMD → search) this resolved **fifteen provably undecidable pairs** — and the aggregate metrics *improved* while it happened (AUC 0.9917 → 0.9951). That combination is what makes it hard to notice: by every number the task reports, things got better.

The same mechanism explains a near miss. `Ethylation on K` is also outside the vocabulary, but the mass fallback is keyed on `(rounded mass, residue)` and contains `{ (28.03, 'K'), 'o' }` — dimethylation. Ethyl-K has the same rounded mass on the same residue, so it maps to the same code, the two encode identically, and that pair was protected **by accident**.

### Not proposed

Changing the tiebreak ordering. A deterministic order is worth more than the bias costs, and randomising it would trade a knowable, reproducible imperfection for an unrepeatable one.

Switching retention-time models. The defect is **vocabulary coverage, not accuracy** — every RT model has a finite vocabulary, and a better model applied to a mis-encoded sequence returns a *more* confident wrong difference.

### Tests

`FdrTest` asserts `HasHydrophobicity` is 1 when the predictor succeeds and 0 when it cannot represent the peptidoform, using a stub predictor that always reports `IncompatibleModifications`. The `trainingInfos` and `ToString` round-trip guards are updated for the new feature.

### Worth your judgement

Point 3 widens which peptidoforms lose the hydrophobicity feature — anything carrying a modification Chronologer does not know. That is the intended semantics (if the model cannot represent the peptidoform, its retention time is not evidence about it), and `HasHydrophobicity` is what makes it safe for the model to learn. But it will change PEP behaviour on modification-rich searches, and you may prefer to gate it behind a parameter. Happy to split that commit out if you would rather take 1 and 2 first.

Related: #2797 (`MaxModificationIsoforms` truncates silently) came out of the same investigation. That one silently *removes* hypotheses; this one silently *manufactures* evidence.

### DIFF
--- a/MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
+++ b/MetaMorpheus/EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs
@@ -4,6 +4,7 @@ using Chromatography.RetentionTimePrediction.SSRCalc;
 using EngineLayer.CrosslinkSearch;
 using EngineLayer.SpectrumMatch;
 using PredictionClients.Koina.SupportedModels.RetentionTimeModels;
+using Omics.SequenceConversion;
 using Proteomics.ProteolyticDigestion;
 using System;
 using System.Collections.Generic;
@@ -16,7 +17,13 @@ namespace EngineLayer.FdrAnalysis
     {
         private static readonly Lazy<ChronologerRetentionTimePredictor> _chronologerInstance =
             new Lazy<ChronologerRetentionTimePredictor>(
-                () => new ChronologerRetentionTimePredictor(),
+                // ReturnNull, not the default RemoveIncompatibleElements. Under the default a modification
+                // outside Chronologer's vocabulary is silently deleted and the residue encoded bare, so the
+                // model predicts the retention time of a DIFFERENT molecule and reports success. That is
+                // particularly damaging within a single scan: two peptidoforms the spectrum cannot separate can
+                // receive different predictions purely because one modification is in the vocabulary and the
+                // other is not. Reporting the failure lets PsmData.HasHydrophobicity mark the value unusable.
+                () => new ChronologerRetentionTimePredictor(SequenceConversionHandlingMode.ReturnNull),
                 System.Threading.LazyThreadSafetyMode.ExecutionAndPublication);
 
         // When non-null, GetChronologer() calls this factory instead of the real Lazy<>.
--- a/MetaMorpheus/EngineLayer/FdrAnalysis/PEPAnalysisEngine.cs
+++ b/MetaMorpheus/EngineLayer/FdrAnalysis/PEPAnalysisEngine.cs
@@ -479,6 +479,7 @@ namespace EngineLayer
             float longestSeq = 0;
             float complementaryIonCount = 0;
             float hydrophobicityZscore = float.NaN;
+            float hasHydrophobicity = 0;
             bool isVariantPeptide = false;
 
             //crosslink specific features
@@ -542,11 +543,13 @@ namespace EngineLayer
                             var dict = isUnmodified
                                 ? FileSpecificTimeDependantHydrophobicityAverageAndDeviation_unmodified
                                 : FileSpecificTimeDependantHydrophobicityAverageAndDeviation_modified;
-                            hydrophobicityZscore = (float)Math.Round(GetRetentionTimeEquivalentZscore(psm, tentativeSpectralMatch.SpecificBioPolymer, dict, RetentionTimePredictor) * 10.0, 0);
+                            hydrophobicityZscore = (float)Math.Round(GetRetentionTimeEquivalentZscore(psm, tentativeSpectralMatch.SpecificBioPolymer, dict, RetentionTimePredictor, out bool retentionTimePredicted) * 10.0, 0);
+                            hasHydrophobicity = Convert.ToSingle(retentionTimePredicted);
                         }
                         else
                         {
                             hydrophobicityZscore = (float)Math.Round(GetMobilityZScore(psm, tentativeSpectralMatch.SpecificBioPolymer) * 10.0, 0);
+                            hasHydrophobicity = 1; // CZE mobility is computed from composition and always available
                         }
                     }
                 }
@@ -620,6 +623,7 @@ namespace EngineLayer
                 LongestFragmentIonSeries = longestSeq,
                 ComplementaryIonCount = complementaryIonCount,
                 HydrophobicityZScore = hydrophobicityZscore,
+                HasHydrophobicity = hasHydrophobicity,
                 IsVariantPeptide = Convert.ToSingle(isVariantPeptide),
 
                 AlphaIntensity = alphaIntensity,
@@ -713,7 +717,22 @@ namespace EngineLayer
                         }
                         fullSequences.Add(bestMatch.SpecificBioPolymer.FullSequence);
 
-                        double predictedHydrophobicity = bestMatch.SpecificBioPolymer is PeptideWithSetModifications pep ? predictor.PredictRetentionTimeEquivalent(pep, out _) ?? 0 : 0;
+                        // A peptidoform the predictor cannot represent must not contribute to the reference
+                        // distribution. `?? 0` would enter it as a predicted hydrophobicity of zero, which is a
+                        // real and extreme value on this scale, and would drag the median and standard deviation
+                        // that every other peptide in this retention-time bin is then scored against.
+                        if (bestMatch.SpecificBioPolymer is not PeptideWithSetModifications pep)
+                        {
+                            continue;
+                        }
+
+                        double? predicted = predictor.PredictRetentionTimeEquivalent(pep, out _);
+                        if (!predicted.HasValue)
+                        {
+                            continue;
+                        }
+
+                        double predictedHydrophobicity = predicted.Value;
 
                         //here i'm grouping this in 2 minute increments becuase there are cases where you get too few data points to get a good standard deviation an average. This is for stability.
                         int possibleKey = (int)(2 * Math.Round(psm.ScanRetentionTime / 2d, 0));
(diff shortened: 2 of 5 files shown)


trishorts:
Moving this to draft — the fix is not yet validated, and there is a specific way it could fail that I should have flagged before opening.

**The concern.** This change replaces a *value* difference with an *availability* difference. For the pair that surfaced the bug — N-terminal acetylation against acetylation on serine 1 — one member's modification is in Chronologer's vocabulary and the other's is not. Before the change they got different `HydrophobicityZScore` values. After it, they get `HasHydrophobicity = 1` and `HasHydrophobicity = 0` respectively.

That is still a usable feature difference, on exactly the same pairs. A tree model can split on it just as readily. So it is quite possible that this fix does not stop hydrophobicity deciding between spectrally-undecidable peptidoforms — it may only change which column does the deciding.

**How it will be checked.** A search is running now on the same 12-file G-PTM-D dataset, with this fix applied on top of the previous configuration so it is the only variable. The acceptance test is the count of PSMs reported ambiguous between a terminal modification and the same modification on the adjacent residue — pairs where every b/y/a and immonium ion is identical, so the spectrum genuinely cannot choose:

- returns to **~233** (the count before the defect began biting) — the fix works
- stays at **~218** — the fix does not reach it, and the diagnosis is wrong
- drops **below 218** — the availability flag became the new discriminator, and the correct fix is different: neither hydrophobicity nor its availability flag should participate in a within-scan comparison at all

I will post the number either way. If it is the third case I will rework this before asking for review, because the right change would then be to exclude retention time from candidate-versus-candidate comparison rather than to annotate it.

Parts 1 and 2 stand on their own regardless — a failed prediction entering the reference distribution as a hydrophobicity of zero is a bug independent of anything to do with ambiguity.

trishorts:
Validation result: **this is the third outcome I described above — the one where the fix is the wrong shape.** Posting it against the PR rather than quietly reworking.

Same 12-file G-PTM-D dataset, this change applied as the only variable:

| | before | + this change |
|---|---|---|
| terminal-vs-adjacent pairs still reported ambiguous | 218 | **211** |
| Ambiguous Peptides Removed | 2,366 | **449** |
| Area Under Curve | 0.99513 | 0.99417 |
| target PSMs q ≤ 0.01 | 58,958 | 58,967 |
| target peptides q ≤ 0.01 | 21,829 | 21,836 |

The reference count, before any of this began, is **233**.

**What went wrong.** The change replaces a *value* difference with an *availability* difference. Before it, the two members of a spectrally-undecidable pair received different `HydrophobicityZScore`. After it they receive `HasHydrophobicity` 1 and 0 — and a gradient-boosted tree splits on that just as readily. The asymmetry was never in how the value was encoded; it is in the vocabulary coverage itself, and it cannot be annotated away. However you express "the model cannot represent this peptidoform," you express it for exactly one member of the pair.

**What did work.** Over-removal fell 2,366 → 449, most of the way back to the 188 seen before any change, with identifications slightly up. So the fix removes the great majority of the damage — it just fails on the specific class it was aimed at.

Worth saying plainly: judged on the numbers this task reports, this change looks good. Removals down five-fold, IDs up, AUC still well above baseline. Only the undecidable-pair count dissents, and no aggregate metric can see it.

**Suggested disposition.**

Parts 1 and 2 stand on their own and I would still like them considered: a failed prediction entering the per-file reference distribution as a hydrophobicity of **zero** corrupts the median and spread that every other peptide in that retention-time bin is scored against, and that is a bug independent of anything to do with ambiguity.

Part 3 (`ReturnNull`) is what creates the flag asymmetry and I would not take it as-is.

**The structural fix** — which I am not going to write, since it is a real change to your FDR path and should be your call — is to give the candidate-versus-candidate comparison inside `Compute_PSM_PEP` a **restricted feature vector** containing only terms the spectrum constrains. Retention time is a peptide-level property; using it, or the fact of its absence, to choose between two peptidoforms that produce identical theoretical fragments is a category error that annotating the value does not repair.

Happy to split this into a parts-1-and-2 PR if that is useful, or to close it if you would rather approach the whole thing differently.

trishorts:
**Rescoped and out of draft.** Acting on the validation result two comments up: the `ReturnNull`
commit is gone, the branch is rebased onto current master (`f095b98b4`), and the description has been
rewritten around what is left. Nothing in the two comments above has been edited — the reasoning that
produced the third outcome is the record and should stay readable.

**What this PR is now:**

1. Failed retention-time predictions no longer enter the per-file, per-RT-bin **reference
   distribution** as a hydrophobicity of zero, where they dragged the median and spread that every
   other peptide in that bin is scored against.
2. `PsmData.HasHydrophobicity`, `[LoadColumn(30)]`, the companion feature `HasSpectralAngle` already
   is to `SpectralAngle`.

**What it is not:** the Chronologer vocabulary problem. That is real and unfixed — under the default
`RemoveIncompatibleElements`, a modification outside the ~22 states Chronologer knows is silently
deleted and the residue encoded bare, so the model predicts the retention time of a different molecule
and reports success. `ReturnNull` was the wrong fix for it, for the reason measured above.

Worth stating plainly, because it is the reason these two parts stand alone: **they fire without
`ReturnNull`.** Chronologer independently declines a sequence longer than 50 residues, shorter than 7,
or carrying a non-canonical amino acid such as selenocysteine, and any predictor can fail outright.
Every one of those currently enters PEP as a hydrophobicity of zero, and every one of them currently
contaminates the reference distribution for its neighbours.

**One thing to weigh before merging.** The only search run that contains these changes also contained
`ReturnNull` and an experimental precursor-mass-error feature that is not in this PR, so I cannot
attribute an identification count to parts 1 and 2 in isolation and have not tried to in the
description. If you would like that measured first, say so and I will hold it.

Full suite green locally on the rebased branch: **2,301 passed, 0 failed, 1 skipped**.

The vocabulary question and the wider issue behind it — that `GetIndicesOfPeptidesToRemove` compares
candidates on an **absolute** 0.05 probability gap, which is not invariant to how sharp the model is,
so improving PEP silently increases ambiguity removal — have been handed to the `pep` project, which
owns this machinery. I am not pursuing either further from here.

Alexander-Sol:

**Summary.** PEP built a `HydrophobicityZScore` from a predicted retention time, coercing a *failed* prediction to `0` via `?? 0`. Since 0 is a real extreme value on that scale, "could not predict" became "disagrees maximally," saturating the z-score to 10 (strong evidence against the candidate). The PR (1) stops failed/unrepresentable peptidoforms from polluting the per-file/per-RT-bin reference median+stdev in `ComputeRetentionTimeEquivalentValues` (`continue` instead of entering 0), and (2) adds `PsmData.HasHydrophobicity` (mirroring `HasSpectralAngle`) so the model can tell "predicted-and-disagrees" from "unpredictable." No confirmed bugs.

### Findings
- **Low (test-coverage gap):** Part 1 — the reference-distribution fix the PR calls the actual bug fix — has no test. The new test only exercises Part 2 (`HasHydrophobicity` 0 vs 1). Nothing asserts an unpredictable peptidoform is *excluded* from the bin median/stdev its neighbours are scored against. The `continue` logic (`PEPAnalysisEngine.cs ~716`) looks correct but is untested. Worth a small unit test.
- **Nit (doc):** `HasHydrophobicity` is 0 in more cases than its XML doc says — in `GetRetentionTimeEquivalentZscore` (`~910-919`), `predictionAvailable` stays false when the file/RT-bin is absent from the reference dict too, not only on predictor failure. Not a bug (that case is also a meaningless saturated 10, so flagging it 0 is arguably more correct); the doc wording is just narrower than the behavior.

### Verified correct
`HasHydrophobicity` registered in both `trainingInfos["standard"]` and `assumedAttributeDirection`; genuinely wired into training via `Concatenate("Features", …)` (`:149`); `[LoadColumn(30)]` is the next free index; `ToString`/test-string ordering consistent; the positive/negative test is a proper controlled comparison (same PSM & reference dicts, only the predictor differs — real SSRCalc3 → non-saturated z-score, `HasHydrophobicity==1`; `NeverPredicts` stub → null → `0`); empty-bin edge cases from the new `continue` don't crash (globalStDev defaults to 1, single-point NaN already handled).

pcruzparri (approved):

PR #2799, opened 2026-09-05
Detach CMD's handlers from the static engine and task events

### LINKED ISSUE #2788: CMD leaks static MetaMorpheusEngine/MetaMorpheusTask event subscriptions, breaking unrelated tests in the same process
`Program.Run` subscribes nine handlers to `MetaMorpheusEngine`'s and `MetaMorpheusTask`'s **static** events and never unsubscribes them (`CMD/Program.cs:154-163` on master). There is no `-=` anywhere in `CMD/`.

Because the events are static, the subscriptions outlive the call. Any code that runs an engine in the same process afterwards fires CMD's handlers, in a context they were not written for.

## What it looks like when it bites

Calling the CLI in-process and then running any engine gives:

```
System.ArgumentNullException : Value cannot be null. (Parameter 'values')
   at System.String.Join(String separator, IEnumerable`1 values)
   at EngineLayer.MetaMorpheusEngine.GetId() in EngineLayer/MetaMorpheusEngine.cs:354
   at MetaMorpheusCommandLine.Program.MyEngine_startingSingleEngineHander(Object sender, SingleEngineEventArgs e) in CMD/Program.cs:451
   at EngineLayer.MetaMorpheusEngine.StartingSingleEngine() in EngineLayer/MetaMorpheusEngine.cs:374
   at EngineLayer.MetaMorpheusEngine.Run() in EngineLayer/MetaMorpheusEngine.cs:307
   at Test.MultiProteaseParsimonyTest.MultiProteaseTest() in Test/MultiProteaseParsimonyTest.cs:83
```

`GetId()` is `string.Join(",", NestedIds)`, and `NestedIds` is null for engines that were not started through a task, so the handler throws inside an event the engine raises on every run.

## Reproduction

Add any test that calls `MetaMorpheusCommandLine.Program.Main(...)` to `Test.csproj`, then run it in the same session as the parsimony tests:

```
dotnet test Test/Test.csproj -c Release --filter "FullyQualifiedName~<YourCmdTest>|FullyQualifiedName~MultiProtease|FullyQualifiedName~TestPSMFdrFiltering_Simulated"
```

I hit this writing coverage for #2724. Four CLI tests passed on their own, and **eleven** unrelated tests then failed in the same run:

- `MultiProteaseParsimony_*` (8 of them)
- `MultiProteaseSamePeptideSameProteinsDifferentProteases`
- `MultiProteaseTest`
- `TestPSMFdrFiltering_Simulated`

All eleven pass in isolation. The only thing joining them to the CLI test is the static subscription.

## Why this is worth fixing rather than working around

1. **It makes CMD effectively untestable in-process.** Every in-process CLI test silently poisons whatever runs after it, and the damage surfaces far from its cause — in someone else's PR, in tests they did not touch. I worked around it by running the CLI as a subprocess, which is fine for that PR, but the next person writing a CMD test will lose the same afternoon to it.
2. **Repeat calls stack up.** Nothing is idempotent, so two calls to `Run` in one process attach every handler twice and each engine event is handled twice over.
3. **It is a known shape in this codebase.** #2709's Avalonia view model subscribes to these same static events and is deliberately `IDisposable` so it can unsubscribe — its description notes that without it "tests passed individually and failed in a suite". CMD has the identical problem and no equivalent cleanup.

## Suggested fix

Either unsubscribe in a `finally` around the body of `Run`, or make the subscriptions idempotent (`-=` then `+=`) and detach before returning. The `IDisposable` subscription holder that #2709 uses would work here too and would give the two front ends one pattern instead of two.

Guarding `GetId()` against a null `NestedIds` would stop the crash, but not the underlying leak — CMD would still be writing console output during unrelated in-process runs.

Found while reviewing #2724; not caused by it. Present on master at `d233c17be`.

acesnik:
🤖 `Program.Run` subscribed nine handlers to `MetaMorpheusEngine`'s and `MetaMorpheusTask`'s static events and never detached them. The subscriptions outlived the call, so anything that ran an engine later in the same process fired CMD's console handlers — and the starting-engine handler reaches `GetId()`, which is `string.Join(",", NestedIds)`, so it threw for any engine a task did not start. Fixes #2788.

## What changed

`CMD/Program.cs` only:

- the nine `+=` lines moved into `SubscribeToEngineAndTaskEvents()`, which detaches before it attaches, so a second run in one process no longer attaches every handler twice and prints everything twice;
- `Run` is now a wrapper around the run body (`RunTasks`) with `finally { UnsubscribeFromEngineAndTaskEvents(); }`, so the early returns for an unreadable task file, a declined Thermo licence and a bad experimental design are covered as well as an exception;
- detaching also resets `MyWriter.Indent` and `InProgress`, the state the handlers keep between events, which would otherwise carry into the next run in the same process.

The run body itself is unchanged apart from the ten wiring lines becoming one call. #2788 suggested reusing the `IDisposable` subscription holder from #2709; CMD's handlers are static methods with no instance to own one, so `try`/`finally` is the same guarantee in the shape this file already has.

## Output

No change to any produced file, and none to console output for an ordinary invocation: the same nine handlers are attached across the run, and the detach happens after the run has returned.

Checked rather than assumed — `CMD.dll --test` (Calibrate → GPTMD → Search) on the fixed build exits 0, produces the usual tree (`allResults.txt`, the three task folders, `AllPSMs.psmtsv` and the rest), and its console output has 3/3 `Starting task:`/`Finished task:`, 24/24 `Starting engine:`/`Finished engine:` pairing LIFO with no residual depth, no adjacent duplicate lines anywhere (a doubled handler would print everything twice), and no warnings or errors. Two consecutive runs differ only in the interleaving of progress percentages, which is `OutProgressHandler` thread ordering and present either way.

I did not build a pre-fix arm for a literal before/after console diff, so that last paragraph is "an ordinary run looks right", not a byte comparison against master.

## Test

`TestCommandLineDetachesItsEventHandlers` in `Test/TestCmd.cs` drives `Program.Main` through a real run — one task file with an unrecognised `TaskType`, which CMD skips, so nothing reads the spectra or the database — then reads the events' backing fields and asserts that no handler declared in `Program` survives. It then runs an engine whose `NestedIds` is null, which is the shape the crash took.

Both assertions fail against the unfixed code; I checked by emptying the body of the unsubscribe helper. The first reports all nine subscriptions:

```
the command line left handlers attached to static engine and task events
Assert.That(CmdHandlersStillAttached(), Is.Empty)
  Expected: <empty>
  But was:  < "MetaMorpheusEngine.WarnHandler -> WarnHandler", ... >
```

and with that assertion set aside, the second reproduces the exception from the issue:

```
System.ArgumentNullException : Value cannot be null. (Parameter 'values')
   at System.String.Join(String separator, IEnumerable`1 values)
   at EngineLayer.MetaMorpheusEngine.GetId()
   at MetaMorpheusCommandLine.Program.MyEngine_startingSingleEngineHander(Object sender, SingleEngineEventArgs e)
   at EngineLayer.MetaMorpheusEngine.StartingSingleEngine()
   at EngineLayer.MetaMorpheusEngine.Run()
```

The whole `TestCmd` fixture passes (22 passed, plus the pre-existing ignore). These were run off Windows through a throwaway harness, since `Test.csproj` targets `net*.0-windows`, so CI is the real check.

The GUI wires up the same static events in its main window constructor, but that window lives as long as the process, so it is not the same defect and I left it alone.


trishorts (dismissed):
Verified rather than taken on trust:

- `Main` reaches the run body only through `Run` (`CMD/Program.cs:36`), so the new `try`/`finally` covers every exit — including the three early returns the description names.
- `-=` against a handler that was never attached is a no-op, so `SubscribeToEngineAndTaskEvents` calling `Unsubscribe` first, and `Run` unsubscribing after an early return that never reached the wiring, are both harmless.
- The `MyWriter.Indent` / `InProgress` reset belongs with the detach: they are the only state the handlers carry between events.

Agreed on leaving the GUI alone — its window is process-lifetime, so the same subscription is not the same defect.

Approving.

nbollis (dismissed):

nbollis (approved):

pcruzparri (approved):

PR #2804, opened 2026-09-08
Archive the TMT design only once it parses, and name the reason a copy failed

trishorts:
Three defects on the multiplex quantification path #2755 added, and the guard wording one of the
three then made wrong. Every one of them needs a non-default setting or a bad design file to reach,
so a stock run is unchanged.

### An unreadable design was still archived beside the results

The copy into the output folder, and its `FinishedWritingFile` announcement, ran **before**
`TmtExperimentalDesign.Read`. A design that did not parse therefore landed in the results folder and
was announced as an output — under a comment saying the folder "records the design it was quantified
under", when nothing had been quantified under anything.

The copy now sits below the read guard. The sibling case is deliberately unchanged: a design that
*reads* is still archived even when a later guard declines to quantify.

### A bare `catch` discarded the reason the archive failed

`catch { Warn("Could not copy the TMT design file…") }` told a user whose copy failed nothing they
could act on. It now reports `e.Message` — the difference between "access is denied" and a shrug.

### Contaminants were filtered asymmetrically, and the guard then misreported it

PSMs went into quantification with `includeContaminants: true` unconditionally, while protein groups
went through `ProteinGroupIsWritten`, which honours `WriteContaminants`. With the box unticked the
quantified set was wider than the written set at the PSM and peptide levels but not the protein
level: contaminant peptides appeared in `RawQuantification.tsv` and `PeptideQuantification.tsv`
while their groups were absent from `ProteinGroupQuantification.tsv`.

The PSM filter now follows `WriteContaminants` too.

One consequence worth naming, found by @acesnik: `SpectralMatch.IsContaminant` is true when *any*
best match's parent is a contaminant, so a peptide shared between a contaminant and a target protein
is dropped by this filter too. No number moves — mzLib's engine already excluded such a match, since
`SingleIdentifiedBioPolymerOrNull` keeps one only when `Distinct()` yields exactly one element and
`PeptideWithSetModifications.Equals` compares `Parent?.Accession`, which
`PeptideSharedBetweenTwoProteins_WarnsThatMatchesWereExcluded` already pins at one excluded match.
What changes is the row set: `PeptideQuantification.tsv` also loses the target-side row of such a
peptide, and those rows were all-zero.

That created a second way to reach an empty quantifiable set, and the existing guard reported it
wrongly: with every reporter-bearing match flagged contaminant, the user was told "No spectral
matches carried reporter ion intensities" when they had, and the filter one line above was what
removed them. The guard now names the filter that emptied the set. The contaminant-inclusive filter
it asks for is evaluated only once the set is already empty, so a run that quantifies never pays
for it.

## Tests

`MultiplexQuantificationTests.cs`, whose guard helper now takes `writeContaminants` alongside the
`writeDecoys` it already had — without it, `Contaminant` did not appear in the file at all and every
test ran the default `true`, so the coverage on that argument was of the arm that does not change.

| Test | Pins |
|---|---|
| `UnreadableDesign_WarnsAndSkips` (extended) | a design that could not be read is not archived |
| `NoPsmCarriesReporterIons_WarnsAndSkips` (existing) | one that *could* be read still is |
| `DesignNamingNoFileInThisRun_WarnsAndSkips` (extended) | nor is one that names no file in this run, which reports through the same `designErrors` |
| `DesignCopyFailure_WarnsAndKeepsGoing` (existing) | the named failure still matches, and does not stop the analysis |
| `ContaminantMatches_AreQuantifiedOnlyWhenTheirGroupsAreWritten` | both directions: ticked, a contaminant quantifies like anything else; unticked, its peptide leaves with its group |
| `EveryReporterIonMatchIsAContaminant_NamesTheFilterRatherThanTheData` | the guard names the filter rather than blaming the data |

All three new assertions were verified by reverting the production change under each:

- block order alone → `UnreadableDesign_WarnsAndSkips` and `DesignNamingNoFileInThisRun_WarnsAndSkips`
  both fail on their archive assertions, from opposite guards
- the contaminant filter alone → `ContaminantMatches_…(False)` fails at `Expected: False / But was:
  True` on the contaminant peptide, and the guard test drops to the protein-group guard's wording
- the guard message alone → `EveryReporterIonMatchIsAContaminant_…` fails with the sentence it
  replaces

Full suite **1983 passed / 0 failed / 1 skipped**.

## Scope

The label-free path has the same asymmetry — two unconditional `includeContaminants: true` filters
feeding tables whose protein groups are written through the same `ProteinGroupIsWritten` — and it is
**not** fixed here. It is not the same change: the write-facing one of the two feeds FlashLFQ's
`peptideSequencesToQuantify`, which mzLib consults
inside peak filtering and MBR peak selection (`FlashLfqEngine.cs:711`, `:948`, `:996`, `:1116`,
`:1400-1428`), so narrowing it can move target peptide intensities rather than only dropping
contaminant rows. That needs its own PR with a measured before/after on a real search.

Filed as #2805, which carries the three candidate fixes and the question they have to settle.

## Credit

The bare-`catch` fix is @pcruzparri's: his review of #2776 flags the three `catch { /* ignore */ }`
blocks in `MainWindow.xaml.cs`, and this is the same swallowed exception one file over. The archive
ordering and the contaminant filter came from reading #2755 back after it merged. The guard wording,
the missing `writeContaminants` coverage and the label-free scope note are all @acesnik's, from his
review of this PR.

Sits alongside #2802, which fixes the individual-file column regression #2755 also shipped — the one
@pcruzparri found. No overlap: that one touches the subset loop, this one the design archive and the
PSM filter.

https://claude.ai/code/session_01F9P6YK6VyQJkgdj63KS4he

### DIFF
--- a/MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
+++ b/MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
@@ -336,21 +336,35 @@ namespace TaskLayer
             // the PSM and peptide levels but not the protein level: with the box unticked, contaminant
             // peptides appeared in RawQuantification.tsv and PeptideQuantification.tsv while their
             // groups were absent from ProteinGroupQuantification.tsv.
-            var filteredPsms = FilteredPsms.Filter(Parameters.AllSpectralMatches,
+            FilteredPsms FilterMatches(bool includeContaminants) => FilteredPsms.Filter(
+                Parameters.AllSpectralMatches,
                 CommonParameters,
                 includeDecoys: false,
-                includeContaminants: Parameters.SearchParameters.WriteContaminants,
+                includeContaminants: includeContaminants,
                 includeAmbiguous: false,
                 includeAmbiguousMods: false,
                 includeHighQValuePsms: false);
 
-            var quantifiablePsms = filteredPsms
-                .Where(psm => psm.IsobaricMassTagReporterIonIntensities is { Length: > 0 })
-                .ToList();
+            static bool CarriesReporterIons(SpectralMatch psm) =>
+                psm.IsobaricMassTagReporterIonIntensities is { Length: > 0 };
+
+            var filteredPsms = FilterMatches(Parameters.SearchParameters.WriteContaminants);
+
+            var quantifiablePsms = filteredPsms.Where(CarriesReporterIons).ToList();
 
             if (quantifiablePsms.Count == 0)
             {
-                Warn("No spectral matches carried reporter ion intensities. Skipping multiplex quantification");
+                // Name the filter that emptied the set. With the contaminant box unticked, every
+                // reporter-bearing match in a run can be a contaminant, and the plain wording then
+                // blames the data for what the filter above removed. Asked only on the way out, so a
+                // run that quantifies never pays for the second filter.
+                bool contaminantsCarriedThemAll = !Parameters.SearchParameters.WriteContaminants
+                    && FilterMatches(includeContaminants: true).Any(CarriesReporterIons);
+
+                Warn(contaminantsCarriedThemAll
+                    ? "Every spectral match that carried reporter ion intensities was a contaminant, and " +
+                      "contaminants are not being written. Skipping multiplex quantification"
+                    : "No spectral matches carried reporter ion intensities. Skipping multiplex quantification");
                 return;
             }
 
--- a/MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
+++ b/MetaMorpheus/TaskLayer/SearchTask/PostSearchAnalysisTask.cs
@@ -292,24 +292,30 @@ namespace TaskLayer
 
             Status("Quantifying multiplex channels...", Parameters.SearchTaskId);
 
+            var tmtFiles = TmtExperimentalDesign.Read(tmtDesignPath, Parameters.CurrentRawFileList, out var designErrors);
+            if (designErrors.Any())
+            {
+                Warn("Error reading TMT design file: " + designErrors.First() + ". Skipping multiplex quantification");
+                return;
+            }
+
             // Copy the design into the output folder, as the label-free path does for ExperimentalDesign,
-            // so a result folder records the design it was quantified under.
+            // so a result folder records the design it was quantified under. AFTER the read, not before:
+            // a design that could not be parsed was not quantified under anything, and archiving it
+            // anyway left it in the results folder and announced it through FinishedWritingFile while
+            // this method was on its way out.
             try
             {
                 string copiedDesign = Path.Combine(Parameters.OutputFolder, Path.GetFileName(tmtDesignPath));
                 File.Copy(tmtDesignPath, copiedDesign, overwrite: true);
                 FinishedWritingFile(copiedDesign, new List<string> { Parameters.SearchTaskId });
             }
-            catch
+            catch (Exception e)
             {
-                Warn("Could not copy the TMT design file to the search task output. That's ok, the search will continue");
-            }
-
-            var tmtFiles = TmtExperimentalDesign.Read(tmtDesignPath, Parameters.CurrentRawFileList, out var designErrors);
-            if (designErrors.Any())
-            {
-                Warn("Error reading TMT design file: " + designErrors.First() + ". Skipping multiplex quantification");
-                return;
+                // Named, not swallowed: a user whose archive failed can act on "access is denied" and
+                // can do nothing with "could not copy".
+                Warn("Could not copy the TMT design file to the search task output: " + e.Message +
+                     ". That's ok, the search will continue");
             }
 
             var tag = IsobaricMassTag.GetIsobaricMassTag(Parameters.SearchParameters.MultiplexModId);
@@ -324,10 +330,16 @@ namespace TaskLayer
             // includeAmbiguous: false is the unambiguous filter the design calls for. A PSM that could be
             // more than one peptide would otherwise have its channel intensities credited to whichever
             // candidate happened to sort first.
+            //
+            // includeContaminants follows WriteContaminants, the same switch ProteinGroupIsWritten reads
+            // below. Passing true unconditionally made the quantified set wider than the written set at
+            // the PSM and peptide levels but not the protein level: with the box unticked, contaminant
+            // peptides appeared in RawQuantification.tsv and PeptideQuantification.tsv while their
+            // groups were absent from ProteinGroupQuantification.tsv.
             var filteredPsms = FilteredPsms.Filter(Parameters.AllSpectralMatches,
                 CommonParameters,
                 includeDecoys: false,
-                includeContaminants: true,
+                includeContaminants: Parameters.SearchParameters.WriteContaminants,
                 includeAmbiguous: false,
                 includeAmbiguousMods: false,
                 includeHighQValuePsms: false);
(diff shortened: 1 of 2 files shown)


acesnik (commented):
🤖 This is the follow-up that makes the design archive honest: before it, an unparseable `TmtDesign.txt` was copied into the results folder and announced through `FinishedWritingFile` on the way out, under a comment claiming the folder records the design the run was quantified under. Moving the copy below the read guard is the right cut, and the new assertion is real — I reverted the block order alone and `UnreadableDesign_WarnsAndSkips` fails at `MultiplexQuantificationTests.cs:606` with `Expected: False / But was: True`, the only failure of the 17. Both defects entered in #2755, which merged 28 minutes before you opened this and after 1.1.9, so neither was ever in a release.

I also chased the case that would have made the contaminant change expensive, and it comes back clean. `SpectralMatch.cs:330` sets `IsContaminant = bestMatches.Any(...)`, so a peptide shared between a contaminant and a target protein is flagged contaminant and is now dropped — but the intensity it contributed to a written target group was already zero, because mzLib's `SingleIdentifiedBioPolymerOrNull` keeps a match only when `Distinct()` yields one element, and `PeptideWithSetModifications.Equals` compares `Parent?.Accession`. `PeptideSharedBetweenTwoProteins_WarnsThatMatchesWereExcluded` already pins that at `AmbiguousSpectralMatchesExcluded == 1`. So no written non-contaminant group loses reporter intensity and no protein-level number moves. What does move is the row set: `PeptideQuantification.tsv` also loses the target-side row of such a peptide, because `peptides` at `:381` comes from `GetIdentifiedBioPolymersWithSetMods()`, which returns every best match rather than the resolved one. Those rows were all-zero, so this is rows and not values — worth a line in the body rather than a change.

Three things I would fix. The guard at `:351` can now lie: with `WriteContaminants` off and every reporter-bearing PSM flagged contaminant, `quantifiablePsms` empties and the user is told "No spectral matches carried reporter ion intensities" — they did carry them, the filter one line above removed them. Before this PR that run fell through to `:374` and got the accurate reason. Same outcome, worse diagnosis.

The asymmetry is also fixed in only one of the two places it exists. The label-free path still passes `includeContaminants: true` at `:573` and `:591`, and its protein groups are written through the same `ProteinGroupIsWritten` (`:1967`) — so with the box unticked `AllQuantifiedPeptides.tsv` keeps contaminants while their groups are absent from `AllProteinGroups.tsv`, exactly the shape you describe. That is the more travelled of the two paths.

And nothing tests the new argument. `Contaminant` does not appear in `MultiplexQuantificationTests.cs` at all, and the helper at `:432` takes `writeDecoys` and `writeHighQValuePsms` but not `writeContaminants`, so all 17 tests run the default `true`. Codecov's green is line coverage of the arm that did not change. A `writeContaminants` parameter mirroring `writeDecoys`, plus one contaminant `Protein`, would make this the only test in the repo pinning the behaviour.

Two smaller ones. `DesignNamingNoFileInThisRun_WarnsAndSkips` still passes, but its stated reason no longer holds — "names no file in this run" reports through `designErrors`, so the copy is skipped there too now; I measured `archived=False`. The string at `:569-570`, "the design itself archived cleanly, so declining is the only thing to report", is now the opposite of what happens. And the fourteen lines of new comment restate what the commit message already says verbatim; house style is to keep the file terse and put the rationale in the commit.

Last, the credit line does not resolve: pcruzparri has no review or comment on #2755. Their review is on #2776, and it does flag the `catch { /* ignore */ }` blocks, so the bare-catch fix is fairly theirs — the archive ordering and the contaminant filter appear in no review on #2755. A body edit, so it costs nothing.

Not verified: `Test.csproj` is `net*.0-windows`, so I ran the fixture through a throwaway net10.0 harness rather than natively, in isolation from the other ~2500 tests, and worked around three Windows-separator path literals to do it. CI's Windows run is the authority and is green on all 13 checks. I did not run a search against a contaminant FASTA, so the contaminant reasoning above is code plus the existing test, not measurement.

trishorts:
🤖 All five addressed, in `365843ccc` and `72a77926c` — except the label-free path, which is filed
rather than folded in. Thank you for reverting the block order and measuring it; that is also what
caught the guard below.

**The guard could lie.** Fixed. It now names the filter that emptied the set, and the
contaminant-inclusive filter it consults is evaluated only once the set is already empty, so a run
that quantifies never pays for it. Your reading of the pre-PR behaviour is confirmed by the revert:
with the filter alone put back, the new guard test does not merely pass differently, it drops to
"No protein group survived the filters" — the accurate reason you said that run used to get.

**The label-free path — agreed, and deliberately not here.** It is not the same change. The
write-facing site feeds FlashLFQ's `peptideSequencesToQuantify`, and mzLib consults that set inside
quantification rather than only on the way out: peak filtering at `FlashLfqEngine.cs:711`, `:948`,
`:996`, `:1116`, and MBR peak selection at `:1400-1428`, where it decides which of two competing
peaks is kept. Narrowing it can therefore move target intensities, not just drop contaminant rows —
so the claim to prove is "no target intensity moves", and that wants a measured before/after on a
real search rather than a one-line switch inside this PR. The other site is FlashLFQ's identification
input, where `includeDecoys: Parameters.SearchParameters.MatchBetweenRuns` one line above establishes
feed-wide/write-narrow as the deliberate pattern.

Filed as #2805 with three candidate fixes ranked by intrusiveness, the open question they have to
settle, and the four other `includeContaminants: true` sites I checked and ruled out. The PR body now
carries the scope note.

**Nothing tested the new argument.** Correct, and it is now the helper's third switch beside
`writeDecoys`. Two tests use it: `ContaminantMatches_AreQuantifiedOnlyWhenTheirGroupsAreWritten`
asserts both directions, and `EveryReporterIonMatchIsAContaminant_NamesTheFilterRatherThanTheData`
pins the guard. Both were verified by reverting the production change under each — the first fails at
`Expected: False / But was: True` on the contaminant peptide, the second with the sentence it
replaces.

**`DesignNamingNoFileInThisRun_WarnsAndSkips`.** You are right, and it was the reason string that had
gone stale rather than the test. It now says why there is one warning — the read fails before
anything is copied — and the fact that replaced the false one is asserted rather than described, so
it cannot go stale the same way: `archived` is `False`, as you measured. With the copy moved back
above the read, it fails alongside `UnreadableDesign_WarnsAndSkips`, the two of them pinning the same
ordering from opposite guards.

**The comment I have left.** It is the one thing here I have not done, and I would rather say so than
quietly trim it: the ordering it explains is exactly the kind a later edit undoes while tidying, and a
commit message is not in front of whoever is doing that. If house style is firm on this I will cut it
to a line — say the word.

**Your contaminant analysis.** Followed it through and it holds — `SingleIdentifiedBioPolymerOrNull`
keeps a match only when `Distinct()` yields one element, so such a match was already excluded and its
contribution to a written target group was already zero. It is now a line in the body, as you
suggested: what moves is the row set, `PeptideQuantification.tsv` loses the all-zero target-side row,
and no protein-level number changes.

**The credit line.** Corrected, and you are right on the facts — there is no review or comment of
@pcruzparri's on #2755 to point at. He sent a draft review rather than posting it, and its finding was
the individual-file column regression, which is #2802's. The body now credits the bare-`catch` fix to
his #2776 review, where the `catch { /* ignore */ }` blocks are flagged, and says plainly that the
archive ordering and the contaminant filter came from reading #2755 back after it merged.

Full suite **1983 passed / 0 failed / 1 skipped** at `72a77926c`.

pcruzparri (approved):

pcruzparri (approved):

Alexander-Sol (approved):

PR #2807, opened 2026-09-10
Name the modifications in the index cache key instead of counting them

trishorts:
> **Project of origin:** `cleavageRules`

## The bug

The peptide index is cached beside the protein database and reused whenever `IndexingEngine.ToString()` matches the `indexEngine.params` file stored next to it — `MetaMorpheusTask.SameSettings` compares the two as literal text.

That key carried only the *counts* of the modifications:

```csharp
sb.AppendLine("Number of fixed mods: " + FixedModifications.Count);
sb.AppendLine("Number of variable mods: " + VariableModifications.Count);
```

So two searches whose modification lists differ but happen to be the same length share a key. The cached index stores decorated `PeptideWithSetModifications`, so the reuse is **wrong, not merely incomplete**:

- swap one variable mod for another of the same count and the run gets a cache hit on an index that contains **no peptides bearing the new modification at all**;
- swap a fixed mod and **every peptide mass in the index is wrong**.

Both complete and report success. And because the cache lives under `<database dir>/DatabaseIndex/`, a poisoned index is shared by every task, every run, and every user of that database on that disk.

### Demonstrated

Two searches over the same database — fixed Carbamidomethyl→Iodoacetamide, variable Oxidation→Phospho:

**Before** (the two keys are byte-identical):

```
Number of fixed mods: 1              Number of fixed mods: 1
Number of variable mods: 1           Number of variable mods: 1

Keys identical? YES — search B silently reuses search A's index
```

**After:**

```
Fixed mods: Carbamidomethyl on C[a4a360dcf322504a]     Fixed mods: Iodoacetamide derivative on C[2669692cb67f9f70]
Variable mods: Oxidation on M[2107fdafa9ba527f]        Variable mods: Phospho on S[84b5ba97ee180fcd]
Silac labels: none                                     Silac labels: none
Turnover labels: none                                  Turnover labels: none

Keys identical? NO — search B rebuilds, correctly
```

## The change

Each modification is rendered as its readable id followed by a short hash of its complete definition (`Modification.ToString()`, i.e. the mods.txt record: target, location restriction, chemical formula, monoisotopic mass, neutral losses, diagnostic ions).

Three decisions worth a look in review:

**Why a hash and not just the id.** A user who edits a custom modification file in place — keeping the name, changing the mass — would otherwise still get a cache hit. The hash covers every field, and keeps covering fields added to `Modification` later. The readable id stays in front of it so a human diffing two params files can see what actually changed.

**Why the list order is preserved rather than sorted.** Sorting looks tempting, so that reordering the list in the GUI does not force a needless rebuild. It is wrong. `GetVariableModificationPatterns` is truncated at `MaxModificationIsoforms` by a `yield break`, so *which* isoforms survive can depend on the order the modifications arrive in. Sorting would let two orders share one key and reuse each other's index — the same bug in a subtler form. Preserving order can only cost a needless rebuild, never a wrong reuse.

**Why SILAC labels are in here too.** `SilacLabels` and `TurnoverLabels` are handed to `Protein.Digest` alongside the modifications and change which peptides land in the index, but appeared nowhere in the key at all. Same defect class.

### Deliberately out of scope

- `Localizeable mods:` is still a sum of counts across proteins. Weaker than it looks, but the `Databases:` line already carries name + `CreationTime`, so a changed database file is caught.
- `CommonParameters.FragmentationParameters` is absent from the key and affects the *fragment* index rather than the peptide list. A separate surface.

### Migration cost

Every cached index on disk misses once and rebuilds. That is the point — those caches were keyed by a predicate that did not determine their contents. Nothing parses `indexEngine.params` (it is only ever compared whole), so unlike `DigestionParams.ToString()` there is no positional-field trap and lines can be replaced freely.

## Tests

`MetaMorpheus/Test/IndexCacheKeyTest.cs`, 26 cases. The helpers are `internal`, reached through the `InternalsVisibleTo("Test")` already declared on EngineLayer.

**Run against unmodified master to confirm they bite: 12 of the 15 that compile there fail.** The three that pass are the ones that should — `IdenticalModificationsGiveTheSameKey`, `TheKeyIsStableAcrossRepeatedCalls`, `AddingAModificationStillChangesTheKey` — because equal inputs must still reuse the cached index, or every search re-indexes for nothing.

Full suite green: **2026 passed, 0 failed, 1 skipped.**

### Mutation testing

Stryker.NET 4.16 over `IndexingEngine.cs`, test set scoped to this fixture. For the four new helper methods: **21 killed, 0 survived, 0 timeout.**

It found a real gap on the first pass — the `,` separator in `DescribeSilacLabels` could be deleted without any test noticing, because no test used more than one label. Three of the 26 tests exist because of that run.

Remaining, all pre-existing and untouched here: one survivor on `CurrentPartition = currentPartition + 1`, and four timeouts on the `WaterMonoisotopicMass` static initializer. Note also that Stryker's safe mode discards every mutant inside `ToString` itself, because a pre-existing untyped `Select` on the `Localizeable mods:` line makes its own mutant uncompilable (`CS0411`) — so the four changed lines in `ToString` are covered by the tests and the before/after above, but not by mutation score.

## Coordination

#2742 rewrites `CheckFiles`/`SameSettings` on the same cache-hit path and documents `ToString()` as "the cache key", but never touches these lines. Whichever lands second may need a small textual rebase inside `IndexingEngine.cs`; there is no semantic conflict.

https://claude.ai/code/session_01SQW1W2eic6fPz5YgUtYE2u

### DIFF
--- a/MetaMorpheus/EngineLayer/Indexing/IndexingEngine.cs
+++ b/MetaMorpheus/EngineLayer/Indexing/IndexingEngine.cs
@@ -7,6 +7,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Security.Cryptography;
 using System.Text;
 using System.Threading.Tasks;
 using Omics;
@@ -64,8 +65,12 @@ namespace EngineLayer.Indexing
             sb.AppendLine("Precursor Index: " + GeneratePrecursorIndex);
             sb.AppendLine("Search Decoys: " + DecoyType);
             sb.AppendLine("Number of proteins: " + ProteinList.Count);
-            sb.AppendLine("Number of fixed mods: " + FixedModifications.Count);
-            sb.AppendLine("Number of variable mods: " + VariableModifications.Count);
+            sb.AppendLine("Fixed mods: " + DescribeModifications(FixedModifications));
+            sb.AppendLine("Variable mods: " + DescribeModifications(VariableModifications));
+            sb.AppendLine("Silac labels: " + DescribeSilacLabels(SilacLabels));
+            sb.AppendLine("Turnover labels: " + (TurnoverLabels == null
+                ? "none"
+                : DescribeSilacLabel(TurnoverLabels.Value.StartLabel) + ">" + DescribeSilacLabel(TurnoverLabels.Value.EndLabel)));
             sb.AppendLine("Dissociation Type: " + CommonParameters.DissociationType);
             sb.AppendLine("Contaminant Handling: " + TcAmbiguity);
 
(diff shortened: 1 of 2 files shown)


nbollis (commented):
Would these mod focused methods be better served in mzlib and used in the Mods structure we have there? Either way works for me.

PR #2809, opened 2026-09-14
Cap glycan box mass and stop repeating per-box work, so N+O glyco searches with large glycan lists can run

trishorts:
## Summary

An N+O glyco search with a realistic glycan list was impractical: 40 O-glycans, 42 N-glycans and up to three O-glycans per peptide build **7.8 million glycan boxes** (203 s and 173 GB allocated before the first scan), and the per-candidate work then scaled with that count. This PR caps the box mass and removes work that was repeated for every candidate, box, graph node or edge although it depended only on the scan, the peptide or the box.

Five commits, each reviewable on its own:

1. **Cap glycan box mass at 4000 Da by default** (`GlycoSearchParameters.MaximumGlycanBoxMass`). Boxes over the cap (total O + N glycan mass) are skipped before their child boxes are built; decoys follow their targets. The cap applies to all three search types: O-, N- and N+O. In an N-glycan search the box is the single N-glycan, so N-glycans heavier than the cap are dropped; with the default this removes one of the 182 shipped N-glycans, HexNAc(7)Hex(8)Fuc(1)NeuAc(4) at 4028 Da. The same search builds 622,039 boxes in 14 s and 12 GB. *This is the one behavior change:* boxes heavier than 4000 Da are no longer searched. Task files without the setting get the default; the existing builder overloads stay uncapped.
2. **Build the glycan-box mass array once.** `FindOGlycan` copied every box mass into a new array for each candidate peptide just to binary-search it (about 12 ms per call at 622k boxes); `FindNGlycan` did the same.
3. **Stop recomputing box-only and scan-only work.** Oxonium intensities once per scan; one stable sort after the candidate loop; a lazily built, thread-safe per-box cache of the localization graph's valid-transition chart, child motif counts and added motifs; `GraphCheck` on cached counts; c/z site fragments grouped once per candidate.
4. **Select the top candidates by counting byte scores instead of sorting** every candidate scoring at least the cutoff. Same ids, same order, ties included.
5. **Cut allocations** found with an allocation trace (183 GB sampled in 5 minutes): return early when no box fits the precursor and fragment only when a box passes `GraphCheck`; allocation-free containment and first-added checks in the cache; structural `int[]` keys instead of comma-joined strings for child boxes; one placeholder `Modification` per motif.

## Results

All on the same two EThcD raw files (180422_10 and 180422_12), N+O search, 40 O- + 42 N-glycans, max 3 O-glycans.

| Workload | Before | After |
|---|---|---|
| O-only, 20,431 Swiss-Prot proteins, file 1, glyco engine | master: 23 min 22 s | 3 min 29 s |
| N+O, 3,883 proteins, file 1, glyco engine | cap only: 39% done after 34 min | 1 min 15 s |
| N+O, 54 proteins, file 1, glyco engine | cap only: 1 h 16 min | 13 s |
| N+O, 3,883 proteins, both files, whole task (7 threads) | — | about 4 min |

A CPU trace after these changes puts the shared fragment-index scoring loop at about half of the remaining CPU; glyco matching itself is now about 10%.

## Correctness

- Every perf commit was checked against the build before it on the 3,883-protein N+O search: **all output files byte-identical** (all PSMs including those above 1% FDR, 1% FDR PSMs, glyco PSMs, protein groups, localization tables), except run-time lines. Repeated runs of one build are byte-identical, so the search is deterministic.
- Randomized tests pin the rewritten pieces to the code they replace: `SelectTopCandidates` against the stable sort with early break (2,000 trials over ties, cutoffs, the byte range and topN −1…1000), and the cache's containment / first-added checks against `LocalizationGraph.TryGetLeft` and `GetDiff` (20,000 trials).
- Glyco tests: 120 of 120 pass.
- Still running: a byte-for-byte comparison of the final build against commit 1 alone (cap only) on the 54-protein search; the cap-only baseline takes well over an hour per file. I will post the result here.

## Overlaps to be aware of

- **#2761** also replaces the candidate sort in `GlycoSearchEngine` (with `DescendingScoreSorter` and a stamped scoring table); commit 4 here does the same job differently, so one of the two should give way.
- **#2742** edits the same `GlycoSearchTask` loop and `GlycoSearchEngine` constructor (`FragmentIndex` types, `indexParams`). The conflict is small; resolved locally in both orders.
- **#2820** is stacked on this PR and adds the cap to the GlycoSearch task window.
- A follow-up (building glycan boxes and the index once per task, and searching spectra files in parallel within the machine's threads and memory) builds on #2742 and will be opened after it merges.

https://claude.ai/code/session_0169ox8g978Suu9AxJMnkDg8


trishorts:
## Correctness check finished: byte-identical to commit 1 (cap only)

The comparison promised in the description is done. N+O search, 40 O- + 42 N-glycans, 54-protein FASTA, both raw files (180422_10 and 180422_12):

| Build | Wall clock |
|---|---|
| `cc40c56c4` (cap only) | 2 h 58 min 43 s |
| `396adaeb0` (this PR's tip) | 55 s |

Every result file is byte-identical between the two, including `AllPSMs.psmtsv`, the glyco PSM file, `_AllProteinGroups.tsv`, both localization tables, and an unfiltered export of every PSM regardless of q-value (7.4 MB; written by a local, uncommitted line added for this check). The only differences are run times in `results.txt`, plus an indexing-engine entry that appears only in the run that built the index rather than reading it from the cache. The intermediate `d174ff701` is also byte-identical to the baseline.

## Correction to the benchmark numbers in the description

The searches behind the Results table were not the semi-specific trypsin searches they look like. The task file named `Protease = "semi-trypsin"`, which current mzLib no longer ships (removed in mzLib #1005), and I substituted `trypsin` with `SearchModeType = "Semi"`. In the glyco search that substitution is **not** equivalent: `SearchModeType = Semi` sends `Protein.Digest` down `ProteinDigestion.SpeedySemiSpecificDigestion`, which only yields maximum-missed-cleavage peptides plus fixed-length truncations meant to be trimmed later by the non-specific search engine. The glyco search never trims them, so it searched a small, distorted subset of the real peptide space.

What this means for this PR:

- **The correctness evidence stands.** Every before/after pair used identical settings, and the changes do not touch digestion, so byte-identical output still shows each commit changes nothing.
- **The speedups are real but measured on a smaller search than a true semi-specific one.** With a genuine semi-specific trypsin restored locally (a `semi-trypsin` entry with specificity `semi`), the same kind of search indexes far more peptides; for example, an O-only search of 130 proteins went from 11 to 348 glyco PSMs at 1% FDR. Absolute run times on real searches will be longer than the table suggests, and the relative gains may differ.
- With real semi-trypsin and settings matched to an independent O-Pair analysis of the same six raw files, this build reproduces it closely: 771 vs 783 O-glyco PSMs at 1% FDR, and on the 628 scans both call, 623 agree on the peptide and 621 on the glycan composition.

The protease problem itself is being fixed separately, in an isolated mzLib PR that restores a real semi-specific trypsin and documents how a `semi` protease differs from `SearchModeType = Semi`, with unit tests.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanBoxLocalizationCache.cs
```diff
@@ -0,0 +1,162 @@
...
+            return true;
+        }
+
+        /// <summary>
+        /// The first element of LocalizationGraph.GetDiff(pre, current), without allocating. GetDiff lists the ids of
+        /// <paramref name="current"/> left after removing those of <paramref name="pre"/>, grouped by id in the order each id first
+        /// appears in current, so its first element is the earliest-appearing id that current holds more times than pre.
+        /// <paramref name="pre"/> must be contained in current, as GetDiff also requires.
+        /// </summary>
+        internal static int FirstAdded(int[] pre, int[] current, out bool anyAdded)
```
RayMSMS: The name of the function could be "AddedGlycan", which is more informative.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanBoxLocalizationCache.cs
```diff
@@ -0,0 +1,162 @@
...
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Whether <paramref name="whole"/> holds every id of <paramref name="part"/>, counting repeats; what
+        /// LocalizationGraph.TryGetLeft(whole, part) returns, without allocating.
+        /// </summary>
+        internal static bool ContainsAll(int[] whole, int[] part)
```
RayMSMS: The naming part: the name of the function could be "IsGlycanCoveredBy"

MetaMorpheus/EngineLayer/GlycoSearch/GlycanBoxLocalizationCache.cs
```diff
@@ -0,0 +1,162 @@
...
+                ChildMotifCounts[y] = new MotifCount(children[y].ModIds);
+                AddedMotif[y] = new string[children.Length];
+                for (int preY = 0; preY <= y; preY++)
+                {
+                    // As BuildValidChart: the later child holds at most one more mod and contains every mod of the earlier child.
+                    if (children[y].NumberOfMods <= children[preY].NumberOfMods + 1
+                        && (children[preY].NumberOfMods == 0 || ContainsAll(children[y].ModIds, children[preY].ModIds)))
+                    {
+                        ValidChart[y][preY] = true;
+                        int added = FirstAdded(children[preY].ModIds, children[y].ModIds, out bool anyAdded);
```
RayMSMS: `added` could be `addedGlycan`

MetaMorpheus/EngineLayer/GlycoSearch/LocalizationGraph.cs
```diff
@@ -65,6 +65,100 @@ public LocalizationGraph(SortedDictionary<int, string> modPos, ModBox modBox, Mo
...
+                    adjNode.CurrentCost = cost;
+                    if (x == 0)
+                    {
+                        adjNode.maxCost = cost;
+                    }
+                    else
+                    {
+                        double maxCost = 0;
+                        var motifInThisPos = modPos_motif[x];
+                        var validRow = cache.ValidChart[y];
```
RayMSMS: `validRow` -> `isValidRow`

MetaMorpheus/EngineLayer/GlycoSearch/LocalizationGraph.cs
```diff
@@ -65,6 +65,100 @@ public LocalizationGraph(SortedDictionary<int, string> modPos, ModBox modBox, Mo
...
+                        var addedMotifRow = cache.AddedMotif[y];
```
RayMSMS: `addMotifRow` -> `addMotifInThisRow`

MetaMorpheus/EngineLayer/GlycoSearch/LocalizationGraph.cs
```diff
@@ -65,6 +65,100 @@ public LocalizationGraph(SortedDictionary<int, string> modPos, ModBox modBox, Mo
...
+                        for (int preY = 0; preY <= y; preY++)
+                        {
+                            if (validRow[preY] && localizationGraph.array[x - 1][preY] != null && (addedMotifRow[preY] == null || addedMotifRow[preY] == motifInThisPos))
```
RayMSMS: Just an option, not a pr blocker. In order to increase readability, we might make a tiny helper named "EdgeAllowed" (glycan check and motif check). But it may cost some computation, especially since this is a hot spot within the nested loop.

MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
```diff
@@ -255,6 +241,62 @@ protected override MetaMorpheusEngineResults RunSpecific()
...
+            }
+
+            System.Runtime.InteropServices.CollectionsMarshal.SetCount(topCandidates, kept);
+            var output = System.Runtime.InteropServices.CollectionsMarshal.AsSpan(topCandidates);
+            foreach (int id in candidateIds)
+            {
+                int score = scores[id];
+                if (score >= lowestKeptScore)
+                {
+                    output[countsByScore[score]++] = id;
```
RayMSMS, line 295: Pretty clever, use the bin to replace sorting. Complexity from NlogN to N

nbollis (dismissed):

MetaMorpheus/EngineLayer/GlycoSearch/GlycanBoxLocalizationCache.cs
trishorts: Renamed to AddedGlycan in 96fa1bcda.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanBoxLocalizationCache.cs
trishorts: Renamed to IsGlycanCoveredBy in 96fa1bcda. I also swapped the two arguments so the call reads the way the name does: IsGlycanCoveredBy(glycan, cover), i.e. the earlier child's ids first and the later child's ids second. Same comparison, same result.

MetaMorpheus/EngineLayer/GlycoSearch/GlycanBoxLocalizationCache.cs
trishorts: Renamed to addedGlycan in 96fa1bcda.

MetaMorpheus/EngineLayer/GlycoSearch/LocalizationGraph.cs
trishorts: Renamed to isValidRow in 96fa1bcda.

MetaMorpheus/EngineLayer/GlycoSearch/LocalizationGraph.cs
trishorts: Renamed to addedMotifInThisRow in 96fa1bcda (the variable was addedMotifRow).

MetaMorpheus/EngineLayer/GlycoSearch/LocalizationGraph.cs
trishorts: Leaving this one as it is, since you flagged the cost yourself: that condition is inside the localization graph's inner loop over preY, which is the hot path this PR is trying to make cheaper, and both checks are single array reads. Happy to add it if you would rather have the readability.

MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs
trishorts: Thanks. Worth noting the counting sort is what makes the tie behaviour exact too: candidates with equal coarse scores keep index order, which is why the output stays byte-identical.

RayMSMS (dismissed):

nbollis (dismissed):

RayMSMS (approved):

zhuoxinshi (approved):

nbollis (approved):

MetaMorpheus/TaskLayer/GlycoSearchTask/GlycoSearchParameters.cs
```diff
@@ -43,6 +44,12 @@ public GlycoSearchParameters()
         public int GlycoSearchTopNum { get; set; }
         public int MaximumOGlycanAllowed { get; set; }
```
nbollis, line 46: **The new default mass cap silently changes the search space**

MaximumGlycanBoxMass now defaults to 4000 Da and is applied to O, N+O, and N-glycan boxes. Existing parameter files that omit this new field can therefore silently lose heavy glycans or compositions that were searchable before, with no warning in the task output.

Suggestion: make the cap opt-in for backward compatibility, or report/warn when entries are excluded.

nbollis (commented):
## TL;DR
The PR improves glycan-box construction, candidate selection, and localization/search caching. Targeted validation passed (build plus 23 selected glyco/cache tests), and specialist review found the cached localization paths equivalent to the uncached behavior. Verdict: request changes for one correctness edge case; the remaining findings are minor behavior/memory concerns.

## Key risks
- A user-configurable mass cap can leave the O/N+O box list empty, but that path dereferences First/Last before checking.
- Static box arrays are replaced for each partition, so cache release only clears the final partition's boxes.
- The new 4000 Da default silently changes existing searches by excluding heavy glycans/compositions.

## Required changes
- Guard the empty GlycanBoxes case in the O/N+O path before First/Last is used.

## Nice-to-haves
- Release caches for every partition-owned box array.
- Make the default mass-cap behavior backward-compatible or emit a clear exclusion warning.

## Test plan
- Build succeeded.
- 23 selected glyco/cache tests passed.
- Specialist review checked cached/uncached localization equivalence, candidate selection stability, mass-cap builders, and lazy cache lifetime.

---
**Additional findings (included as summary notes):**

- **Empty GlycanBoxes can throw in the O/N+O path**

The new MaximumGlycanBoxMass filter can produce an empty GlycanBoxes array when the cap is below every available combination. MatchNGlycopeptide guards the analogous empty NGlycans case, but MatchGlycopeptide still calls GlycanBoxes.First().Mass/GlycanBoxes.Last().Mass before FindOGlycan's bounds checks, so O-only and N+O searches can throw InvalidOperationException instead of returning no matches.

Suggestion: guard GlycanBoxes.Length == 0 before First/Last is used. (path: MetaMorpheus/EngineLayer/GlycoSearch/GlycoSearchEngine.cs, position not available)
- **Cache release only reaches the last partition's box array**

GlycanBox.OGlycanBoxes/NOGlycanBoxes are static and reassigned by each GlycoSearchEngine construction. In a multi-partition run, GSMs from earlier partitions retain their own GlycanBox instances, but the single ReleaseSearchCaches call operates only on the final static array. Earlier arrays therefore retain their caches, reducing the intended memory release.

Suggestion: release every partition-owned array after its readers finish. (path: MetaMorpheus/TaskLayer/GlycoSearchTask/GlycoSearchTask.cs, position not available)

PR #2811, opened 2026-09-15
Add 5'- and 3'-terminus columns to OSM TSV output

nbollis:
## Motivation

OSM (`.osmtsv`) output does not currently report the terminus chemistry of each matched oligo, even though every `OligoWithSetMods` already carries its `FivePrimeTerminus` / `ThreePrimeTerminus`, and mzLib's reader (`OsmFromTsv`) already understands the `5'-Terminus` and `3'-Terminus` columns declared in `SpectrumMatchFromTsvHeader`. This change closes that gap for OSM output only — `.psmtsv` files are intentionally unchanged.

## Summary of changes

- `PsmTsvWriter.AddPeptideSequenceData` now emits two oligo-only columns, `5'-Terminus` then `3'-Terminus`, immediately after `Essential Sequence`.
- The columns are gated on `GlobalVariables.AnalyteType == AnalyteType.Oligo` (same pattern as the ProForma gate), so the header and every data row include/exclude them together and protein/proteoform PSM output is untouched.
- Values are the terminus chemical-formula strings (`ThisChemicalFormula.Formula`) resolved across ambiguous matches with the established convention: a single value when the term is identical across hypotheses, pipe `|`-separated when it differs.

## Testing notes

- Unit tests added in `PsmTsvWriterTests` covering: column placement immediately after Essential Sequence, absence from Peptide/Proteoform output, header/data-row alignment, ambiguous pipe-joining, and an mzLib round-trip (write `.osmtsv` → `SpectrumMatchTsvReader.ReadOsmTsv` → assert each row's termini match).
- End-to-end RNA search test (`SearchTaskWithRna`) asserts the written `AllOSMs.osmtsv` header carries both columns.
- Verified with a clean build; targeted suites pass (30/30 across `PsmTsvWriterTests` + `SearchTaskWithRna`).

## Migration steps

None. This only adds columns to newly written `.osmtsv` files. Readers map columns by name, and older files without these columns still parse (mzLib falls back to residue-derived termini when the columns are absent).

## Screenshots

N/A — TSV output change; no UI change.


MetaMorpheus/EngineLayer/SpectrumMatch/PsmTsvWriter.cs
```diff
@@ -206,6 +207,14 @@ internal static void AddPeptideSequenceData(Dictionary<string, string> s, Spectr
             if (GlobalVariables.AnalyteType == AnalyteType.Proteoform)
                 s[SpectrumMatchFromTsvHeader.ProForma] = pepWithModsIsNull ? " " : Resolve(pepsWithMods.Select(b => b.ToProFormaString())).ResolvedString;
             s[SpectrumMatchFromTsvHeader.EssentialSequence] = pepWithModsIsNull ? " " : sm.EssentialSequence != null ? sm.EssentialSequence : Resolve(pepWithModsIsNull ? null : pepsWithMods.Select(b => b.EssentialSequence(ModsToWritePruned))).ResolvedString;
+            // 5'- and 3'-termini are oligo-only columns
+            if (GlobalVariables.AnalyteType == AnalyteType.Oligo)
+            {
+                s[SpectrumMatchFromTsvHeader.FivePrimeTerminus] = pepWithModsIsNull ? " " :
+                    Resolve(pepsWithMods.Select(b => (b as OligoWithSetMods)?.FivePrimeTerminus?.ThisChemicalFormula?.Formula)).ResolvedString;
```
trishorts, line 214: Non-blocking: the two partial branches behind the red `codecov/patch` are these `?.` null paths (`as OligoWithSetMods` / `FivePrimeTerminus` / `ThisChemicalFormula`). Under the `AnalyteType.Oligo` gate every match is an `OligoWithSetMods` with a terminus, so those branches are effectively unreachable. Fine to leave as defensive code; just noting why patch coverage is red.

trishorts (approved):
Looks good. Header and data rows share the same `AddPeptideSequenceData` dictionary, so the new oligo-gated columns can't misalign, and the psmtsv output is untouched. Built locally and ran `PsmTsvWriterTests` + `SearchTaskWithRna`: 45/45 pass.

The red `codecov/patch` is partial-branch coverage only (5 patch lines, 5 hit, 2 partials); see the inline note. Not blocking.

zhuoxinshi (approved):

RayMSMS (approved):

PR #2812, opened 2026-09-15
Load semi-trypsin settings as trypsin + Semi, refuse seed digestion outside the non-specific search, and let the Glyco window choose semi-specific digestion

trishorts:
**Draft until an mzLib release contains smith-chem-wisc/mzLib#1303** (then the mzLib package references are bumped to it). Reworked to match the cut-down smith-chem-wisc/mzLib#1304: no `SpecificTermini`; everything uses the existing `SearchModeType`.

## Why

In mzLib, `SearchModeType` and `FragmentationTerminus` together decide what digestion returns:

| SearchModeType | FragmentationTerminus | Digestion returns |
|---|---|---|
| Full | any | fully specific peptides |
| Semi | Both | semi-specific peptides (since mzLib #1303) |
| Semi | N or C | seeds for the non-specific engine to trim |
| None | any | singleN / singleC seeds |

Only the non-specific search engine can use seeds. Before mzLib #1303, Semi + Both also returned seeds, so a semi-specific Glyco (or Classic/Modern) search silently lost most of its identifications. mzLib #1005 had removed the `semi-trypsin` protease in favour of trypsin + Semi, so settings files naming it stopped loading, and the substitute was broken. This PR makes MetaMorpheus use trypsin + Semi correctly everywhere and stops the silent failures.

## Changes (one commit each)

1. **Settings files naming `semi-trypsin` load again.** The `IDigestionParams` converter in `tomlConfig`, which every task-settings load path uses, reads them as the shipped fully specific protease with the same motifs (`K|,R|`, found by motif because mzLib #1186 renames trypsin) plus `SearchModeType` Semi, keeping every other setting, and warns. A non-specific search file keeps its Semi/None and terminus; a user-defined `semi-trypsin` is used as defined; any other unknown protease still fails.
2. **End-to-end glyco test.** Trypsin + Semi on the one-spectrum leukosialin data identifies the semi-tryptic TTGSLEPSSGASGPQVSSVK (three O-glycans); a fully specific control does not. Fails against mzLib 1.0.590, passes with #1303.
3. **Glyco window: "Semi-specific digestion" checkbox.** Saves Semi or Full, always with terminus Both. Tooltip says it replaces semi-trypsin.
4. **Crosslink, GPTMD and calibration windows stop resetting the search mode to Full on save**; they keep what the task loaded. All four windows show `SpecificProtease` (not singleN/singleC). The rules are in `GuiFunctions.Util.TaskWindowSearchMode` so they can be tested.
5. **Tasks that would search seeds they cannot use are refused.** `DigestionSearchModeCheck`: Semi + N/C or None, in anything but the non-specific search. `EverythingRunnerEngine` checks all tasks before running any and refuses with a warning (the #2759 pattern: a throw reaches the GUI as a crash); `RunTask` throws the same message as a backstop.
6. **Index cache:** the key for Semi + Both gains a line, so an index of seeds cached before mzLib #1303 is never reused. Other keys are unchanged.
7. **mzIdentML `Enzyme/@semiSpecific`** is now written (it was always omitted: `semiSpecificSpecified` was never set) and is true for trypsin + Semi.

## Tests

All written first and seen failing (against master, a do-nothing stub, or mzLib 1.0.590), with mutation checks where the test passed by construction:

- `LegacySemiTrypsinSettingsTests` (11): the two orphaned `GlycoTestData/*_ETD*.toml` files through both readers, digest equivalence with the old Semi protease, seed modes, user-defined and unknown proteases.
- `GlycoSemiSpecificSearchTests` (2)
- `TaskWindowSearchModeTests` (15)
- `DigestionSearchModeRefusalTests` (46): every task type × search mode × terminus, the runner refusal, the backstop.
- `IndexCacheKeySemiSpecificTests` (5)
- `MzIdentMlSemiSpecificTests` (7)

Full MetaMorpheus suite against a local mzLib containing #1303: 2,094 passed, 0 failed, 1 skipped.

## Not in this PR

- The unreachable semi-trypsin recommendation prompt (`UpdateGUISettings.UseSemiTrypsinRecommendedSettings`) is left alone: removing it means changing persisted GUI settings.
- One-sided semi-specific digestion (only the N- or only the C-terminus specific) was dropped with mzLib #1304's `SpecificTermini`.


RayMSMS (dismissed):

MetaMorpheus/GUI/TaskWindows/GlycoSearchTaskWindow.xaml.cs
```diff
@@ -196,7 +197,8 @@ private void UpdateFieldsFromTask(GlycoSearchTask task)
             MaxPeptideLengthTextBox.Text = task.CommonParameters.DigestionParams.MaxLength == int.MaxValue ? "" : task.CommonParameters.DigestionParams.MaxLength.ToString(CultureInfo.InvariantCulture);
             if (task.CommonParameters.DigestionParams is DigestionParams digestionParams)
             {
-                proteaseComboBox.SelectedItem = digestionParams.Protease;
```
nbollis, line 200: The checkbox is derived only from SearchModeType, while the save path always writes ForSemiSpecificChoice(...), which forces FragmentationTerminus.Both and maps an unchecked state to Full. A loaded (Semi, N/C) seed digestion is therefore changed to (Semi, Both), and a loaded None digestion is changed to Full+Both. This bypasses the new run-time refusal and silently changes which peptides are searched.

Suggested fix: Detect and preserve or refuse unrepresentable mode/terminus combinations when loading/saving; do not coerce them to Full+Both without a visible warning.

MetaMorpheus/TaskLayer/SearchTask/MzIdentMLWriter.cs
```diff
@@ -576,7 +576,11 @@ public static void WriteMzIdentMl(IEnumerable<SpectralMatch> psms, List<EngineLa
                 {
```
nbollis, line 579: The PR unconditionally sets semiSpecificSpecified=true, so fully-specific and non-specific searches now serialize semiSpecific="false" where previous output omitted the attribute. This is a format-wide compatibility change unrelated to the targeted semi-specific case and may affect consumers that distinguish omission from explicit false.

Suggested fix: Keep omission for unaffected searches, or document and verify this deliberate format compatibility change with downstream readers.

MetaMorpheus/GUI/TaskWindows/XLSearchTaskWindow.xaml.cs
```diff
@@ -280,13 +281,17 @@ private void SaveButton_Click(object sender, RoutedEventArgs e)
             int MaxPeptideLength = string.IsNullOrEmpty(MaxPeptideLengthTextBox.Text) ? int.MaxValue : (int.Parse(MaxPeptideLengthTextBox.Text, NumberStyles.Any, CultureInfo.InvariantCulture));
             int MaxModificationIsoforms = (int.Parse(maxModificationIsoformsTextBox.Text, CultureInfo.InvariantCulture));
```
nbollis, line 286: The XL, GPTMD, and calibration windows now preserve loaded SearchModeType and FragmentationTerminus, but expose no control or indicator for them. A user can see an ordinary protease selection while the saved task retains an unrepresentable seed/non-specific mode and is later refused at run time.

Suggested fix: Show a read-only digestion-mode indicator or provide a way to resolve unsupported loaded modes before saving.

nbollis (commented):
Review of MetaMorpheus#2812 at b8ddb8819771f1375febe583ab7fbdb4e97e029c. Three deduplicated findings were recorded: two minor correctness/compatibility concerns and one non-blocking UI usability concern. All three resolve to inline diff positions. No GitHub comments were posted; explicit approval is required.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -46,6 +46,98 @@ public enum MyTask
...
+
+        /// <summary>
+        /// Semi-specific proteases mzLib used to ship, with the cleavage motifs each had. mzLib #1005 (February 2026) removed
+        /// "semi-trypsin", the only one, because a semi-specific search is asked for with <c>SearchModeType = Semi</c> on
+        /// the fully specific protease. Settings files written before that still name it, including the O-Pair Search
+        /// paper's glyco settings.
+        /// </summary>
+        private static readonly (string Name, string Motifs)[] RemovedSemiSpecificProteases =
+        {
+            ("semi-trypsin", "K|,R|"),
```
nbollis, line 108: is this necessary given the changes to the mzlib implementation?

MetaMorpheus/GUI/TaskWindows/GlycoSearchTaskWindow.xaml.cs
trishorts: Agreed. Saving still writes a mode a glyco search can use: Full + Both, or Semi + Both, from the checkbox. But the window no longer does it silently. When the loaded task asks for seeds (SearchModeType None, or Semi with FragmentationTerminus N or C), a warning under the checkbox names the loaded setting and says that saving will change it to fully or semi-specific digestion at both termini, as set by the box.

Such a task can only come from a hand-edited settings file. Before this PR, the window reset every task to Full + Both on save with no message at all.

The rule for which settings give seeds is now one method, `DigestionSearchModeCheck.AsksForSeeds`. The run-time refusal and the window warnings both use it, and a test checks that the warnings appear for exactly the settings the refusal rejects.

MetaMorpheus/TaskLayer/SearchTask/MzIdentMLWriter.cs
trishorts: Agreed. `semiSpecificSpecified` is now true only when the search is semi-specific, so fully specific and non-specific searches write the same Enzyme element as before this PR, with no semiSpecific attribute. Only the case that was wrong changes: a semi-specific search now writes `semiSpecific="true"`. The tests assert that the attribute is absent in the other cases.

MetaMorpheus/GUI/TaskWindows/XLSearchTaskWindow.xaml.cs
trishorts: Agreed. The crosslink, GPTMD and calibration windows now show a read-only warning when the loaded task asks for seeds. It names the setting, says that saving keeps it and that the task will be refused when run, and says to set SearchModeType Full, or Semi with FragmentationTerminus Both, in the task's settings file. These windows have no control for it, so the warning is all they can offer. It is hidden for every other task.

trishorts:
@nbollis all three findings are addressed in 568552ee5 (mzIdentML: 6de3d53b6; window warnings: 568552ee5); the replies are in each thread. The branch is also rebased onto master ee506c655 (#2742, #2809, #2811, #2820).

@RayMSMS the push dismissed your approval. What changed since you approved: task windows now show a warning for a loaded digestion they can't represent; mzIdentML writes semiSpecific only for semi-specific searches; the seed rule is shared between the refusal and the windows. Could you take another look?

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -47,6 +47,98 @@ public enum MyTask
...
+        /// protease, so it becomes <c>Semi</c>. Its FragmentationTerminus is kept; in glyco, crosslink and classic settings
+        /// it is Both, which asks mzLib for the semi-specific peptides themselves (mzLib #1303). A file that already said
+        /// Semi or None is a non-specific search asking for seeds, so its search mode and terminus are kept as they are.</para>
+        /// <para><b>Replacement.</b> Found by motif, not by name, because the name of the no-proline-rule trypsin is itself
+        /// changing (mzLib #1186). A protease the user has defined under the old name is in the dictionary and is used as
+        /// defined; nothing is translated.</para>
+        /// <para>Settings are rebuilt through the DigestionParams constructor rather than patched, so that Protease and
+        /// SpecificProtease stay consistent with the search mode.</para>
+        /// </remarks>
+        private static DigestionParams ReadProteinDigestionParams(TomlTable table)
```
nbollis, line 130: Use the normal toml reader instead of a new one

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -658,6 +750,13 @@ public MyTaskResults RunTask(string output_folder, List<DbForTask> currentProtei
                 var stopWatch = new Stopwatch();
                 stopWatch.Start();
 
+                // backstop for callers that bypass EverythingRunnerEngine, which refuses these tasks with a warning first
+                string digestionRefusal = DigestionSearchModeCheck.GetRefusal(this, displayName);
```
nbollis: we usually assume tasks are valid here, moving validation to the GUI level like the rests of the tasks will make it so this is not needed.

MetaMorpheus/TaskLayer/DigestionSearchModeCheck.cs
```diff
@@ -0,0 +1,58 @@
...
+        public static bool AsksForSeeds(DigestionParams digestionParams) =>
+            digestionParams.SearchModeType == CleavageSpecificity.None
+            || (digestionParams.SearchModeType == CleavageSpecificity.Semi && digestionParams.FragmentationTerminus is FragmentationTerminus.N or FragmentationTerminus.C);
+
+        /// <summary>
+        /// The message explaining why this task cannot run with its digestion settings, or null when it can.
+        /// </summary>
+        /// <param name="task">The task to check.</param>
+        /// <param name="taskName">The name the user knows the task by, included in the message when given.</param>
+        public static string GetRefusal(MetaMorpheusTask task, string taskName = null)
```
nbollis: This is good, but should likely be in our task validator instead of a new structure.

MetaMorpheus/EngineLayer/Indexing/IndexingEngine.cs
```diff
@@ -85,6 +85,11 @@ public override string ToString()
             sb.AppendLine("digestionTerminus: " + CommonParameters.DigestionParams.FragmentationTerminus);
             sb.AppendLine("maxModsForEachPeptide: " + CommonParameters.DigestionParams.MaxMods);
             sb.AppendLine("cleavageSpecificity: " + CommonParameters.DigestionParams.SearchModeType);
+            // Semi + Both digested into C-terminal seeds before mzLib #1303 and into semi-specific peptides since, under the same
+            // settings. This line keeps an index of seeds cached by an older MetaMorpheus from ever matching a new search.
+            if (CommonParameters.DigestionParams.SearchModeType == Omics.Digestion.CleavageSpecificity.Semi
+                && CommonParameters.DigestionParams.FragmentationTerminus == Omics.Fragmentation.FragmentationTerminus.Both)
+                sb.AppendLine("semiSpecificDigestion: peptides, not seeds (mzLib #1303)");
```
nbollis: Is this what you want in the manuscript prose? It is more about the code than the search itself.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
trishorts: Yes, it is still needed. mzLib 1.0.591 has no `semi-trypsin` row, so without it, a settings file that names semi-trypsin throws `KeyNotFoundException` in the `Protease` converter, and the GUI refuses the file. That includes the O-Pair Search paper's glyco settings. What #1303 changed is that the translation is now correct: trypsin with `SearchModeType = Semi` and `FragmentationTerminus = Both` digests into the same peptides semi-trypsin gave (`LegacySemiTrypsinSettingsTests` checks this). A name alias alone would not be enough, because those files say `SearchModeType = Full`; their semi-specificity came from the protease.

The replacement is looked up by cleavage motif rather than by name because #1186 is renaming the no-proline trypsin, so a hard-coded `trypsin` would quietly pick up the proline rule once that merges.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
trishorts: Done in 6ab9b127c. The separate reader is gone. The converter now calls the normal `Get<DigestionParams>()`. Only when the file names semi-trypsin does it first rewrite `Protease`/`SpecificProtease` and switch `SearchModeType` from Full to Semi in the parsed table (Nett's `TomlTable.Update`), and warn. Every other file goes straight through the normal reader. The legacy-settings tests are unchanged and still pass. With the search-mode switch removed, 6 of them fail.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
trishorts: Agreed, and removed in 6ab9b127c. `RunTask` no longer checks, and neither does `EverythingRunnerEngine`. The check now runs where the other run-level checks do:
- **GUI:** `TaskValidator.CheckDigestionSearchMode(task, name)`, called for every task from `RunAllTasks_Click`, next to the experimental-design check. It takes the task itself rather than text-box strings, because a task loaded from a toml never goes through a task window.
- **CMD:** `Program.RefuseSeedDigestion`, next to `ResolveExperimentalDesign`. It prints the reason and returns exit code 6. CMD cannot reach `TaskValidator`, which lives in the WPF project.

One consequence worth noting for #2709: the Avalonia GUI starts `EverythingRunnerEngine` directly, so it would need the same one-line check before a run.

MetaMorpheus/TaskLayer/DigestionSearchModeCheck.cs
trishorts: Done. `DigestionSearchModeCheck` is deleted. The rule (`AsksForSeeds`) and the message (`GetSeedDigestionRefusal`) are now on `MetaMorpheusTask`, so the WPF validator, the command line and the task windows' warnings all use the same text. The GUI check itself is in `TaskValidator`.

MetaMorpheus/EngineLayer/Indexing/IndexingEngine.cs
trishorts: Removed in 8d17eb9be. To clarify: `IndexingEngine.ToString()` does not reach the prose. It is written only to `indexEngine.params`, which is the cache key `SameSettings` compares. The line's job was to stop a cached index of seeds, written before #1303, from matching a new Semi + Both search.

It is no longer needed. #2742 moved the index files to `.bin` with a format header, so no released MetaMorpheus's cached index can match any more. Every master build that writes `.bin` already uses mzLib 1.0.591. `IndexCacheKeySemiSpecificTests` is deleted with it.

trishorts:
@nbollis all five comments are addressed in 8d17eb9be (index cache-key line) and 6ab9b127c (everything else), with replies in each thread. In short:
- The seed-digestion check moved to `TaskValidator` (GUI Run button) and to the command line, next to the experimental-design check.
- `DigestionSearchModeCheck` and the `EverythingRunnerEngine`/`RunTask` checks are removed.
- The semi-trypsin translation now goes through the normal toml reader.
- The index cache-key line is removed.

On "other tasks": the crosslink, GPTMD and calibration windows have no semi-specific control. If we want one there, that would be a separate PR.

@RayMSMS your approval was dismissed by the earlier push; could you take another look when you have a moment?

pcruzparri (commented):
Code review — 9 finding(s), 4 at severity >= High, posted as inline comments.

MetaMorpheus/GUI/TaskWindows/CalibrateTaskWindow.xaml.cs
```diff
@@ -289,13 +294,17 @@ private void SaveButton_Click(object sender, RoutedEventArgs e)
...
                 digestionParamsToSave = new DigestionParams(
                     protease: protease.Name,
                     maxMissedCleavages: maxMissedCleavages,
                     minPeptideLength: minPeptideLength,
                     maxPeptideLength: maxPeptideLength,
                     maxModificationIsoforms: maxModificationIsoforms,
-                    maxModsForPeptides: maxModsPerPeptide);
+                    maxModsForPeptides: maxModsPerPeptide,
+                    searchModeType: searchModeType,
+                    fragmentationTerminus: fragmentationTerminus);
```
pcruzparri, line 307: **High** — Calibration save drops `InitiatorMethionineBehavior`

**Problem:** This constructor call gains `searchModeType` and `fragmentationTerminus` but still never passes `initiatorMethionineBehavior`, so it falls back to mzLib's default of `Variable`. The window does load the setting — `UpdateFieldsFromTask` puts it into `InitiatorMethionineBehaviorComboBox` (line 87), which is populated at 248-250 and shown at `CalibrateTaskWindow.xaml:213` — so the value is on screen and then dropped on the way back out. GPTMD, Glyco and XL all pass this argument; calibration is the only one that doesn't.

**Impact:** Open a calibration task whose `InitiatorMethionineBehavior` is `Retain` or `Cleave`, press Save without touching anything, and it comes back `Variable`. N-terminal methionine is then handled differently, so calibration runs against a different peptide set than the settings file asked for, with nothing to say so. That is the same failure this PR is fixing for the search mode, one argument over.

**Suggested fix:** Add `initiatorMethionineBehavior: (InitiatorMethionineBehavior)InitiatorMethionineBehaviorComboBox.SelectedIndex` to the call, matching `GPTMDTaskWindow.xaml.cs:459`.

**Affected test scenarios:** Nothing covers it today — `TaskWindowSearchModeTests` asserts the search mode and terminus survive a load/save round trip, but no test asserts the same for the other digestion fields. A case that builds `DigestionParams` with `InitiatorMethionineBehavior.Retain`, replays this window's argument list and asserts the value comes back fails on the current head and passes with the fix.

MetaMorpheus/GUI/TaskWindows/XLSearchTaskWindow.xaml.cs
```diff
@@ -280,13 +285,17 @@ private void SaveButton_Click(object sender, RoutedEventArgs e)
...
             DigestionParams digestionParamsToSave = new DigestionParams(
                 protease: protease.Name,
                 maxMissedCleavages: MaxMissedCleavages,
                 minPeptideLength: MinPeptideLength,
                 maxPeptideLength: MaxPeptideLength,
                 maxModificationIsoforms: MaxModificationIsoforms,
-                initiatorMethionineBehavior: InitiatorMethionineBehavior);
+                initiatorMethionineBehavior: InitiatorMethionineBehavior,
+                searchModeType: searchModeType,
+                fragmentationTerminus: fragmentationTerminus);
```
pcruzparri, line 298: **High** — Crosslink save resets `MaxModsForPeptide` to 2

**Problem:** This call picks up `searchModeType` and `fragmentationTerminus` but still doesn't pass `maxModsForPeptides`, so it falls back to mzLib's default of 2. Unlike the calibration case there is no control for it anywhere in this window, so the value is never on screen — it is read from the task, ignored, and written back as 2. Calibrate, GPTMD and Glyco all pass it.

**Impact:** A crosslink task configured with `MaxModsForPeptide` of 3 or more silently drops to 2 the first time someone opens the window and saves. Fewer modified peptides are considered and identifications are lost, with nothing in the UI to show the setting was ever there.

**Suggested fix:** Pass `maxModsForPeptides: TheTask.CommonParameters.DigestionParams.MaxMods`, alongside the `TaskWindowSearchMode.Preserve` call just above — there is no text box to read it from, so it has to come from the loaded task.

**Affected test scenarios:** None today. A case that builds `DigestionParams` with `maxModsForPeptides: 5`, replays this window's argument list and asserts it comes back 5 fails on the current head.

MetaMorpheus/GUI/TaskWindows/GlycoSearchTaskWindow.xaml.cs
```diff
@@ -366,14 +372,18 @@ private void SaveButton_Click(object sender, RoutedEventArgs e)
...
                 protease: protease.Name,
                 maxMissedCleavages: MaxMissedCleavages,
                 minPeptideLength: MinPeptideLength,
                 maxPeptideLength: MaxPeptideLength,
                 maxModificationIsoforms: MaxModificationIsoforms,
                 maxModsForPeptides: MaxModPerPep,
-                initiatorMethionineBehavior: InitiatorMethionineBehavior);
+                initiatorMethionineBehavior: InitiatorMethionineBehavior,
+                searchModeType: searchModeType,
+                fragmentationTerminus: fragmentationTerminus);
```
pcruzparri, line 386: **High** — All four windows reset `GeneratehUnlabeledProteinsForSilac` on save

**Problem:** None of the four task windows pass `generateUnlabeledProteinsForSilac`, `keepNGlycopeptide` or `keepOGlycopeptide`, so all three fall back to mzLib's constructor defaults (`true`, `false`, `false`). `SearchTaskWindow` is the only window that passes the SILAC flag. The same omission is at `CalibrateTaskWindow.xaml.cs:307`, `GPTMDTaskWindow.xaml.cs:566` and `XLSearchTaskWindow.xaml.cs:298`; I am raising it here because the shipped test data makes this window the demonstrable case.

**Impact:** `Test/GlycoTestData/GlycoSearchTaskconfig_ETD_Run3.toml` sets `GeneratehUnlabeledProteinsForSilac = false`. Load that task in this window, press Save without touching anything, and it comes back `true` — MetaMorpheus then also searches unlabeled proteins, changing SILAC quantification. Worth noting that `ReadProteinDigestionParams` in this same PR is careful to preserve all three flags when it translates a settings file, and then the next save throws them away.

**Suggested fix:** Pass all three through from the loaded `DigestionParams` in each of the four windows. Given how often this keeps recurring, a clone-with-overrides on `DigestionParams` — mzLib already has `Clone(FragmentationTerminus?)` — would make the fifth window correct by default rather than by remembering to name every field.

**Affected test scenarios:** None. Loading `GlycoSearchTaskconfig_ETD_Run3.toml`, replaying this window's argument list and asserting the flag survives fails on the current head.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -47,6 +47,144 @@ public enum MyTask
...
+        private static Protease FindShippedFullySpecificProtease(string motifs)
+        {
+            static string Signature(IEnumerable<DigestionMotif> m) => string.Join(";", m
+                .Select(x => $"{x.InducingCleavage}|{x.PreventingCleavage}|{x.CutIndex}|{x.ExcludeFromWildcard}")
+                .OrderBy(s => s, StringComparer.Ordinal));
+
+            string wanted = Signature(DigestionMotif.ParseDigestionMotifsFromString(motifs));
+            return ProteaseDictionary.Dictionary.Values
+                .Where(p => p.CleavageSpecificity == CleavageSpecificity.Full && Signature(p.DigestionMotifs) == wanted)
+                .OrderBy(p => p.Name, StringComparer.Ordinal)
```
pcruzparri, line 194: **High** — Legacy protease can resolve to a user's custom protease

**Problem:** `FindShippedFullySpecificProtease` searches `ProteaseDictionary.Dictionary.Values`, which `GlobalVariables.LoadDigestionAgents` populates from the user's editable custom proteases file as well as the embedded table, and then breaks ties with `.OrderBy(p => p.Name, StringComparer.Ordinal)`. Ordinal ordering puts every uppercase letter ahead of every lowercase one, so a user-defined fully specific protease with motifs `K|,R|` and a capitalised name wins over `trypsin`. The method name says "Shipped", but the search is not restricted to shipped entries.

**Impact:** mzLib's own `proteases.tsv` documents the custom-file format with the commented example `MyCustomProtease  K|,R|  full`. With that entry present, loading a legacy `semi-trypsin` settings file resolves to `MyCustomProtease` rather than `trypsin` — I ran this against the branch to confirm. The warning text then names the wrong protease, and the wrong name propagates into results and the mzIdentML `Enzyme` element. If the custom entry carries different protease-specific modifications, the search itself differs too.

**Suggested fix:** Restrict the scan to the embedded proteases, or treat more than one motif match as an error rather than resolving it alphabetically — a silent arbitrary pick is the worst of the three options.

**Affected test scenarios:** `LegacySemiTrypsinSettingsTests` covers a user-defined protease *named* `semi-trypsin`, but not one under a different name with the same motifs. A case that registers `MyCustomProtease` with motifs `K|,R|`, loads `GlycoSearchTaskconfig_ETD.toml` and asserts the protease is `trypsin` fails on the current head.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -47,6 +47,144 @@ public enum MyTask
...
+        /// <remarks>
+        /// Only the non-specific search engine, which a Search task runs for <see cref="SearchType.NonSpecific"/>, can use seeds
+        /// (it makes its own N and C passes). Classic and Modern search, Glyco, crosslink, GPTMD and calibration score digestion
+        /// products as they are, so given seeds they finish normally and report far fewer, wrong identifications.
+        /// </remarks>
+        /// <param name="taskName">The name the user knows the task by, included in the message when given.</param>
+        public string GetSeedDigestionRefusal(string taskName = null)
+        {
+            if (CommonParameters?.DigestionParams is not DigestionParams digestionParams // RNA digestion has no protein seed request
+                || this is SearchTask { SearchParameters.SearchType: SearchType.NonSpecific } // the non-specific search engine trims seeds
```
pcruzparri, line 82: **Non-blocking** — `SpectralAveragingTask` is refused over digestion it never uses

**Problem:** The only task-type exemption here is a non-specific `SearchTask`. `SpectralAveragingTask` is a `MetaMorpheusTask` whose `RunSpecific` reads nothing but its `SpectralAveragingParameters`, yet `CommonParameters` always carries a real `DigestionParams` (`CommonParameters.cs:94`). So an averaging task whose settings block happens to ask for seeds is refused over a setting it ignores.

**Impact:** `GetSeedDigestionRefusal` returns the full "Cannot proceed... seed peptides that only the non-specific search can use" message for an averaging task, and `TaskValidator.CheckDigestionSearchMode` then blocks the run. The message is meaningless for a task that only averages spectra. `EverythingRunnerEngine` already special-cases `SpectralAveragingTask` for an unrelated check, which is the precedent for exempting it here.

**Out of scope note:** reaching this needs a hand-edited or repurposed averaging toml, since the generated default is `Full`. It is a gap rather than something a user hits by accident, so it should not block the merge.

**Suggested fix:** Add `|| this is SpectralAveragingTask` to the guard, or invert it so the check applies only to tasks that actually digest.

**Affected test scenarios:** `DigestionSearchModeRefusalTests.EveryTaskAndDigestion` enumerates six task kinds and omits this one. Adding `SpectralAveragingTask` to that matrix with `expectedRefused: false` fails on the current head.

MetaMorpheus/TaskLayer/SearchTask/MzIdentMLWriter.cs
```diff
@@ -572,11 +572,17 @@ public static void WriteMzIdentMl(IEnumerable<SpectralMatch> psms, List<EngineLa
             int protease_index = 0;
             foreach (DigestionAgent protease in proteases)
             {
+                // The search's own specificity when the caller gives it: trypsin is Full even in a semi-specific search,
+                // whose semi-specificity comes from SearchModeType. Otherwise what the protease itself says.
+                bool isSemiSpecific = semiSpecific ?? protease.CleavageSpecificity == CleavageSpecificity.Semi;
```
pcruzparri, line 577: **Non-blocking** — caller override flattens per-enzyme `semiSpecific`

**Problem:** The line this replaces computed specificity per enzyme inside `foreach (DigestionAgent protease in proteases)`. Now a non-null `semiSpecific` from the caller applies to every entry in the list, so each enzyme's own `CleavageSpecificity` is discarded.

**Impact:** Passing two enzymes — fully specific trypsin and a semi-specific protease — with `semiSpecific: true` writes `semiSpecific="true"` on both, misreporting trypsin to downstream readers. I confirmed this against the branch. It is latent today, since `PostSearchAnalysisTask.WriteProteinResults` is the only production caller and always passes a single-element list, but this is a public method taking `List<DigestionAgent>` and the repo has multi-enzyme fixtures, so the first multi-enzyme caller inherits it.

**Suggested fix:** Apply the caller's value only when the list holds one entry, or take the value per enzyme.

**Affected test scenarios:** `MzIdentMlSemiSpecificTests` only ever passes one protease. A case with two enzymes of different specificity and an explicit `semiSpecific: true`, asserting the fully specific one keeps its own value, fails on the current head.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -47,6 +47,144 @@ public enum MyTask
...
+            Warn($"These settings name the protease \"{removed.Name}\", which is no longer available. They were read as \"{replacement.Name}\" " +
+                 $"with SearchModeType {digestionParams.SearchModeType} and FragmentationTerminus {digestionParams.FragmentationTerminus}, which digests the same peptides. " +
+                 "Save the task to update its settings.");
+            return digestionParams;
+        }
+
+        /// <summary>The shipped fully specific protease with exactly these cleavage motifs, or null if there is none.</summary>
+        private static Protease FindShippedFullySpecificProtease(string motifs)
+        {
+            static string Signature(IEnumerable<DigestionMotif> m) => string.Join(";", m
```
pcruzparri: **Non-blocking** — motif-signature formula duplicated in production and test

**Problem:** This four-field ordinal-sorted formula is reproduced verbatim as `MotifSignature` in `LegacySemiTrypsinSettingsTests.cs:49`, so the test checks the production code against its own copy of the same logic rather than against an independent expectation.

**Impact:** If `DigestionMotif` gains a field, or the format needs to change, only one copy is likely to be updated and the test keeps passing against a stale duplicate of the bug. Hand-rolling the comparison is fair enough — `DigestionMotif` overrides neither `Equals` nor `ToString`, and `DigestionAgent.Equals` compares `Name` only — it is writing it twice that is worth fixing.

**Suggested fix:** Move it to `EngineLayer/Util/MzlibExtensions.cs`, which already exists for this kind of mzLib-type comparison helper, and call it from both places.

**Affected test scenarios:** No behaviour change, so no new case — the existing `LegacySemiTrypsinSettingsTests` assertions keep working once both sites call the shared helper.

MetaMorpheus/GUI/TaskWindows/CalibrateTaskWindow.xaml.cs
```diff
@@ -78,7 +79,11 @@ private void UpdateFieldsFromTask(CalibrationTask task)
 
             if (task.CommonParameters.DigestionParams is DigestionParams digestionParams)
             {
-                ProteaseComboBox.SelectedItem = digestionParams.Protease; //protease needs to come first or recommended settings can overwrite the actual settings
+                ProteaseComboBox.SelectedItem = TaskWindowSearchMode.ProteaseToShow(digestionParams); //protease needs to come first or recommended settings can overwrite the actual settings
+                // a loaded task that asks for seed peptides cannot be shown here as it is, so say what happens to it
+                string searchModeWarning = TaskWindowSearchMode.PreserveWarning(digestionParams);
+                SearchModeWarningTextBlock.Text = searchModeWarning;
```
pcruzparri, line 85: **Non-blocking** — warning `TextBlock` hand-copied into eight files

**Problem:** This two-line `Text`/`Visibility` toggle appears verbatim in all four task windows (`CalibrateTaskWindow.xaml.cs:85`, `GPTMDTaskWindow.xaml.cs:98`, `GlycoSearchTaskWindow.xaml.cs:205`, `XLSearchTaskWindow.xaml.cs:160`), and the matching `TextBlock` with `Foreground="DarkRed" FontWeight="DemiBold" MaxWidth="450"` is repeated in all four `.xaml` files. `App.xaml` already holds the shared-style idiom this codebase uses for repeated visual treatment.

**Impact:** Eight files have to change in lockstep for anything later — a theme-aware colour for dark mode, a different wrap width, an icon — with no compiler error to catch a missed one. The copies have already drifted: only the calibration one carries `Grid.Row`/`Grid.Column`.

**Suggested fix:** A `WarningTextBlockStyle` in `App.xaml` plus a null-to-visibility converter, or a single shared `SetWarning(TextBlock, string)` helper.

**Affected test scenarios:** None — presentation only. To be clear, the warnings themselves are a good addition; this is only about how they are wired.

MetaMorpheus/Test/GuiTests/TaskWindowSearchModeTests.cs
```diff
@@ -0,0 +1,206 @@
...
+        private static IEnumerable<TestCaseData> SeedRequests()
+        {
+            yield return new TestCaseData(CleavageSpecificity.Semi, FragmentationTerminus.N);
+            yield return new TestCaseData(CleavageSpecificity.Semi, FragmentationTerminus.C);
+            yield return new TestCaseData(CleavageSpecificity.None, FragmentationTerminus.N);
+            yield return new TestCaseData(CleavageSpecificity.None, FragmentationTerminus.Both);
+        }
+
+        /// <summary>
+        /// Review of #2812 (nbollis): a hand-edited glyco task with Semi + N, or None, was rewritten to Semi + Both or
```
pcruzparri, line 132: **Non-blocking** — comments carry PR numbers, ticket numbers and reviewer names

**Problem:** This line and `:151` open with "Review of #2812 (nbollis): ...", and the same pattern is in `MzIdentMlSemiSpecificTests.cs:63`. Elsewhere the diff cites mzLib #1005, #1186 and #1303 plus a dated "(February 2026)" in `MetaMorpheusTask.cs`, a 20-line changelog-style `<remarks>` block in `LegacySemiTrypsinSettingsTests.cs`, and a bare `#2759` in `DigestionSearchModeRefusalTests.cs`.

**Impact:** Once this merges the branch is gone and these point a reader at a review conversation they have to go and resolve. `MetaMorpheusTask.cs:75` is the sharpest case — it describes #1186 as still in flight, so it goes stale the moment that lands and nothing here will prompt anyone to update it.

**Suggested fix:** Keep the explanations, drop the provenance. The peptides-versus-seeds rule and why the replacement is found by motif rather than by name are both worth documenting; the ticket numbers and reviewer names are the parts that rot.

**Affected test scenarios:** None — comments only.

RayMSMS (dismissed):

trishorts:
Rebased onto master `cb93955d1` to clear the merge conflict — head is now `b10032fb3`.

The conflict was in `MainWindow.RunAllTasks_Click`: #2780 inserted the "at least one spectra file checked" guard and `spectraFilePathsToUse` at the same anchor where this branch inserted the `TaskValidator.CheckDigestionSearchMode` loop. Resolved by keeping both, digestion check first.

That is the only change: `git diff 8ef2f07d5 b10032fb3` is exactly #2780's two files, and this branch's own diff against its base is unchanged. GUI builds; the five test classes this PR adds plus `CalibrationTests` (which #2780 extends) are 131/131 passing locally.

@RayMSMS sorry — the push dismissed your approval again. Nothing in this PR's own code moved.

Note: @pcruzparri's review from earlier today is not addressed here; replies to those are coming separately.

nbollis (dismissed):

RayMSMS (dismissed):

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
trishorts: Fixed in 6bd87594b. You're right that the candidate set is the merged dictionary and that the ordinal tie-break lets a custom name win -- every capitalised name sorts ahead of `trypsin`.

One refinement from digging into it: because the match is on the exact motif signature, any protease that wins digests *identically* by construction. What the signature did not compare is the protease's cleavage modification (`DigestionAgent.CleavageMod`), and that is the real difference -- a custom protease with trypsin's motifs plus a cleavage mod would win the tie-break and then silently modify every terminus it cut.

The signature now carries the cleavage modification on both sides, null on the wanted side because the removed proteases modified nothing when they cleaved. New test `LegacySemiTrypsin_IsNotReplacedByAProteaseThatModifiesEveryCleavage` builds exactly that protease under a name that sorts first; it fails without the change.

Preferring shipped proteases over custom ones outright would be the more complete fix, but it needs `GlobalVariables` to record which names came from the custom file -- `LoadDigestionAgents` currently discards `CustomDigestionAgentLoadResult`, and consuming it is the separate custom-files PR that the comment there already points at.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
trishorts: Fixed in 9b1d63cde, and I would rate it above non-blocking: the check runs over every task before the run starts, so an averaging task carrying a seed-asking digestion from an old settings file would refuse the whole run, including the search it was staged for.

`GetSeedDigestionRefusal` now exempts `MyTask.Average`, which mirrors the protein-database check two lines above it in `RunAllTasks_Click` ("will not throw this if averaging is the only task"). Averaging does not digest, so its `DigestionParams` are never read.

The task-kind table in `DigestionSearchModeRefusalTests` gains Averaging; without the exemption its Semi+N, Semi+C and None+Both cases fail, and Full+Both passes either way since it never asked for seeds.

MetaMorpheus/GUI/TaskWindows/CalibrateTaskWindow.xaml.cs
trishorts: Confirmed, and it is a real bug -- the window loads the value into `InitiatorMethionineBehaviorComboBox` at line 87 and then does not write it back.

It is not one this PR introduces. Against master the call was already `new DigestionParams(protease, maxMissedCleavages, minPeptideLength, maxPeptideLength, maxModificationIsoforms, maxModsForPeptides)`; this branch appends `searchModeType` and `fragmentationTerminus` and changes nothing else about it. The omission predates the branch.

This PR is scoped to making semi-trypsin work again for glyco searches, so I am leaving it. It belongs with the other two save-path findings in one PR covering all four windows.

MetaMorpheus/GUI/TaskWindows/XLSearchTaskWindow.xaml.cs
trishorts: Confirmed, including that there is no control for it in this window -- grep finds no `MaxMods` in `XLSearchTaskWindow.xaml` or its code-behind, so the value is read from the task, ignored, and written back as mzLib's default of 2.

Same situation as the calibration one: pre-existing on master, where the call already omitted `maxModsForPeptides`, and this branch only appends `searchModeType` and `fragmentationTerminus`. Out of scope here, in scope for the follow-up PR.

MetaMorpheus/GUI/TaskWindows/GlycoSearchTaskWindow.xaml.cs
trishorts: Confirmed for `generateUnlabeledProteinsForSilac`, and pre-existing on master in all four windows for the same reason as the other two save-path findings: the calls already omitted it and this branch only appends `searchModeType` and `fragmentationTerminus`.

On `keepNGlycopeptide` and `keepOGlycopeptide` specifically -- since those names look glyco-relevant and this PR is about glyco, I checked whether resetting them matters here. They are real `DigestionParams` constructor arguments, but grepping `GUI/`, `TaskLayer/GlycoSearchTask/` and `EngineLayer/GlycoSearch/` finds no reader for either property. Nothing consumes them today, so a glyco search is unaffected by the reset.

MetaMorpheus/TaskLayer/SearchTask/MzIdentMLWriter.cs
trishorts: Fair reading of the signature, but it cannot happen from the only call site: `PostSearchAnalysisTask` passes `[CommonParameters.DigestionParams.DigestionAgent]`, a single-element list, and that is the only caller of `WriteMzIdentMl` in the repo. With one enzyme there is nothing to flatten.

The parameter exists because the search's specificity is not the protease's: since mzLib #1005 a semi-specific search uses fully specific trypsin with `SearchModeType = Semi`, so `protease.CleavageSpecificity` reports Full and the old line wrote `false` for every semi-specific search.

Happy to add a comment noting that the per-enzyme fallback would need to come back if MetaMorpheus ever writes more than one enzyme -- say the word and I will.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
trishorts: Noted, and a genuine weakness in principle.

In practice the test's copy serves a different assertion than the production formula: it asserts that whatever protease was chosen cuts exactly where semi-trypsin cut. That held, and stayed meaningful, when I changed the production signature in this round to include the cleavage modification -- which is precisely the case that would have exposed the coupling if it were load-bearing.

Leaving it as is. Folding it in would mean exposing the production signature for tests, which is more API surface than the check is worth.

MetaMorpheus/GUI/TaskWindows/CalibrateTaskWindow.xaml.cs
trishorts: Agreed on the duplication, with one clarification: the shared *logic* is already factored out -- `TaskWindowSearchMode.PreserveWarning` is the single rule all four windows call. What repeats is the two-line `Text`/`Visibility` toggle and the matching `TextBlock` markup.

A shared control would remove it, but that is a GUI refactor touching four windows and their XAML, and this PR is scoped to making semi-trypsin work again for glyco. Not doing it here.

MetaMorpheus/Test/GuiTests/TaskWindowSearchModeTests.cs
trishorts: Taking this one as declined for now, and I want to be straight about why rather than silently skipping it.

The reviewer names genuinely should not be in the source, and I do not disagree with the finding. But the change does nothing for the behaviour this PR exists to restore, and the PR has already drifted wider than its title; the call on this branch is to stop adding edits that are not about semi-trypsin working in glyco. A comment-only push would also restart all eight CI checks for cosmetics.

On the PR numbers, I would push back even in a follow-up: the mzLib references (#1005, #1186, #1303) are the only record of why `semi-trypsin` disappeared, and why the replacement is found by cleavage motif rather than by name. Neither is recoverable from the code. The dated "(February 2026)" I would drop.

If you would rather have the reviewer names out before this merges, say so and I will strip them -- it is two comments.

trishorts:
@pcruzparri's round is answered — replies in all nine threads. Two of the nine are fixed, pushed as `9b1d63cde` and `6bd87594b` on top of today's rebase:

- **`FindShippedFullySpecificProtease` could resolve to a user's custom protease.** The signature now compares the cleavage modification as well as the motifs. The motif match already guaranteed identical cleavage, so the modification was the only way a custom protease could actually change the digestion — and it would have modified every terminus it cut. New test builds that protease under a name that wins the ordinal tie-break; it fails without the change.
- **A spectral averaging task was refused over digestion it never uses.** `MyTask.Average` is now exempt, mirroring the protein-database check in `RunAllTasks_Click`. Worth more than non-blocking: the check runs over every task before the run starts, so one averaging task could refuse the whole run.

The other seven are answered in-thread rather than changed. Briefly: the three save-path Highs (`InitiatorMethionineBehavior`, `MaxModsForPeptide`, the SILAC and keepN/keepO flags) are all pre-existing on master — those constructor calls were already incomplete and this branch only appends `searchModeType` and `fragmentationTerminus` to them. They deserve one PR covering all four windows' save paths. The mzIdentML per-enzyme point cannot be reached from the only call site, which passes a single-element list. The rest are cleanup that would widen a PR that is already wider than its title.

Full suite on `6bd87594b`: 2240 passed, 1 skipped, 0 failed.

@RayMSMS @nbollis this needs approvals again — today's rebase onto master (#2780 had touched the same lines in `RunAllTasks_Click`) dismissed Ray's.

RayMSMS (approved):

pcruzparri (approved):

PR #2813, opened 2026-09-15
Keep the glycopeptide and SILAC digestion flags when a file has its own settings

trishorts:
> **Project of origin:** `NOSearch`

## What changes

**Digestion flags now survive file-specific settings.** When a spectra file has its own settings, even ones that only change a mass tolerance, `MetaMorpheusTask.SetAllFileSpecificCommonParams` rebuilds the task's `DigestionParams` from an explicit list of fields. Three fields were missing from that list, so they silently fell back to their constructor defaults for those files:

| Flag | Reset to | Effect for files with file-specific settings |
|---|---|---|
| `KeepNGlycopeptide` | `false` | a digest restricted to N-glycosylation-motif peptides became unrestricted |
| `KeepOGlycopeptide` | `false` | a digest restricted to Ser/Thr-containing peptides became unrestricted |
| `GeneratehUnlabeledProteinsForSilac` | `true` | a SILAC search set not to quantify unlabeled peptides quantified them anyway (`PostSearchAnalysisTask` reads this flag) |

The rebuild now carries all three from the task.

## What does not change

- File-specific overrides (protease, missed cleavages, peptide lengths, max mods) still apply exactly as before.
- Files without file-specific settings are not rebuilt and are unaffected.

## How it is tested (test-first)

New `Test/FileSpecificDigestionParamsTests.cs`. **5 of its 6 cases fail on master.** The sixth is the all-defaults combination, which cannot show the bug.

- **Each flag survives:** in every combination, including the non-default ones, a file-specific missed-cleavage override leaves the flag at the task's value.
- **Whole-object check:** a file-specific change unrelated to digestion (a precursor tolerance) leaves the digestion parameters **equal** to the task's. The task sets every digestion field away from its default, and `DigestionParams.Equals` compares every field, so a digestion setting added later and forgotten in this rebuild fails the test as well.

**Full MetaMorpheus test suite: 2,014 passed, 0 failed.**

## Related

- **Conflict:** smith-chem-wisc/MetaMorpheus#2812 (draft) adds `specificTermini` to the same constructor call; whichever merges second needs a one-line rebase.
- **Nearby:** #2650 edits `SetAllFileSpecificCommonParams` further down (deconvolution), not these lines.

https://claude.ai/code/session_01V1qU8gPh15GvhNYqTBdBxs


nbollis (approved):

RayMSMS (approved):

zhuoxinshi (approved):

pcruzparri (commented):
Automated code review — 0 finding(s) at severity >= High; 1 non-blocking, posted as inline comments.

PR #2814, opened 2026-09-15
Search glyco spectra files in parallel, and move threads from finished files to files still searching

trishorts:
> **Project of origin:** `NOSearch` (glyco search performance)

> **Rebased onto master `ee506c655`** now that #2742 and #2809 are merged: three commits, `26c98bd12`, `7657b0426` and `aa515528f`.
>
> Changes made to fit master:
> - `GlycanSearchSpace.Build` now applies #2809's glycan mass cap to N-glycan searches as well (`970809597`), as master's constructor does.
> - Both `GlycoSearchEngine` constructors take master's `IEnumerable<IBioPolymerWithSetMods>` peptide index and #2742's `candidates` argument.
> - When files are searched one after another, the task keeps #2742's two-round search for a partitioned index. Its two round methods keep their own thread loop: they run one file at a time, as on master. Files are searched in parallel only when the index is not partitioned, so the shared thread budget is only used by the single-pass `Run`.
> - `GlycoSearchEngineSearchesThroughTheBudgetItIsGiven` passes an empty `nestedIds` list rather than null. The command-line program's engine-started handler stays subscribed after a CMD test, and when the verbosity is normal it joins the ids.

## Why

A glyco N+O search on several spectra files searched them one after another, each on the whole thread budget. Most of a glyco search is per-scan CPU work that scales well to about 16 threads and poorly beyond, so several files at once on a divided budget uses the machine better. Dividing the budget once, though, leaves it mostly idle at the end of a run: files finish at different times, and the last one keeps searching on its own fraction. On six files the largest searched alone on 5 of 32 threads for over a minute.

## What changes

**Glycan boxes once per task** (`26c98bd12`). The glycans and glycan boxes depend only on the glycan databases and settings, so `GlycanSearchSpace.Build` makes them once and every file's engine shares them.

**Files in parallel** (`7657b0426`). When every file shares one unpartitioned index (same index-relevant settings, no partitioning), `GlycoSearchTask` searches files at once. `FileParallelism.Decide` (pure, unit-tested) sets how many: bounded by the number of files, a minimum of 4 threads per file, 80% of free memory measured after loading the first file, and a new `MaximumSpectraFilesInParallel` setting (0 = no cap, 1 = one after another). Results are gathered in file order, exactly as the sequential search adds them. The decision and its reason go into the manuscript prose.

**A shared thread budget** (`aa515528f`). `SearchThreadBudget` (EngineLayer/Util) is one budget shared by every file a task searches:
- a file is entitled to an equal share among the files still handing out work;
- threads no file below its share wants go to whoever asks, so none sit idle;
- a file over its share gives a thread back, after its current scan, only when every thread is in use and another file is below its share.

So a finished file's threads go to the files still searching, and a file that starts later (files finish loading at different times) gets its share from files already searching. `GlycoSearchEngine` hands scans out through `SearchThreadBudget.Run`: workers take the next unsearched scan from a shared counter instead of the fixed stride. Only the loop header changed; per-worker scratch and the per-scan body are untouched. Without a budget the engine uses one of its own with `MaxThreadsToUsePerFile` threads, so every other caller behaves as before. Progress is now counted exactly with `Interlocked`, as `IndexingEngine` does, instead of a shared `double` incremented from every thread.

## Results do not change

Each scan index is handed out exactly once and writes only its own `GlobalGsms[scanIndex]` slot, so which thread searched a scan does not matter.

- `SearchingSpectraFilesInParallelMatchesSearchingThemInTurn` asserts `AllPSMs.psmtsv`, `no_glyco.psmtsv`, `_AllProteinGroups.tsv` and both localization tables are byte-identical between files in parallel and one after another.
- On the real data below, every result file is byte-identical to the build before each commit. Only `AutoGeneratedManuscriptProse.txt` differs (run time, free memory, and the reworded thread line).

After the rebase, an N+O search of two of the files (180422_10 and 180422_11, 32 threads) wrote byte-identical result files on master `ee506c655` and on this branch, which searched them in parallel (2 files, 16 threads each to load, one shared budget of 32). The task settings file differs only by the new `MaximumSpectraFilesInParallel = 0` line, and the prose only by the two lines this PR adds. Wall time 1:02 on master, 0:55 here.

Master now includes #2742, so the index build no longer reorders equal-mass peptides between runs.

## Timing

N+O glyco search: 6 Thermo glycoproteomics files (1.2 to 2.1 GB each), 130 human proteins plus decoys, trypsin with semi-specific digestion, 43 O-glycans and 42 N-glycans, up to 3 O-glycans, 32-thread budget on a 64-thread machine, E: on a spinning disk.

| Build | 6 files | Result files |
|---|---|---|
| #2742 + #2809 + files in parallel | 5:38 | baseline |
| + #2809's lazy child boxes | 4:30 | byte-identical |
| + shared thread budget (this PR) | **3:12** | byte-identical |

When each file's glyco search finished:

| Finished | Fixed split | Shared budget |
|---|---|---|
| 1st | 1:08 | 0:45 |
| 3rd | 2:06 | 1:32 |
| last | 3:26 | 2:03 |

Output: 26,256 PSMs and 1,221 glyco PSMs at 1% FDR in every run.

The semi-specific digestion was run with a local mzLib build containing smith-chem-wisc/mzLib#1303, which makes `trypsin` + `SearchModeType = Semi` produce the semi-tryptic peptides instead of seeds. Nothing in this PR depends on it.

## Prose

Before:
`spectra files searched in parallel = 6 of 6, with 5 threads each from a budget of 32 (limited by the number of spectra files);`

After:
`spectra files searched in parallel = 6 of 6 (limited by the number of spectra files), each loaded with 5 threads and searched with threads shared from a budget of 32 that move from files that finish to files still searching;`

## Tests

- `GlycoFileParallelismTests`: the split decision, the thread-count clone, and parallel-vs-sequential byte identity.
- `SearchThreadBudgetTests` (written first, failing against a skeleton):
  - the share rules, one call at a time so they do not depend on timing: one file takes the whole budget, a later file gets its fair share, a finished file hands its threads on, leftover threads from rounding are not idle, more files than threads;
  - `Run` hands out every item exactly once at 1, 3 and 8 threads, never runs more workers than the budget, a run still working gets a finished run's threads, a run that starts later gets its share while the other is still working (a mutation that stops workers giving threads back fails this one), and a worker exception ends the run with every thread returned;
  - `GlycoSearchEngine` searches through the budget it is given.
- Glyco, index partitioning and index engine tests: 214 passed. After the rebase (mzLib 1.0.591): the full suite, 2,156 passed, 1 skipped.

## Known limits

- Files still loading use their own threads from the initial split, outside the shared budget, so the machine can briefly run more threads than the budget while files load.
- `IndexingEngine`'s protein digestion and the other search engines still use the stride loop; they could use `SearchThreadBudget.Run` later, but that is left out to keep this PR to glyco.


MetaMorpheus/TaskLayer/GlycoSearchTask/GlycoSearchTask.cs
```diff
@@ -93,128 +97,249 @@ protected override MyTaskResults RunSpecific(string OutputFolder, List<DbForTask
...
+                ProseCreatedWhileRunning.Append("memory free after building the index and loading the first spectra file = " + (freeBytes / 1e9).ToString("0.0") + " GB, estimated "
+                    + (bytesPerFile / 1e9).ToString("0.0") + " GB for each further file; \n");
+
+                // Each file's own copies of its parameters with the divided thread count, built here on one thread (see above).
+                var loadParamsPerFile = combinedParamsPerFile.Select(p => p.CloneWithNewMaxThreadsToUsePerFile(plan.ThreadsPerFile)).ToArray();
+                var searchParamsPerFile = indexParamsPerFile.Select(p => p.CloneWithNewMaxThreadsToUsePerFile(plan.ThreadsPerFile)).ToArray();
+
+                var gsmsPerFile = new List<GlycoSpectralMatch>[fileCount][];
+                Status("Searching files...", taskId);
+                Parallel.For(0, fileCount, new ParallelOptions { MaxDegreeOfParallelism = plan.FilesInParallel }, spectraFileIndex =>
```
pcruzparri, line 202: **Non-blocking** — `Parallel.For` can queue files on one worker while another sits idle

**Problem:** `Parallel.For` splits the files into one block per worker, and each worker claims files from its block in chunks that double in size (1, then 2, then 4…). A claimed file can't be taken by another worker. Once a block holds 3 or more files, the second claim reserves two files at once. Leftover files form extra blocks no worker starts on, and a worker that finishes early takes two of those at once too. With 6 files and `MaximumSpectraFilesInParallel = 2`, worker 1 claims file 0, then {1, 2}; worker 2 claims file 3, then {4, 5}. If file 4 is slow, worker 1 finishes file 2, finds file 5 already claimed, and exits, so file 5 waits behind file 4 while a file slot sits empty.

**Impact:** Results don't change, but it brings back the idle tail this PR is meant to remove. The freed threads move to the file still searching, but one file scales poorly past ~16 threads, and the queued file can't start loading until the one ahead of it finishes. It can only happen when fileCount ≥ 2 × FilesInParallel + 2 (e.g. 10+ files on a 16-thread budget, 18+ on 32, or a memory-capped run), so the 6-files-at-6 benchmark wouldn't show it.

**Suggested fix:** Hand files out one at a time, only when a worker is free (needs `using System.Collections.Concurrent;`):
```csharp
Parallel.ForEach(Partitioner.Create(Enumerable.Range(0, fileCount), EnumerablePartitionerOptions.NoBuffering),
    new ParallelOptions { MaxDegreeOfParallelism = plan.FilesInParallel }, spectraFileIndex => { ... });
```
I checked this builds at `aa515528` and the parity test still passes.

**Affected test scenarios:** Not covered. `SearchingSpectraFilesInParallelMatchesSearchingThemInTurn` runs 2 files at 2 in parallel. A test with ~6 small files and `MaximumSpectraFilesInParallel = 2` that records when each file starts and finishes would show it. Byte-identity passes either way.

pcruzparri (commented):
Automated code review — 0 findings at severity >= High; 1 non-blocking finding, posted as an inline comment.

MetaMorpheus/EngineLayer/CommonParameters.cs
```diff
@@ -355,6 +355,63 @@ public CommonParameters CloneWithNewTotalPartitions(int totalPartitions)
             return clone;
         }
 
+        /// <summary>
+        /// Copy with a different MaxThreadsToUsePerFile, for dividing a task's thread budget across spectra files searched at
+        /// once. Returns a new instance for the same reason as <see cref="CloneWithNewTotalPartitions"/>: the task's own
+        /// CommonParameters may be what a file was handed. Construct copies on one thread: the constructor reads and clears the
+        /// process-wide custom product-type list.
+        /// </summary>
+        public CommonParameters CloneWithNewMaxThreadsToUsePerFile(int maxThreadsToUsePerFile)
```
nbollis, line 364: A new clone method would be pretty fragile for updates to common params. Could we instead just make the other clone method take in an additional optional parameter? Or use a normal clone method then set the property outside?

Alexander-Sol:

**Summary.** Parallelizes glyco search across spectra files with a work-stealing thread budget (`SearchThreadBudget`) shared by all concurrently-searching files, plus supporting refactors (glycan boxes built once via `GlycanSearchSpace`, a pure `FileParallelism.Decide`). The concurrency design is careful and preserves the per-scan ownership model that guarantees determinism: each scan index is handed out exactly once via an `Interlocked` counter and writes only its own `GlobalGsms[scanIndex]` slot, so results are independent of thread interleaving. Test coverage is unusually strong for a concurrency change (mutation-kill test + a direct parallel-vs-sequential byte-identity test).

### Findings
- **Low (only actionable item) — thread reservation leaks if `Task.Factory.StartNew` throws.** `SearchThreadBudget.cs:114-135`: `TryTakeThread()` increments `HeldThreads`/`_threadsInUse` before `StartNew`; if `StartNew` throws (catastrophic thread/OOM exhaustion, exacerbated by `LongRunning` = dedicated OS thread), the exception unwinds to the outer `finally`, and `Leave()` (`:293`) removes the share without decrementing `_threadsInUse` for the un-launched worker — permanently subtracting it from the budget. Repeated failures could shrink the budget toward 0 → stall. Cheap fix: `share.ReturnThread()` on the throw path before rethrowing.
- **Low (cosmetic):** `results.txt` per-engine summary block ordering is now finish-order-dependent (run-to-run variation). The `lock (resultTexts)` at `MyTaskResults.cs:70` correctly prevents list corruption; no *data* output is affected (byte-identity test covers `AllPSMs.psmtsv`, `no_glyco.psmtsv`, `_AllProteinGroups.tsv`, both localization tables — none derive from `resultTexts`).
- **Low:** brief CPU oversubscription while files load (loading uses each file's own `ThreadsPerFile` outside the shared budget) — already disclosed in "Known limits."
- **Low (perf):** `NextItem()` takes `_lock` once per scan per worker (`:241-250`); measurable lock traffic for very short scans, but end-to-end numbers (3:12 vs 5:38) net positive.
- **Nit (pre-existing, not introduced):** sequential-path progress at `GlycoSearchTask.cs:341` uses integer division with no `*100` (reports 1%), unlike the parallel path's `100 * done / fileCount`. Confirmed identical on `origin/master`.

### Verified correct
Exactly-once hand-out / determinism (`Interlocked.Increment`, single-owner-per-slot `MergeIntoGlobalGsms:399`), no thread over-budget (`_threadsInUse < TotalThreads` under `_lock`, asserted by `RunNeverUsesMoreThreadsThanTheBudget`), no missed-wakeup hang (`Monitor.Wait(_lock, 50ms)` + `PulseAll`), cross-file exception isolation (`AwWorkerExceptionEndsTheRun` asserts `ThreadsInUse == 0`), `MyFileManager` `ConcurrentDictionary`, `LazyInitializer` glycan-box caches, two-round/partitioned path unchanged, file-ordered result gathering.

MetaMorpheus/EngineLayer/CommonParameters.cs
trishorts: Agreed, and taken — that was a third copy of the constructor's argument list, which is exactly the "one more place to forget" you're describing.

Your second suggestion, in the form that doesn't widen the API: `CloneWithNewMaxThreadsToUsePerFile` now calls the method that already copies everything and sets the one setting that differs on the copy. The setter stays `private` — that is no obstacle from inside the class.

```csharp
public CommonParameters CloneWithNewMaxThreadsToUsePerFile(int maxThreadsToUsePerFile)
{
    CommonParameters clone = CloneWithNewTotalPartitions(TotalPartitions);
    clone.MaxThreadsToUsePerFile = maxThreadsToUsePerFile;
    return clone;
}
```

I didn't take the optional-parameter version because the call site would then read `CloneWithNewTotalPartitions(p.TotalPartitions, threads)`, which says the opposite of what it does.

`CloneWithNewMaxThreadsToUsePerFilePreservesEveryOtherSetting` still passes unchanged: it walks every public property, requires the clone to carry it (by reference, for reference types), and requires the original not to be mutated. So a setting added to `CommonParameters` now only has to be added to `CloneWithNewTotalPartitions`. (`CloneWithNewTerminus` still keeps its own copy of the list — pre-existing, and I've left it alone.)

`2df1f339f`

MetaMorpheus/TaskLayer/GlycoSearchTask/GlycoSearchTask.cs
trishorts: Taken, with a test — you're right that this is the idle tail the PR exists to remove, just moved from the thread budget to the file queue.

The loop moved into `FileParallelism.ForEachFile(fileCount, filesInParallel, searchFile)` (next to `Decide`, which already owns how many files run at once), and it uses your partitioner:

```csharp
Parallel.ForEach(
    Partitioner.Create(Enumerable.Range(0, fileCount), EnumerablePartitionerOptions.NoBuffering),
    new ParallelOptions { MaxDegreeOfParallelism = Math.Max(1, filesInParallel) },
    searchFile);
```

New test `ForEachFileLeavesNoFileWaitingWhileAFileSlotIsFree`: six files at two in parallel, file 1 blocks until the test releases it, and every other file must start while it is blocked. It was **red before the change** — it timed out with file 2 never started, which is the claimed-in-a-chunk case you described — and passes in about a tenth of a second after it. It asserts the property ("no file waits while a slot is free") rather than the chunk arithmetic, so it won't go stale if the TPL changes how it chunks.

Thanks for checking that it builds and that the parity test still passes; confirmed here too. The `GlycoFileParallelismTests`, `SearchThreadBudgetTests` and clone fixtures are green locally, including `SearchingSpectraFilesInParallelMatchesSearchingThemInTurn`; the full suite is left to CI on this push. Results are still gathered in file order, so nothing about the output changes.

`275cce491`

trishorts:
Thanks for the review — the reservation leak is real and is fixed, along with the progress nit.

**Thread reservation leaked when a worker can't be started (`SearchThreadBudget.cs`).** Exactly as you describe: `TryTakeThread()` takes the thread, and if `StartNew` throws, the worker whose `finally` would return it never runs, so `Leave()` drops the share with the thread still counted. `Run` now returns it before rethrowing. Fixed in `c7dc435b3`.

Testing it needed a seam, since a failed thread start isn't reachable on a healthy machine: the launch is now an `internal Func<Action, Task> StartWorkerThread` field defaulting to the same `Task.Factory.StartNew(..., LongRunning, TaskScheduler.Default)`, and the test replaces it. `AThreadIsReturnedWhenItsWorkerCannotBeStarted` was red first with `ThreadsInUse == 1`, and also asserts the budget still hands out all of its threads to the next file afterwards — the part that actually matters, since a run whose budget has shrunk to nothing waits forever.

**Progress nit (`GlycoSearchTask.cs`).** Pre-existing, as you say, but it is one line in the method this PR restructures and the parallel path made the two inconsistent, so it is fixed here: `74c9d036b`.

The other three I've left as they are:

- **`results.txt` block order.** Agreed it's cosmetic — `resultTexts` feeds only the "Engine Results" prose block, and every data file (`AllPSMs.psmtsv`, `no_glyco.psmtsv`, `_AllProteinGroups.tsv`, both localization tables) is gathered in file order and covered by the parity test. Ordering the prose would mean giving engine result strings a file key they don't currently carry, which is more than this is worth.
- **Oversubscription while files load.** Known and disclosed: loading uses each file's own `ThreadsPerFile` outside the shared budget, and it lasts only as long as the load.
- **`NextItem()` lock traffic.** Measured net positive end to end (5:38 → 3:12 on the six-file semi-tryptic N+O benchmark), and a glyco scan is long enough that the lock isn't the bottleneck. A lock-free fast path is a reasonable follow-up if profiling ever asks for it.

Also rebased onto master for #2815 before making these changes.

Branch head is now `74c9d036b`.

trishorts:
﻿Local suite number, to finish what I said above: **2,336 passed, 1 skipped, 0 failed** (7m36s).

That run covers `c7dc435b3`, `2df1f339f` and `275cce491`. It was already running when I made the one-line progress change in `74c9d036b`, so that commit is covered by the build and by CI here, not by that number.

PR #2815, opened 2026-09-15
Find the scoring window's end once per scan, and skip a mass read OpenSearchMode never uses

trishorts:
> **Project of origin:** `NOSearch` (glyco search performance)

> **Rebased onto master `ee506c655`** now that #2742 is merged: two commits, `54f3fb1bb` and `805cecbf0`.
>
> Master's modern search now holds the peptide index as `List<IBioPolymerWithSetMods>` (#2759), so the helpers this PR adds take that type too. They only read `MonoisotopicMass`.

## Why

The glyco and crosslink engines score through `ModernSearchEngine.IndexedScoring`. For every fragment bin a scan hits, it counts every peptide no heavier than the precursor plus 1 Da. Profiling a semi-tryptic N+O glyco search, `IndexedScoring` was **56% of all CPU**, and most of that was reading `PeptideWithSetModifications.MonoisotopicMass` off peptide objects scattered across the heap:

1. **The end of the window was found per bin by mass.** Each bin ran a binary search that read a peptide's mass at every probe, repeated for every bin of every scan.
2. **Every candidate reaching the score cutoff had its mass read** to ask the mass-difference acceptor. Glyco and crosslink use `OpenSearchMode`, whose `Accepts` returns 0 for any mass, so the read decided nothing.

## What changes

**One peptide id per scan** (`54f3fb1bb`). The indexing engine sorts the peptide index by mass (undefined masses first), and peptide ids are positions in it. So "mass at or below the bound" is the same as "id at or below the last id at or below the bound".
- `LastPeptideIdAtOrBelow` finds that id once per scan.
- `LastBinPositionAtOrBelowId` finds each bin's end with an integer search over the bin's own ascending ids. That is the same position the mass search found, with no peptide read.
- The forward walk after the mass search never moved the end (the search already returns the last entry at or below the bound), so the new path does not need it.
- An index that is not in mass order (`IsSortedForBinSearch`, checked once per index) keeps the per-bin mass search.
- The lower bound and the scoring loops are unchanged.

**No mass read for `OpenSearchMode`, and one order check** (`805cecbf0`).
- When the acceptor is exactly `OpenSearchMode`, the candidate's mass is not read. It is the exact type only, because a subclass may override `Accepts`.
- The index-order check ran on every search thread at once, each reading every peptide. It now runs once, under a lock.

## Results do not change

On the search below, every result file is byte-identical before and after each commit (only the run time in the prose differs): 26,256 PSMs and 1,221 glyco PSMs at 1% FDR.

After the rebase, the same N+O search on two of the files (180422_10 and 180422_11, 32 threads) wrote byte-identical results on master `ee506c655` and on this branch; only run times and paths differ.

## Timing

N+O glyco search: 6 Thermo files, 130 human proteins plus decoys (1.66 M peptides in the index), trypsin with semi-specific digestion, 43 O-glycans and 42 N-glycans, up to 3 O-glycans, 32 threads.

| Build | 6 files | Longest file's glyco search |
|---|---|---|
| before | 3:12 | 2:03 |
| + one peptide id per scan | 2:41 | 1:43 |
| + no mass read for OpenSearchMode, one order check | **1:21** | **0:23** |

These were measured with #2809 and #2814 also applied (files searched in parallel on a shared thread budget), which is what the "before" row is. The change is in the per-scan scoring every glyco and crosslink search does, so it applies without them; I have not timed it on #2742 alone. The semi-specific digestion used a local mzLib build with smith-chem-wisc/mzLib#1303; nothing here depends on it.

## Tests

`IndexedScoringPeptideIdCutoffTests`, written first (the helpers failed against throwing skeletons; the single-check test failed with 16 checks instead of 1):
- `LastPeptideIdAtOrBelow` and `LastBinPositionAtOrBelowId` against linear scans: exact ties, values just either side, undefined masses, beyond both ends, empty.
- `IsSortedForBinSearch` accepts `IndexingEngine.SortByMonoisotopicMass` output and rejects an index with two entries swapped.
- `IndexedScoring` against a **verbatim copy of the previous method** over 800 windows on 300 random bins (including empty bins and bins holding every id): the same ids added in the same order and the same scoring table, for a sorted index, an unsorted index, LowCID, a tolerance acceptor, and an `OpenSearchMode` subclass that rejects everything.
- 16 threads starting together check the index order once.

Mutations that fail the tests: reading a tie at the bound as outside it (fails 3), and `is OpenSearchMode` in place of the exact type (fails the subclass test).

On this branch after the rebase (mzLib 1.0.591): the full suite passed, 2,143 tests, 1 skipped. Before the rebase, on #2742 with mzLib 1.0.589: glyco, crosslink, index-partitioning, index-engine and these tests, 228 passed.


RayMSMS (approved):

nbollis (approved):

pcruzparri (approved):

Alexander-Sol:
## Agent review — Approve

**Summary.** A performance refactor of `ModernSearchEngine.IndexedScoring` (the hot loop for glyco/crosslink first-pass scoring). It replaces the per-bin, mass-reading binary search for the window's upper end with a single per-scan peptide-id bound plus a cheap integer search per bin, and skips the mass-difference-acceptor call when the acceptor is exactly `OpenSearchMode`. The transformation is result-preserving; the PR carries an unusually strong test pinning the new path against a verbatim copy of the old one. No correctness bugs found.

### Correctness verification (the core question)
The id-based window is exactly equivalent to the old mass-based window under two invariants that hold in production:
- **Index is mass-sorted** — built via `IndexingEngine.SortByMonoisotopicMass`; `IsSortedForBinSearch` (NaN→−∞, non-decreasing) always holds, so the fast path engages. If it ever didn't, `PeptideIndexIsSortedForBinSearch` returns false and the code falls back to the byte-identical old mass search.
- **Bins hold ascending peptide ids** (`IndexingEngine.cs:161-166`). Given both, `{ids with MassForBinSearch ≤ bound}` is exactly the id prefix `0..lastPeptideIdInWindow`, whose last position both searches return. The removed forward-walk was provably a no-op in the sorted path; it is *retained* in the unsorted `else if` branch (`MSE_head.cs:630-642`), so that path is byte-for-byte unchanged.
- **OpenSearchMode skip** (`:588/653/671`): `OpenMassDiffAcceptor.Accepts` returns 0 unconditionally, so `>= 0` is always true — skipping is exact. Guard is `GetType() == typeof(OpenSearchMode)` (exact type), so an overriding subclass still gets called.
- **Concurrency:** `_binSearchOrder` is `volatile`, published after full construction, updated under lock with a double-check; `_binSearchOrderChecks` via `Interlocked`. All current callers pass a constant `base.PeptideIndex`, so the single-slot cache always hits after the first scan.

### Findings
- **Low:** the single-slot order cache (`MSE_head.cs:554-571`) is latent-fragile — if an engine were ever reused across two interleaved `peptideIndex` objects, each switch would re-run the O(n) `IsSortedForBinSearch` under the global lock. No current caller does this; worth a comment noting the invariant.
- **Nit:** `LastPeptideIdAtOrBelow` re-implements the upper-bound search shape of `BinarySearchBinForPrecursorIndex`; acceptable since they operate on different inputs and factoring would obscure the hot path.

### Test coverage
Strong and adversarial: helpers pinned against linear scans (ties, ±1e-6, undefined masses, out-of-range, empty); `IndexedScoring` pinned against a verbatim reference over 800 windows × 300 bins across sorted+HCD, sorted+LowCID, unsorted+HCD, a tolerance acceptor, and an `OpenSearchMode` subclass; plus a 16-thread barrier test asserting the order check runs exactly once. Reports byte-identical result files on a real 6-file N+O search.

PR #2816, opened 2026-09-15
Optionally write an SDRF-Proteomics file with the search results

trishorts:
> **Project of origin:** `sdrf`

A search knows everything an SDRF says about the **assay**: enzyme, modifications, tolerances,
dissociation type, instrument, software version. This adds an opt-in `WriteSdrf` search parameter
(a checkbox in the search task window, off by default). When it is on, `experiment.sdrf.tsv` is
written into the search's own output folder alongside the results, so searches can be pooled and
mined across experiments.

**Counterpart mzLib PRs, both merged and released:**
smith-chem-wisc/mzLib#1174 (`SdrfBuilder`, `SdrfCoverage`) and smith-chem-wisc/mzLib#1207
(`SdrfValidator` made public). This branch builds against the **mzLib 1.0.589** already pinned on
master, so it touches no `.csproj` and no package version.

## What it does

- **All SDRF logic stays in mzLib.** `PostSearchAnalysisTaskSdrf.cs` is the adapter. It hands
  `SdrfBuilder` the facts MetaMorpheus already holds and writes what comes back.
  `PostSearchAnalysisTask` becomes `partial` so the adapter can sit in its own file instead of
  growing a 1,700-line class.
- **One row per spectra file, from per-file parameters.** Per-file overrides of protease,
  tolerances and dissociation type are real, and task-level values would flatten them.
  `comment[data file]` names the original acquisition, not a `-calib` intermediate.
- **Sample metadata comes from `ExperimentalDesign.tsv`, and missing metadata never blocks the run.**
  If the file is absent or unusable, `SearchTask` warns *before* the search starts, naming what the
  SDRF will not be able to fill. After writing, the coverage report names the columns that came out
  uninformative. Those cells say `not available`, which the specification permits. The degenerate
  design `PostSearchAnalysisTask` makes up when the file is missing is **never** written as though
  it were curated.
- **It does not assert what it does not know.** Acquisition method is read from the search (DIA →
  `PRIDE:0000450`) rather than hardcoded to DDA. A labelled search (TMT, iTRAQ, SILAC, turnover) is
  never described as `label free sample`.
- **The writer cannot take a finished search down.** Unlike its sibling writers, it is wrapped in
  `EngineCrashed`: an SDRF is metadata about results, not results.

## Not in scope: labelled experiments

SDRF wants one row per sample **per channel**, but this writer emits one row per file. For a
labelled search `comment[label]` is therefore left unresolved, and both a pre-run warning and the
coverage report say so. SILAC has no channel-to-sample map in MetaMorpheus. TMT now does
(`TmtDesign.txt`, #2774), so expanding TMT rows from it is the natural follow-up. It is a separate
change because it alters the row model, not just a column.

## Testing

`SdrfOutputTests`, 14 tests. Most drive real searches on a copied spectra file in an isolated folder:
- Off by default, and off writes nothing.
- Opting in without a design file, or with a malformed one, still runs the search and still writes
  the SDRF.
- The written file reads back through mzLib, with this search's own protease and software version.
- It names the original data file.
- It carries every column SDRF requires, and **mzLib's own `SdrfValidator` reports no errors on it**.
- Acquisition method and label resolve as described above.
- A search whose SDRF cannot be written still writes its results and leaves `SdrfWriter_crash.txt`.
- An unreadable data file gives no instrument rather than a failure.
- A labelled search is warned before the run.
- The organism is the first target protein's NCBI taxonomy ID.

**One limitation found while writing that last test.** A FASTA search never fills
`characteristics[organism]`. MetaMorpheus passes `LoadProteinFasta` explicit UniProt regexes, which
skips mzLib's header detection and with it the `OX=` regex, so `NcbiTaxonomyId` stays null. #2782
records the same gap and deliberately leaves that call unchanged, because parsing `OX=` changes
other output. So this PR does not change it either: the organism resolver is tested directly, and
the column reads `not available` for FASTA searches until the loader parses `OX=`.

The checkbox sits in **Output Options** (review, `ef446f2df`).

Locally, on `net10.0` Debug against mzLib 1.0.589 with no local package feed:
- Full solution builds with 0 errors, GUI included.
- Full suite **2,018 passed / 0 failed / 1 skipped**, run on the first three commits.
- The last two commits change one test and some wording (tooltip, comments, one warning string).
  The SDRF, `SearchTask` and `PostSearchAnalysis` subsets were re-run after them: **117 passed /
  0 failed**.


MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml
```diff
@@ -492,6 +492,19 @@
                                     </ToolTip>
                                 </ToolTipService.ToolTip>
                             </CheckBox>
+                                <CheckBox x:Name="WriteSdrfCheckBox" Content="Write SDRF experimental-design file">
```
nbollis: This should go in the output options section instead

trishorts: Moved in ef446f2df: it now sits in Output Options, after "Write target-decoy FASTA file".

nbollis (dismissed):

Alexander-Sol:
## Agent review — Request changes

**Summary.** Adds an opt-in `WriteSdrf` parameter that writes `experiment.sdrf.tsv` next to results, delegating format/vocabulary/validation to mzLib. The adapter is thoughtfully defensive (failures wrapped in `EngineCrashed`, gaps warned pre-run and reported post-write, no asserting unknown values), wiring is clean, and test coverage is strong. One confirmed bug defeats a headline goal.

### 🔴 HIGH (confirmed) — per-file parameter lookup never matches; per-file overrides silently flattened to task-level
`PostSearchAnalysisTaskSdrf.cs:94-96`
```csharp
var common = FileSpecificParameters
    ?.FirstOrDefault(f => string.Equals(f.FileName, fileName, StringComparison.OrdinalIgnoreCase))
    .Parameters ?? CommonParameters;
```
`fileName` is `Path.GetFileName(rawFilePath)` (bare, e.g. `run1.mzML`), but `FileSpecificParameters[i].FileName` holds the **full path** (`MetaMorpheusTask.cs:674/689` store `currentRawDataFilepathList[i]`; `SearchTask.cs:619` passes that through). So `string.Equals(fullPath, bareName)` is never true, `FirstOrDefault` returns the default value-tuple, `.Parameters` is null, and it always falls back to `CommonParameters`.

**Scenario:** a 2-file search where `run2.mzML` has a file-specific `.toml` with a different protease/precursor tolerance/dissociation type. Both SDRF rows report the task-level protease and tolerances; `run2`'s real overrides never appear — exactly the "silently flatten real differences between rows" outcome the surrounding comment claims to prevent. Output stays schema-valid, so no validator/coverage check catches it.

**Fix:** match against `rawFilePath` (the full path), not `fileName`. (No NRE here — the value-tuple element type makes it a wrong-value-source bug, not a crash.)

**Test gap that hides it:** every end-to-end `SdrfOutputTests` case uses a single spectra file with no file-specific TOML, so nothing asserts per-file overrides reach the correct row. A 2-file test with a differing file-specific toml, asserting per-row `comment[cleavage agent details]`, would catch it.

### Minor
- **Low (plausible):** pre-run `WarnAboutSdrfGaps` calls `Directory.GetParent(...).ToString()` (NRE if no parent) and `ReadExperimentalDesign` outside any try/catch, gated only on `WriteSdrf` — a new pre-search failure surface. Throw risk is low in practice.
- **Low:** the instrument reader opened at `PostSearchAnalysisTaskSdrf.cs:200` (`MsDataFileReader.GetDataFile(...).GetSourceFile()`) is never disposed. One short-lived handle per file.
- **Low:** `ResolveModifications` (`:262-273`) keys on `IdWithMotif` only, ignoring `ModificationType` — can match the wrong mod / emit duplicates if two known mods share an `IdWithMotif`.

### Verified correct
Back-compat (new bool defaults false, old TOMLs load), GUI round-trip, null-safety of the `sampleInfo?` chains, 0/1-based replicate/fraction conversion, decoy-excluded organism resolution, "don't assert what you don't know" for label/acquisition method, and number formatting delegated to mzLib (no culture-sensitive `ToString` in the adapter).

trishorts:
Thanks — the HIGH one is real, and it is fixed in 74a5f91 along with a test that fails without the fix.

**🔴 per-file parameter lookup — confirmed and fixed.** `MetaMorpheusTask` stores each entry as `(currentRawDataFilepathList[i], …)` (`MetaMorpheusTask.cs:673,688`), i.e. the **full path**, and `SearchTask.cs:619` hands that same list through, so comparing it to `Path.GetFileName(rawFilePath)` never matched. `FirstOrDefault` returned the default value-tuple, `.Parameters` was null, and every row fell back to `CommonParameters` — no crash, schema-valid output, and a file's own `.toml` silently gone. Now matched on `rawFilePath`, which is the same string from the same list.

**Test gap — agreed, and that is why it survived.** Every end-to-end case in `SdrfOutputTests` searched one file with no file-specific toml, so nothing could tell the two sources of parameters apart. `APerFileParameterOverrideReachesThatFilesRow` searches two files, gives the second an `Asp-N` toml, and asserts each row's `comment[cleavage agent details]` separately. Confirmed red on the old lookup (the Asp-N row read "Trypsin"), green on the new one.

The three Low findings I am **not** changing in this PR, deliberately — each is recorded so it does not get lost:

- **`WarnAboutSdrfGaps` throw surface.** Accepted as written: the risk is a spectra path with no parent directory, which cannot happen for a file MetaMorpheus has already accepted into a run. Worth a guard, but not worth re-opening an approved diff.
- **Undisposed `MsDataFileReader`.** Real, one short-lived handle per file, released at collection. It belongs with the other reader-lifetime call sites rather than alone here.
- **`ResolveModifications` keyed on `IdWithMotif` only.** Real — `ModificationType` (the tuple's `Item1`) is dropped, so two known mods sharing an `IdWithMotif` would both match. It needs a decision about what the SDRF should write when that collision happens, not a one-line patch.

Also merged current master, so the branch is no longer behind.

@nbollis sorry — this dismissed your approval; the only production change since it is the one line above.

PR #2820, opened 2026-09-16
Let the glyco task window set the maximum glycan mass (stacked on #2809)

trishorts:
## Summary

**Stacked on #2809.** Review only the last commit (`feat(gui): let the glyco task window set the maximum glycan mass`); everything below it belongs to #2809, which this PR builds on and should merge after.

#2809 adds `GlycoSearchParameters.MaximumGlycanBoxMass` (default 4000 Da), but it could only be changed by editing the task `.toml`. This adds it to the GlycoSearch task window.

- A **Maximum Glycan Mass (Da)** box in the Glyco Search section, right under *Maximum OGlycan Allowed*, which it works together with (both set how many glycan combinations are built).
- The tooltip says the limit **applies to every search type: O-glycan, N-glycan, and N-O-glycan**, and is on the total glycan mass placed on one peptide (the single N-glycan in an N-glycan search). #2809 now applies the cap in N-glycan searches too, so the tooltip is accurate.
- Save refuses a value that is not a positive number: a cap of zero or less drops every glycan, and the search would silently find nothing.

## Tests

- `TestToml.GlycoSearchTaskMaximumGlycanBoxMassRoundTrips`: the value survives save and reload, and a task file saved before the setting existed loads the 4000 Da default.
- GUI builds with no errors; toml + glyco tests: 152 of 152 pass.
- The window code itself is WPF code-behind, which the Test project does not reference. I have not yet opened the window to check the new row visually.


RayMSMS (dismissed):

MetaMorpheus/GUI/TaskWindows/GlycoSearchTaskWindow.xaml.cs
```diff
@@ -286,6 +287,13 @@ private void SaveButton_Click(object sender, RoutedEventArgs e)
                 return;
             }
 
+            // A cap of zero or less would drop every glycan, so the search would silently find nothing.
```
nbollis: Use the double control text box with allow negative to false instead of this parsing. It would prevent it from being entered into the box instead of checking after save.

nbollis (changes requested):

MetaMorpheus/GUI/TaskWindows/GlycoSearchTaskWindow.xaml
```diff
@@ -142,6 +143,21 @@
                                     <local:IntegerTexBoxControl x:Name="TbMaxOGlycanNum" Width="45"/>
                                 </StackPanel>
                                 <CheckBox x:Name="CkbOxoniumIonFilt" Content="OxoniumIonFilt" Grid.Row="3" Grid.Column="1" VerticalAlignment="Center"/>
+                                <StackPanel Orientation="Horizontal" Grid.Row="4" ToolTipService.ShowDuration="999999" ToolTipService.InitialShowDelay="500">
```
RayMSMS: Image

Very small issue with the visual design.
Move the MaxGlycanMass down a little bit, just for the box overlapping

MetaMorpheus/GUI/TaskWindows/GlycoSearchTaskWindow.xaml.cs
trishorts: Done in db8325db4. The save-time parse and message box are gone. The box is a `DoubleTextBoxControl`, which already refuses a minus sign as you type (`IntegerTexBoxControl` is the one with `AllowNegative`). I also set `LowerBound="0"`, and the value is now read with `double.Parse`, like the other double fields in this window.

MetaMorpheus/GUI/TaskWindows/GlycoSearchTaskWindow.xaml
trishorts: Fixed in db8325db4. The row now has a 5 px top margin, so the 4000 box no longer touches the O-glycan box above it.

RayMSMS (approved):

nbollis (approved):

zhuoxinshi (approved):

PR #2821, opened 2026-09-16
Search Parameter Inheritance

nbollis:
Glyco and XL search params now inherit from SearchParameters. 

This required a few renames, so a backwards compatibility structure for the inheritance was also created.


MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -1944,28 +1944,58 @@ protected static void SanitizeBioPolymerDatabase<TBioPolymer>(List<TBioPolymer>
...
+
         /// <summary>
-        /// Legacy TOML compatibility when ProductMassTolerance_LowRes is omitted, the helper falls back to ProductMassTolerance to keep constant result.
+        /// Reads task TOMLs with backwards compatibility for renamed parameter sections/properties.
         /// </summary>
-        /// <typeparam name="TTask"></typeparam>
-        /// <param name="filePath"></param>
-        /// <returns></returns>
-        public static TTask ReadTaskTomlWithLowResFallback<TTask>(string filePath) where TTask : MetaMorpheusTask
+        public static TTask ReadTaskTomlWithBackwardsCompatibility<TTask>(string filePath) where TTask : MetaMorpheusTask
```
trishorts, line 1952: **Blocking:** this replaced `ReadTaskTomlWithLowResFallback`, but the body no longer does the low-res fallback. An old TOML without `ProductMassTolerance_LowRes` used to get `ProductMassTolerance` copied into it; now it keeps the default. That's the one failure in `Test on Windows (core)`:

```
Failed TestLowResToleranceConstruction_Default2_Omit
  Expected: <MzLibUtil.PpmTolerance>
  But was:  <MzLibUtil.AbsoluteTolerance>
  at GlycoSearchEngineTest.cs:line 334
```

Could you put the `CommonParameters` block back, ahead of the task-type branches? Since every task type now goes through this method, it would then cover all of them, not just glyco.

MetaMorpheus/TaskLayer/MetaMorpheusTask.cs
```diff
@@ -1944,28 +1944,58 @@ protected static void SanitizeBioPolymerDatabase<TBioPolymer>(List<TBioPolymer>
...
-                    task.CommonParameters.ProductMassTolerance_LowRes = task.CommonParameters.ProductMassTolerance;
+                    string legacyKey = raw.ContainsKey("GlycoSearchParameters") ? "GlycoSearchParameters"
+                        : raw.ContainsKey(nameof(SearchTask.SearchParameters)) ? nameof(SearchTask.SearchParameters)
+                        : null;
+                    if (legacyKey != null)
+                        glycoTask._glycoSearchParameters = raw.Get<TomlTable>(legacyKey).Get<GlycoSearchParameters>();
                 }
-            }
 
+                if (raw.ContainsKey(nameof(GlycoSearchTask._glycoSearchParameters)))
```
trishorts, line 1978: Non-blocking: aliases only get applied when the section is under the new `_glycoSearchParameters` key. If the section was loaded from a legacy key above (`GlycoSearchParameters` / `SearchParameters`), that table's `DoQuantification` / `DoMbrAnalysis` / `WritePrunedDataBase` / `WriteSpectrumLibrary` are never mapped, so they silently fall back to the defaults. Applying the aliases to whichever table was actually loaded would cover both cases.

trishorts (commented):
I like the direction, but one thing needs fixing before this can go in: the rewritten TOML reader drops the `ProductMassTolerance_LowRes` fallback, and `TestLowResToleranceConstruction_Default2_Omit` fails on CI because of it (details inline). The other two inline comments are non-blocking.

MetaMorpheus/TaskLayer/XLSearchTask/XLSearchParameters.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace TaskLayer
 {
-    public class XlSearchParameters
+    public class XlSearchParameters : SearchParameters
```
trishorts, line 7: Non-blocking question: with the inheritance, XL (and glyco) TOMLs now serialize every `SearchParameters` property, e.g. `DoLabelFreeQuantification = true`, `WriteMzId`, `DoLocalizationAnalysis` and `SilacLabels`, and nothing in the XL task reads them. Someone editing a saved XL TOML could change those and see no effect. Is that acceptable for now, or should the unused ones be ignored on write for these task types?

PR #2822, opened 2026-09-17
Move Max Threads out of Fragment Ion Search Parameters in the calibrate, XL and glyco task windows

### LINKED ISSUE #1887: Relocate "Max Threads" Parameter in Search Task GUI
It's currently in "Fragment Ion Search Parameters". I don't see a great place to put it, but it doesn't make sense for it to be where it currently is located.

For all other task guis, it's located in "Search Parameters"

acesnik:
🤖 Closes #1887.

Max Threads is a per-file threading knob, not a fragment-ion setting, but the calibrate, crosslink and glyco task windows still keep it inside a "Fragment Ion Search Parameters" group box. This moves it into "Search Parameters" in each, which is where the search and G-PTM-D windows already put it.

The search task itself — the window the issue names — was already fixed in passing by #2627, which extracted the fragment-ion group into `FragmentationParamsControl` and left Max Threads behind in the main Search Parameters grid (`SearchTaskWindow.xaml:300`). So this PR picks up the issue's body rather than its title; the three remaining windows are what it was actually reporting.

Per window:

- `CalibrateTaskWindow.xaml` — Max Threads sat three levels down, in Advanced Options → Search Parameters → Fragment Ion Search Parameters, as that group box's only child. It moves to the always-visible Search Parameters grid on the first window (new row 5, left column), mirroring the search task, and the emptied group box is deleted.
- `XLSearchTaskWindow.xaml`, `GlycoSearchTaskWindow.xaml` — Max Threads moves to the top of the Search Parameters group, above the expander, so it reads without expanding. The group box content becomes a `StackPanel` wrapping that row plus the existing `Expander`; the expander's contents are otherwise untouched, and their indentation is deliberately left alone to keep each diff at 31 lines instead of 150.

Label text is normalised to "Max Threads:" at `Width="150"` to match the search task row. `MaxThreadsTextBox` / `maxThreadsTextBox` keep their names, so no code-behind moves.

No behaviour change: same controls, same names, same bindings, same defaults — only the parent container differs. Nothing here touches search output.

Verified: all five task-window XAML files parse, and `dotnet build GUI/GUI.csproj -p:EnableWindowsTargeting=true` on macOS (.NET 10 arm64) gives 0 errors and 98 warnings, none naming the edited files, with `MarkupCompilePass2` regenerating BAML for all three windows. Since a `GroupBox` takes a single child, the new `StackPanel` wrappers would have failed markup compilation had they been malformed. **The layout was not verified visually** — WPF will not run on macOS, so the new rows want a look on Windows before merge. Two spots to eyeball: the calibrate grid now has a sixth star-sized row, and the XL/glyco Fragment Ion group's second column is left holding only "Minimum score allowed".


trishorts (approved):
Looks good. The new calibrate row 5 is free, and the control names are unchanged, so the code-behind is unaffected. As you noted, the layout still wants a quick visual check on Windows.

PR #2824, opened 2026-09-17
Let the search task window set match-between-runs and normalization for SILAC

### LINKED ISSUE #1795: GUI: MBR and Normalization checkboxes are disabled for SILAC

acesnik:
🤖 Closes #1795. SILAC quantification runs through FlashLFQ exactly like LFQ does, but the GUI hides the match-between-runs and normalization controls whenever the SILAC radio button is selected, so those settings are unreachable from the search task window.

## The checkboxes were never deliberately disabled for SILAC

The binding is collateral damage from a 2019 rename. Before the quantification radio group existed, one master checkbox named `checkBoxQuantification` meant "do quantification at all", and the FlashLFQ options bound to it. The SILAC PR itself, 2f5f187b (#1574, April 2019), deleted `x:Name="checkBoxQuantification"` and put the `QuantType` radio group in its place — but left the FlashLFQ options still bound to `ElementName=checkBoxQuantification`, a name that no longer existed anywhere in the file. WPF resolves a dangling `ElementName` by falling back to the default, `IsEnabled="True"`, so from April to October 2019 MBR and normalization were, accidentally, available under SILAC.

0798f84f ("disabled SILAC turnover. fixed binding reference names", #1734, October 2019) then swept ~10 stale `checkBox*` references onto their new capitalized names, and `checkBoxQuantification` happened to map onto the `CheckBoxLFQ` radio button rather than onto "quantification is enabled":

```diff
-<CheckBox x:Name="CheckBoxMatchBetweenRuns" ... IsEnabled="{Binding IsChecked, ElementName=checkBoxQuantification}">
+<CheckBox x:Name="CheckBoxMatchBetweenRuns" ... IsEnabled="{Binding IsChecked, ElementName=CheckBoxLFQ}">
```

That is what actually disabled them for SILAC — two months before this issue was filed that December. The `Visibility` bindings that make the controls *vanish* rather than grey out came later still, when #2605 removed SpectralRecovery and collapsed a `CheckBoxLFQ`-or-`CheckBoxLFQwSpectralRecovery` MultiBinding down to a single binding. No commit anywhere makes a deliberate decision about SILAC + MBR.

The block is also leaky today. `SearchTaskWindow.xaml.cs` loads and saves `MatchBetweenRuns`, `Normalize` and `MbrFdrThreshold` unconditionally, and `Visibility` does not change `IsChecked` — so opening a TOML with `MatchBetweenRuns = true` and switching to SILAC leaves MBR **on, invisible, and saved back out**. `PostSearchAnalysisTask` passes all of these to `FlashLfqEngine` with no SILAC branch, and the CLI/TOML path has no guard at all.

## Both settings are sound for SILAC

- **Normalization.** Every factor `IntensityNormalizationEngine` computes is a single scalar per `SpectraFileInfo`, applied to every isotopic envelope of every peak in that file (`IntensityNormalizationEngine.cs:220-232`, `:303-315`, `:105-114`). MetaMorpheus keeps light and heavy identifications on the *same* `SpectraFileInfo` during the FlashLFQ run and only splits into pseudo light/heavy files afterwards, so both channels get the identical scalar and H/L ratios are mathematically unchanged; only absolute intensities move. With the usual flat experimental design (one condition, one biorep, one fraction) it is a no-op — `NormalizeFractions` returns immediately when unfractionated, and a single biorep compared against itself yields a factor of 1.
- **Match between runs.** MBR keys on `ModifiedSequence` and searches the acceptor file at the donor's `PeakfindingMass` (`FlashLfqEngine.cs:645`, `:1212-1213`), and skips the donor's own file (`:915-918`). Light and heavy share a base sequence but differ by ≥4 Da, so a transfer cannot cross channels. It is actively useful here: if a file has an MS2 ID for only one channel, the other channel's full sequence still counts as missing, so MBR goes and finds it.

## ⚠️ This changes SILAC output when MBR is on

Turning MBR on for SILAC exposes a real defect, so this PR fixes it in the same change rather than shipping the GUI half alone.

FlashLFQ leaves random-retention-time MBR decoy peaks in `FlashLfqResults.Peaks` so it can estimate MBR FDR, and it filters them out again when it builds the peptide tables (`FlashLFQResults.cs:147-149`). `SilacConversions.CalculateSilacIntensities` recomputes SILAC intensities straight off `Peaks` with no such filter, and it merges peaks by full sequence:

```csharp
string fullSequence = peak.Identifications.First().ModifiedSequence;
if (sequenceToPeakDictionary.ContainsKey(fullSequence))
{
    ChromatographicPeak previousPeak = sequenceToPeakDictionary[fullSequence];
    previousPeak.IsotopicEnvelopes.AddRange(peak.IsotopicEnvelopes);
}
```

A random-RT decoy is constructed from its donor's identification (`FlashLfqEngine.cs:1261`), so it carries the **same** `ModifiedSequence` as the target peak and gets `AddRange`d straight into it. Its envelopes then count toward the channel's summed intensity. The decoys are not removed beforehand: when PEP analysis does not run — which needs >100 MBR peaks and >20 decoys, i.e. not a small SILAC run — `CalculateFdrForMbrPeaks` deliberately keeps them ("better to err on the safe side and not remove the decoys", `FlashLfqEngine.cs:1445-1452`).

The fix skips decoy and random-RT peaks when building that lookup.

`TestSilacIntensitiesIgnoreMbrDecoyPeaks` covers it: light and heavy in one file at a 2:1 ratio, plus a random-RT decoy built from the light donor. I confirmed it is not vacuous by mutating the guard away — the light channel then reads **3030 instead of 30** (`Expected: 30 +/- 1E-06, But was: 3030.0d`), i.e. the decoy's 3 x 1000 envelopes summed straight into the target. Restoring the guard makes it pass again.

**This is inert for existing SILAC runs.** Decoy identifications are only handed to FlashLFQ when MBR is on (`PostSearchAnalysisTask.cs:330`, `includeDecoys: Parameters.SearchParameters.MatchBetweenRuns`) and random-RT peaks only exist when MBR ran, so with MBR off the new predicate is always false and not a single intensity moves. That argument is from the code, not from a measured before/after on real spectra — worth a second pair of eyes.

## Not changed

Two more controls are hidden under SILAC by the same accident. I left them out to keep this reviewable, but they are real:

- **`PeakFindingToleranceTextBox`** — `QuantifyPpmTol` *is* applied to SILAC, so a SILAC user currently cannot set the peakfinding tolerance at all.
- **`CheckBoxUseSharedPeptidesForQuant`** — worth fixing properly first rather than exposing: `SilacConversions.cs:453` calls `CalculateProteinResultsMedianPolish(true)` with `useSharedPeptides` hardcoded, so `UseSharedPeptidesForLFQ` is ignored for SILAC protein quant. Un-hiding the checkbox would show a control that does nothing.

One gap remains in the filter. FlashLFQ also drops MBR peaks failing the q-value threshold, but `MbrChromatographicPeak.MbrQValue` is `internal` to mzLib, so MetaMorpheus cannot apply that part. Making it public is a one-word mzLib change; happy to open it if you want the filter completed.

Multiplex is correctly excluded throughout — it returns before FlashLFQ ever runs (`PostSearchAnalysisTask.cs:257-265`).


MetaMorpheus/GUI/TaskWindows/SearchTaskWindow.xaml
```diff
@@ -1,4 +1,4 @@
-<Window x:Class="MetaMorpheusGUI.SearchTaskWindow"
+﻿<Window x:Class="MetaMorpheusGUI.SearchTaskWindow"
```
trishorts, line 1: Nit: this adds a UTF-8 BOM to line 1, which is unrelated to the change. Could you drop it?

trishorts (approved):
Looks good. `boolOrConverter` and `boolOr2VisConverter` are already declared as resources in this window, and `CheckBoxSILAC` exists, so the MultiBindings will resolve at runtime. The decoy/random-RT guard in `CalculateSilacIntensities` matches what FlashLFQ does when it builds its own peptide tables, and it's a no-op with MBR off. One nit inline.

PR #2825, opened 2026-09-17
Name the spectra file in calibration warnings

### LINKED ISSUE #1646: report the name of the file that failed to calibrate
image

acesnik:
🤖 Closes #1646. Calibration emits its warnings once per spectra file and the text was identical between files, so a run over several files filled the notification list with interchangeable rows — `Calibration failure! Could not find enough high-quality PSMs. Required 20, saw 12` — with nothing to say which file each one came from.

Every per-file warning in `CalibrationTask` now leads with the file name. From the new test's run:

```
warnFileName.mzML: Could not find enough PSMs to calibrate with; opening up tolerances to 20 ppm precursor and 60 ppm product
warnFileName.mzML: Calibration failure! Could not find enough high-quality PSMs. Required 16, saw 0
```

That covers the widened-tolerance warning, all three `Calibration failure!` variants (PSMs, MS1 datapoints, MS2 datapoints) and the "Calibration for .mgf files is not supported" message. `WarnForWiderTolerance` and `CalibrationWarnMessage` take the name as an optional parameter, and a shared `SpectraFilePrefix` helper yields an empty prefix when none is supplied, so the message bodies are unchanged and the existing assertions on them still hold. Only the notification strings change — no calibration behavior, no output files, no values.

`CalibrationWarningsNameTheSpectraFile` calibrates a yeast file against a GAPDH-only database, which widens the tolerances, then gives up, exercising both warning paths in one run. Mutating `SpectraFilePrefix` to return `""` makes it fail on the prefix assertion while both preconditions still pass, so the failure is attributable to the prefix rather than to the run producing no warnings.

Verified on macOS arm64 / .NET 10: the solution builds clean with no new warnings. `Test.csproj` can't run on macOS, so the tests ran through a throwaway harness compiling `CalibrationTests.cs` against EngineLayer and TaskLayer — `CalibrationWarningsNameTheSpectraFile`, `ExperimentalDesignCalibrationTest` (both cases) and `CalibrationTestNoPsms` (both search types), 5/5 pass. Windows CI will be the real check.


trishorts (approved):
Looks good. The prefix is additive, the message bodies are unchanged, and the new test pins the prefix on both warning paths.

PR #2827, opened 2026-09-17
Don't annotate an MS3 child scan against a peptide it isn't evidence for

### LINKED ISSUE #2826: Cross-link MS3 child scans are annotated against a peptide whose stub masses they did not match
🤖 Split out of #2717 at @trishorts's request. Pre-existing behaviour, not introduced there.

`ScoreChildScan` in `MetaMorpheus/EngineLayer/CrosslinkSearch/CrosslinkSearchEngine.cs` chooses the fragments to annotate a child scan with in two steps. The first branch, at `:733`, handles a genuine cleaved signature ion:

```csharp
if (Crosslinker.Cleavable && childScan.TheScan.MsnOrder == 3 && (shortMassAlphaMs3 || longMassAlphaMs3))
```

`shortMassAlphaMs3` and `longMassAlphaMs3` are both computed against `mainPeptide`, so the test is "this child's precursor is this peptide plus one of the crosslinker stubs". When it fails, control reaches the `else`, and for an MS3 child that means `:767`:

```csharp
//It tried to corver the situation that the MS3 scan have the same precursor mass as the parent MS2.  
//This is so rare that I only see it in one data. May need unit test in the future.
childProducts = CrosslinkedPeptide.XlGetTheoreticalFragments(CommonParameters.MS3ChildScanDissociationType,
    Crosslinker, new List<int> { possibleSite }, otherPeptide.MonoisotopicMass, mainPeptide).First().Item2;
```

So a scan rejected by the first branch *because it is not this peptide's signature ion* is annotated against that peptide anyway, and its matched ions feed that peptide's score. Nothing declines to annotate: the `return null` that would do it is commented out at `:771-774`.

`TestMs2Ms3` exercises this, which is how it surfaced. @trishorts instrumented it while reviewing #2717, and its child scan 4 is beta's signature ion:

```
DSSO  CleaveMassShort=54.01056  CleaveMassLong=85.982635  CleaveDissociationTypes=[CID,HCD]
parent scan 2  MsnOrder=2  PrecursorMass=3437.5264
  child scan 4  MsnOrder=3  PrecursorMass=1519.7202   (child - parent) = -1917.8062
```

When alpha is scored, both stub tests miss, so scan 4 falls to `:767` and is annotated against alpha. Alpha is not present in that scan under any reading, so the matched ion is coincidental. The fallback's own comment describes a different situation — an MS3 sharing the parent MS2's precursor mass — which is not the case here either: the child precursor is 1917.8 Da from the parent's.

Two things to settle, and a PR proposing an answer to both is linked below.

1. Should the fallback decline to annotate — the commented-out `return null` — when an MS3 child's precursor matches neither the parent's nor either stub mass?
2. If it should still annotate, against which masses? Either way the comment wants rewriting to describe the case it is actually handling.

#2717 changed the dissociation-type gate inside `XlGetTheoreticalFragments`, so it changed *which* masses `:767` localizes when a search runs under LowCID, but not the fact that it annotates. Fixing this is independent of that, and should not hold #2717 up.

Line numbers are as of the current `master` merge on #2717.

acesnik:
🤖 Closes #2826.

**This changes crosslink search output.** Measured below rather than argued. The headline scores do not move, but child-scan annotations and `MS3ChildScore` do.

## The problem

`ScoreChildScan` chose fragments for a child scan with a two-way `if/else`. The first branch handled a genuine cleaved signature ion — the child's precursor matching `mainPeptide` plus a crosslinker stub. The `else` then split on `MsnOrder == 2`, and for anything else annotated the child against `mainPeptide` regardless. So a scan rejected by the first branch *because it is not this peptide's signature ion* was annotated against that peptide anyway, and its matched ions fed that peptide's score. Nothing declined to annotate; the `return null` that would have was commented out.

The fallback's own comment described a different case entirely — an MS3 sharing the parent MS2's precursor mass — which was not what it was handling.

## The fix

A four-way chain ending in `return null`: stub-signature match, then `MsnOrder == 2`, then the same-precursor-as-parent case the comment describes, then nothing.

Deliberately narrower than the author's commented-out intent. Their `else { return null; }` sat *after* the `MsnOrder == 2` split, so restoring it literally would also have caught MS2 children whose deconvoluted precursor differs from the parent's by more than 0.001 Da. The MS2-child path is untouched here.

## Measured output change

Two-arm comparison on `XlTestData/10226.mzML`, the `TestMs2Ms3` fixture: BSA, DSSO, CID parent with LowCID MS3 children. Each arm run repeatedly, every printed value byte-identical within an arm, and `EngineLayer.dll` hashes flipped on every swap so each arm genuinely recompiled.

All four children of parent scan 2 are MS3, and none is within 1500 Da of the parent's precursor. Scans 6 and 7 are the **alpha** peptide's DSSO stub signatures — 31.973 Da apart, the short/long cleave-mass gap. Scans 4 and 5 are **beta's**, 31.969 Da apart.

| | before | after |
|---|---|---|
| alpha `ChildMatchedFragmentIons` | `{4→1, 5→1, 6→10, 7→5}` | `{6→10, 7→5}` |
| beta `ChildMatchedFragmentIons` | `{4→5, 5→3, 6→0, 7→0}` | `{4→5, 5→3}` |
| alpha `MS3ChildScore` | 17.08808334947279 | 15.086720142850464 |
| `XLTotalScore` | 68.340401234277635 | unchanged |
| `Score`, `BetaPeptide.Score`, `DeltaScore` | | unchanged, bit-identical |

Alpha was matching one ion apiece against *beta's* two signature spectra, and beta was carrying two empty entries against alpha's. The alpha entries are the coincidental matches this PR is about, and removing them is the whole of the −2.0014 change in `MS3ChildScore`. The beta entries held zero ions, so dropping them changes only the dictionary count and the TSV round trip.

Worth flagging separately: `XLTotalScore` does not move because `MS3ChildScore` is not feeding it on this path. A cleaved-MS3 experiment's child evidence not reaching the score that ranks CSMs looks like its own bug, but not one to fix here.

## Test

The four child-scan assertions in `TestMs2Ms3` now check the whole dictionary instead of `First()`. The positional form is what let the old assertions read as "alpha peptide signature ion … scan 4" when scan 4 is in fact beta's; the dictionary form states which scans are annotated for whom.

`Is.EqualTo` on a `Dictionary<int, int>` was probed rather than assumed: under NUnit 4 it is order-insensitive but compares keys, values and count, and — the case that matters here — a superset does **not** satisfy it. A four-entry actual against a two-entry expectation fails. So these assertions pin the *removal* of the spurious entries, not merely the survival of the real ones.

With the fix, `TestMs2Ms3` passes. With the fix reverted and the new assertions kept, **all four fail**, each on element count, with `EngineLayer.dll`'s hash confirmed to change on every swap so no arm was measuring a stale binary. It is a real regression test, not a vacuous one.

Run through a `net10.0` harness on macOS/arm64, since `Test.csproj` is `net*-windows` and cannot host a run here; the harness took the method verbatim apart from the `XlTestData\10226.mzML` path separator. The full suite on Windows is still CI's job.

## Known gap

The third branch — `Math.Abs(childScan.PrecursorMass - parentScan.PrecursorMass) < 0.001` — never fires on this fixture, and nothing in the test tree appears to produce an MS3 child sharing its parent MS2's precursor. It is carried over from the author's commented-out intent to preserve the case their comment describes, and it is unexercised rather than newly tested. Removing it instead would be defensible; I kept it because the comment says that data exists.

## Interaction with #2717

#2717 moves the same two alpha assertions from 1 to 2, for the same scan 4 that this PR removes from alpha entirely. Whichever lands second will need its share of `XLTest.cs` reconciled, and the numbers here support what #2717's own description already says: that 1 → 2 move was coincidental, and this removes the coincidence rather than enshrining it.


trishorts (approved):
Looks good. The chain now declines to annotate an MS3 child that is neither this peptide's stub signature nor a same-precursor MS3, and both callers already skip a null result. The dictionary-equality assertions pin the removal of the spurious entries.

Issue #2823, opened 2026-09-17
GPTMD acceptance filters default to off


acesnik:
🤖 The GPTMD acceptance filters from #2542 do nothing unless a user selects them. That's worth an explicit decision rather than an inherited default.

`GptmdParameters()` initializes `GptmdFilters` to an empty list (`TaskLayer/GPTMDTask/GPTMDParameters.cs:14`), and `GptmdEngine` rejects a candidate only when some filter says no (`EngineLayer/Gptmd/GptmdEngine.cs:154`):

```csharp
if (Filters.Any(f => !f.Passes(newPep, psm, score, originalScore, matchedIons, modSiteInPeptideIndex, pepWithSetMods.Length, mod)))
    continue;
```

An empty list means no filtering, i.e. the pre-#2542 behaviour. The task window comes up clear as well: `PopulateChoices()` constructs all four view models selected, then `UpdateFieldsFromTask()` overwrites `IsSelected` from the task's empty saved list (`GUI/TaskWindows/GPTMDTaskWindow.xaml.cs:215`). As a side effect the `isSelected = true` default in `GuiFunctions/ViewModels/GptmdFilterViewModel.cs:18` never takes effect.

Changing the default is not a cleanup — enabling any filter changes which modifications are written to the GPTMD database, and therefore the results of every search run against that database. `ImprovedScoreFilter` is the mildest candidate if we want one on.

Two smaller things in the same area:

- The TOML converter silently drops filter names it doesn't recognize (`TaskLayer/MetaMorpheusTask.cs:142-153`, `.Where(t => t != null)`), so a typo in `GptmdFilters` gives an unfiltered run with no warning.
- I didn't find a test covering the `GptmdFilters` TOML round-trip — `GPTMDengineTest.cs:428` passes the list into the engine but doesn't exercise serialization.

Issue #1190, opened 2018-05-25
don't let gptmd look for amino acid substitutions on amino acids associated with the protease


trishorts:
e.g. G->R is the second most prominent substitution in mouse. however, this would result in an unpredictable mass shift because the new peptide would get truncated at R, removing an unknown portion of the original peptide. There would be multiple mass shifts associated with this substitution. Therefore, it should not be allowed in a simple G-PTM-D type discovery of amino acid substitutions.

acesnik:
🤖 Worth recording that the GPTMD filters added in #2542 don't cover this case. All four are position-based — they ask whether matched ions cover or flank the modified residue (`EngineLayer/Gptmd/IGptmdFilter.cs:49-154`) — and none is aware of the protease.

The distinction matters because this is a different failure. Coverage filters assume the mass shift is right and only the position is in doubt. Here the shift itself is wrong: if `G->R` introduces a tryptic site, the real variant peptide is truncated, so no spectrum has mass `peptide + 156.101 - 57.021`. The hypothesis can therefore only ever be matched by something that isn't a `G->R` substitution, which is what makes it pure false-positive surface.

Rather than dropping these from the mod list up front, the filter interface added in #2542 looks like a better home for it, because it is position-aware and so doesn't over-exclude. `IGptmdFilter.Passes` already receives the candidate peptide and the site, `IBioPolymerWithSetMods` exposes `DigestionParams` (`mzLib/Omics/IBioPolymerWithSetMods.cs:39`), and `DigestionAgent.GetDigestionSiteIndices(string)` (`mzLib/Omics/Digestion/DigestionAgent.cs:77`) will enumerate cut sites for whichever agent the task is using. A filter could apply the substitution to the peptide sequence plus a residue of flanking context, compare the site indices before and after, and reject the candidate when they differ — which covers creation, destruction and the not-before-proline case in one rule, and works for RNases too without hardcoding K/R.

mzLib now pins the semantics of all three on the sequence-variant side, in smith-chem-wisc/mzLib#1098: `S07` sub-destroys-cut-site, `S08` sub-creates-cut-site, `S09` sub-blocks-cut-before-proline (`Test/DatabaseTests/VariantCorpus/VariantCorpusTests.cs:338,348,360`).

PR #2828, opened 2026-09-17
Add a GPTMD filter that rejects substitutions which move a cleavage site

acesnik:
🤖 #1190 observes that an amino acid substitution at or near a protease recognition residue cannot be discovered reliably by a mass-difference search, and names `G->R` as the second most prominent substitution in mouse. Where such a substitution destroys or blocks the cleavage that defines one of the peptide's termini, the peptide being scored is not produced at all. Where it creates a site inside the peptide, the uncleaved form survives as a missed-cleavage peptide, but the signal splits across it and the two cleaved products. Either way the plain "peptide mass plus substitution delta" hypothesis is unreliable, which is the false-positive surface that issue describes.

`CleavageSiteFilter` is a new `IGptmdFilter` that rejects those candidates. The GPTMD task window offers it as "Cleavage Site Preserved".

## Why a filter rather than a shorter mod list

#1190 proposes excluding these substitutions from the searchable set. The filter interface added in #2542 is a better place for the rule, because it sees the specific peptide and the specific site and can ask the narrow question: does applying *this* substitution *here* change the digestion? Excluding them up front is position-blind — under trypsin it would have to drop every `X->K`, `X->R`, `K->X`, `R->X` and `X->P`, a large share of the 378 entries in `substitutions.txt`, including cases that are perfectly well posed. `K->R` preserves cleavage entirely, and a `G->R` in a peptide that already ends in R moves nothing.

## How it decides

The filter substitutes the candidate residue into a window of the parent sequence and compares `DigestionAgent.GetDigestionSiteIndices` before and after. Any difference means the digestion moved, so the candidate is rejected. Going through the digestion agent rather than a hardcoded K/R table covers site creation, site destruction and the not-before-proline rule with one comparison, and holds for RNases without change.

The window extends past each end of the peptide by twice the longest `InducingCleavage + PreventingCleavage` motif length. That is further than a cut decision ever reads from its own location in `DigestionMotif.Fits`, so no site the substitution could move falls outside it. Both windows share the same boundaries, so any edge effect cancels and the only difference between them is the substituted residue.

Substitutions are recognised the way mzLib recognises them when it converts them to sequence variants — a `nucleotide substitution` modification type with an `X->Y` identifier, per `VariantApplication.ConvertNucleotideSubstitutionModificationsToSequenceVariants`. Everything else returns `true` immediately, so this is inert for ordinary PTMs.

The filter sits on the single code path in `GptmdEngine` that target and decoy candidates both take, so neither is treated specially.

## Scope, and one deliberate piece of strictness

Both of these are choices rather than consequences, so they are worth stating plainly.

It only applies to fully specific searches. A `SearchModeType` other than `CleavageSpecificity.Full` returns immediately, because a semi- or non-specific peptide is not required to end at a cleavage site and moving one therefore says nothing about whether the candidate is real.

It rejects any substitution that moves a site, including the create-a-site case where the uncleaved form is still producible within the missed-cleavage allowance. That is stricter than rejecting only the impossible. I went that way because those are exactly the candidates #1190 is about, and because a filter whose purpose is stringency seems like the wrong place to be generous — rejecting only provably impossible peptides would let the whole `G->R` class back in. If you would rather have the precise version it is a small change: compare the peptide's own start and end against the substituted site set and count the internal sites against `MaxMissedCleavages`. Happy to switch if you think that is the better default.

## Output impact

**Selecting this filter changes which modifications are written to the GPTMD database, and therefore the results of every search run against that database.** That is the intent, but it is not a no-op.

Nothing changes for anyone who does not select it. Like the other four filters it is opt-in, and `GptmdParameters.GptmdFilters` is empty by default — #2823 tracks whether that default should change.

I have not run this against real data, so I cannot say how many candidates it removes in practice. That seems worth measuring before it is considered for on-by-default.

## Tests

Eleven cases in `CleavageSiteFilterTests`, all built on real digested peptides so the filter reads a genuine `DigestionAgent` rather than a stub:

- `T->R` inside `PEPTIDEK` creates a cut site — rejected
- `K->A` at the peptide's last residue destroys the site that ends it — rejected
- `A->P` at the N terminus of `AAAR` under `trypsin|P` blocks the preceding cut — rejected
- `T->S` moves no site — accepted
- `K->R` keeps the site — accepted
- `A->P` under plain `trypsin`, whose motif is `K|,R|` and has no proline rule — accepted
- a modification that is not a substitution — accepted
- a candidate with no digestion agent — accepted
- site 1, which is the peptide N terminus rather than a residue — accepted
- a semi-specific search — accepted, pinning the specificity guard
- `T->R` under `maxMissedCleavages: 2` — still rejected

The last one does not reach any code the first case doesn't; `Passes` never consults `MaxMissedCleavages`. It is there to pin the stringency choice described above so it cannot change silently, and it would catch a future version that started taking missed cleavages into account. I verified separately that `PEPTIDEKAAAR` really does digest to both `PEPTIDEK` and the full-length form at that setting, so the premise behind it holds.

I confirmed by mutation that the tests fail against unfixed code: forcing `Passes` to `return true` fails exactly the three rejecting cases, and deleting the specificity guard fails exactly the semi-specific case. The 23 pre-existing `GptmdFilterTests` still pass.

`Test/Test.csproj` targets `net10.0-windows` and cannot run on macOS, so these were run through a throwaway NUnit harness that references `EngineLayer` and compiles only `GptmdFilterTests.cs`. The full suite runs in CI.


MetaMorpheus/GUI/TaskWindows/GPTMDTaskWindow.xaml.cs
```diff
@@ -1,4 +1,4 @@
-using EngineLayer;
+﻿using EngineLayer;
```
trishorts, line 1: Nit: this adds a UTF-8 BOM to the first line (same in `Test/GptmdFilterTests.cs`). It's probably an editor artifact. Could you drop it so the diff only has the payload?

trishorts (approved):
Looks good. I checked the two things the filter depends on in real GPTMD runs: `GptmdEngine` passes `pepSeqIndex + 2` as the site, which matches the `- 2` here, and the `newPep` it hands the filter comes from `Localize`, which keeps `Protein` and `_digestionParams`, so the filter isn't silently inert outside the tests. The `GetDigestionSiteIndices` before/after set comparison is sound, since both windows share their edges. The stricter create-a-site rejection seems like the right default for an opt-in stringency filter. One small nit inline.

PR #2833, opened 2026-09-17
Enable RNA label-free quantification

nbollis:
## Motivation  RNA searches could expose label-free quantification settings, but RNA mode hid the controls and FlashLFQ crashed while calculating isotope distributions because RNA chemical formulas were not passed to the quantification engine.  ## Summary  - Expose the existing quantification controls in the Search Task window for RNA mode without enabling quantification by default. - Add RNA integration coverage for explicit label-free quantification. - Pass each identified bio-polymer chemical formula to FlashLFQ through Identification.OptionalChemicalFormula. - Preserve mzLib unchanged.  ## Testing  - Release solution build passed with 0 errors. - RNA transcriptomics tests passed: 2/2. - Re-ran the original four RAW files with the failing configuration; quantification completed and produced AllQuantifiedOligos.tsv, AllQuantifiedPeaks.tsv, and AllQuantifiedTranscriptGroups.tsv without Quantification_crash.txt.


Alexander-Sol (approved):
It works great for now. I opened a  PR in FlashLFQ, https://github.com/smith-chem-wisc/mzLib/pull/1319. This change is compatible with that PR.

However, we should revisit the implementation after that PR is merged and mzLib is released to make things less ad hoc

Issue #1330, opened 2018-08-03
HTML reports


rmillikin:
I thought this idea was large enough where it should be it's own issue even though it's part of #1124 .

@acesnik wants some HTML reports and I think a lot of users really want some way to easily/graphically interpret the data coming out of MetaMorpheus/FlashLFQ. 

One way to do it is to have static plots that describe things we think the user would want, i.e. ppm error over time or PSMs per file or boxplots of protein abundance. 

Another way to do it is to have some kind of dynamic system that would let the user plot things that they want - i.e. define a statistic to make a box plot out of, or plot any x vs y (i.e. ppm error vs RT).

So here is an issue to centralize some ideas of what would be nice for users to have (hopefully with some input from users). I suspect this issue will never really be "closed" but is an open-ended place to suggest improvements.

We are already working on a MS2 spectrum annotator within MetaMorpheus so that particular type of plot probably doesn't belong here. We'd also have to load spectra files which I think isn't really easily doable in HTML.

Some ideas for static plots:
- [ ] Summary statistics and histogram for PSMs by file and overall
- [ ] Summary statistics and histogram for sequence variations when we get there
- [ ] Summary statistics and histogram for modified peptides and unique PTMs discovered

acesnik:
Another way to think about this is answering typical questions one would have about each task: did calibration work? Did GPTMD add any modifications? Did search find any of those modifications? How many peptides were found in search? What portion of spectra resulted in a PSM? How many peptides were identified in consolation events? How many peptides were quantified? What peptides were elevated or lower in abundance in my condition of interest (volcano plot with tooltip displays)?

acesnik:
Could use Markdown instead of writing HTML directly https://github.com/Knagis/CommonMark.NET

rmillikin:
here's some ideas I was thinking.. feel free to contribute
- Histogram of precursor PPM tolerances (around 0 Da mass-difference notch only)
- Histogram of fragment PPM tolerances
- Histogram of precursor charges
- Histogram of fragment charges
- Precursor PPM error vs. RT
- Fragment PPM error vs. RT
- Histograms of count of different PTMs seen at 1% FDR

For multifile runs:
- Bar chart of PSMs per file (maybe also include precursors fragmented per file in the same bar chart)

acesnik:
Where's that from?

rmillikin:
[...]

PR #2835, opened 2026-09-18
Support shared protein and RNA non-specific search handling

nbollis:
Companion PR to Mzlib #1318

## Motivation

The mzLib digestion refactor now represents protein and RNA digestion through shared effective-agent and terminus semantics. MetaMorpheus's non-specific search engine still assumed peptide-only indexes and interpreted N/C termini without considering RNA's five-prime/three-prime terminology.

## Summary

- Generalize non-specific search indexing, scoring, trimming, and product handling from peptides to IBioPolymerWithSetMods.
- Materialize updated candidates as either PeptideWithSetModifications or OligoWithSetMods.
- Normalize RNA singleN/singleC effective agents to five-prime/three-prime search termini for product types, fragmentation, seed trimming, and terminal modifications.
- Route RNA semi-specific searches through five-prime and three-prime passes rather than protein N/C passes.
- Make FDR reconciliation safe when a major category is empty.
- Update digestion-parameter test doubles for SpecificDigestionAgent.
- Update search and spectral-library tests for the current mzLib sequence/modification contract.

## Testing notes

- Built MetaMorpheus against the local mzLib package normalized by NuGet as 1.8.4 from the requested 1.8.04.
- Full MetaMorpheus suite: **2,133 passed, 1 skipped, 0 failed**.
- Targeted analyte-aware search tests: 61 passed.
- mzLib RNA effective-agent tests: 25 passed.

The local working tree contains uncommitted package-reference and local NuGet configuration changes used only for local verification; these are intentionally excluded from this PR.


trishorts:
Read this because it is the PR mzLib's red `integration` job is waiting on — your two `SpecificDigestionAgent` ignores are the fix for all 26 MetaMorpheus failures that appear once mzLib master can compile MetaMorpheus again (measured on mzLib #1330's run: 2243 passed / **26 failed**, every one of them tracing to `Failed to convert TOML object with key 'SpecificDigestionAgent'`). Write-up is on [mzLib #1330](https://github.com/smith-chem-wisc/mzLib/pull/1330#issuecomment-5768011353). Those two lines are right and needed.

**One line next to them is not:** `.IgnoreProperty(p => p.SpecificProtease)`.

### It silently rewrites a non-specific search's enzyme to trypsin

`SpecificProtease` is not a derived convenience like `SpecificDigestionAgent` — for `SearchModeType == None` it is the **only** record of the enzyme the user chose. `RecordSpecificProtease()` parks the named protease there and swaps `Protease` for `singleN`/`singleC` (`DigestionParams.cs:185-192`), so the written toml is:

```toml
Protease = "singleN"
SpecificProtease = "Lys-C"
SearchModeType = "None"
```

Three places on master read it back as such: `ReadProteinDigestionParams` says so in its own remarks and reads the key by hand (`MetaMorpheusTask.cs:142-144`), `TaskWindowSearchMode.ProteaseToShow(loaded) => loaded.SpecificProtease` (`GuiFunctions/Util/TaskWindowSearchMode.cs:48`), and `IndexingEngine.ToString()` puts it in the index cache fingerprint (`EngineLayer/Indexing/IndexingEngine.cs:89`).

Ignoring the property takes it out of the round trip in both directions. And it does not come back **empty** — `DigestionParams`'s parameterless ctor has already seeded it:

```csharp
public DigestionParams() : this("trypsin") { }
```

so it comes back **trypsin**, whatever the user actually ran. Measured against Nett 0.15.0 on a class with this exact shape (parameterless ctor seeding the specific agent, `{ get; private set; }`, one get-only property):

| toml config | read back |
|---|---|
| `SpecificProtease` not ignored (master today) | `Protease=singleN`  `Specific=Lys-C` ✅ |
| `SpecificProtease` ignored (this PR) | `Protease=singleN`  `Specific=trypsin` ❌ — and the key is dropped on write |
| get-only property, not ignored | `ArgumentException : Property set method not found.` (the 26 failures) |

**Failure scenario:** save a non-specific search that uses Lys-C; reload the toml. The GUI protease box shows trypsin, the index fingerprint says `specificProtease: trypsin`, and missed cleavages are counted against trypsin's sites. No warning, and the toml on disk no longer contains the answer, so it is not recoverable by re-reading it.

`SpecificProtease` also needs no silencing: it is typed `Protease`, which already has a `TomlString` converter (`MetaMorpheusTask.cs`, `ConfigureType<Protease>`). Only the get-only `SpecificDigestionAgent` does.

### `SpecificRnase` is the same bug, and this PR is what makes it live

`RnaDigestionParams.SpecificRnase` is new in #1318, so nothing has ever persisted it and ignoring it loses nothing *today*. But it is typed `DigestionAgent`, which has no converter — that is presumably why you reached for the ignore — and **this PR is the one that adds RNA `singleN`/`singleC`**, i.e. exactly the case where `SpecificRnase` stops equalling `Rnase`. So it will start losing the named rnase the moment the feature is used.

### Suggested

- Drop `.IgnoreProperty(p => p.SpecificProtease)`. Keep both `SpecificDigestionAgent` ignores.
- For RNA, either type `SpecificRnase` as `Rnase` in mzLib so the existing `ConfigureType<Rnase>` converter applies, or add a `ConfigureType<DigestionAgent>` converter. Keeping the ignore works only while nobody saves an RNA non-specific task.
- Worth a test either way, because nothing currently covers it: write a `SearchModeType = None` `DigestionParams` to toml, read it back, assert `SpecificProtease.Name` survived. It passes on master and fails on this branch. Happy to push that to this branch or send it as a follow-up — say which you prefer.

